Issue #117, opened 2011-12-05
Show polyline normals


mrmalee:
Hi, first off its great that Tiled now has polyline and polygon objects, however, I really need to know what direction my lines are facing, if there could be an option to draw line normals (front and perpendicular) that would make my life so much easier. Right now, I build my lines, compile my game and have to fix up any lines that don't face my desired direction. If I could see this in Tiled I wouldn't have to go through this tedious task.

Thanks.

jamesrehabstudio:
Sorry to resurrect an old post, but this is a feature I would like/need. I'd try adding it myself, but I can't write C++.

bjorn:
@jamesrehabstudio I'll just note that @MegaJiXiang implemented this feature for orthogonal maps in MegaJiXiang/tiled@7c391f4f184cb74f1e38bbb4ab1fba97b62c923e. Eventually this feature will make it into Tiled, but I wouldn't wait for it (though I would of course [support me on Patreon](https://www.patreon.com/bjorn?ty=h) to help me get around to it earlier :-)).

bjorn:
Apparently the MegaJiXiang account was deleted but their patch that adds rendering of normals is still available at https://github.com/mapeditor/tiled/pull/182/changes/7c391f4f184cb74f1e38bbb4ab1fba97b62c923e.

PR #182, opened 2012-03-28
Adds several new features (unique IDs, hierarchical cloning, and more)

ghost:
Hi Björn and others!  I thought I'd have more time to spend on this but I won't.

-unique IDs assigned to every object
-property types (set to link for magical linking ability)
-hierarchical cloning (using link properties)
-drawing of normals on line segments and polygons
-perhaps something else

I won't have time to do these:
-verify that there are no drawing issues (shouldn't be too hard for you guys)
-put in my new optimized version of assigning IDs.  I'll do it for you guys later and it will simply replace what I have now with nothing noticed by the users.
-make anything work with the plugins (why are plugins part of the actual project?)

Hope you guys enjoy the new stuff! :)

### DIFF
--- a/src/libtiled/map.cpp
+++ b/src/libtiled/map.cpp
@@ -208,6 +208,13 @@ Map *Map::clone() const
         o->addLayer(layer->clone());
     o->mTilesets = mTilesets;
     o->setProperties(properties());
+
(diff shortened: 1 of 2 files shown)


bjorn:
Closing this because it's not ever going to be in a state that can be merged. I think many of the features implemented here already found equivalents in Tiled later, though there is still #117 open regarding rendering of polygon normals.

PR #465, opened 2013-07-06
Allow user to save with any file extension in MacOSX.

### LINKED ISSUE #262: Save Tiled maps as any extension
In Tiled (0.8.1), I'm forced to save maps as the extension TMX. I would like to be able to save Tiled maps as any explicit typed extension. I would like it be the same for how someone would use Notepad to save a document; it defaults to the TXT extension (TMX in the case for Tiled) but still allows for anyone to save the document to any extension they so please.

Thanks!

gwdp:
This should resolve #262.

### DIFF
--- a/src/tiled/mainwindow.cpp
+++ b/src/tiled/mainwindow.cpp
@@ -750,6 +750,9 @@ bool MainWindow::saveFileAs()
             }
         }
     }
+    //This will allow user to save the file without an extension OR with any extension.
+    filter += QLatin1String(";;");
+    filter += tr("Any files (*)");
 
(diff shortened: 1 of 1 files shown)


bjorn:
This patch won't work because Tiled would no longer be able to determine in which format to save the file. Anyway I'm not convinced it's a good idea to allow saving with any extension. The standard TextEdit application on macOS doesn't allow it either (the behaviour of automatically adding the extension is part of the native file chooser).

Maybe this patch is also no longer necessary, since TMX files can be saved as .xml since a while as well.

PR #593, opened 2014-01-19
Map Object resizing, v3!

mauvecow:
Depends on pull requests #591 and #592, so you might want to get through them first to make the comparison less unsightly. Replaces pull #485. Does the following:

- Map renderer supports scaling of cells.
- Map reader and object tile creation will initialize the size of the object to the tile's size.
- Removed handles from mapobjectitem.cpp.
- In objectselectiontool.cpp, added a set of 8 ResizeHandles, and changed the Corner enum to AnchorPosition, representing the edge centers as well.
- When a single object is selected, resizing is handled in object space.
- When multiple objects are selected, resizing is handled in world space.

Caveats:
- Negative scaling factors are not allowed.

### DIFF
--- a/src/libtiled/hexagonalrenderer.cpp
+++ b/src/libtiled/hexagonalrenderer.cpp
@@ -234,7 +234,7 @@ void HexagonalRenderer::drawTileLayer(QPainter *painter,
             const Cell &cell = layer->cellAt(rowTile);
 
             if (!cell.isEmpty())
-                renderer.render(cell, rowPos, CellRenderer::BottomLeft);
+                renderer.render(cell, rowPos, QSizeF(0, 0), CellRenderer::BottomLeft);
 
             rowPos.rx() += p.tileWidth;
         }
--- a/src/libtiled/isometricrenderer.cpp
+++ b/src/libtiled/isometricrenderer.cpp
@@ -72,10 +72,13 @@ QRectF IsometricRenderer::boundingRect(const MapObject *object) const
         const Tile *tile = object->cell().tile;
         const QSize imgSize = tile->image().size();
         const QPoint tileOffset = tile->tileset()->tileOffset();
-        return QRectF(bottomCenter.x() + tileOffset.x() - imgSize.width() / 2,
-                      bottomCenter.y() + tileOffset.y() - imgSize.height(),
-                      imgSize.width(),
-                      imgSize.height()).adjusted(-1, -1 - nameHeight, 1, 1);
+        const QSizeF objectSize = object->size();
+        const QSizeF scale(objectSize.width() / imgSize.width(), objectSize.height() / imgSize.height());
+
+        return QRectF(bottomCenter.x() + (tileOffset.x() * scale.width()) - objectSize.width() / 2,
+                      bottomCenter.y() + (tileOffset.y() * scale.height()) - objectSize.height(),
+                      objectSize.width(),
+                      objectSize.height()).adjusted(-1, -1 - nameHeight, 1, 1);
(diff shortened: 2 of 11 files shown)


bjorn:
I've merged master into this branch and pushed the result to `wip/mapobjectresizev3`. Fixing this up and making it ready to merge back to master will be one of my objectives for Tiled 0.12, which I will work on in the coming month.

So far, it looks like you've done an amazing job, @mauvecow! I'm sorry it has taken me so long to finally work towards merging this.

I'll document here the issues I'm aware of that I think should be looked at before merging it into master:

- [ ] Switching map orientation leaves the handles in the wrong place (this wasn't possible back when you worked on this).
- [ ] There's this problem related the name display for non-tile objects.
- [ ] It's too easy to get into the situation where you resize your object down to 0-size, and then you can't make it bigger anymore. I think either it should allow making it bigger starting from 0-size or we should prevent the size from going below 1.
- [ ] The resize behavior looks quite broken for isometric maps.

Either before or after merging to master, but before releasing, I'd prefer to also tackle the following:

- [ ] Add a modifier key for keeping the aspect ratio (`Ctrl` in Inkscape).
- [ ] Add a modifier key for putting the resize origin in the center (`Shift` in Inkscape).
- [ ] The resize handles should look a lot nicer.
- [ ] The mouse cursors are not pointing in the right direction for single rotated objects. I'm thinking if we make the handles arrows themselves we do not actually need these mouse cursors (like in Inkscape).

ntwest:
This is exciting news! I just discovered tiled and have been having great fun making maps for D&D. With object resizing this will be a first-class RPG map editor!

mauvecow:
It's been so long I've actually sort of forgotten everything I'd done here. (the project that uses this is, uh, still underway, just taking forever)

Sounds like you've got a fair degree of it handled. Is there anything you want me to review? It's kind of a bad time for me to hit this up quickly though.

bjorn:
@mauvecow I may have some questions later. It'll still be one week till I really have time to work on this. Of course I understand you will have forgotten most of this patch by now, and so have I only faint memories of earlier review rounds we made. You've had this patch in progress since almost 2 years now. :-/

bjorn:
I've pushed the progress so far to the `wip/mapobjectresizev3`. Not many issues down, but I've got a much better idea of the situation and could fix at least one of the problems with resizing (commit 9ff1514c1932164b3154e0bf6a7b4f2c0a2d9596). Unfortunately, Tiled has grown several inconsitencies related to this over the years which have to be dealt with now.

bjorn:
I pushed some mostly visual improvements in change 333df0eb3c2099eedb1f05890fd4b9ae4f8f8f39. Now it looks like this:

selection_002

bjorn:
It's merged! The isometric projection as well as the varying object origins (top-left for most objects, bottom-left for tile objects, bottom-center for tile objects on isometric maps...) made it all quite challenging, but I'm relatively happy with the resulting code (commit 42b564daf596296a73353855bc3b946a69b0e0f1).

So all known issues with this patch have been resolved, but some more testing will be in order. The Windows daily build including these changes is already available and an OS X one should be coming soon, as well as rebuilds of the other daily packages.

Feature-wise, it's still lacking a mode to keep the aspect ratio, which I think is quite important when working with images.

bjorn:
`Control` modifier added for keeping aspect ration in commit 66db60164714104a2e51eb36d2066ff4588d5e7d. Snap toggle moved to `Alt` during resize.

kheftel:
woohoo!

mauvecow:
Glad to hear it finally made it in. Now to finish the game I was working on that used it...

PR #627, opened 2014-03-24
Adjustable tilelayer horizontal and vertical pixel offset

asermax:
This is an initial implementation of the feature described on the title. The only renderer that currently renders the map with the offseted layers is the isometric one.
The intention was to cover the feature described in #4, but adding the flexibility to use an arbitrary offset instead of only a positive vertical one.

Some caveats:

* For **positive vertical** and **negative horizontal** offsets, the current behavior when drawing is to *push* the rest of the tile layers on the opossite direction. Object layers are independent of the offsets since they use pixel positions, which will cause the objects to "fall out of place" on those cases (they are actually staying in place while everything else moves).
* I tried to adpt the bounding and exposed rects to the offsets in a given map and the current layer, but I'm not completely sure I did a good work, so that may hit on performance (maybe?).

There's no hurry, check the changes whenever you've got some time!

### DIFF
--- a/src/libtiled/isometricrenderer.cpp
+++ b/src/libtiled/isometricrenderer.cpp
@@ -280,13 +280,21 @@ void IsometricRenderer::drawTileLayer(QPainter *painter,
 void IsometricRenderer::drawTileSelection(QPainter *painter,
                                           const QRegion &region,
                                           const QColor &color,
-                                          const QRectF &exposed) const
+                                          const QRectF &exposed,
+                                          const TileLayer* layer) const
 {
+    const QMargins mapOffset = map()->offset();
+    QRect offsettedExposed = exposed.toAlignedRect();
+
+    offsettedExposed.adjust(0, 0,
+                mapOffset.left() + mapOffset.right(),
+                mapOffset.top() + mapOffset.bottom());
+
     painter->setBrush(color);
     painter->setPen(Qt::NoPen);
     foreach (const QRect &r, region.rects()) {
-        QPolygonF polygon = tileRectToScreenPolygon(r);
-        if (QRectF(polygon.boundingRect()).intersects(exposed))
+        QPolygonF polygon = tileRectToScreenPolygon(r, layer);
+        if (QRectF(polygon.boundingRect()).intersects(offsettedExposed))
             painter->drawConvexPolygon(polygon);
     }
 }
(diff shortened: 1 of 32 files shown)


LearnCocos2D:
Any update here?

I tried to do this shifting the layer by a fixed offset in code, but editing this kind of map in Tiled without the layer offset quickly becomes super-confusing. Here's an example using hex tiles:

screen shot 2015-04-28 at 17 19 29

In code it's as simple as shifting the position of each tile layer upwards by +22 points. Even mouse picking works just by removing/adding the layer's offset before/after conversion.

bjorn:
@LearnCocos2D I'd still like to get around to merging this pull request sometime, and of course it would be nice if @asermax would have some more time to clean up the commits. Given the increasing support I receive on Patreon I imagine it'll be something I'll get around to in the coming months.

asermax:
@LearnCocos2D from a practical point of view, you *could* clone this branch, compile it and get a somewhat working offset functionality for isometric maps (if you need something to work with right now and need a temporary solution).

As per the cleaning/refactoring/rewriting of this change, the project for which I started working on this feature is kinda stalled, so this isn't a high priority for me right now.
I still intend to work on it if I get the oportunity, but can't promise anything, sorry :(

LearnCocos2D:
As a matter of fact, I *did* clone it, compile it, and left it sitting there. Good to know it worked (at least on one machine), so that if it becomes super-urgent I’m prepared. ;)

Just not very keen on working myself into a huge C++ codebase and a new IDE right now.

alvinsight:
Hi guys, 
Any updates on this ? 
Now that the patreon funding reached a milestone, would this be PR anywhere on the roadmap ?
Or, is is there anyway to vote for this feature ? 
Thank you !

bjorn:
@alvinsight Thanks for asking. I've added it to the roadmap for Tiled 0.14. Unfortunately there is currently no real way to vote, so I just make adjustments based on feedback.

Chaoseiro:
@bjorn maybe you could use [Trello](https://trello.com/) as a roadmap, it has the 'Vote' feature, you just need to enable anonymous voting, so people without an account can also vote :)

bjorn:
@Chaoseiro Let's give it a try. There's now a [Tiled Roadmap board](https://trello.com/b/yl3PAtN0/tiled-roadmap) on Trello with voting by public members enabled (I don't think anonymous is an option, but I'd rather see who's voting anyway).

@alvinsight Please try to cast your vote there!

I considered whether I should maybe restrict the voting to patrons, but this would be quite a management overhead. Maybe it's better to allow everybody to vote anyway.

bjorn:
Now the votes are totally skewed for this issue, for sure since it's the only issue that links to the roadmap! Anyway, i'll make sure to get to it once I'm done with Tiled 0.13. :-)

bjorn:
In working on this, I ended up with a rather different implementation which I've now merged into `master`. I decided not to touch the map renderer at all, since the offset is independent of the map orientation. Rather, it is applied by the `TileLayerItem` and `ObjectGroupItem` instances (indeed I also extended this offset support to object layers). Then I adjusted all the tools as well as the tile and object selection visualizations to take into account the layer offset.

All the changes are visible by looking at the merge commit 7f1adb4b2d17ce94fdd371f71168a29529a40031.

alvinsight:
Hi Bjorn, thanks for these changes, unfortunately I won't have time to give it a proper test until the end of next month, but it's looking great, looking forward to seeing this in a build !

bjorn:
@alvinsight Ah, that's quite a while away. I think the Tiled 0.14 release will be before that, so hopefully you still find a bit of time earlier or maybe I get some feedback from others in the meantime. :-)

Now that the feature is merged to master, it is already available from the daily builds.

bjorn:
@LearnCocos2D Note that this change does not change the layer rendering order, it just shifts layers by their offset. As such, your hexagonal map screenshot is not so easy to reproduce since the high trees would overlapped by all tiles in higher layers. Fixing that is unfortunately rather non-trivial and would interfere with other features if not done carefully.

Unfortunately it is a somewhat rising trend that with each feature that gets added, several other features get either impossible or just very hard to implement. I guess this is a common problem for complex software, but I do feel like it is becoming time to re-think the whole data structure...

PR #824, opened 2014-11-12
Suggest transparent color for indexed image

thoduv:
I think a lot of people still use indexed colors images for tilesets. The usual behavior for transparency is that the first color of the color table is the transparent. So this just sets the transparent color button in the "New tileset" dialog to the first color of color table (palette) when selecting an indexed colors images.

This is the first time I use git and github, and I wanted to start with a very small suggestion, so I hope this is alright...

Thanks !

### DIFF
--- a/src/tiled/newtilesetdialog.cpp
+++ b/src/tiled/newtilesetdialog.cpp
@@ -191,6 +191,10 @@ void NewTilesetDialog::browse()
 
         if (!mNameWasEdited)
             mUi->name->setText(QFileInfo(f).completeBaseName());
+            
+        QImage img(f);
+        if(!img.hasAlphaChannel() && img.colorTable().count() > 0)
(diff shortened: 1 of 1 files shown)


bjorn:
When I load a GIF file using QImage("test.gif") it anyway ends up having a 32-bit format. I've tested this with both Qt 4 and Qt 5. Fortunately it seems to correctly support GIF transparency already.

When I load an 8-bit PNG, it also already uses the alpha channel when present.

The main problem I see with the patch is that in the case where it is not what the user wants but still he uses 8-bit images with no alpha channel, it will continuously reset the transparent color to the first color in the palette even though that is not necessarily the color he would like to use.

I think a better way to make it easier to pick the transparent color is to allow the user to pick it from the image (like an eye dropper tool). So I'll reject this pull request if you don't mind, and maybe eventually we'll implement this alternative suggestion.

PR #850, opened 2014-12-30
Updated the German translation

Ablu:


bjorn:
@wayfu Please get a free couple of minutes before Sunday, because I'd like to release Tiled 0.11 then if possible. Of course, it's fine too if you are busy, we can skip the review or somebody else could do it too.

wayfu:
I'm here and on it!

Ablu:
I applied fixes to all issues and commented on one for a maybe better resolution.

It would be cool if you could comment in the pullrequest view in the future. This makes it easier to look at comments which were done to old commits/lines. This makes tracking wether all issues were fixed a lot easier (also it sends me the surrounding lines in the mail so i can work directly from my mail client :P).

Thanks a lot for the review so far. I will squash the commits once the last issue is resolved.

wayfu:
Sorry! The github client/site interface is really unintuitive to me :(

Ablu:
Yes it says that since you commented to the commit which is linked to this pull request. But as you can see all the comments are gone now since i changed the commit. So the best is to go to the diff view of the pullrequest (by clicking "Files changed") and comment there.

Those comments will be linked to the line. So updates to the commits do not matter.

Anyway. This should be mergable now. Thanks a lot for your suggestions/review :)

Have a nice weekend,
Ablu

wayfu:
Got it. :) Thanks!

bjorn:
Another downside of commenting on the patches individually is that only the patch author gets notified about it, while for comments to the PR I will also get notifications. So now I have no idea what corrections you have made. Oh well. :-)

Anyway, thanks a lot guys for the updated translation!

PR #852, opened 2015-01-02
Update Czech translation

encukou:
New translations for a new release!


bjorn:
Looking good as far as I can tell, thanks for the update! :-)

Issue #128, opened 2011-12-21
BucketErase / Magic Wand


mrmalee:
Would love to have the ability to bucket erase a group of tiles, or even have a photoshop magic wand style selector which selects all of the same type of tile (contiguous or separated) which then allows you to delete or maybe even swap id's.

bjorn:
Magic Wand has now been implemented. It does not include a mode to "select all same-tiles, regardless of adjacency" and no tile swapping, but I think those requests are covered by other issues already (like #7 and #866).

PR #854, opened 2015-01-04
Implement Magic Wand Tool (#128)

HenryJia:
This implements the magic wand tool of #128


src/tiled/magicwandtool.cpp
```diff
@@ -0,0 +1,244 @@
...
+{
+    MapDocument *document = mapDocument();
+
+    if (mSelectedRegion != document->selectedArea()) {
+        QUndoCommand *cmd = new ChangeSelectedArea(document, mSelectedRegion);
+        document->undoStack()->push(cmd);
+    }
+
+    brushItem()->setTileRegion(mSelectedRegion);
+    document->setSelectedArea(mSelectedRegion);
```
bjorn: This line seems wrong. The selected area is already applied to the document by the `ChangeSelectedArea` undo command created and pushed on the undo stack above.

Also, this seems like the wrong moment to set the tile region of the brush item. Shouldn't that be done in the `tilePositionChanged` function to provide a preview of the area that's going to be selected?

src/tiled/magicwandtool.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+    if (!mStamp)
+        return;
+
+    for (int x = 0; x < mStamp->width(); x++)
+        for (int y = 0; y < mStamp->height(); y++)
+            if (!mStamp->cellAt(x, y).isEmpty())
+                mRandomList.append(mStamp->cellAt(x, y));
+}
+*/
```
bjorn: Please remove all this commented out stuff. It's irrelevant to the magic wand tool.

bjorn:
To fix the AppVeyor continue-integration build you should add `magicwandtool.cpp` and `magicwandtool.h` to the files list in `src/tiled/tiled.qbs`.

HenryJia:
OK, Done.

src/tiled/magicwandtool.h
```diff
@@ -0,0 +1,77 @@
...
+
+    //void updateStatusInfo();
+
+private:
+    enum SelectionMode {
+        Replace,
+        Add,
+        Subtract,
+        Intersect
+    };
```
bjorn: You have these selection modes here but you're not using them. If the tool is working so far, why not try and implement that functionality like in the `TileSelectionTool`?

HenryJia:
OK

HenryJia:
I'm not fully sure how the selection modes work in the actual program (as in how you change it)

bjorn:
> I'm not fully sure how the selection modes work in the actual program (as in how you change it)

Please see if you can just derive this by reading the code in `TileSelectionTool`.

HenryJia:
Oh I see, through modifiers on the keyboard.

HenryJia:
I've added the keyboard modifiers. However the shift (add) modifiers doesn't work because computefillRegion only computes it within the selected region.

bjorn:
> I've added the keyboard modifiers. However the shift (add) modifiers doesn't work because computefillRegion only computes it within the selected region.

Alright, that will require some changes to `computeFillRegion` then. Essentially, it needs to be optional whether it takes the selection into account because it does still need to do that for the fill tool. I will try to push a change for this sometime soon.

HenryJia:
I forgot to say, the magic wand tool will also need a new icon for it. Currently I've just got the icon set to the same one as bucket fill.

bjorn:
> I forgot to say, the magic wand tool will also need a new icon for it. Currently I've just got the icon set to the same one as bucket fill.

Fortunately, the GIMP also has a magic wand tool so we can just use the same icon. Most of the other icons are also taken from the GIMP.

HenryJia:
Hmm, I don't know where to get the GIMP icons.

toddcarnes:
You should be able to find the source at gimp.org

HenryJia:
I've added the magic wand icon.

What shortcut key should I use or this?

src/tiled/magicwandtool.h
```diff
@@ -0,0 +1,66 @@
...
+
+    void mousePressed(QGraphicsSceneMouseEvent *event);
+    void mouseReleased(QGraphicsSceneMouseEvent *event);
+
+    void languageChanged();
+
+protected:
+    void tilePositionChanged(const QPoint &tilePos);
+
+    //void updateStatusInfo();
```
bjorn: Please remove this commented line.

src/tiled/magicwandtool.cpp
```diff
@@ -0,0 +1,95 @@
...
+    MapDocument *document = mapDocument();
+
+    QRegion selection = document->selectedArea();
+
+    if(modifiers == Qt::ShiftModifier)
+        selection += mSelectedRegion;
+    else if(modifiers == Qt::ControlModifier)
+        selection -= mSelectedRegion;
+    else if(modifiers == (Qt::ControlModifier | Qt::ShiftModifier))
+        selection &= mSelectedRegion;
```
bjorn: Please always put a space between the `if` and the `(`. It just happens to be my code style.

HenryJia:
Done

bjorn:
Thanks for your work on this tool! I've squashed your commits and pushed it as 57c27c1cf2107713ee9ead2c8504f485ffe7a3bb. I've then fixed the `computefillRegion` problem in change 4a6322f134e4d31923a46d259e5e80e8814d8231.

PR #858, opened 2015-01-10
French translation update for the next Tiled version! :D

Bertram25:
Hexagonal maps FTW!

Have a nice Week-End and a nice release. :)


bjorn:
Thanks Bertram!

Unfortunately something unexpected came up this weekend so I will have a minimal amount of spare time only on Sunday, so it remains to be seen whether I can still do a release. I will do my best at least!

Bertram25:
Np. Take the time needed :)

PR #859, opened 2015-01-10
Updated Japanese translation file.

toddcarnes:


.gitignore
```diff
@@ -11,6 +11,7 @@ ui_*.h
 *.pro.user*
 qrc_*
 docs/html
+.directory
```
bjorn: Why?

bjorn:
@teraapi Could you review this update, please?

.gitignore
toddcarnes: Because it was pulling in the .directory for on my system.

bjorn: What is the .directory?

toddcarnes: A normally invisible file that the file manager on kubuntu used to track
files in the directory.

bjorn: Ok, then please set this as a global ignore, for example by putting it in ~/.gitignore and doing:

    git config --global core.excludesfile ~/.gitignore

toddcarnes: I will do that. Thank you for the information. I wasn't aware I could do
that. :)

bjorn: You're welcome. :-)

teraapi:
Looks good to me. :))

bjorn:
Alright, thanks @teraapi for the review and @toddcarnes for the update. I've applied the translation in commit 2de09c5aa6d2067efa9bf37fdffda5dd128ddae6 (dropping the change to .gitignore).

PR #860, opened 2015-01-11
Added a Tile reference for MapObjects

Juniperbrew:
TMXMapReader looks for the "gid" attribute when reading map objects. If it finds one it gives the MapObject a reference to the relevant Tile. If the gid value can't be parsed to an integer it will crash.


bjorn:
Looks fine, thanks!

Issue #453, opened 2013-06-09
Map -> Add External Tileset doesn't remember last used folder


leiradel:
Every time an external tileset is added, the file open dialog starts in the same directory where the last map was opened.

I think it would be nice if that dialog started in the same directory where the last external tileset was added from.

bjorn:
@HenryJia maybe you'd like to tackle this one.

My idea would be to store a separate directory for external tilesets in the Preferences. The preferences already has a mechanism for remember the last used directory for different purposes, so you can extend the `Preferences::FileTypes` enum with an `ExternalTileset` entry and use that similar to how the other entries are used when external tilesets are being added.

PR #861, opened 2015-01-11
Adds a separate directory for adding external tilesets and exporting tilesets #453 

HenryJia:


src/tiled/tilesetdock.cpp
```diff
@@ -758,20 +758,20 @@ void TilesetDock::exportTileset()
     if (!tileset)
         return;
 
-    const QLatin1String extension(".tsx");
-    QString suggestedFileName = QFileInfo(mMapDocument->fileName()).path();
-    suggestedFileName += QLatin1Char('/');
-    suggestedFileName += tileset->name();
-    if (!suggestedFileName.endsWith(extension))
-        suggestedFileName.append(extension);
```
bjorn: Please don't just rip out the suggested file name logic. You'll need to combine the stored path with the suggested file name.

src/tiled/mainwindow.cpp
```diff
@@ -1187,14 +1187,19 @@ void MainWindow::addExternalTileset()
     if (!mMapDocument)
         return;
 
-    const QString start = fileDialogStartLocation();
+    Preferences *prefs = Preferences::instance();
+
+    const QString start = prefs->lastPath(Preferences::ExternalTileset);
```
bjorn: Please still use `fileDialogStartLocation()` if the last path is empty (the first time people use it).

bjorn:
Apart from my inline comment, please also remove the merge commit from the pull request.

HenryJia:
All done.

bjorn:
Why did you close this, @HenryJia?

HenryJia:
Oops I think I must've accidentally deleted it whilst removing some branches that I created that weren't any use. I'll re open it

HenryJia:
I've reopened it and restored the branch

src/tiled/mainwindow.cpp
```diff
@@ -1187,14 +1187,23 @@ void MainWindow::addExternalTileset()
     if (!mMapDocument)
         return;
 
-    const QString start = fileDialogStartLocation();
+    Preferences *prefs = Preferences::instance();
+
+    QString start = prefs->lastPath(Preferences::ExternalTileset);
+
+    if(start.isEmpty())
```
bjorn: Please put a space between the `if` and `(`.

src/tiled/preferences.cpp
```diff
@@ -340,6 +340,9 @@ static QString lastPathKey(Preferences::FileType fileType)
     case Preferences::ExportedFile:
         key.append(QLatin1String("ExportedFile"));
         break;
+    case Preferences::ExternalTileset:
+        key.append(QLatin1String("External Tileset"));
```
bjorn: Please remove the space from the settings key.

src/tiled/tilesetdock.cpp
```diff
@@ -758,20 +758,31 @@ void TilesetDock::exportTileset()
...
+    QString suggestedFileName = prefs->lastPath(Preferences::ExternalTileset);
+
+    for(int i = (suggestedFileName.length() - 1); suggestedFileName.at(i) != QLatin1Char('/')
+        && suggestedFileName != empty; i--)
+        suggestedFileName.remove(i, 1);
+
+    if(suggestedFileName == empty)
+        suggestedFileName += QLatin1Char('/');
     suggestedFileName += tileset->name();
-    if (!suggestedFileName.endsWith(extension))
```
bjorn: Why did you remove this condition? I think it should stay to avoid adding the extension twice when it is part of the tileset name.

src/tiled/tilesetdock.cpp
```diff
@@ -758,20 +758,31 @@ void TilesetDock::exportTileset()
...
-    suggestedFileName += QLatin1Char('/');
+    QLatin1String empty(".");
+    QString suggestedFileName = prefs->lastPath(Preferences::ExternalTileset);
+
+    for(int i = (suggestedFileName.length() - 1); suggestedFileName.at(i) != QLatin1Char('/')
+        && suggestedFileName != empty; i--)
+        suggestedFileName.remove(i, 1);
+
+    if(suggestedFileName == empty)
+        suggestedFileName += QLatin1Char('/');
```
bjorn: What does this do? If the suggestedFileName equals ".", it will add a slash? And what happened to the code that would suggest using the path of the map file? I think that should stay in case no path was saved yet for `ExternalTileset`.

src/tiled/tilesetdock.cpp
```diff
@@ -758,20 +758,31 @@ void TilesetDock::exportTileset()
...
+
     const QLatin1String extension(".tsx");
-    QString suggestedFileName = QFileInfo(mMapDocument->fileName()).path();
-    suggestedFileName += QLatin1Char('/');
+    QLatin1String empty(".");
+    QString suggestedFileName = prefs->lastPath(Preferences::ExternalTileset);
+
+    for(int i = (suggestedFileName.length() - 1); suggestedFileName.at(i) != QLatin1Char('/')
+        && suggestedFileName != empty; i--)
+        suggestedFileName.remove(i, 1);
```
bjorn: Please just keep the original approach that was using `QFileInfo::path`, because this is a little strange.

HenryJia: The for loop is required because when you suggest a path using the lastPath it will also include the file name of the last path inside the string. The foor loop loops from the end until it reaches a / to remove that and then add the Tileset's name to it.

What do you mean by that? How do you want me to change it? I don't really see an easier way to do it.

bjorn: I would prefer if you changed it back to the way it was done before:
```
suggestedFileName = QFileInfo(suggestedFileName).path();
```

src/tiled/tilesetdock.cpp
```diff
@@ -758,20 +758,33 @@ void TilesetDock::exportTileset()
...
+
+    if(suggestedFileName == empty)
+        QString suggestedFileName = QFileInfo(mMapDocument->fileName()).path();
+        suggestedFileName += QLatin1Char('/');
```
bjorn: The `if` condition here seems broken. Why do you expect the `suggestedFileName` to be "."? I think you want to put this check above the "extracting of the path" part, and just use `suggestedFileName.isEmpty()`.

bjorn:
I've fixed up the commit a little and pushed it as 38932930c4c7bfb6ae0662e0b5c553632b5eedbd.

I'm sorry about some of my earlier comments, since they have not been entirely correct. In hindsight, it was not necessary to fall back to the current map file path since `Preferences::lastPath` was already doing this.

Thanks for your contribution!

HenryJia:
Your welcome!
Thanks for letting me work on this too, I've learnt quite a lot from working on this project.

PR #865, opened 2015-01-18
Some change to support custom data read/write in origin tmx file.

czsilence:
Hi, here are some change to support custom data read/write in origin tmx file. detail:
1. a CustomXmlElementWriteCallback member in MapWriter, and a CustomXmlElementReaderCallback member in MapReader, to read/write custom data;
2. a new interface class CustomDataManager in Map, to store custom data read in CustomXmlElementReadCallback and write back by CustomXmlElementWriteCallback.
3. plugin developer can implement CustomXmlElementWriteCallback/CustomXmlElementReaderCallback/CustomDataManager in the plugin project.
4. I build tiled in win32 with qmake -spec win32-msvc2012, and found the temp file not ignore, so I add to .gitignore.

example:
1. define a custom node to ref to custom resource

2. add object with custom properties

3. finally show like this, with a custom plugin

I'm a game developer in china, and this is my first pull request in github, hope it's useful for others.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,16 @@ tests/test_mapreader
 
 # Ronn html outputs
 docs/*.1.html
+
+#qmake some
+.qmake.stash
+.localized
+
+#visual studio
+debug
+release
+*.ilk
+*.pdb
+*_manifest.rc
+*_manifest.res
+*.embed.manifest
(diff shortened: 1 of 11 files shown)


czsilence:
@bjorn , thanks for reply and advice, I think you are right, the patch will cause more complex for TMX file format and plugin. I will update it if I have better idea. Thanks again.

PR #866, opened 2015-01-19
Added functionality to swap tiles

theHacker:
During development of my game I am frequently changing the graphics file of the tileset. It is pretty anonying that I destroy the map when moving around the tiles on the graphic. First I wrote a script to switch two ids in the map.

Because it is not very fun to look up ids which to switch, I decided that it would be nice if I can do this in Tiled directly. I found no such function... so I wrote one myself :-)
If exactly two tiles are selected you can selected "Swap tiles" in the contextmenu to swap these tiles on the current TileLayer.


src/tiled/tilesetview.cpp
```diff
@@ -620,6 +630,24 @@ void TilesetView::editTileProperties()
...
+    if (selectionModel()->selectedIndexes().size() != 2)
+        return;
+
+    TileLayer *tileLayer = dynamic_cast<TileLayer*>(mMapDocument->currentLayer());
+    if (!tileLayer)
+        return;
+
+    const TilesetModel *model = tilesetModel();
+    Tile *tile1 = model->tileAt(selectionModel()->selectedIndexes()[0]);
+    Tile *tile2 = model->tileAt(selectionModel()->selectedIndexes()[1]);
```
bjorn: You're retrieving the list of selected indexes three times in this function. Please put it in a local variable.

src/tiled/swaptiles.cpp
```diff
@@ -0,0 +1,64 @@
...
+                swapCell.tile = mTile2;
+                mTileLayer->setCell(x, y, swapCell);
+            }
+            else if (cell.tile == mTile2) {
+                Cell swapCell = cell;
+                swapCell.tile = mTile1;
+                mTileLayer->setCell(x, y, swapCell);
+            }
+        }
+    }
```
bjorn: Hmm, this fails to call `MapDocument::emitRegionChanged` to trigger a repaint. I don't think it makes sense to determine this region in detail, but just emit it for the region of the layer (`Layer::bounds`).

src/tiled/tilesetview.cpp
```diff
@@ -582,6 +583,15 @@ void TilesetView::contextMenuEvent(QContextMenuEvent *event)
             Utils::setThemeIcon(tileProperties, "document-properties");
             connect(tileProperties, SIGNAL(triggered()),
                     SLOT(editTileProperties()));
+
+            // Enable "swap" if there are exactly 2 tiles selected
+            bool exactlyTwoTilesSelected =
+                    (selectionModel()->selectedIndexes().size() == 2);
+
+            QAction *tileSwap = menu.addAction(tr("Swap tiles"));
```
bjorn: Please name the menu item "Swap Tiles" rather than "Swap tiles".

bjorn:
Overall a nice and clean patch! One question though, why did you decide to only swap the tiles for the current layer? This is going to be painful for people who want to swap the tiles globally and have many layers, which seems to be part of your main use-case.

Also, I guess it would make sense to support this on object layers for tile objects as well.

theHacker:
Thx :smile:

To answer your question why only for tile layers and the current layer... I never worked with Qt before and it was my first time reading Tiled's source code as well, so my goal was to change so few as possible to avoid breaking something. Because I only have exactly one layer (a tile layer) I stuck with `currentLayer()`.

When doing this on all layers you would imagine that the tiles would be swapped on all layers - object layers and tile layers all at once?

bjorn:
> When doing this on all layers you would imagine that the tiles would be swapped on all layers - object layers and tile layers all at once?

Sorry for the late answer, but yes that's what I would expect to happen.

In general, I'm a little concerned that this action will be mistaken for a way to rearrange the tileset itself, rather than modifying the map. But I'll look into giving this a try and merging it sometime soon.

bjorn:
Rebased on current master and pushed as 2c6cde6ae1621b7286960178c2e024b3addd3c79. The behavior is still the same (works on current tile layer only), but I'll change it to apply to the whole map.

bjorn:
I made the swap affect the entire map in change b52338ca20158d51d600314d5d1a139cdb93e441.

tangorri:
Hi, interested by TileSet rearrangement, can't find any way to use these on interface

bjorn:
@tangorri Currently you can't modify the tileset image in Tiled. The change that was merged here is meant to help users who modified their tileset image externally, to adjust the maps they have already created.

Issue #277, opened 2012-09-13
Add CSV export plugin


Asmageddon:
Would really help with making flixel maps.

PR #869, opened 2015-01-20
Extend CSV plugin to export all tile layers.

SoyYuma:
- Layers get exported in different files.
- Filenames are constructed as
    #base_#layer.csv
  where #base is the name chosen in the
  save-file dialog (without extension), and
  #layer is the name for the layer.

Related to #277

### DIFF
--- a/src/plugins/csv/csvplugin.cpp
+++ b/src/plugins/csv/csvplugin.cpp
@@ -24,7 +24,9 @@
 #include "tile.h"
 #include "tilelayer.h"
 
+#include <QDir>
 #include <QFile>
+#include <QFileInfo>
 
 #if QT_VERSION >= 0x050100
 #define HAS_QSAVEFILE_SUPPORT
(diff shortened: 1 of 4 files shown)


SoyYuma:
I just made this small extension to the CSV plugin. Now it can export all tile-based layers into separate files.

It's not the most elegant code I've ever seen but I think it's quite useful for us Flixel guys ;-)

src/plugins/csv/csvplugin.cpp
```diff
@@ -43,62 +45,64 @@ CsvPlugin::CsvPlugin()
 
 bool CsvPlugin::write(const Map *map, const QString &fileName)
 {
+    // Extract file name without extension and path
+    QFileInfo fileInfo(fileName);
+    fileInfo.setCaching(false);
```
bjorn: Why would you worry about the caching?

src/plugins/csv/csvplugin.cpp
```diff
@@ -43,62 +45,64 @@ CsvPlugin::CsvPlugin()
...
+    const QString fileNameWoExtension = fileInfo.baseName();
```
bjorn: I'd replace `baseName` with `completeBaseName` to avoid cutting off more than just the `.csv` part in the case of multiple dots.

src/plugins/csv/csvplugin.cpp
```diff
@@ -43,62 +45,64 @@ CsvPlugin::CsvPlugin()
...
+        const QString layerName = tileLayer->name();
+        const QString layerFileName = fileNameWoExtension + QString("_") + layerName + QString(".csv");
+        const QString layerFileNameWPath = QDir(filePath).filePath(layerFileName);
+            
 #ifdef HAS_QSAVEFILE_SUPPORT
-    QSaveFile file(fileName);
+        QSaveFile file(layerFileNameWPath);
 #else
-    QFile file(fileName);
+        QFile file(layerFileName);
```
bjorn: This should also use `layerFileNameWPath`. Please just name that `layerFilePath`, btw.

src/plugins/csv/csvplugin.cpp
```diff
@@ -43,62 +45,64 @@ CsvPlugin::CsvPlugin()
 
 bool CsvPlugin::write(const Map *map, const QString &fileName)
 {
+    // Extract file name without extension and path
+    QFileInfo fileInfo(fileName);
+    fileInfo.setCaching(false);
+    const QString fileNameWoExtension = fileInfo.baseName();
+    const QString filePath = fileInfo.path();
```
bjorn: To avoid confusion with actual file paths, I think this should just be called `path`.

bjorn:
I think it's nice functionality, but what about the check on the file already being there? Preferably I'd still warn the user as usual the first time he tries to export and it would overwrite existing files, but I don't really see a good way of doing that.

Also, maybe it would make sense to keep the previous behavior in the case of a single tile layer? Otherwise people who have been using this plugin will suddenly have the layer name be appended to the file they're trying to export. On the other hand, that'd be inconsistent...

src/plugins/csv/csvplugin.cpp
SoyYuma: Ehm... I got confused with Qt docs. I wanted to prevent QFileInfo from accessing the actual filesystem (I just want QFileInfo to strip out parts of the path). I'll take line 50 out.

src/plugins/csv/csvplugin.cpp
SoyYuma: Makes sense. I'll change it.

src/plugins/csv/csvplugin.cpp
SoyYuma: Oops, I forgot to read below the #else. I'll change it too.

src/plugins/csv/csvplugin.cpp
SoyYuma: Cool, I'll change it.

SoyYuma:
For my use case (just finished making a small game), I would have hated having to press enter to confirm I wanted to overwrite files. It's so quick to just go to File->Export and done. It's so quick as it is now. I'd ask you to reconsider to silently overwrite files.

If you keep wanting to warn users when overwriting files, we could do this:
  1) get all the file paths that we would use
  2) check if any of those already exist
  3) if some exist: open yes-no dialogue with the warning about overwriting files
  4) if none exist (or user selects yes in the previous dialogue): execute the export

The logic of the plugin would be a bit more complex, but I'll happily do it if you prefer it this way.

In the case of a single tile layer, it makes sense to keep the same name as the TMX file (without the layer name) so, as you say, we don't make previous users sad. But it's also ugly (mind my OCD) to have a special name convention for single-layer maps. I think in the end it's a matter of preferences. I'll gladly change it to whatever you prefer :-)

Ablu:
How about doing it like gimp does it? It asks whether you want to override
the first time and then does it silently the next time.

Regards, Ablu

bjorn:
@Ablu That's how it works since Tiled 0.11 actually. I think @SoyYuma didn't realize this either.

The overwrite warning happens only when using "Export" the first time, or when using "Export As...". When you trigger "Export" a second time it silently does whatever it was told to do before.

@SoyYuma Your approach seems fine to me, so it would be nice if you could implement that. One tricky bit is that it can't be done just by the plugin because it does not know whether it should silently overwrite files. I also think it would be good to implement this outside of the plugin so that the logic can be used with other plugins as well. So maybe add a new virtual function to the MapWriterInterface that takes the file name and the map, and just returns the files is plans to write?

src/plugins/csv/csvplugin.cpp
bjorn: Alright. Actually, I think neither of the operations you're doing will cause it to access the file system anyway, since they just operate on the path.

Ablu:
Ah. sorry then :)

SoyYuma:
@bjorn You're right, I wasn't aware it was silently overwriting the second time when using "Export".
I'll have a look and will make the changes you propose. Can't do anything until next week, though.

Thanks for you time, guys!

bjorn:
I'm sorry for the silence on this PR for so long. Thanks for all the fixes, I think you took care of all my concerns and I only see some small code style issues that I'd rather just correct myself than nitpick about it (unless you want me to nitpick). I'll be looking into merging this in the coming evenings.

SoyYuma:
No worries! I don't mind correcting the style issues. Although some times it's easier to make the changes directly than to point out the style issues :-)

bjorn:
Alright, merged and I pushed some adjustments in change 39c8a6aab3ab48a9217610ec06cc001eab4592e1. Most significantly I decided to keep the old behavior for maps with just a single tile layer.

Thanks for the contribution!

SoyYuma:
Awesome! Thanks for integrating this change :-)

PR #879, opened 2015-01-29
Added an auto tile to layer function.

tylerellis21:
Added an auto tile to layer function.
Say you have two tile sets defined in the editor named (the name in the editor):
Grassland.ground
Grassland.object
And two layers named:
Object
Ground
What this will do is is look at the tile set in which the tile comes from and try to place it in the correct layer based on the extension in the name (not the image actual file). So say I have the Object layer selected but I am using a tile from the Grassland.ground what this would do is place the tile into the ground layer instead since it would look at the .ground and try to find the layer it needs to go to. It's not case sensitive and it currently used the '.' character to split the name. If the layer is not found it will default to the currently selected layer.

It could probably be named better but AutoLayer is what I thought of.

### DIFF
--- a/src/tiled/preferences.cpp
+++ b/src/tiled/preferences.cpp
@@ -99,6 +99,10 @@ Preferences::Preferences()
     mAutoMapDrawing = boolValue("WhileDrawing");
     mSettings->endGroup();
 
+    mSettings->beginGroup(QLatin1String("AutoLayer"));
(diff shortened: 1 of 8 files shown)


tylerellis21:
Any feedback is appreciated.

tylerellis21:
Going to implement this into the properties of the tile instead.

PR #890, opened 2015-02-18
Automapping Tile Objects

Seanba:
Tile Objects can be added to Object Layers through the automapping
mechanism.
Also, fixed bug in "DeleteTiles" functionality with respect to Objects
(tiled or otherwise).

### DIFF
--- a/src/tiled/automapper.cpp
+++ b/src/tiled/automapper.cpp
@@ -469,10 +469,22 @@ void AutoMapper::autoMap(QRegion *where)
                 TileLayer *dstTileLayer = dstLayer->asTileLayer();
                 if (dstTileLayer)
                     dstTileLayer->erase(region);
-                else
+                else {
+                    // Region has to be put into object space
+                    const int w = mMapDocument->map()->tileWidth();
+                    const int h = mMapDocument->map()->tileHeight();
+                    const QRect& tilesRect = region.boundingRect();
+                    QRect objRect;
+                    objRect.setTop(tilesRect.top() * h);
+                    objRect.setLeft(tilesRect.left() * w);
+                    objRect.setBottom((tilesRect.bottom() + 1) * h);
+                    objRect.setRight((tilesRect.right() + 1) * w);
+                    QRegion objRegion;
+                    objRegion += objRect;
                     eraseRegionObjectGroup(mMapDocument,
                                            dstLayer->asObjectGroup(),
-                                           region);
+                                           objRegion);
+                }
             }
         }
     }
--- a/src/tiled/automapper.cpp
+++ b/src/tiled/automapper.cpp
@@ -469,22 +469,10 @@ void AutoMapper::autoMap(QRegion *where)
                 TileLayer *dstTileLayer = dstLayer->asTileLayer();
                 if (dstTileLayer)
                     dstTileLayer->erase(region);
-                else {
-                    // Region has to be put into object space
-                    const int w = mMapDocument->map()->tileWidth();
-                    const int h = mMapDocument->map()->tileHeight();
-                    const QRect& tilesRect = region.boundingRect();
-                    QRect objRect;
-                    objRect.setTop(tilesRect.top() * h);
-                    objRect.setLeft(tilesRect.left() * w);
-                    objRect.setBottom((tilesRect.bottom() + 1) * h);
-                    objRect.setRight((tilesRect.right() + 1) * w);
-                    QRegion objRegion;
-                    objRegion += objRect;
+                else
                     eraseRegionObjectGroup(mMapDocument,
                                            dstLayer->asObjectGroup(),
-                                           objRegion);
-                }
+                                           region);
             }
         }
     }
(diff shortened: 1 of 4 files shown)


bjorn:
@stefanbeller Do you have time to review this?

src/tiled/automappingutils.cpp
```diff
@@ -42,7 +43,10 @@ void eraseRegionObjectGroup(MapDocument *mapDocument,
         // tile objects. polygons and polylines are not covered correctly by this
         // erase method (we are in fact deleting too many objects)
         // TODO2: toAlignedRect may even break rects.
-        if (where.intersects(obj->bounds().toAlignedRect()))
+		
+		// Note: intersects returns false in some conditions so check if corners are contained 
+        const QRect& objRect = obj->bounds().toAlignedRect();
+        if (where.intersects(objRect) || where.contains(objRect.topLeft()) || where.contains(objRect.bottomLeft()))
```
bjorn: Hmm, `intersects` should only return false if the intersection is empty. This happens also when either rectangle is empty, even though one is contained in the other. So, it should be enough to check only one corner.

src/tiled/automappingutils.cpp
```diff
@@ -73,7 +77,8 @@ const QList<MapObject*> objectsInRegion(ObjectGroup *layer,
         // QRegion::intersects() returns false for empty regions even if they are
         // contained within the region, so we also check for containment of the
         // top left to include the case of zero size objects.
-        if (where.intersects(rect) || where.contains(rect.topLeft()))
+        // (Also, for Tile objects we need to check the bottom left)
+        if (where.intersects(rect) || where.contains(rect.topLeft()) || where.contains(rect.bottomLeft()))
```
bjorn: It should not matter whether you check the top left or bottom left for tile objects, since their size is 0.

src/tiled/automapper.cpp
```diff
@@ -469,10 +469,22 @@ void AutoMapper::autoMap(QRegion *where)
...
+                    const int w = mMapDocument->map()->tileWidth();
+                    const int h = mMapDocument->map()->tileHeight();
+                    const QRect& tilesRect = region.boundingRect();
+                    QRect objRect;
+                    objRect.setTop(tilesRect.top() * h);
+                    objRect.setLeft(tilesRect.left() * w);
+                    objRect.setBottom((tilesRect.bottom() + 1) * h);
+                    objRect.setRight((tilesRect.right() + 1) * w);
+                    QRegion objRegion;
+                    objRegion += objRect;
```
bjorn: Hmm, this seems not exactly correct in two ways:

* It should use the `MapRenderer` to convert the region into object space, though I'm not sure if the current API suffices for this. Maybe it could convert each corner of each rectangle in the region and include the bounding rect of the result into the new region.

* It's converting a bounding rect of the entire region rather than converting each rectangle separately, causing a potentially larger area to be erased than was intended?

Seanba: I'm using your MapRenderer suggestion but also moving that functionality into the eraseRegionObjectGroup call. Testing looks good. Should have another pull request soon.

bjorn: Alright.

Btw, please don't open a new pull request. Rather just update this one (usually done by force-pushing).

src/tiled/automappingutils.cpp
Seanba: The problem here was that tile objects use the bottom left corner as their origin but I still think there are issues with this approach. Tile objects actually get bounding rectangles with a negative width and height of 1 which I think is complicating things. My next pull request will address this.

Seanba:
I am putting together another pull request. Thanks for the feedback.

Seanba:
I don't know what I'm doing with Github. :/
Re-opening this pull request with modified files.

src/tiled/automapper.cpp
Seanba: Yes, I figured out deleting the pull request was the wrong thing to do. I've re-opened this one and the modified files look correct. Thanks.

Seanba:
Not sure what's with that build failure. Is it because I had briefly closed the pull request?

bjorn:
> Not sure what's with that build failure. Is it because I had briefly closed the pull request?

That's what the AppVeyor page indicates, yes. Nothing to worry about.

src/tiled/automappingutils.cpp
```diff
@@ -42,7 +43,22 @@ void eraseRegionObjectGroup(MapDocument *mapDocument,
         // tile objects. polygons and polylines are not covered correctly by this
```
stefanbeller: Does the comment here need updating?

bjorn: Hmm, it does not look like it to me, because it already says the correct thing is done for tile objects. The patch seems to make the comment more correct than before.

bjorn:
I've pushed this fix as commit 521a40a5dac6552c943f8639f761a8b1e7535b82, thanks!

Seanba:
Awesome. Thanks, Bjorn.

PR #893, opened 2015-02-25
Update tiled_ru.ts

semtiko:
Sorry for very long delay. I hope it's done.


bjorn:
Better late than never, thanks for the update! :-)

PR #906, opened 2015-03-19
Fix Japanese translation

mogemimi:
Hello,
this fixes some grammatical errors and adjustments to make the Japanese translation seem more natural.


bjorn:
@mogemimi Thanks for making these improvements!

@teraapi Since I can't judge these fixes, what do you say about them as original translator?

PR #916, opened 2015-04-12
Added commandline configuration to terraingenerator

atheros:
```
Usage: terraingenerator [options]

Options:
  -h --help        : Display this help.
  -v --version     : Display the version.
     --overwrite   : Target is overwritten rather than extended.
  -e --embed-image : Tile images will be embedded in the TSX file instead
                     of being saved as a separated PNG file.
  -c --combine T1[ T2 [Tn ...]]
                   : Specify the terrains to combine together (all combinations).
  -s --source TS1[ TS2 [TSn ...]]
                   : Add source tilesets, order is not important.
  -o --output OUT  : Specify output tileset filename.
  -p --priority T1[ T2 [Tn ...]]
                   : Add terrain names to priority list (T1 < T2 < Tn).
```

Example usage:
```
terraingenerator -o test.tsx --priority LavaRock Dirt2 Dirt Sand Water Hole Grass PlowedSoil Lava --source LavaRock.tsx Dirt2.tsx Dirt.tsx Water.tsx Hole.tsx Grass.tsx PlowedSoil.tsx Lava.tsx --overwrite --combine Dirt Dirt2 Water Grass --combine Dirt2 Grass LavaRock --combine Lava LavaRock --combine PlowedSoil Grass --combine Dirt Grass Hole
```

### DIFF
--- a/src/terraingenerator/main.cpp
+++ b/src/terraingenerator/main.cpp
@@ -1,6 +1,7 @@
 /*
  * main.cpp
  * Copyright 2012, Thorbjørn Lindeijer <thorbjorn@lindeijer.nl>
+ * Copyright 2015, Przemysław Grzywacz <nexather@gmail.com>
  *
  * This file is part of the Terrain Generator tool.
  *
@@ -45,13 +46,17 @@ struct CommandLineOptions {
         : showHelp(false)
         , showVersion(false)
         , overwrite(false)
+        , embedImage(false)
     {}
 
     bool showHelp;
     bool showVersion;
     bool overwrite;
+    bool embedImage;
     QString target;
     QStringList sources;
+    QStringList terrainPriority;
+    QList<QStringList> combineList;
 };
 
 } // anonymous namespace
@@ -60,11 +65,21 @@ static void showHelp()
 {
     // TODO: Make translatable
     qWarning() <<
-            "Usage: terraingenerator [options] [target] [sources...]\n\n"
+            "Usage: terraingenerator [options]\n\n"
             "Options:\n"
-            "  -h --help      : Display this help\n"
-            "  -v --version   : Display the version\n"
-            "     --overwrite : Target is overwritten rather than extended";
+            "  -h --help        : Display this help.\n"
+            "  -v --version     : Display the version.\n"
+            "     --overwrite   : Target is overwritten rather than extended.\n"
+            "  -e --embed-image : Tile images will be embedded in the TSX file instead\n"
+            "                     of being saved as a separated PNG file.\n"
+            "  -c --combine T1[ T2 [Tn ...]]\n"
+            "                   : Specify the terrains to combine together (all combinations).\n"
+            "  -s --source TS1[ TS2 [TSn ...]]\n"
+            "                   : Add source tilesets, order is not important.\n"
+            "  -o --output OUT  : Specify output tileset filename.\n"
+            "  -p --priority T1[ T2 [Tn ...]]\n"
+            "                   : Add terrain names to priority list (T1 < T2 < Tn).\n"
+    ;
 }
 
 static void showVersion()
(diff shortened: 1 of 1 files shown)


src/terraingenerator/main.cpp
```diff
@@ -45,13 +46,18 @@ struct CommandLineOptions {
         : showHelp(false)
         , showVersion(false)
         , overwrite(false)
+        , embedImage(false)
+
```
bjorn: Please no empty line here. :-)

src/terraingenerator/main.cpp
```diff
@@ -73,28 +89,132 @@ static void showVersion()
...
+                inSource = false;
+            } else {
+                // Append the source file.
+                argCount++;
+                options.sources.append(arg);
+                //qDebug() << "adding source" << arg;
+            }
+
+            // If we should still process input list, skip the rest of the loop.
+            if (inSource) continue;
```
bjorn: Rather than doing a condition here, just put the `continue` right after `inSource = false` (same for `inPriority` and `inCombine` below).

src/terraingenerator/main.cpp
```diff
@@ -73,28 +89,132 @@ static void showVersion()
...
         if (arg == QLatin1String("--help") || arg == QLatin1String("-h")) {
             options.showHelp = true;
         } else if (arg == QLatin1String("--version")
                 || arg == QLatin1String("-v")) {
             options.showVersion = true;
+        } else if (arg == QLatin1String("-o")
+                || arg == QLatin1String("--output")) {
+            i++;
+            if (i >= arguments.size()) {
+                qWarning() << "Missing argument to" << arg << "option";
```
bjorn: You'll probably want to put a `return false;` here, to avoid crashing on the line below and because the tool should abort if the command line parameters are invalid (there were other warnings above for which I think it also makes sense to just return false immediately after printing the message).

src/terraingenerator/main.cpp
```diff
@@ -73,28 +89,132 @@ static void showVersion()
...
 
-static void parseCommandLineArguments(CommandLineOptions &options)
+static bool parseCommandLineArguments(CommandLineOptions &options)
 {
     const QStringList arguments = QCoreApplication::arguments();
+    bool inPriority = false;
+    bool inSource = false;
+    bool inCombine = false;
+    int argCount = 0;
+    QStringList combine;
```
bjorn: Instead of having this combine variable, you could just append an empty `QStringList` to `options.combineList` and use `options.combineList.last().append(...)` to add to it. That avoids the need to add it later as well as to clear it.

src/terraingenerator/main.cpp
```diff
@@ -73,28 +89,132 @@ static void showVersion()
...
+                // Some other option is starting here.
+                if (argCount == 0) {
+                    qWarning() << "Missing arguments for --input option.";
+                }
+                inSource = false;
+            } else {
+                // Append the source file.
+                argCount++;
+                options.sources.append(arg);
+                //qDebug() << "adding source" << arg;
```
bjorn: Please remove these debug output lines.

src/terraingenerator/main.cpp
```diff
@@ -237,10 +357,15 @@ int main(int argc, char *argv[])
 
     a.setOrganizationDomain(QLatin1String("mapeditor.org"));
     a.setApplicationName(QLatin1String("TerrainGenerator"));
-    a.setApplicationVersion(QLatin1String("1.0"));
+    a.setApplicationVersion(QLatin1String("1.1"));
```
bjorn: Version does not increase since there was no release yet. :-)

src/terraingenerator/main.cpp
```diff
@@ -237,10 +357,15 @@ int main(int argc, char *argv[])
...
 
     CommandLineOptions options;
-    parseCommandLineArguments(options);
+
+    // Parse commandline arguments.
```
bjorn: Please no useless comments that repeat the source code.

src/terraingenerator/main.cpp
```diff
@@ -260,6 +385,11 @@ int main(int argc, char *argv[])
         return 1;
     }
 
+    // Check terrain priorities.
+    if (options.terrainPriority.isEmpty()) {
+        qFatal("No terrain priorities set!");
```
bjorn: Maybe mention `-p` option like you did with `-c` option on empty combine list.

src/terraingenerator/main.cpp
```diff
@@ -319,27 +460,31 @@ int main(int argc, char *argv[])
...
-    terrainOrder << "PlowedSoil";
-    terrainOrder << "Lava";
+    // Setup terrain priorities.
     TerrainLessThan lessThan;
     int priority = 0;
-    foreach (const QString &terrainName, terrainOrder) {
+    foreach (const QString &terrainName, options.terrainPriority) {
         lessThan.terrainPriority.insert(terrainName, priority);
-        ++priority;
+        priority++;
```
bjorn: Please don't arbitrarily change code around. It was fine in prefix notation since the result is not used.

src/terraingenerator/main.cpp
```diff
@@ -319,27 +460,31 @@ int main(int argc, char *argv[])
...
     }
 
     qDebug() << "Terrains found:" << terrains.keys();
 
+    // Check if all terrains from priority list were found and loaded.
+    foreach (const QString &terrainName, lessThan.terrainPriority.keys())
+        if (!terrains.contains(terrainName))
+            qWarning() << "Terrain" << terrainName << "from priority list not found.";
```
bjorn: This is a nice check, but it is actually more important to make sure that all terrains mentioned in combine lists are present. It looks like the tool would currently crash in such a case.

src/terraingenerator/main.cpp
```diff
@@ -73,28 +88,116 @@ static void showVersion()
...
+                continue;
+            }
+        }
+
+        // Process priority option.
+        if (inPriority) {
+            if (arg.at(0) == QLatin1Char('-')) {
+                // Some other option is starting here.
+                if (argCount == 0) {
+                    qWarning() << "Missing arguments for --input option.";
```
bjorn: Error message should mention `--priority` instead of `--input`.

src/terraingenerator/main.cpp
```diff
@@ -73,28 +88,116 @@ static void showVersion()
...
 
     for (int i = 1; i < arguments.size(); ++i) {
         const QString &arg = arguments.at(i);
+
+        // Process source option.
+        if (inSource) {
+            if (arg.at(0) == QLatin1Char('-')) {
+                // Some other option is starting here.
+                if (argCount == 0) {
+                    qWarning() << "Missing arguments for --input option.";
```
bjorn: This should be `--source` instead of `--input`.

src/terraingenerator/main.cpp
```diff
@@ -360,110 +498,30 @@ int main(int argc, char *argv[])
...
+                        //qDebug()<<topLeft->name()<<"\t"<<topRight->name()<<"\t"<<bottomLeft->name()<<"\t"<<bottomRight->name();
+                        process.append(TileTerrainNames(topLeft->name(),
+                                                        topRight->name(),
+                                                        bottomLeft->name(),
+                                                        bottomRight->name()));
+                    }
                 }
             }
         }
+        terrainList.clear();
```
bjorn: Instead of reusing and clearing the same `terrainList` every time, please just make the variable more local.

src/terraingenerator/main.cpp
```diff
@@ -360,110 +498,30 @@ int main(int argc, char *argv[])
...
-                    process.append(TileTerrainNames(topLeft->name(),
-                                                    topRight->name(),
-                                                    bottomLeft->name(),
-                                                    bottomRight->name()));
+        // Construct a vector with all terrain combinations to process
+        foreach (Terrain *topLeft, terrainList) {
+            foreach (Terrain *topRight, terrainList) {
+                foreach (Terrain *bottomLeft, terrainList) {
+                    foreach (Terrain *bottomRight, terrainList) {
+                        //qDebug()<<topLeft->name()<<"\t"<<topRight->name()<<"\t"<<bottomLeft->name()<<"\t"<<bottomRight->name();
```
bjorn: Please remove the debug lines, even when they are commented out.

bjorn:
Alright, thanks for the fixes! Please squash everything into the original commit now, then I can accept the merge request.

bjorn:
Really nice improvement, thanks!

Let's aim for shipping this tool with Tiled 0.13. I'd rather not hastily ship it with Tiled 0.12 at this point and then we have time enough to set up a good example and documentation.

PR #926, opened 2015-04-30
Update tiled_ru.ts

semtiko:
Thank you for your work!

### DIFF
--- a/translations/tiled_ru.ts
+++ b/translations/tiled_ru.ts
@@ -11,7 +11,7 @@
     <message>
         <location line="+83"/>
         <source>Donate</source>
-        <translation type="unfinished"></translation>
+        <translation>Спонсировать</translation>
     </message>
     <message>
         <location line="+7"/>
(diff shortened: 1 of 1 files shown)


translations/tiled_ru.ts
```diff
@@ -2357,44 +2357,44 @@ Line %1, column %2</source>
...
         <location line="+11"/>
         <source>Image too Big</source>
-        <translation type="unfinished"></translation>
+        <translation>Изображение слишком большое</translation>
     </message>
     <message>
         <location line="+1"/>
         <source>The resulting image would be %1 x %2 pixels and take %3 GB of memory. Tiled is unable to create such an image. Try reducing the zoom level.</source>
-        <translation type="unfinished"></translation>
+        <translation>Конечное изображение имеет размеры %1 x %2 пикселей и требует %3 GB памяти. Tiled не удалось создать изображение. Попробуйте увеличить масштаб.</translation>
```
bjorn: Google translate translates the last part as "Try to zoom in.", but reducing the zoom level means zooming out. Is this translation wrong or Google translate being silly?

bjorn:
Thank you for the update! I checked a few translations in Google Translate and found one case where I think the translation may be wrong, but I have personally no way to judge it so please verify.

semtiko:
Sorry for long reply. In russian language "try to zoom in" it's like a word game and can be translated in few ways and my way is just a simple way, but without missing something important. Don't worry )

bjorn:
Alright, thanks again!

PR #927, opened 2015-05-02
Implemented 'TODO'. Made cmd line parser's help msg translatable.

IMMZ:
A little impromevents to the project. One 'TODO' implemented.

### DIFF
--- a/src/tiled/commandlineparser.cpp
+++ b/src/tiled/commandlineparser.cpp
@@ -72,8 +72,8 @@ bool CommandLineParser::parse(const QStringList &arguments)
         if (arg.length() == 1) {
             // Traditionally a single hyphen means read file from stdin,
             // write file to stdout. This isn't supported right now.
-            qWarning().nospace() << "Bad argument " << index
-                                 << ": lonely hyphen";
+            qWarning().nospace() << tr("Bad argument ") << index
+                                 << tr(": lonely hyphen");
             showHelp();
             return false;
         }
@@ -87,7 +87,7 @@ bool CommandLineParser::parse(const QStringList &arguments)
             }
 
             if (!handleLongOption(arg)) {
-                qWarning().nospace() << "Unknown long argument " << index
+                qWarning().nospace() << tr("Unknown long argument ") << index
                                      << ": " << arg;
                 mShowHelp = true;
                 break;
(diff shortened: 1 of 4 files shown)


src/tiled/commandlineparser.cpp
```diff
@@ -118,13 +118,13 @@ bool CommandLineParser::parse(const QStringList &arguments)
...
+    qWarning().nospace() << QObject::tr("Usage") << ":\n"
                          << "  " << qPrintable(mCurrentProgramName)
-                         << " [options] [files...]\n\n"
-                         << "Options:";
+                         << " [" << QObject::tr("options") << "]"
+						 << " [" << QObject::tr("files") << "...]\n\n"
+                         << QObject::tr("Options") << ":";
 
-    qWarning("  -h %-*s : Display this help", mLongestArgument, "--help");
+    qWarning("  -h %-*s : " + QObject::tr("Display this help").toUtf8(), mLongestArgument, "--help");
```
bjorn: Thanks for trying to handle this TODO! However, it was still on my list because it is a little more involved. Here's my feedback:

Please use [`QCoreApplication::translate`](http://doc.qt.io/qt-5/qcoreapplication.html#translate) or use [`Q_DECLARE_TR_FUNCTIONS`](http://doc.qt.io/qt-5/qcoreapplication.html#Q_DECLARE_TR_FUNCTIONS) on the `CommandLineParser` to make sure the translations are given the right context "CommandLineParser". If you just call `QObject::tr`, the translations will get the context "QObject", which isn't helpful.

Also, please don't use `toUtf8()` to produce printable strings but rather [`qPrintable(...)`](http://doc.qt.io/qt-4.8/qtglobal.html#qPrintable). This is because you can't assume the console encoding is UTF-8.

Try to use less small strings that are later put together. It is better for the translator when the sentences are kept more intact, because it gives more context and more freedom in doing the translation. You can use parameters instead. The source strings here should probably be:

"Usage:\n  %1 [options] [files...]"
"Options:"
"Display this help"

Finally, note that this is only part of the command line output. The texts printed below in the `foreach` can be found in the `CommandLineHandler` subclass in `main.cpp` and they should then also be made translatable (and hence I would prefer if `Q_DECLARE_TR_FUNCTIONS(CommandLineParser)` was used at the `CommandLineParser` class, for which you need to `#include <QCoreApplication>`).

PR #929, opened 2015-05-02
Fix desktop Exec field.

grandpaul:
The desktop entry has support for a mime type, that means the application can
handle files of the listed mime types as a parameter. So we should add %f to
Exec key in the desktop file.


bjorn:
Hey, thanks for the improvement and for your packaging efforts!

PR #930, opened 2015-05-02
Implemented 'TODO'. Made cmd line parser translatable.

IMMZ:
Implemented 'TODO'. Made cmd line parser translatable.

### DIFF
--- a/src/tiled/commandlineparser.cpp
+++ b/src/tiled/commandlineparser.cpp
@@ -72,8 +72,7 @@ bool CommandLineParser::parse(const QStringList &arguments)
         if (arg.length() == 1) {
             // Traditionally a single hyphen means read file from stdin,
             // write file to stdout. This isn't supported right now.
-            qWarning().nospace() << "Bad argument " << index
-                                 << ": lonely hyphen";
+            qWarning() << qPrintable(tr("Bad argument %1: lonely hyphen").arg(index));
             showHelp();
             return false;
         }
@@ -87,8 +86,7 @@ bool CommandLineParser::parse(const QStringList &arguments)
             }
 
             if (!handleLongOption(arg)) {
-                qWarning().nospace() << "Unknown long argument " << index
-                                     << ": " << arg;
+                qWarning() << qPrintable(tr("Unknown long argument %1: %2").arg(index).arg(arg));
                 mShowHelp = true;
                 break;
             }
@@ -100,8 +98,8 @@ bool CommandLineParser::parse(const QStringList &arguments)
         for (int i = 1; i < arg.length(); ++i) {
             const QChar c = arg.at(i);
             if (!handleShortOption(c)) {
-                qWarning().nospace() << "Unknown short argument " << index
-                                     << '.' << i << ": " << c;
+                qWarning()
+                    << qPrintable(tr("Unknown short argument %1.%2: %3").arg(index).arg(i).arg(c));
                 mShowHelp = true;
                 break;
             }
@@ -118,13 +116,10 @@ bool CommandLineParser::parse(const QStringList &arguments)
 
 void CommandLineParser::showHelp()
 {
-    // TODO: Make translatable
-    qWarning().nospace() << "Usage:\n"
-                         << "  " << qPrintable(mCurrentProgramName)
-                         << " [options] [files...]\n\n"
-                         << "Options:";
+    qWarning().nospace() << qPrintable(tr("Usage:\n  %1 [options] [files...]").arg(mCurrentProgramName))
+                         << "\n\n" << qPrintable(tr("Options:"));
 
-    qWarning("  -h %-*s : Display this help", mLongestArgument, "--help");
+    qWarning("  -h %-*s : %s", mLongestArgument, "--help", qPrintable(tr("Display this help")));
 
     foreach (const Option &option, mOptions) {
         if (!option.shortName.isNull()) {
--- a/src/tiled/commandlineparser.h
+++ b/src/tiled/commandlineparser.h
@@ -22,6 +22,7 @@
 #ifndef COMMANDLINEPARSER_H
 #define COMMANDLINEPARSER_H
 
+#include <QCoreApplication>
 #include <QStringList>
 #include <QVector>
 
@@ -54,6 +55,7 @@ void MemberFunctionCall(void *object)
  */
 class CommandLineParser
 {
+    Q_DECLARE_TR_FUNCTIONS(CommandLineParser);
 
 public:
     CommandLineParser();
--- a/src/tiled/main.cpp
+++ b/src/tiled/main.cpp
@@ -91,17 +91,17 @@ CommandLineHandler::CommandLineHandler()
     option<&CommandLineHandler::showVersion>(
                 QLatin1Char('v'),
                 QLatin1String("--version"),
-                QLatin1String("Display the version"));
+                tr("Display the version"));
 
     option<&CommandLineHandler::justQuit>(
                 QChar(),
                 QLatin1String("--quit"),
-                QLatin1String("Only check validity of arguments"));
+                tr("Only check validity of arguments"));
 
     option<&CommandLineHandler::setDisableOpenGL>(
                 QChar(),
                 QLatin1String("--disable-opengl"),
-                QLatin1String("Disable hardware accelerated rendering"));
+                tr("Disable hardware accelerated rendering"));
 
     option<&CommandLineHandler::setExportMap>(
                 QChar(),


src/tiled/commandlineparser.h
```diff
@@ -55,6 +56,8 @@ void MemberFunctionCall(void *object)
 class CommandLineParser
 {
```
bjorn: Please remove this initial newline while you're at it. :-)

src/tiled/commandlineparser.cpp
```diff
@@ -72,8 +72,8 @@ bool CommandLineParser::parse(const QStringList &arguments)
         if (arg.length() == 1) {
             // Traditionally a single hyphen means read file from stdin,
             // write file to stdout. This isn't supported right now.
-            qWarning().nospace() << "Bad argument " << index
-                                 << ": lonely hyphen";
+            qWarning().nospace() << tr("Bad argument ") << index
+                                 << tr(": lonely hyphen");
```
bjorn: It seems like you're not aware of how to use string arguments. This should really be:
```c++
qWarning().nospace() << tr("Bad argument %1: lonely hyphen").arg(index);
```
Please also do the "Usage:" line that way with the program name.

src/tiled/commandlineparser.cpp
```diff
@@ -118,13 +118,11 @@ bool CommandLineParser::parse(const QStringList &arguments)
 
 void CommandLineParser::showHelp()
 {
-    // TODO: Make translatable
-    qWarning().nospace() << "Usage:\n"
-                         << "  " << qPrintable(mCurrentProgramName)
-                         << " [options] [files...]\n\n"
-                         << "Options:";
+    qWarning().nospace() << tr("Usage:\n  ") << qPrintable(mCurrentProgramName)
+                         << tr(" [options] [files...]\n\n")
```
bjorn: Please do not include `\n\n` at the end of the translatable string, since they're easy to miss for translators. For those it is really better to have them outside of the translatable part.

src/tiled/commandlineparser.cpp
```diff
@@ -118,13 +118,11 @@ bool CommandLineParser::parse(const QStringList &arguments)
...
+                         << tr("Options") << ":";
 
-    qWarning("  -h %-*s : Display this help", mLongestArgument, "--help");
+    qWarning(qPrintable(tr("  -h %-*s : Display this help")), mLongestArgument, "--help");
```
bjorn: Similarly, here I would go with your initial approach of making only "Display this help" translatable, since the rest should not be touched.

src/tiled/commandlineparser.h
IMMZ: fixed

src/tiled/commandlineparser.cpp
IMMZ: Fixed. Not sure If I do ok with program name by removing qPrintable.

src/tiled/commandlineparser.cpp
IMMZ: '\n\n' is extracted out of translation.

src/tiled/commandlineparser.cpp
IMMZ: Done.

src/tiled/commandlineparser.cpp
```diff
@@ -72,8 +72,7 @@ bool CommandLineParser::parse(const QStringList &arguments)
         if (arg.length() == 1) {
             // Traditionally a single hyphen means read file from stdin,
             // write file to stdout. This isn't supported right now.
-            qWarning().nospace() << "Bad argument " << index
-                                 << ": lonely hyphen";
+            qWarning().nospace() << tr("Bad argument %1: lonely hyphen").arg(index);
```
bjorn: This will need to be wrapped with `qPrintable` to avoid it being quoted in the output. Also, the `nospace()` can be removed since there aren't multiple pieces streamed to the output anyway. So this should be:
```
qWarning() << qPrintable(tr("Bad argument %1: lonely hyphen").arg(index));
```
Other lines need to be changed similarly.

src/tiled/commandlineparser.cpp
```diff
@@ -118,13 +116,12 @@ bool CommandLineParser::parse(const QStringList &arguments)
...
-                         << "  " << qPrintable(mCurrentProgramName)
-                         << " [options] [files...]\n\n"
-                         << "Options:";
+    qWarning().nospace() << tr("Usage:\n  %1").arg(mCurrentProgramName)
+                         << tr(" [options] [files...]") << "\n\n"
+                         << tr("Options") << ":";
 
-    qWarning("  -h %-*s : Display this help", mLongestArgument, "--help");
+    qWarning(qPrintable(QLatin1String("  -h %-*s : ") + tr("Display this help")),
+             mLongestArgument, "--help");
```
bjorn: I think this would be clearer:
```
qWarning("  -h %-*s : %s", mLongestArgument, "--help", qPrintable(tr("Display this help")));
```

IMMZ: Fixed.

src/tiled/commandlineparser.cpp
IMMZ: Fixed.

src/tiled/commandlineparser.cpp
```diff
@@ -100,8 +98,8 @@ bool CommandLineParser::parse(const QStringList &arguments)
         for (int i = 1; i < arg.length(); ++i) {
             const QChar c = arg.at(i);
             if (!handleShortOption(c)) {
-                qWarning().nospace() << "Unknown short argument " << index
-                                     << '.' << i << ": " << c;
+                qWarning()
+                    << qPrintable(tr("Unknown short argument %1.%2: %3").arg(index).arg(i).arg(c));
```
bjorn: There's a `%` sign missing for the third argument.

src/tiled/commandlineparser.cpp
```diff
@@ -118,13 +116,11 @@ bool CommandLineParser::parse(const QStringList &arguments)
 
 void CommandLineParser::showHelp()
 {
-    // TODO: Make translatable
-    qWarning().nospace() << "Usage:\n"
-                         << "  " << qPrintable(mCurrentProgramName)
-                         << " [options] [files...]\n\n"
-                         << "Options:";
+    qWarning().nospace() << qPrintable(tr("Usage:\n  %1").arg(mCurrentProgramName))
+                         << tr(" [options] [files...]") << "\n\n"
```
bjorn: Please merge the above two strings like "Usage:\n  %1 [options] [files...]".

src/tiled/commandlineparser.cpp
IMMZ: bjorn: it 'drawn' by white color. I have no ideas why. It really exists in the file.

bjorn: Ah, silly highlighter...

src/tiled/commandlineparser.cpp
IMMZ: Done.

src/tiled/commandlineparser.cpp
```diff
@@ -118,13 +116,10 @@ bool CommandLineParser::parse(const QStringList &arguments)
 
 void CommandLineParser::showHelp()
 {
-    // TODO: Make translatable
-    qWarning().nospace() << "Usage:\n"
-                         << "  " << qPrintable(mCurrentProgramName)
-                         << " [options] [files...]\n\n"
-                         << "Options:";
+    qWarning().nospace() << qPrintable(tr("Usage:\n  %1 [options] [files...]").arg(mCurrentProgramName))
+                         << "\n\n" << tr("Options") << ":";
```
bjorn: Hmm, I may have missed this before, but the `tr("Options")` also needs to be wrapped in `qPrintable`.

src/tiled/commandlineparser.cpp
IMMZ: Done.

bjorn:
Thanks for all the fixes, I think it looks pretty good now!

I've branched off Tiled 0.12, so that I could merge this into master. Since it touches the translatable strings and Tiled 0.12 is already in string freeze, this will be for Tiled 0.13.

bjorn:
Hey @IMMZ, the Bulgarian translator asked whether it would be possible to make also the output of the `tmxrasterizer` and `tmxviewer` tools translatable. Is this something you'd be interested to work on?

IMMZ:
@bjorn, I will look at it in the nearest time.

bjorn:
@IMMZ Cool! There is no rush of course. Note that it won't be enough to just pass the strings through `tr`, since the tools also still need to install the translators. To start with you could just copy the `languagemanager.{h,cpp}` files, but it may be nice to share this functionality in a small static library.

IMMZ:
@bjorn, there is need to extract functionality of language manager to small static library and use this library in all (tiled, rasterizer, viewer) executables? And how should be library called?

bjorn:
@IMMZ I mentioned this part as optional, but if you want to give it a try, maybe call it `tiledlocalization`.

PR #938, opened 2015-05-10
Objecttype

ludoza:
I wanted to pass on the current selected object type, so I added a replacement.


bjorn:
Looks like a fine addition to me. I've squashed your commits together and pushed them to the `0.12` branch as b06f22a so that it will get included in the next release.

PR #939, opened 2015-05-11
Added PropertyBrowser to Tiled Collision Editor

Seanba:
The PropertyBrowser (more correctly, the PropertiesDock) was edited slightly to allow the Collision Editor to use it for adding and manipulating properties for the ObjectGroup that it adds to tiles.

A couple of issues:
- It would be nice to be able to select objects even if the Tileset is
external but the MapView is disabled in this case.
- If I try to keep the MapView enabled then the toolbar actions have to
be disabled, except for the ObjectSelection tool. I couldn't get this to
work -- but I'm new to Qt.
- The ObjectSelection tool is no longer just about selection but allows
you to position, scale, and rotate the object that is currently selected
which further complicates this.
- It would also be nice to be able to go from selecting an Object to the
Object Group easily. Right now this is only achieved by selecting a new
tile (in which case the ObjectGroup for that tile is selected in the
PropertyBrowser until you select an object.


bjorn:
This looks like a very good start and I'm sure we can tackle those issues.

Could you look into cleaning up your git history and doing a force-push to update this pull request with just a clean version of change 4beed1c1ef26a55a2aa0302d2c8115a7ce5d38ce?

bjorn:
@Seanba Do you think you'll manage to get rid of those "Automapping Tile Objects" related commits and the merge commit to clean up this pull request? I can try to give you some instructions if not.

About the "ObjectGroup" selection problem, are you sure there is a use-case in being able to select this? I consider the object group mostly just an implementation detail.

The other problems you mentioned seem to be about having a "read-only" mode for the selection tool. Also the problem with disabling some tools explicitly is probably that the tools update whether they are enabled by themselves already. It's something I could look into.

In the end though, I should really make it so that there is no need to disable editing external tilesets in the first place. This is something I want to do anyway eventually, and I think the ability to select objects could wait until then.

Seanba:
Bjorn, sorry for the late reply. I did go through and clean up the commits but there's a stubborn bit of history with the first Automapping pull request I made that I don't know how to get rid of. I could just make a separate fork and start off clean with this pull request if that's easier.

For "ObjectGroup" selection, I like to use the color property here for my other tools in the development chain so it would be nice to be able to select and edit it.

Thanks.

bjorn:
> Bjorn, sorry for the late reply. I did go through and clean up the commits but there's a stubborn bit of history with the first Automapping pull request I made that I don't know how to get rid of. I could just make a separate fork and start off clean with this pull request if that's easier.

I'll try to give you a few instructions to get you out of the mess. I'm assuming you have a remote called 'bjorn' pointing to my repository, rename that as needed. I'm also assuming you have no local changes that you want to keep, because this will throw them away.

Reset your master to match with my master branch:

    git checkout master
    git fetch bjorn
    git reset --hard bjorn/master

Cherry-pick commit ffef1649f0719a6a52f1f7a242cdfd0608f969bd onto this master branch:

    git cherry-pick ffef1649f0719a6a52f

Push it to your own repository (needs force-push):

    git push -f

This will clean up this pull request, since it is referring to the master branch in your fork.

Another small issue that I found while testing your change, is that it should not be possible to toggle the visibility of the collision objects. It makes little sense anyway, but the main reason is that currently you would not be able to select the object to toggle it back to visible. I guess this will need to be some kind of option passed on to the properties view.

bjorn:
Pushed to `master` as 086e1c7e77f21712d4d1d9ec832973dd786c57dd and improved a little in change c3374357bd5d75e45daaf98be519e92f6b47ce1b. Thanks!

I've made no attempts at improving the "read only" mode of this editor, since I instead would like to make this mode obsolete by supporting the editing of external tilesets anyway. That will be quite a difficult change still, though.

PR #967, opened 2015-05-30
Added a single tilesheet per layer option

Ohmnivore:
Now allows to optionally save a .tmx file in which:
* Layer elements have a tileset property (if it's not a totally empty layer) which identify the tileset used for the layer. It doesn't refrain you from using various tilesets in a single layer, but if you do that then you don't need this option anyways:
 ```
 <data encoding="csv" tileset="test_tiles">
 ```
* Tiles' ids are saved, not the gids

There is an option (defaults to false) to enable this saving mode. It can be toggled in the new map creation dialog, as well as the map properties pane.

### DIFF
--- a/src/libtiled/mapwriter.cpp
+++ b/src/libtiled/mapwriter.cpp
@@ -94,6 +94,7 @@ private:
     QDir mMapDir;     // The directory in which the map is being saved
     GidMapper mGidMapper;
     bool mUseAbsolutePaths;
+    const Map *mMap;
 };
 
 } // namespace Internal
@@ -213,6 +214,7 @@ void MapWriterPrivate::writeMap(QXmlStreamWriter &w, const Map *map)
         firstGid += tileset->tileCount();
     }
 
+    mMap = map;
     foreach (const Layer *layer, map->layers()) {
         const Layer::TypeFlag type = layer->layerType();
         if (type == Layer::TileLayerType)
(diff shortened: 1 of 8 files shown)


src/libtiled/mapwriter.cpp
```diff
@@ -426,11 +428,34 @@ void MapWriterPrivate::writeTileLayer(QXmlStreamWriter &w,
...
             for (int x = 0; x < tileLayer->width(); ++x) {
-                const unsigned gid = mGidMapper.cellToGid(tileLayer->cellAt(x, y));
+                const Cell c = tileLayer->cellAt(x, y);
+                unsigned gid = mGidMapper.cellToGid(c);
                 w.writeStartElement(QLatin1String("tile"));
+                if (mMap->singleSheet())
+                    if (c.isEmpty())
+                        gid = 0;
+                    else
+                        gid = c.tile->id();
```
bjorn: So in this mode, it is in fact not possible for a layer to have "empty" cells, since they are indistinguishable from the first tile in the tileset.

bjorn:
I guess this patch is missing changes to the map reader, because as-is the maps written like this will not read properly back into Tiled.

Even if that was fixed though, this isn't something I'd be likely to merge because it's superfluous. There is nothing added here apart from a little efficiency when reading a map. And if efficiency was that important, you'd be better off implementing your own map format either as an export plugin or as some convertion tool.

Issue #975, opened 2015-06-10
Change different properties of flipping = Exception


Znevs:
If select both object and change flipping Horizontal flag, then application is crashed. Sorry for bad English.

![q405dyp](https://cloud.githubusercontent.com/assets/531764/8139501/9082b162-1154-11e5-99a8-d16573d57f22.png)

bjorn:
Hmm, strange, I am unable to reproduce this issue. Does it always happen when flipping multiple objects at the same time, or is it maybe a more specific case? Also, did you try older versions of Tiled to see whether this is a new issue?

IMMZ:
I'm getting crash too. Steps to reproduce:
1) Create a new map.
2) Put 2 tiled objects on it.
3) Set 'horizontal flipped = true' for one of them (any you want)
4) Select both.
5) Try to change horizontal flipping again
CRASH!
I've made a fix and created a pull request #978 It would be good to see any critique or see it's sucesfully merged.

bjorn:
@IMMZ Thanks for finding the extra step needed to reproduce this issue, as well as for the fix!

PR #978, opened 2015-06-13
#975 Fixed crash while changing different properties at once.

IMMZ:
applyMapObjectValue(...) returns command only if flipping state is different from the user set. In sutuations when user selects multiply objects and try to set an option for them applyMapObjectValue(...) returns 0 for some already checked/unchecked flipping options (the state is same as user set). This '0' is the reason of crash and now was fixed.


bjorn:
The fix looks fine, but since it should go to the `0.12` branch I cherry-picked it there as commit 4c2c4a1. I've also shortened the code a little and put a bit more information in the commit message.

Thanks!

Issue #981, opened 2015-06-14
Straight polyline disappear when you try to resize it


Darky-Lucera:
Version 0.12.3

If you have an horizontal straight polyline and you use any of the 'to down' arrows (down, left-down, right-down) the polyline disappear and the Y property is equal to -2147483647,00 (0x80000001?). 
There is no problem is you use any of the 'to up' arrows (up, left-up, right-up), then only X is modified.

If you have a vertical straight polyline and you use any of the 'to right' arrows (right, right-down, right-up) the polyline disappear and the X property is equal to -2147483647,00 (0x80000001?). 
There is no problem is you use any of the 'to left' arrows (left, left-down, left-up), then only Y is modified.

There is no problem if the polyline is not a straight line.

bjorn:
Hmm, of course it is not actually possible to stretch a single horizontal line downwards. I'll probably have to deal with this case separately. Thanks for the report!

IMMZ:
The problem really exists. The best possible fix I think is to prevent user from incorrect resizing and don't annoy him by showing incorrect variants. My pull request with fix => https://github.com/bjorn/tiled/pull/982

PR #982, opened 2015-06-14
#981 Hide extra sizers from straight polylines.

IMMZ:
Fix for #981. Now there are only two handles for straight lines which prevent user from incorrect resizing.


src/tiled/objectselectiontool.cpp
```diff
@@ -825,6 +825,24 @@ void ObjectSelectionTool::updateHandleVisibility()
...
         mResizeHandles[i]->setVisible(showHandles && mMode == Resize);
 
+    // Hide extra handles for the straight line polygons.
+    foreach (MapObject *mapObj, mapDocument()->selectedObjects()) {
+        if (mapObj->shape() == MapObject::Polyline &&
+            mapObj->polygon().size() == 2) {
+            mResizeHandles[TopLeftAnchor]->setVisible(false);
+            mResizeHandles[BottomLeftAnchor]->setVisible(false);
+            mResizeHandles[TopRightAnchor]->setVisible(false);
+            mResizeHandles[BottomRightAnchor]->setVisible(false);
```
bjorn: Hmm, this seems wrong to me. You're hiding all corner anchors as soon as you see a polyline with only two points. However, there can be a diagonal polyline with two points, for which these handles make perfect sense.

Another problem with the patch is that it doesn't fix the same problem for polylines with more than two points that all line up either vertically or horizontally. This means the check should probably be done on the bounding box.

Finally, you're going over all selected objects and discard some handles as soon as one of them would be a short polyline. However, when there is a short polyline selected in combination with some other object, or even just two short polylines in different positions, then all handles make sense and work fine.

Now, it's too late for me to consider the correct fix. Feel free to have another go at it, but otherwise I will look into it later.

IMMZ: I'll take some tries again, so you can leave it to me and pay an attention to more important issues/features. Thanks for words, it seems I hurried with the patch.

IMMZ:
1. I'm interested in the way it should work. There is a screenshot below of my point of view. Is it OK or has some issues? If it's not good then it would be nice if someone will make an offer how it should work.
Current situation: (OK means that situation is 'handled')
 * Straight horizontal lines - OK. (**Behavior**: two anchors shown)
 * Straight vertical lines - OK. (**Behavior**: two anchors shown)
 * Straight right diagonal lines - OK. (**Behavior**: two anchors shown)
 * Straight left diagonal  lines - OK. (**Behavior**: two anchors shown)
 * Segmented straight lines with two or more points (horizontal, vertical, diagonal) - OK.  (**Behavior**: two anchors shown)
Selection of more than one shapes - OK (**Behavior**: All anchors shown)
tmx

2. What about resizing when 'Snap to grid' is ON and the shape already has it's minimal size (width or height = 1 cell)? Next resizing makes one (or both) of shape's sides invisible. The screenshot below shows this situation. Should it be extracted to separate issue here at github? I offer just to block resizing in such situations.
Something like:
if (snapToGridEnabled && (selectedShape.width() == tileWidth || selectedShape.height() == tileHeight()))
{
  // Block resizing.
}
block

bjorn:
> * Straight right diagonal lines - OK. (Behavior: two anchors shown)
> * Straight left diagonal lines - OK. (Behavior: two anchors shown)

These two are not OK. There is no reason not to simply show all handles for these cases.

> * Segmented straight lines with two or more points (horizontal, vertical, diagonal) - OK. (Behavior: two anchors shown)

Almost correct. Again, for diagonal straight lines there is no reason not to show all anchors.

> * Selection of more than one shapes - OK (Behavior: All anchors shown)

It is not always correct to show all anchors in this case. Basically, when the bounding box of all selected objects has 0-width or 0-height, and it is not a single selected rectangle, ellipse or tile object, then there is no way to stretch it in horizontal or vertical direction respectively and the handles for these could be hidden.

> What about resizing when 'Snap to grid' is ON and the shape already has it's minimal size (width or height = 1 cell)?

Snap to grid has nothing to do with the minimum size, which is still 0. Also, it is not the size that is snapped to the grid, but the corner or side that you're dragging around. This way, you can easily snap one side but not another side, for example.

IMMZ:
Still don't know how it should work properly. Maybe [this](https://gist.github.com/IMMZ/81b9ac7fb56b2d59eac8#file-gistfile1-diff) patch is OK?

bjorn:
> Maybe this patch is OK?

Could you update this pull request instead of referring to a pasted diff? It would make reading the code a lot more comfortable.

IMMZ:
Done.

src/tiled/objectselectiontool.cpp
```diff
@@ -825,6 +825,44 @@ void ObjectSelectionTool::updateHandleVisibility()
     for (int i = 0; i < AnchorCount; ++i)
         mResizeHandles[i]->setVisible(showHandles && mMode == Resize);
 
+    if (hasSelection) {
+        int selectedObjectsCount = mapDocument()->selectedObjects().count();
+        MapObject *firstSelectedObj =mapDocument()->selectedObjects().at(0);
+        qreal totalHeight = 0.0, totalWidth = 0.0;
```
bjorn: Summing up the totals won't work correctly. It's not the sum of the widths that matters, but the width of the bounding box around all the selected objects. You can see this being calculated in `ObjectSelectionTool::updateHandles` as `boundingRect`.

src/tiled/objectselectiontool.cpp
```diff
@@ -825,6 +825,44 @@ void ObjectSelectionTool::updateHandleVisibility()
...
+        }
+
+
+        /*
+         * Check whether we haven't selected single rectangle, ellipse or tile or we have selected
+         * multiply objects.
+         */
+        if ((selectedObjectsCount == 1 &&
+                firstSelectedObj->shape() == MapObject::Polyline &&
+                firstSelectedObj->polygon().first().y() == firstSelectedObj->polygon().last().y()) ||
```
bjorn: This check on the y coordinate is wrong. You should check the bounding box, since you may have a polyline that has for example three points, with the first and the last on the same y but the middle node lower or higher.

Also, a `MapObject::Polygon` can have 0 width or height just as well and can't be resized then either, so it's wrong to only check for polylines.

IMMZ:
Ok. I'll try to handle this next time.

bjorn:
I'm sorry I can't tell you exactly how to solve it and rather just have to point out what will not work in your patch. The thing is that if I would spend more time and consider exactly how this can be fixed, then I could just as well write the patch. :-)

PR #983, opened 2015-06-14
allow setting transparency color for tileset image collection

nobody-special666:
This patch moves the transparency related widgets out of the Image group-box so it can be set when creating a image collection tileset. It also applies the transparency color when you add an image from the tileset dock.

screenshot
screenshot2

### DIFF
--- a/src/tiled/newtilesetdialog.ui
+++ b/src/tiled/newtilesetdialog.ui
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>387</width>
-    <height>314</height>
+    <height>345</height>
    </rect>
   </property>
   <property name="windowTitle">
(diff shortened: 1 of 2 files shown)


bjorn:
Ah, I thought nobody used this feature anymore. ;-)

Anyway, good idea to still have it there since there is no reason not to support it. However, your patch is incomplete since the color would not only have to be applied in the Tilesets dock, but also when the map is loaded. I think the part that actually applies the color should be moved into the Tileset class.

nobody-special666:
I fixed setting the transparent color when saving and loading a tileset and added the transparency color to the propertybrowser. Please review and commit

src/libtiled/mapreader.cpp
```diff
@@ -456,13 +465,6 @@ void MapReaderPrivate::readTilesetImage(Tileset *tileset)
 
     const QXmlStreamAttributes atts = xml.attributes();
     QString source = atts.value(QLatin1String("source")).toString();
-    QString trans = atts.value(QLatin1String("trans")).toString();
-
-    if (!trans.isEmpty()) {
-        if (!trans.startsWith(QLatin1Char('#')))
-            trans.prepend(QLatin1Char('#'));
-        tileset->setTransparentColor(QColor(trans));
-    }
```
bjorn: Hmm, you can't just remove this code, since that breaks backwards compatibility. In general I'm not sure whether this change is really worth breaking compatibility at all. :-/

bjorn:
Maybe it could just be fixed while keeping compatibilty, by writing the `trans` attribute on each `image` element of the different tiles.

nobody-special666:
I didn't just remove that code for transparency. I moved it out of readTilesetImage() where it only applies to loading a tileset image and moved it into readTileset() where it applies to both tileset image and collection. I tested it loading both types of tilesets and it works.

bjorn:
By moving it from `readTilesetImage`, which reads the `image` element inside a `tileset` element, to `readTileset`, which reads the `tileset` element, you're now expecting the `trans` attribute to be on the `tileset` element instead of the `image` element. This breaks compatibility.

nobody-special666:
Solution: Then allow the property in both places and update the tileset transparency from both places. The front-end only allows 1 transparency color per tileset anyways so it should match at both levels.

nobody-special666:
ok the trans property has been moved back to the image tag

nobody-special666:
Is there still a problem preventing this commit from being merged?

bjorn:
Yes, the fact that I still need an almost full time job to make a living so it could take some days for me to get around to doing another review.

PR #988, opened 2015-06-17
Fixed possible memory leaks.

IMMZ:
There are some classes which used as base for other classes and haven't got virtual destructors. It could be the reason of memory leaks in some cases. This pull request fixes this situation.

### DIFF
--- a/src/libtiled/logginginterface.h
+++ b/src/libtiled/logginginterface.h
@@ -46,6 +46,8 @@ class LoggingInterface
       INFO, ERROR
     };
 
+    virtual ~LoggingInterface() {}
+
     virtual void log(OutputType type, const QString) = 0;
 };
 
--- a/src/libtiled/mapreader.h
+++ b/src/libtiled/mapreader.h
@@ -54,7 +54,7 @@ class TILEDSHARED_EXPORT MapReader
 {
 public:
     MapReader();
-    ~MapReader();
+    virtual ~MapReader();
 
     /**
      * Reads a TMX map from the given \a device. Optionally a \a path can


src/libtiled/hexagonalrenderer.h
```diff
@@ -69,6 +69,7 @@ class TILEDSHARED_EXPORT HexagonalRenderer : public OrthogonalRenderer
 
 public:
     HexagonalRenderer(const Map *map) : OrthogonalRenderer(map) {}
+    virtual ~HexagonalRenderer() {}
```
bjorn: Not needed because `MapRenderer` already declared the destructor virtual.

src/libtiled/orthogonalrenderer.h
```diff
@@ -41,6 +41,7 @@ class TILEDSHARED_EXPORT OrthogonalRenderer : public MapRenderer
 {
 public:
     OrthogonalRenderer(const Map *map) : MapRenderer(map) {}
+    virtual ~OrthogonalRenderer() {}
```
bjorn: Same here, `MapRenderer` has already declared it virtual.

src/libtiled/logginginterface.h
```diff
@@ -46,6 +46,8 @@ class LoggingInterface
       INFO, ERROR
     };
 
+    virtual ~LoggingInterface() {}
```
bjorn: Nobody ever calls `delete` on a `LoggingInterface` pointer, and any class implementing this interface is also inheriting from `QObject` which already has a virtual destructor. But anyway I think this addition is fine.

src/libtiled/mapreader.h
```diff
@@ -54,7 +54,7 @@ class TILEDSHARED_EXPORT MapReader
 {
 public:
     MapReader();
-    ~MapReader();
+    virtual ~MapReader();
```
bjorn: While the `TmxMapReader` indeed derives from this class, it is never used with polymorphism. Still, I think this is a fine addition as well just to avoid running into possible issues later.

bjorn:
Please edit the commit to only add the virtual to the destructors that weren't already virtual.

IMMZ:
Fixed.

bjorn:
Thanks. :-)

PR #989, opened 2015-06-17
TODO solved. Opening last files is now is optional.

IMMZ:
Now on every opening user will be asked for opening last files. It gives him the choice to open them or not.

### DIFF
--- a/src/tiled/main.cpp
+++ b/src/tiled/main.cpp
@@ -286,7 +286,7 @@ int main(int argc, char *argv[])
     if (!commandLine.filesToOpen().isEmpty()) {
         foreach (const QString &fileName, commandLine.filesToOpen())
             w.openFile(fileName);
-    } else {
+    } else if (Preferences::instance()->openLastFilesOnStartup()) {
         w.openLastFiles();
     }
 
--- a/src/tiled/preferences.cpp
+++ b/src/tiled/preferences.cpp
@@ -118,6 +118,11 @@ Preferences::Preferences()
     }
     mSettings->setValue(QLatin1String("RunCount"), mRunCount);
     mSettings->endGroup();
+
+    // Retrieve startup settings
+    mSettings->beginGroup(QLatin1String("Startup"));
+    mOpenLastFilesOnStartup = boolValue("OpenLastFiles", true);
+    mSettings->endGroup();
 }
 
 Preferences::~Preferences()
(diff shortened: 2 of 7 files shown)


bjorn:
Well, this isn't the solution I was looking for since asking the user all the time will get very annoying fast.

A much better solution in my opinion is to have an "Open last files on startup" checkbox in the preferences dialog. It should be enabled by default, but it would allow people who don't like this behavior to disable it.

IMMZ:
Done.

src/tiled/preferences.cpp
```diff
@@ -435,6 +440,15 @@ void Preferences::setPatron(bool isPatron)
     emit isPatronChanged();
 }
 
+void Preferences::setOpenLastFilesOnStartup(bool open)
+{
+    if (mOpenLastFilesOnStartup != open)
+    {
+        mOpenLastFilesOnStartup = open;
+        mSettings->setValue(QLatin1String("Startup/OpenLastFiles"), open);
+    }
```
bjorn: Please watch the code style here and follow the same structure as other setters:
```c++
    if (mOpenLastFilesOnStartup == value)
        return;

    mOpenLastFilesOnStartup = value;
    mSettings->setValue(QLatin1String("Startup/OpenLastFiles"), value);
```
(also I consider `value` a better name here than `open`)

src/tiled/preferences.cpp
```diff
@@ -118,6 +118,11 @@ Preferences::Preferences()
     }
     mSettings->setValue(QLatin1String("RunCount"), mRunCount);
     mSettings->endGroup();
+
+    // Retrieve startup settings
+    mSettings->beginGroup(QLatin1String("Startup"));
+    mOpenLastFilesOnStartup = boolValue("OpenLastFiles");
```
bjorn: This would default the setting to `false`, but it should really be enabled by default, so `boolValue("OpenLastFiles", true);`

src/tiled/main.cpp
```diff
@@ -286,9 +286,8 @@ int main(int argc, char *argv[])
     if (!commandLine.filesToOpen().isEmpty()) {
         foreach (const QString &fileName, commandLine.filesToOpen())
             w.openFile(fileName);
-    } else {
-        w.openLastFiles();
-    }
+    } else if (Preferences::instance()->openLastFilesOnStartup())
+            w.openLastFiles();
```
bjorn: Please correct the indentation here, as well as keeping the `{` and `}`.

IMMZ:
All notices fixed.

bjorn:
Code looks fine now, but please make the following adjustment to the preference dialog layout (just putting the checkbox in the "Saving and Loading" section):

selection_016

IMMZ:
Done.

bjorn:
Nice, thanks for the improvement!

PR #991, opened 2015-06-18
TODO solved. Select next property after deleting one.

IMMZ:
This patch selects next property when user deletes one.

### DIFF
--- a/src/tiled/propertiesdock.cpp
+++ b/src/tiled/propertiesdock.cpp
@@ -223,9 +223,16 @@ void PropertiesDock::removeProperty()
 
     const QString name = item->property()->propertyName();
     QUndoStack *undoStack = mMapDocument->undoStack();
+    QList<QtBrowserItem *> items = item->parent()->children();
+    if (items.count() > 1) {
+        int currentItemIndex = items.indexOf(item);
+        if (item == items.last()) {
+            mPropertyBrowser->setCurrentItem(items.at(currentItemIndex - 1));
+        } else {
+            mPropertyBrowser->setCurrentItem(items.at(currentItemIndex + 1));
+        }
+    }
     undoStack->push(new RemoveProperty(mMapDocument, mMapDocument->currentObjects(), name));
-
-    // TODO: Would be nice to automatically select the next property
 }
 
 void PropertiesDock::renameProperty()


bjorn:
Hmm, in testing this I noticed when I press Delete once, sometimes two or three properties get deleted. It seems like the `removeProperty` function is called multiple times, and the fact that the selection was lost the first time was preventing it from deleting multiple properties, but no longer after this patch.

So probably the code in `PropertiesDock::event` is wrong and something should be changed about how this shortcut is handled.

src/tiled/propertiesdock.cpp
```diff
@@ -223,9 +223,15 @@ void PropertiesDock::removeProperty()
 
     const QString name = item->property()->propertyName();
     QUndoStack *undoStack = mMapDocument->undoStack();
+    if (item->parent()->children().count() > 1) {
+        int currentItemIndex = item->parent()->children().indexOf(item);
+        if (item == item->parent()->children().last()) {
+            mPropertyBrowser->setCurrentItem(item->parent()->children().at(currentItemIndex - 1));
+        } else {
+           mPropertyBrowser->setCurrentItem(item->parent()->children().at(currentItemIndex + 1));
```
bjorn: After seeing `item->parent()->children()` for the 5th time, it's really asking for the introduction of a local variable for it.

IMMZ: Fixed.

src/tiled/propertiesdock.cpp
```diff
@@ -223,9 +223,16 @@ void PropertiesDock::removeProperty()
 
     const QString name = item->property()->propertyName();
     QUndoStack *undoStack = mMapDocument->undoStack();
+    QList<QtBrowserItem *> items = item->parent()->children();
+    if (items.count() > 1) {
+        int currentItemIndex = items.indexOf(item);
+        if (item == items.last()) {
+            mPropertyBrowser->setCurrentItem(items.at(currentItemIndex - 1));
+        } else {
+           mPropertyBrowser->setCurrentItem(items.at(currentItemIndex + 1));
```
bjorn: Ah, and this line has the wrong indentation.

IMMZ: Ooops. Fixed.

IMMZ:
Oh. Unexpected behavior. I'll try to investigate this.

IMMZ:
Manual gui testing shows that with 'delete/bkspc' there always two properties deleted. I looked at code and made next conclusion: the problem happens in PropertiesDock::event(...). Every time we press the delete/bkspc at less two events occured - at first ShortcutOverride event and then KeyPress event. I have no idea why there are handled two events type. Qt Assistant tells that 'ShortcutOverride' is the 'key press in a child' and because of this we have two KeyPress events. I can make a mistake at this point. That's why every time we press the key we get 'removeProperty()' called twice. Code works as expected if we remove 'case QEvent::KeyPress' and I have a questions:
1. Can we safely remove one case from switch?
2. If we can safely remove it (see question 1) - should this patch be separated from current changes with automatic selection of next property or they should be in a single commit and single pull request?

bjorn:
> Can we safely remove one case from switch?

Well, the reason the `QEvent::ShortcutOverride` is there is because otherwise the Delete action from the menu will take preference (if it is enabled, like with a tile or object selection active). So we can't remove that. Yet, just removing `QEvent::KeyPress` doesn't solve the problem because the shortcut override event still gets delivered twice for some reaon.

I pushed a fix which I think is the correct way (though I could find no real documentation on how ShortcutOverride is supposed to be used) in commit db0f372.

> If we can safely remove it (see question 1) - should this patch be separated from current changes with automatic selection of next property or they should be in a single commit and single pull request?

I can now just accept this pull request. :-)

PR #993, opened 2015-06-21
Bugfix for #992 Property browser improved to be translated immediately.

### LINKED ISSUE #992: Property browser isn't translated immediately.
Steps to reproduce:
1. Just open some map or select some object to see it's properties.
2. Open preferences and change lang.
All GUI elements change their language but properties browser doesn't. It will change the lang only if you'll select any other object but it should do it immediately as other widgets.

IMMZ:
Property browser improved to be translated immediately. Property browser is updated when some object (different from current one) is set. So it's not good to modify setObject(...) and then call it. I decided to use LanguageChange event and split up setObject(...)'s content into some funcs which used shared by setObject(...) and retranslateUi(...). This patch was created to fix #992.

### DIFF
--- a/src/tiled/propertybrowser.cpp
+++ b/src/tiled/propertybrowser.cpp
@@ -107,45 +107,10 @@ void PropertyBrowser::setObject(Object *object)
     if (mObject == object)
         return;
 
-    // Destroy all previous properties
-    mVariantManager->clear();
-    mGroupManager->clear();
-    mPropertyToId.clear();
-    mIdToProperty.clear();
-    mNameToProperty.clear();
-    mCustomPropertiesGroup = 0;
-
+    removeProperties();
     mObject = object;
 
-    if (!mObject)
-        return;
-
-    mUpdating = true;
-
-    // Add the built-in properties for each object type
-    switch (object->typeId()) {
-    case Object::MapType:               addMapProperties(); break;
-    case Object::MapObjectType:         addMapObjectProperties(); break;
-    case Object::LayerType:
-        switch (static_cast<Layer*>(object)->layerType()) {
-        case Layer::TileLayerType:      addTileLayerProperties();   break;
-        case Layer::ObjectGroupType:    addObjectGroupProperties(); break;
-        case Layer::ImageLayerType:     addImageLayerProperties();  break;
-        }
-        break;
-    case Object::TilesetType:           addTilesetProperties(); break;
-    case Object::TileType:              addTileProperties(); break;
-    case Object::TerrainType:           addTerrainProperties(); break;
-    }
-
-    // Add a node for the custom properties
-    mCustomPropertiesGroup = mGroupManager->addProperty(tr("Custom Properties"));
-    addProperty(mCustomPropertiesGroup);
-
-    mUpdating = false;
-
-    updateProperties();
-    updateCustomProperties();
+    addProperties();
 }
 
 void PropertyBrowser::setMapDocument(MapDocument *mapDocument)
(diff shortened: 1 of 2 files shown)


bjorn:
Hmm, just a small note, you do not need to open an issue about something just to refer to it in a pull request.

src/tiled/propertybrowser.cpp
```diff
@@ -857,6 +835,55 @@ QtVariantProperty *PropertyBrowser::createProperty(PropertyId id, int type,
     return property;
 }
 
+void PropertyBrowser::addNewProperties(Object* object) {
```
bjorn: Please keep in mind the coding style used in this project, which places the opening `{` for functions on the next line.

src/tiled/propertybrowser.h
```diff
@@ -151,8 +154,11 @@ private slots:
                                       const QString &name,
                                       QtProperty *parent);
 
+    void addNewProperties(Object* object);
+    void removeCurrentProperties();
```
bjorn: Naming-wise, I'd call these functions simply `addProperties` and `removeProperties`. The "new" and "current" additions are not adding any information. Also, the `addNewProperties` function does not need to take the `Object*` parameter, but should just use `mObject`.

src/tiled/propertybrowser.cpp
```diff
@@ -107,43 +107,13 @@ void PropertyBrowser::setObject(Object *object)
...
-
-    // Add a node for the custom properties
-    mCustomPropertiesGroup = mGroupManager->addProperty(tr("Custom Properties"));
-    addProperty(mCustomPropertiesGroup);
-
-    mUpdating = false;
-
+    addNewProperties(object);
     updateProperties();
     updateCustomProperties();
```
bjorn: I'm pretty sure these two function calls should be moved to the `addNewProperties` function, to make sure the properies have the current values and that the custom properties are still there, after the language has been changed.

src/tiled/propertybrowser.cpp
```diff
@@ -857,6 +835,55 @@ QtVariantProperty *PropertyBrowser::createProperty(PropertyId id, int type,
...
+    case Object::TilesetType:
+        addTilesetProperties();
+        break;
+    case Object::TileType:
+        addTileProperties();
+        break;
+    case Object::TerrainType:
+        addTerrainProperties();
+        break;
+    }
```
bjorn: Please keep the original formatting for this whole switch block. I had it nicely aligned.

IMMZ:
Done. Sorry for code style breaking - I've forgotten that Eclipse CDT formatter works with me in co-op:)

IMMZ:
Done. Sorry for code style breaking - I've forgotten that Eclipse CDT formatter works with me in co-op:)

src/tiled/propertybrowser.cpp
```diff
@@ -857,6 +833,47 @@ QtVariantProperty *PropertyBrowser::createProperty(PropertyId id, int type,
...
+        }
+        break;
+    case Object::TilesetType:           addTilesetProperties(); break;
+    case Object::TileType:              addTileProperties(); break;
+    case Object::TerrainType:           addTerrainProperties(); break;
+    }
+
+    // Add a node for the custom properties
+    mCustomPropertiesGroup = mGroupManager->addProperty(
+            tr("Custom Properties"));
```
bjorn: I'd prefer to not wrap this line (I realize it's technically one character too long to fit in a terminal, but I had decided not to wrap it before).

src/tiled/propertybrowser.cpp
```diff
@@ -107,45 +107,13 @@ void PropertyBrowser::setObject(Object *object)
...
-    mPropertyToId.clear();
-    mIdToProperty.clear();
-    mNameToProperty.clear();
-    mCustomPropertiesGroup = 0;
-
+    removeProperties();
     mObject = object;
 
     if (!mObject)
         return;
```
bjorn: Could you move this check also into `addProperties`? Then it can also be removed in `retranslateUi`.

IMMZ:
Done.

bjorn:
Great, thanks!

PR #996, opened 2015-06-23
Added 'File => Save All' functionality.

IMMZ:
Useful feature to save all opened documents at once. Useful when game developer opens all his game's levels in an editor then edit them in similar way and wants to save them with one click.

### DIFF
--- a/src/tiled/mainwindow.cpp
+++ b/src/tiled/mainwindow.cpp
@@ -293,6 +293,7 @@ MainWindow::MainWindow(QWidget *parent, Qt::WindowFlags flags)
             SLOT(clearRecentFiles()));
     connect(mUi->actionSave, SIGNAL(triggered()), SLOT(saveFile()));
     connect(mUi->actionSaveAs, SIGNAL(triggered()), SLOT(saveFileAs()));
+    connect(mUi->actionSaveAll, SIGNAL(triggered()), SLOT(saveAll()));
     connect(mUi->actionExportAsImage, SIGNAL(triggered()), SLOT(exportAsImage()));
     connect(mUi->actionExport, SIGNAL(triggered()), SLOT(export_()));
     connect(mUi->actionExportAs, SIGNAL(triggered()), SLOT(exportAs()));
@@ -814,6 +815,29 @@ bool MainWindow::saveFileAs()
     mMapDocument->setWriterPluginFileName(writerPluginFilename);
 
     return saveFile(fileName);
+}
+
+void MainWindow::saveAll()
+{
+    foreach (MapDocument *mapDoc, mDocumentManager->documents()) {
+        if (!mapDoc->isModified())
+            continue;
+
+        QString fileName(mapDoc->fileName());
+        QString error;
+
+        if (fileName.isEmpty()) {
+            mDocumentManager->switchToDocument(mapDoc);
+            if (!saveFileAs())
+                return;
+        } else if (!mapDoc->save(fileName, &error)) {
+            mDocumentManager->switchToDocument(mapDoc);
+            QMessageBox::critical(this, tr("Error Saving Map"), error);
+            return;
+        }
+
+        setRecentFile(fileName);
+    }
 }
 
 bool MainWindow::confirmSave(MapDocument *mapDocument)
@@ -1446,6 +1470,7 @@ void MainWindow::updateActions()
 
     mUi->actionSave->setEnabled(map);
     mUi->actionSaveAs->setEnabled(map);
+    mUi->actionSaveAll->setEnabled(map);
     mUi->actionExportAsImage->setEnabled(map);
     mUi->actionExport->setEnabled(map);
     mUi->actionExportAs->setEnabled(map);
--- a/src/tiled/mainwindow.h
+++ b/src/tiled/mainwindow.h
@@ -119,6 +119,7 @@ public slots:
     void openFile();
     bool saveFile();
     bool saveFileAs();
+    void saveAll();
     void export_(); // 'export' is a reserved word
     void exportAs();
     void exportAsImage();
--- a/src/tiled/mainwindow.ui
+++ b/src/tiled/mainwindow.ui
@@ -21,16 +21,7 @@
   </property>
   <widget class="QWidget" name="centralWidget">
    <layout class="QVBoxLayout" name="verticalLayout">
-    <property name="leftMargin">
-     <number>0</number>
-    </property>
-    <property name="topMargin">
-     <number>0</number>
-    </property>
-    <property name="rightMargin">
-     <number>0</number>
-    </property>
-    <property name="bottomMargin">
+    <property name="margin">
      <number>0</number>
     </property>
    </layout>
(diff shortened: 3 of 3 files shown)


src/tiled/mainwindow.cpp
```diff
@@ -1446,6 +1466,7 @@ void MainWindow::updateActions()
 
     mUi->actionSave->setEnabled(map);
     mUi->actionSaveAs->setEnabled(map);
+    mUi->actionSaveAll->setEnabled(mDocumentManager->documentCount() > 1);
```
bjorn: There is no reason to disable this action when there is only one document. :-)

src/tiled/mainwindow.cpp
```diff
@@ -816,6 +817,25 @@ bool MainWindow::saveFileAs()
...
+        QString error;
+        if (!mapDoc->save(fileName, &error)) {
+            QMessageBox::critical(this, tr("Error Saving Map"), error);
+            return false;
+        }
+
+        setRecentFile(fileName);
+    }
+
+    return true;
```
bjorn: I think you can remove the bool return value. Nobody cares if it worked out.

src/tiled/mainwindow.cpp
```diff
@@ -816,6 +817,25 @@ bool MainWindow::saveFileAs()
     return saveFile(fileName);
 }
 
+bool MainWindow::saveAll()
+{
+    foreach (MapDocument *mapDoc, mDocumentManager->documents()) {
+        QString fileName(mapDoc->fileName());
+        if (fileName.isEmpty())
```
bjorn: This is the case for new documents that were not saved yet, right? I think in this case, we should probably switch the current document and pop up a dialog to choose the file name (ie. call `saveFileAs`).

IMMZ:
Done.

IMMZ:
Now 'saveFileAs' called for new files and 'saveFile' for exisiting ones.

src/tiled/mainwindow.cpp
```diff
@@ -816,6 +817,27 @@ bool MainWindow::saveFileAs()
...
 }
 
+void MainWindow::saveAll()
+{
+    foreach (MapDocument *mapDoc, mDocumentManager->documents()) {
+        QString fileName(mapDoc->fileName());
+        bool successfullSave = true;
+        QString error;
+        if (fileName.isEmpty()) {
+            successfullSave = saveFileAs();
```
bjorn: You can't just call `saveFileAs`, you first need to set the `mapDoc` as the current document using `mDocumentManager->switchToDocument(mapDocument);`.

src/tiled/mainwindow.cpp
```diff
@@ -816,6 +817,27 @@ bool MainWindow::saveFileAs()
...
+            error = tr("Target filename is empty.");
```
bjorn: This error makes little sense. When the `saveFileAs` does not return true, the user already got an appropriate error message or pressed cancel, in which case the `saveAll` action should just abort.

src/tiled/mainwindow.cpp
```diff
@@ -816,6 +817,27 @@ bool MainWindow::saveFileAs()
...
+        } else
+            successfullSave = mapDoc->save(fileName, &error);
```
bjorn: Please do include the `{` and `}` in such cases. Qt style is to leave them out consistently for whole the `if/else` chain, or to include all of them when necessary.

src/tiled/mainwindow.cpp
```diff
@@ -816,6 +817,27 @@ bool MainWindow::saveFileAs()
...
+
+        if (!successfullSave) {
+            QMessageBox::critical(this, tr("Error Saving Map"), error);
+            continue;
```
bjorn: It would probably be better to abort on error.

IMMZ:
Done.

bjorn:
You have to give me some time to review. :-)

src/tiled/mainwindow.cpp
IMMZ: Abort full cycle?

bjorn: Yes, when there is an error, abort the save-all action.

IMMZ: Done.

IMMZ:
It seems that I added an icon (similar as 'save file' in Qt Designer. Remove it?

IMMZ:
Removed icon. I added it mistakenly.

bjorn:
Ok. Did you miss my earlier comments about the error message and the coding style?

IMMZ:
No, I didn't. I got all messages to email. Fixed code style issues.

src/tiled/mainwindow.cpp
```diff
@@ -816,6 +817,25 @@ bool MainWindow::saveFileAs()
...
+    foreach (MapDocument *mapDoc, mDocumentManager->documents()) {
+        QString fileName(mapDoc->fileName());
+        QString error;
+
+        if (fileName.isEmpty()) {
+            mDocumentManager->switchToDocument(mapDoc);
+            if (!saveFileAs())
+                return;
+        } else if (!mapDoc->save(fileName, &error)) {
+            QMessageBox::critical(this, tr("Error Saving Map"), error);
```
bjorn: I guess before showing this error, it also makes sense to switch to this document.

bjorn:
One problem found while testing: the action has the same shortcut (Ctrl+Shift+S) as the "Save As..." action. I guess we should decide where this shortcut should go.

IMMZ:
As I know usually programs don't set up shortcut for 'Save as' and use 'Ctrl+Shift+S' for save all but last word is on you as author of Tiled. If follow by GIMP I should change the shortcut for Save all or remove it...

bjorn:
Let's leave the shortcut for "Save All" empty for now. I checked GIMP, Inkscape and Sublime Text and all of them used Ctrl+Shift+S for "Save As". In the long run, shortcuts should of course be user-customizable.

IMMZ:
Ok. Fixed with no shortcut.

src/tiled/mainwindow.cpp
```diff
@@ -816,6 +817,25 @@ bool MainWindow::saveFileAs()
     return saveFile(fileName);
 }
 
+void MainWindow::saveAll()
+{
+    foreach (MapDocument *mapDoc, mDocumentManager->documents()) {
+        QString fileName(mapDoc->fileName());
+        QString error;
+        mDocumentManager->switchToDocument(mapDoc);
```
bjorn: Ah, please don't switch to the document unnecessarily. This may cause UI flickering and will make things slower since there's a lot happening when switching documents. You just need to move this line into the two conditional branches below.

IMMZ:
Is this variant OK?

bjorn:
No, this variant does not work, since you need to do the switch _before_ calling into `saveFileAs` or showing the message box.

IMMZ:
Fixed.

bjorn:
Hmm, one thing I was just wondering. Is 'Save All' really expected to blindly save all files, or should it be a little smarter and only save files that have unsaved changes?

bjorn:
I just checked with Sublime Text, and that one only saves the files that have unsaved changes.

IMMZ:
No problems, now Tiled acts the same way as Sublime Text.

bjorn:
Awesome. Sorry for all the back and forth, but I think it's perfect now! :-)

PR #998, opened 2015-06-24
Build fixes

jakepetroules:
You can now build the NSIS installer using `qbs -p dist` or `qbs --all-products`

It will not be built during a default invocation of Qbs (think `make dist` vs `make`/`make all`)

### DIFF
--- a/dist/win/dist.qbs
+++ b/dist/win/dist.qbs
@@ -1,0 +1,36 @@
+import qbs
+import qbs.FileInfo
+
+NSISSetup {
+    builtByDefault: false
+    condition: qbs.toolchain.contains("mingw")
+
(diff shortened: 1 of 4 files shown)


bjorn:
That looks really awesome, thanks!

bjorn:
Seems I merged this one too quick:

> ERROR: C:\Qt\Tools\QtCreator\share\qtcreator\qbs\share\qbs\modules\cpp\GenericGCC.qbs:94 product.version must be a string in the format x[.y[.z[.w]] where each component is an integer
Command exited with code 1

bjorn:
This is because `BUILD_INFO_VERSION` is not always a version string with that format. I've tried to push a fix in 51b06cc88f4c32c72619e074ade18947757af39e.

PR #1003, opened 2015-06-25
Fixed comment on magic wand tool.

IMMZ:
Fixed comment on magic wand tool.


bjorn:
Hmm, I would just have included this in your same-tile selection tool pull request, but alright. :-)

Issue #941, opened 2015-05-12
Allow setting the working directory for commands


bjorn:
It would be nice if the working directory could be specified for the custom commands.

Suggested magic variables:
- `%executablePath` - the path of the executable that's supposed to be run
- `%mapPath` - the path of the current map file

Originally requested at issue #940.

IMMZ:
Implementation is on the way.

**Questions**:
1. @bjorn, How to get '%executablepath%' if user only specified binary file without full path (for example, 'geany %mapfile')?
2. Has 'xterm' some parameter to specify working dir? Can't google some...

bjorn:
> How to get '%executablepath%' if user only specified binary file without full path (for example, 'geany %mapfile')?

One way would be to search the directories in `PATH` for that executable, but this is tricky and depends on the operating-system. So I'd suggest that in this case, `%executablePath` is simply an empty string.

> Has 'xterm' some parameter to specify working dir? Can't google some...

Why would you need this? You're supposed to set the working directory using [`QProcess::setWorkingDirectory`](http://doc.qt.io/qt-4.8/qprocess.html#setWorkingDirectory).

IMMZ:
I thought with "QProcess::setWorkingDirectory(...)' xterm itself runs with working directory, not the process set by user, so it's possible it's not true and I didn't touch 'terminal running' code.

bjorn:
> I thought with "QProcess::setWorkingDirectory(...)' xterm itself runs with working directory, not the process set by user, so it's possible it's not true and I didn't touch 'terminal running' code.

That might very well be and depends on the behavior of xterm, but it's outside the scope of Tiled to resolve such an issue.

bjorn:
Hmm, sorry, I just realized there is an "Execute in Terminal" option in the context menu, though it currently only shows up with Qt 4. I'll push a fix to bring it back for Qt 5.

bjorn:
@IMMZ I quickly checked, and the working directory of the process run with `xterm -e` is matching with the working directory in which `xterm` was run. Just check for example the output in the window when doing:

    cd /tmp
    xterm -hold -e pwd

So at least when using `xterm`, there is no problem here.

zcabjro:
Any reason this wouldn't work for map path? (command.cpp)
```C++

QString Command::finalCommand() const
{
    QString finalCommand = command;

    // Perform variable replacement
    if (MapDocument *mapDocument = DocumentManager::instance()->currentDocument()) {
[...]

PR #1004, opened 2015-06-27
Implemented working directory support for commands.

IMMZ:
Implemented a feature requested in #941.

### DIFF
--- a/src/tiled/command.cpp
+++ b/src/tiled/command.cpp
@@ -22,6 +22,7 @@
 #include "documentmanager.h"
 #include "mapdocument.h"
 
+#include <QDir>
 #include <QMessageBox>
 #include <QSettings>
 
@@ -50,6 +51,46 @@ QString Command::finalCommand() const
     return finalCommand;
 }
 
+QString Command::finalWorkingDir() const
+{
+    QString finalWorkingDir = workingDir;
+
+    MapDocument *mapDoc = DocumentManager::instance()->currentDocument();
+    if (mapDoc) {
+        const QString mapPath = mapDoc->path();
+        finalWorkingDir.replace(QLatin1String("%executablepath"),
+                                QLatin1String("\"%1\"")).arg(executablePath());
+        finalWorkingDir.replace(QLatin1String("%mappath"),
+                                QLatin1String("\"%1\"")).arg(mapPath);
+
+    }
+    return finalWorkingDir;
+}
+
+QString Command::executableFileName() const
+{
+    QString finalCmd = finalCommand();
+    QString executableFilename;
+    QStringList stringsToCheck;
+    stringsToCheck << QLatin1String("'") << QLatin1String("\"") << QLatin1String("\\\"");
+    foreach (const QString &s, stringsToCheck) {
+        if (finalCmd.contains(s)) {
+            executableFilename = finalCmd.mid(1, finalCmd.lastIndexOf(s));
+            return executableFilename;
+        }
+    }
+    executableFilename = finalCmd.mid(0, finalCmd.indexOf(QLatin1String(" ")));
+    return executableFilename;
+}
+
+QString Command::executablePath() const
+{
+    QString executableFilename(executableFileName());
+    if (executableFilename.contains(QDir::separator()))
+        return QFileInfo(executableFilename).path();
+    return QString();
+}
+
 void Command::execute(bool inTerminal) const
 {
     // Save if save option is unset or true
@@ -71,6 +112,7 @@ QVariant Command::toQVariant() const
     hash[QLatin1String("Enabled")] = isEnabled;
     hash[QLatin1String("Name")] = name;
     hash[QLatin1String("Command")] = command;
+    hash[QLatin1String("WorkingDir")] = workingDir;
     return hash;
 }
 
@@ -81,6 +123,7 @@ Command Command::fromQVariant(const QVariant &variant)
     const QString namePref = QLatin1String("Name");
     const QString commandPref = QLatin1String("Command");
     const QString enablePref = QLatin1String("Enabled");
+    const QString workingDirPref = QLatin1String("WorkingDir");
 
     Command command;
     if (hash.contains(enablePref))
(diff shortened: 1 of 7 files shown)


src/tiled/command.cpp
```diff
@@ -50,6 +51,34 @@ QString Command::finalCommand() const
     return finalCommand;
 }
 
+QString Command::finalWorkingDir() const
+{
+    QString finalWorkingDir;
```
bjorn: This can't be right, it looks like `finalWorkingDir` will always return an empty string.

src/tiled/mapdocument.cpp
```diff
@@ -197,6 +198,11 @@ MapDocument *MapDocument::load(const QString &fileName,
     return mapDocument;
 }
 
+QString MapDocument::path() const
+{
+    return mFileName.mid(0, mFileName.lastIndexOf(QDir::separator()));
```
bjorn: Please use `QFileInfo(mFileName).path()` instead.

src/tiled/command.cpp
```diff
@@ -50,6 +51,34 @@ QString Command::finalCommand() const
...
+    }
+    return finalWorkingDir;
+}
+
+QString Command::executablePath() const
+{
+    QString finalcommand = finalCommand();
+
+    // removing parameters from command string
+    QString executableFilename(finalcommand.mid(0, finalcommand.lastIndexOf(QLatin1String(" ")) - 1));
```
bjorn: The beginning of the parameters is assumed to be the last space character? That would only cover the single-parameter case. I'm not sure how this can be done properly... maybe executable and parameters should be separated already in the UI?

src/tiled/command.cpp
```diff
@@ -50,6 +51,34 @@ QString Command::finalCommand() const
...
+
+    if (executableFilename.contains(QDir::separator()))
+        return executableFilename.mid(0, executableFilename.lastIndexOf(QDir::separator()));
```
bjorn: Here too you should use `QFileInfo(executableFilename).path()`.

src/tiled/command.cpp
IMMZ: Parameters always should start after a space. So there are possible spaces in the path to executable itself and in those cases path should be surrounded with ' / ". Maybe next strategy won't be tricky?
See if the string contains ' / ". YES => use last " as executable path. NO => get substring until first space.
Also we can try to create some validator for string field to replace user's string to some parseable format. I don't sure too what way would be good here...

bjorn: Of course handling quotes is a little bit more tricky, since you'd also need to handle `\"` within the executable path. Still of course a small function could be written for this.

src/tiled/command.cpp
```diff
@@ -50,6 +51,46 @@ QString Command::finalCommand() const
...
+{
+    QString finalWorkingDir = workingDir;
+
+    MapDocument *mapDoc = DocumentManager::instance()->currentDocument();
+    if (mapDoc) {
+        const QString mapPath = mapDoc->path();
+        finalWorkingDir.replace(QLatin1String("%executablepath"),
+                                QLatin1String("\"%1\"")).arg(executablePath());
+        finalWorkingDir.replace(QLatin1String("%mappath"),
+                                QLatin1String("\"%1\"")).arg(mapPath);
```
bjorn: Hmm, did you actually test this? I tried out your patch and couldn't get it to work, and it's because this code doesn't actually put any executable or map path into the final working directory. The `QString::arg` function returns a copy rather than doing an in-place replacement.

I think you shouldn't use `QString::arg` here anyway. One problem is that when for example no `%executablepath` is present in the working directory, `arg` will complain to standard error about a missing argument since it will be called on a string without `%1`. But also it is simply not needed. Just use `QString::replace` to replace the variables directly:

```cpp
        finalWorkingDir.replace(QLatin1String("%executablepath"), executablePath());
        finalWorkingDir.replace(QLatin1String("%mappath"), mapPath);
```

src/tiled/command.cpp
```diff
@@ -50,6 +51,46 @@ QString Command::finalCommand() const
...
+    }
+    executableFilename = finalCmd.mid(0, finalCmd.indexOf(QLatin1String(" ")));
+    return executableFilename;
+}
+
+QString Command::executablePath() const
+{
+    QString executableFilename(executableFileName());
+    if (executableFilename.contains(QDir::separator()))
+        return QFileInfo(executableFilename).path();
```
bjorn: I don't think this check is necessary. I tried the case without separator and it seemed to return ".", which just means current directory and would be a correct return value in this case. So you can write this function like:
```c++
QString Command::executablePath() const
{
    return QFileInfo(executableFileName()).path();
}
```

src/tiled/commanddatamodel.cpp
```diff
@@ -250,7 +257,9 @@ Qt::ItemFlags CommandDataModel::flags(const QModelIndex &index) const
     Qt::ItemFlags f = QAbstractTableModel::flags(index);
 
     if (isNormalRow) {
-        f |= Qt::ItemIsDragEnabled | Qt::ItemIsDropEnabled;
+        // no reasons to support support 'Drag&Drop' for working dir
+        if (index.column() != WorkingDirColumn)
+            f |= Qt::ItemIsDragEnabled | Qt::ItemIsDropEnabled;
```
bjorn: Why this? I think it's a strange omission and would rather see `dropMimeData` be fixed to pass the working directory to the copy.

src/tiled/command.cpp
```diff
@@ -50,6 +51,46 @@ QString Command::finalCommand() const
...
+
+QString Command::executableFileName() const
+{
+    QString finalCmd = finalCommand();
+    QString executableFilename;
+    QStringList stringsToCheck;
+    stringsToCheck << QLatin1String("'") << QLatin1String("\"") << QLatin1String("\\\"");
+    foreach (const QString &s, stringsToCheck) {
+        if (finalCmd.contains(s)) {
+            executableFilename = finalCmd.mid(1, finalCmd.lastIndexOf(s));
```
bjorn: I understand the idea, but this implementation has several problems.

* Since you use `contains`, you may arbitrarily cut off the first character of the executable name if any argument is quoted. Use `startsWith` instead. You'll probably want to use `trimmed` on the `finalCmd` as well.
* Even if the found quote was the first, you're then taking the string until the last quote. But you should take the next quote instead, so using `indexOf` and passing the right `from` value.
* The final quote you're searching for is `"\\\""`, which has two characters. As such, it would be incorrect to always cut only a single starting character when using `mid` on the `finalCmd`.

Issue #1005, opened 2015-06-27
Layers moving button is updated incorrectly.


IMMZ:
Steps to reproduce:
1. Create map.
2. Create layer A. Buttons 'move up / move down' are disabled.
3. Create layer B.
4. Choose upper one (B) and delete it.
5. Now the bottom one (A) is active but 'move up' button is active this time with only single layer on the map which is not correct behaviour.

bjorn:
Good catch. Fortunately, it does nothing when you click it. :-)

Are you working on a fix?

IMMZ:
Not yet :) Just left it up here, maybe someone wants to accept a challenge.

PR #1006, opened 2015-06-28
Fixed issue #1005.

IMMZ:
Fortunately, I found time to look into #1005. Not sure is it a correct fix. In updateActions() at the moment of deleting mapDocument still returns 2 layers (when should be one), so the guaranteed deletion would be after layerRemoved signal as I think.

### DIFF
--- a/src/tiled/mapdocumentactionhandler.cpp
+++ b/src/tiled/mapdocumentactionhandler.cpp
@@ -175,6 +175,8 @@ void MapDocumentActionHandler::setMapDocument(MapDocument *mapDocument)
     updateActions();
 
     if (mMapDocument) {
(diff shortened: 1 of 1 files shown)


bjorn:
The fix looks fine to me, but could you use a more descriptive commit message? Rather than just saying "Fixed issue #1005.", the message should look more like this:

```
Fixed enabled state of 'move layer up' action

After deleting the top-most layer, the 'move layer up' action was not 
set to disabled like it should.

Fixes #1005
```

The main thing being, that the fact that it fixes some github issue is a minor detail, while the other stuff is the relevant information to include about the change.

IMMZ:
Fixed.

Issue #997, opened 2015-06-23
Tileset properties doesn't list margin and spacing


AMDmi3:
And actually more properties which are only available on tileset creation.

bjorn:
Right, they could be added as read-only for now, until  #773 is implemented.

bjorn:
@IMMZ Since you were looking for more things to do, maybe this would be something for you. Note that for now I'd prefer them to be just read-only, since before allowing the user to change them I want to reconsider the design a bit.

The complete list of missing properties is:

* Source Image
* Tile Size
* Margin
* Spacing
* Transparent Color (though you can skip this one, since it's covered by #983)

Also, please don't add these properties for the case of a 'Collection of Images' tileset, since in that case they're all irrelevant. You can recognize this case by whether the `Tileset::imageSource()` is an empty string.

IMMZ:
@bjorn Implementation should add 5 properties to the properties dock for a tileset and all of them read-only?

bjorn:
@IMMZ Yes, almost. It will be 4 properties once you skip the one I said you can skip. :-)

IMMZ:
@bjorn Sorry, my fault :) Ok, I'll look at it this week.

PR #1009, opened 2015-06-29
Added new read-only properties for tilesets.

IMMZ:
Pull request for #997 
Added four read-only properties:
1. Source Image;
2. Tile Size;
3. Margin;
4. Spacing;
These properties are added only if the tileset isn't the type of 'Collection of images'.

### DIFF
--- a/src/tiled/propertybrowser.cpp
+++ b/src/tiled/propertybrowser.cpp
@@ -494,6 +494,21 @@ void PropertyBrowser::addTilesetProperties()
     QtProperty *groupProperty = mGroupManager->addProperty(tr("Tileset"));
     createProperty(NameProperty, QVariant::String, tr("Name"), groupProperty);
     createProperty(TileOffsetProperty, QVariant::Point, tr("Drawing Offset"), groupProperty);
+
+    // Next properties we should add only for non 'Collection of Images' tilesets
+    const Tileset *currentTileset = dynamic_cast<const Tileset*>(mObject);
+    if (!currentTileset->imageSource().isEmpty()) {
+        QtVariantProperty *srcImgProperty =
+                createProperty(SourceImageProperty, QVariant::String, tr("Source Image"), groupProperty);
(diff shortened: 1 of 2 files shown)


src/tiled/propertybrowser.cpp
```diff
@@ -494,6 +494,24 @@ void PropertyBrowser::addTilesetProperties()
     QtProperty *groupProperty = mGroupManager->addProperty(tr("Tileset"));
     createProperty(NameProperty, QVariant::String, tr("Name"), groupProperty);
     createProperty(TileOffsetProperty, QVariant::Point, tr("Drawing Offset"), groupProperty);
+
+    // Next properties we should add only for non 'Collection of Images' tilesets
+    if (mObject->typeId() != Object::TilesetType)
+        return;
```
bjorn: If you're going to return here, be sure to move the `addProperty(groupProperty);` up so that it still gets executed.

bjorn: Wait a second, this whole check should be removed. In the `addTilesetProperties` function, you can already safely assume that the `mObject` is a `Tileset` instance.

IMMZ:
Done.

src/tiled/propertybrowser.cpp
```diff
@@ -946,6 +961,10 @@ void PropertyBrowser::updateProperties()
         const Tileset *tileset = static_cast<const Tileset*>(mObject);
         mIdToProperty[NameProperty]->setValue(tileset->name());
         mIdToProperty[TileOffsetProperty]->setValue(tileset->tileOffset());
+        mIdToProperty[SourceImageProperty]->setValue(tileset->imageSource());
+        mIdToProperty[TileSizeProperty]->setValue(tileset->tileSize());
+        mIdToProperty[MarginProperty]->setValue(tileset->margin());
+        mIdToProperty[SpacingProperty]->setValue(tileset->tileSpacing());
```
bjorn: You'll run into a crash if you conditionally create these properties but unconditionally set their value. Please do test your change. :-)

IMMZ:
Sh#&. I need some tester :)  Fixed.

bjorn:
Thanks. :-)

PR #1010, opened 2015-06-29
Animation driver TODO solved.

IMMZ:
Now animation driver isn't started when there are no animated tiles. There is also added a check that driver isn't running already to avoid extra searching of animated tiles each times user calls 'mAnimationDriver->start()' more than once without correspondend 'stop()'.
**Warning**: due to technical problems (Very old PC) the patch wasn't tested.


bjorn:
You can't implement it like that because the driver would fail to run once you add some animation.

I would leave this TODO for now, since there are many ways to go here and it will need some careful consideration.

IMMZ:
OK.

PR #1011, opened 2015-06-29
FIXME solved. Refactored disabling of dragging directories in the maps dock.

IMMZ:
FIXME solved. Now dragging of directories in the maps dock is disabled in correctly way via creating a new model derived from 'QFileSystemModel' and reimplementing its 'flags(...)'.

### DIFF
--- a/src/tiled/mapsdock.cpp
+++ b/src/tiled/mapsdock.cpp
@@ -41,6 +41,26 @@
 using namespace Tiled;
 using namespace Tiled::Internal;
 
+/**
+ * Class represents the file system model with disabled dragging of directories.
+ */
+class FileSystemModel : public QFileSystemModel
+{
+public:
+    explicit FileSystemModel(QObject *parent = 0):
+        QFileSystemModel(parent)
+    {
+    }
+
+    Qt::ItemFlags flags(const QModelIndex &i) const override
+    {
+        Qt::ItemFlags flags = QFileSystemModel::flags(i);
+        if (isDir(i))
+            flags &= ~Qt::ItemIsDragEnabled;
+        return flags;
+    }
+};
+
 MapsDock::MapsDock(MainWindow *mainWindow, QWidget *parent)
     : QDockWidget(parent)
     , mDirectoryEdit(new QLineEdit)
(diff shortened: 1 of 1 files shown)


src/tiled/filesystemmodel.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+FileSystemModel::FileSystemModel(QObject *parent):
+    QFileSystemModel(parent)
+{
+}
+
+Qt::ItemFlags FileSystemModel::flags(const QModelIndex &i) const
+{
+    Qt::ItemFlags flags = QFileSystemModel::flags(i);
+    if (this->isDir(i))
```
bjorn: No need for `this->` here.

src/tiled/filesystemmodel.h
```diff
@@ -0,0 +1,43 @@
...
+
+#include <QFileSystemModel>
+
+namespace Tiled {
+namespace Internal {
+
+/**
+ * Class represents the file system model with disabled dragging of directories.
+ */
+class FileSystemModel: public QFileSystemModel
```
bjorn: Rather than putting this in its own set of files, I'd just have this class in `mapsdock.cpp`. There is no need to change `mapsdock.h` even, since the variable can stay a `QFileSystemModel*`.

bjorn:
Looks fine otherwise.

IMMZ:
Fixed.

src/tiled/mapsdock.cpp
```diff
@@ -41,6 +41,26 @@
...
+ */
+class FileSystemModel: public QFileSystemModel
+{
+public:
+    explicit FileSystemModel(QObject *parent = 0):
+        QFileSystemModel(parent)
+    {
+    }
+
+    virtual Qt::ItemFlags flags(const QModelIndex &i) const
```
bjorn: Ah, instead of repeating the `virtual` keyword, which does nothing because the method is already virtual, it would be more useful to mark this method at the end with `override`, which makes the compiler verify that it's overriding a virtual method.

src/tiled/mapsdock.cpp
```diff
@@ -41,6 +41,26 @@
 using namespace Tiled;
 using namespace Tiled::Internal;
 
+/**
+ * Class represents the file system model with disabled dragging of directories.
+ */
+class FileSystemModel: public QFileSystemModel
```
bjorn: Please put a space in front of the colon here.

IMMZ:
Done.

bjorn:
Great, thanks! :-)

PR #1012, opened 2015-06-30
Refresh assets in collection of Tilesets

EdenIndustries:


bjorn:
Please instead of adding another commit that reverts the previous commit, just remove the commit and force-push to your master branch again. If you need some instructions on how to do this with git just say so.

If you want to open multiple pull requests for independent features, you should create multiple branches for them in your fork.

About the actual changes, could you make sure it follows the same coding style as the existing code? This means mostly:

* Not indenting with tabs but with 4 spaces
* Don't put "Get" in front of a getter method (and start them with a lowercase character)
* Put a space between `for` and `(`.

This will make the code easier to read for me.

src/libtiled/mapreader.cpp
```diff
@@ -53,6 +53,9 @@ using namespace Tiled::Internal;
 namespace Tiled {
 namespace Internal {
 
+	QString mTilesetSourceFiles[MAX_TILESET_SOURCE_FILES];
+	int mNumTilesetSourceFiles;
```
bjorn: I use the `m` prefix for member variables, but you have just made these ones globals. Were they meant to be member variables?

Also, you could probably just merge these two variables into a single `QStringList`.

src/libtiled/mapreader.h
```diff
@@ -95,6 +97,9 @@ class TILEDSHARED_EXPORT MapReader
      */
     QString errorString() const;
 
+	int GetNumTilesetSourceFiles();
+	QString* GetTilesetSourceFiles();
```
bjorn: Since these functions are not modifying the object, they should be marked as `const`.

bjorn:
Design-wide, this approach isn't very nice.

There is no need to keep track of the tile source files explicitly in the `MapReader`, since this information will already be available by just iterating of the `Tile` instances in the `Tileset` later.

Also, the `TmxMapReader` should not go and call into the tileset manager to add these references, because that way we don't know when we can stop watching the files. It also is simply not the job of this class. Instead, the `TilesetManager` can start watching these files itself when a new tileset is added to it (`TilesetManager::addReference`) and then it can also stop watching these files once the last reference goes away (in `TilesetManager::removeReference`).

bjorn:
We'll also need some solution for starting to watch newly added tiles. Possibly some explicit `tilesAdded(const QList<Tile *> &tiles)` and similar `tilesRemoved` kind of calls could be made to the `TilesetManager` from the `AddRemoveTiles` undo command.

src/libtiled/mapreader.h
IMMZ: Also, according to qt and Tiled styles, these funcs should start with lowercase and without 'get' prefix.

bjorn: Well, I had noted that already in another comment. But actually all this code is simply not necessary, as I wrote in later comments. So it should just be removed rather than fixed.

PR #1013, opened 2015-07-01
Let travis build with g++ 4.8

Ablu:
lets see if this works

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,11 @@ matrix:
           env: QMAKESPEC=unsupported/linux-clang
 
 before_install:
+  - sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
   - sudo apt-get update -qq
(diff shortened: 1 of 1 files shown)


Ablu:
Giving up for now. Running out of time. Need to look at it later again.

bjorn:
Thanks for trying, @Ablu!

bjorn:
What about, instead of trying to tell qmake to use different compiler executable names, you'd tell Ubuntu to change the gcc symlink to the newer version? I think there's some update-alternatives thingy for this.

Ablu:
Ok this should succeed now once `-std=c++11x` is specified for qt4 builds (qt4's qmake does not support `CONFIG += c++11`

bjorn:
Yeah, i can fix that, but I hope you can still need to make it not break the clang build. :-)

Ablu:
Fixed :)

bjorn:
Awesome! :-)

PR #1014, opened 2015-07-04
AppVeyor: install NSIS using the Chocolatey package manager.

jakepetroules:
Another small improvement to CI config. I'm working on getting Qbs available through Chocolatey as well so then you'll be able to easily use a newer version without relying on Qt Creator.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,8 +7,7 @@ environment:
 configuration: Release
 
 install:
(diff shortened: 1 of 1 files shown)


jakepetroules:
Also, note that the latest Qbs is available through Homebrew, and Travis supports OS X and has Homebrew pre-installed. ;)

bjorn:
Nice one!

I realized Travis supports OS X but so far I just didn't get around to looking into setting up OS X builds there. Maybe you'll be faster!

PR #1015, opened 2015-07-06
Usability improvement: double click on polygon.

IMMZ:
**Implements**: Changing 'object selection tool' to 'polygon edit tool' when clicking on polygon.
**Implementation details**: When user makes double click on the polygon or the polyline the current tool is changed to '*polygon edit tool*'. Current tool could be only '*object selection tool*' as other tools are used to create objects and we shouldn't prevent user to create something. Also patch includes shape selection after double click for increasing the usability as the shape isn't selected after double click and user should manually select it to be able to change it with new tool.
**Testing**: A base test which could tell us that it **works**. Moreover tested some cases:
* Double click when current tool is not 'object selection tool';
* Double click on point with no shape;
* Double click on non-polygon & non-polyline shapes;
No bugs noticed in all these cases.

### DIFF
--- a/src/tiled/documentmanager.cpp
+++ b/src/tiled/documentmanager.cpp
@@ -386,6 +386,9 @@ void DocumentManager::currentIndexChanged()
         mapScene->setSelectedTool(mSelectedTool);
         mapScene->enableSelectedTool();
         mSceneWithTool = mapScene;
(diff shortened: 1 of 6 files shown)


IMMZ:
Some changes on the lines look strange... Seems like *.patch file issues... or possible something wrong I made with git.

Issue #1016, opened 2015-07-07
Prevent user from creating "nan" polylines through resizing


LearnCocos2D:
I just created a polyline that ended up like so in the TMX file:

```
  <object id="2" x="nan" y="148" width="nan">
   <polyline points="nan,0 nan,-4"/>
  </object>
```

Properties view:

![screen shot 2015-07-07 at 11 32 05](https://cloud.githubusercontent.com/assets/214599/8543251/d5840cac-249b-11e5-96ee-c454ecee2020.png)

While I can select that object from the objects list it can still create a range of problems, specifically at runtime depending on the implementation of the TMX reader/renderer. Since the line isn't shown or selectable on the map it's also not standing out as a potential problem.

How to create a nan polyline? Good question! :)

It took me a moment to reproduce. As prerequisite you need a polyline with 3 points. Then use the select objects tool and select the line. This should bring up the resize arrows on every side and corner. Try to squeeze down the path to a point. Most of the times this will end up drawing a single dot, that's not the problem. The problem appears because in some way (can't really say for sure how) it's possible to drag all the points together to the same location or even "slightly farther" - basically this happened: I had a single line segment left, then dragged the top side down over and under the other remaining segment point when the line vanished entirely and gave nan coords. 

Or perhaps it's just a random floating point inaccuracy that adds up when you keep resizing the line?

Note: I have not tested this but I think it's quite possible this problem could occur with other object types too.

bjorn:
This is probably related to #981. A fix was attempted at #982 but there remained some problems with it and @IMMZ decided to tackle other issues for now. I'll try to fix it for the next release, unless somebody beats me to it.

PR #1018, opened 2015-07-09
Avoid Inf values in scaling factors

Carisius:
This solves #1016 and #981 issues

### DIFF
--- a/src/tiled/objectselectiontool.cpp
+++ b/src/tiled/objectselectiontool.cpp
@@ -1202,8 +1202,12 @@ void ObjectSelectionTool::updateResizingSingleItem(const QPointF &resizingOrigin
         const QPointF relPos = pos - origin;
         const QPointF startDiff = start - origin;
 
-        QSizeF scalingFactor(qMax((qreal)0.01, relPos.x() / startDiff.x()),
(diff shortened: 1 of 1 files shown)


Carisius:
What do you think about this patch? The nan values come when any coordinate of startDiff is 0 (small change?) and dividing by 0 makes that coord of scalingFactor become Inf then other operation on that Inf value, maybe multiplying by 0 is from where the nan comes. I've noticed there is another float division not checked in the same file, in updateResizingItems, I've not changed that one. This patch only avoids odd values result of division by zero with float numbers because those divisions are unchecked. It works, solving the mentioned issues though maybe it would be better to hide the extra sizers as tryed #982, well or maybe not.

bjorn:
Sorry I still need to find some time to review this and consider the options, but I certainly plan do it before releasing Tiled 0.13. Thanks in any case for looking into this!

bjorn:
I think the fix is a good start, but I'll lower the limit a little to avoid preventing resize too soon and I'll apply it also to the case of resizing multiple items, since the same problem exists there.

bjorn:
I decided to do it slightly differently, by checking the values afterwards, see commit a927db8.

Let me know if you see any problems with that approach.

Issue #1019, opened 2015-07-11
Failed to build for DEB


modos189:
I'm trying to make DEB package for 12.2 and 12.3 releases using the command
dpkg-buildpackage -rfakeroot
but I get an error about the lack of library libtiled.so.1, which is present in the debian/tiled-daily/usr/lib/tiled/

On the launchpad repository same error, the build log: https://launchpadlibrarian.net/207253905/buildlog_ubuntu-vivid-amd64.tiled_0.12.2-3_BUILDING.txt.gz

Please tell me how to fix and build :)

modos189:
I corrected the path to the library and created pull: https://github.com/bjorn/tiled/pull/1021

bjorn:
As such that patch is not something I can accept since it would break the library references for the default `LIBDIR` which is simply `$${PREFIX}/lib`.

PR #1021, opened 2015-07-12
Fix library paths

modos189:
More: https://github.com/bjorn/tiled/issues/1019


bjorn:
This is probably something you should do with a patch file, or it should be somehow automatic based on the directory in which the library gets installed. I know Debian likes to put the libtiled.so in a `lib/tiled` subdirectory, but this does not happen by default and for example Arch Linux doesn't do this.

There is technically no reason why this library should not be in `lib` since in principle other applications could be written that use it as well.

modos189:
Thank you for the explanation

PR #1025, opened 2015-07-16
Improved Windows icon

zOadT:
Small cosmetic commit: The tiled-icon scales poorly on windows to resolution 16x16, so I added this resolution (and while I was at it also some other resolutions :D ) to the .ico file. It's still not perfect, but I think it's an improvement.

Note: Windows caches thumbnails, so the new icon may not be shown on systems where the old icon is already present (but you can rename it to see the change)


bjorn:
That's great, thanks!

PR #1029, opened 2015-07-17
Parse objects with floating point positions and dimensions

kinglime:
I changed the object's positions and dimensions from ints to floats. I ran the unit tests and they passed.

### DIFF
--- a/util/java/libtiled-java/src/tiled/core/MapObject.java
+++ b/util/java/libtiled-java/src/tiled/core/MapObject.java
@@ -29,6 +29,7 @@
 package tiled.core;
 
 import java.awt.*;
+import java.awt.geom.Rectangle2D;
 import java.util.Properties;
 import java.io.File;
 import java.io.IOException;
@@ -41,7 +42,7 @@ public class MapObject implements Cloneable
 {
     private Properties properties = new Properties();
     private ObjectGroup objectGroup;
-    private Rectangle bounds = new Rectangle();
+    private Rectangle2D.Double bounds = new Rectangle2D.Double();
     private Shape shape = new Rectangle();
     private String name = "Object";
     private String type = "";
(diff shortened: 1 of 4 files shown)


bjorn:
I'll do a detailed review later, but I just thought I'd point out that double should be fine. Tiled is using double as well internally.

kinglime:
@bjorn Oh ok! I literally had them all as doubles but then for some reason I decided to change them to floats... I think it was because the xml writer would only take a float not a double. Anyways I could back and change them all once again if you want.

bjorn:
> I think it was because the xml writer would only take a float not a double.

Hmm, I guess you are talking about the custom XML writer I wrote. You could just add an override for double to it.

kinglime:
OK. It is late here for me but I will make the changes tomorrow.

kinglime:
@bjorn Sorry for taking so long but I've changed the floats to doubles. Again I ran the tests and they passed.

util/java/libtiled-java/src/tiled/core/MapObject.java
```diff
@@ -151,24 +152,62 @@ public Image getImage(double zoom) {
...
+        } else {
+            if (newv < oldv) {
+                if (bounds.height >= 0) {
+                    bounds.height += newv - Double.MAX_VALUE;
+                    if (bounds.height < 0) bounds.height = Double.MAX_VALUE;
+                }
+                newv = Double.MAX_VALUE;
+            }
+        }
+        bounds.y = newv;
```
bjorn: Why is this so complicated? Why not 
```java
bounds.x += dx;
bounds.y += dy;
```

bjorn: Note that doubles don't wrap around like integers do, so I don't think anything more meaningful can really be done here.

kinglime:
@bjorn Basically Rectangle had a method that 2DRectabgle did not have. I decided to just view the source for that method and copied it/modified it. If it only takes two lines that's great! I guess I was just being cautious.

bjorn:
@kinglime I'm sure Rectangle had that method only for keeping the behavior somewhat stable in the face of integer number wrapping. Since is not needed with floats or doubles, Rectangle2D does not have that function. Will you update the PR? Please squash all changes into a single commit.

kinglime:
@bjorn I will

kinglime:
@bjorn Do I need to make any other changes?

bjorn:
@kinglime No, I think it looks fine now, I just need to get around to run a quick test before integrating it.

kinglime:
@bjorn OK great!

bjorn:
Tested and seems to work fine! Sorry about forgetting to merge this before the Tiled 0.13 release. At least it will go into Tiled 0.14 now.

Thanks!

PR #1046, opened 2015-07-28
Update Czech translations

encukou:
Here are new translations for 0.13!


bjorn:
Looking great, thanks!

PR #1060, opened 2015-08-23
fr translation update.

Bertram25:
With my best regards. :D

### DIFF
--- a/translations/tiled_fr.ts
+++ b/translations/tiled_fr.ts
@@ -77,32 +77,32 @@
     <message>
         <location filename="../src/tiled/main.cpp" line="-172"/>
         <source>Display the version</source>
-        <translation type="unfinished"></translation>
+        <translation>Afficher la version</translation>
     </message>
     <message>
         <location line="+5"/>
         <source>Only check validity of arguments</source>
-        <translation type="unfinished"></translation>
+        <translation>Vérifier seulement la validité des arguments</translation>
     </message>
     <message>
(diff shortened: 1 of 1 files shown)


translations/tiled_fr.ts
```diff
@@ -2595,7 +2596,7 @@ Voulez-vous le remplacer ?</translation>
     <message>
         <location line="-1223"/>
         <source>Ctrl+Shift+O</source>
-        <translation type="unfinished"></translation>
+        <translation>Ctrl+Màj+O</translation>
```
bjorn: Just as a reminder, you should not translate the names of keys in shortcuts, otherwise Qt will not be able to determine which keys are meant. Qt will take care of translating key names. I'll fix this one when cherry-picking the translation update to the 0.13 branch.

bjorn:
Cherry-picked to `0.13` branch as commit 46011b5. Many thanks! :-)

translations/tiled_fr.ts
Bertram25: Truly sorry for that one. I remember seeing several other shortcuts translated and thus thought the rule had changed. Thanks for fixing my mistakes. :)

bjorn: Really no problem! I indeed noticed one earlier "Màj" and have fixed that one as part of the update. If you know about any other translated shortcuts, please do tell or submit a fix. :-)

PR #1072, opened 2015-09-08
Qbs build improvements

jakepetroules:


util/java/libtiled-java/libtiled-java.qbs
```diff
@@ -0,0 +1,16 @@
...
+
+    Group {
+        fileTagsFilter: ["java.jar"]
+        qbs.install: true
+        qbs.installDir: "jar"
+    }
+
+    Export {
+        Depends { name: "java" }
+        java.manifestClassPath: [product.targetName + ".jar"]
```
bjorn: It's great that these .jar files can be compiled using Qbs, but it really shouldn't be required as part of building Tiled. Currently, on a system without a JDK installed, you get:
```
ERROR: /usr/share/qtcreator/qbs/share/qbs/modules/java/JavaModule.qbs:349 The following properties are not set. Set them in your profile or product:
java.jdkPath
java.compilerVersion
java.compilerVersionMajor
java.compilerVersionMinor
java.compilerVersionUpdate
The following properties have invalid values:
java.compilerVersion: must be a version number with 4 to 4 components
java.compilerVersionMajor: must be an integer >= 1
java.compilerVersionMinor: must be an integer >= 0
java.compilerVersionPatch: must be an integer >= 0
java.compilerVersionUpdate: must be an integer >= 0
```
So at the least it should auto-disable based on whether a Java compiler is available.

I'm currently undecided on whether this product should build by default at all, since it is not required by Tiled and it is low quality and out of date. Even if it is built by default, it should probably not have an install rule.

util/java/tmxviewer-java/tmxviewer-java.qbs
```diff
@@ -0,0 +1,15 @@
...
+
+    entryPoint: "TMXViewer"
+
+    files: ["src/**/*.java"]
+
+    Group {
+        fileTagsFilter: ["java.jar"]
+        qbs.install: true
+        qbs.installDir: "jar"
+    }
```
bjorn: The Java-based TMXViewer application is just a small example and the Qt-based tmxviewer is much more suitable to ship as basic TMX viewer (and it already installed). I don't think this one should install anywhere.

src/tmxrasterizer/tmxrasterizer.qbs
```diff
@@ -20,8 +20,10 @@ QtGuiApplication {
     Group {
         qbs.install: true
         qbs.installDir: {
-            if (qbs.targetOS.contains("windows") || qbs.targetOS.contains("osx"))
+            if (qbs.targetOS.contains("windows"))
                 return ""
+            else if (qbs.targetOS.contains("darwin"))
+                return "Tiled.app/Contents/MacOS"
```
bjorn: I guess the same change would apply to the `tmxviewer` and the `automappingconverter`?

src/tmxrasterizer/tmxrasterizer.qbs
```diff
@@ -8,7 +8,7 @@ QtGuiApplication {
     Depends { name: "libtiled" }
 
     cpp.includePaths: ["."]
-    cpp.rpaths: ["$ORIGIN/../lib"]
+    cpp.rpaths: qbs.targetOS.contains("darwin") ? ["@loader_path/../Frameworks"] : ["$ORIGIN/../lib"]
```
bjorn: Since this depends on the install location of `libtiled`, does it make sense to move it to the `Export` part in `libtiled.qbs`? Or is this not possible / desired?

bjorn:
@jakepetroules Thanks a lot for these improvements! I haven't tested it on OS X yet, but I look forward to deprecating `qmake` once all platforms can be built with Qbs!

src/tmxrasterizer/tmxrasterizer.qbs
jakepetroules: It's the responsibility of the caller to specify where it looks for its libraries, not the callee.

The fact that libtiled specifies where it's installed, rather than Tiled.app specifying libtiled as an bundle-embedded library dependency, is currently a limitation of Qbs.

src/tmxrasterizer/tmxrasterizer.qbs
jakepetroules: I guess? I omitted it because I didn't see those two binaries in your normal DMG distribution.

util/java/tmxviewer-java/tmxviewer-java.qbs
jakepetroules: OK.

util/java/libtiled-java/libtiled-java.qbs
jakepetroules: I'll make Java an optional dependency.

jakepetroules:
> @jakepetroules Thanks a lot for these improvements! I haven't tested it on OS X yet, but I look forward to deprecating qmake once all platforms can be built with Qbs!

All platforms can't be built with Qbs yet? Why not? What's missing?

jakepetroules:
I asked AppVeyor to update to Qt Creator 3.5 which includes Qbs 1.4.2 (http://help.appveyor.com/discussions/problems/3023-qt-creator-is-out-of-date)

Alternatively, you may be able to use my Qbs package for Chocolatey but I don't know if there's a way to install unreviewed packages: https://chocolatey.org/packages/qbs

bjorn:
> All platforms can't be built with Qbs yet? Why not? What's missing?

The one you're working on. :-)

Seems AppVeyor is going to upgrade today, so I won't bother to try installing qbs through chocolatey for now.

It could very well be that automappingconverter and tmxviewer were not included in the dmg before. This was simply because I didn't know what would be the proper way to ship these tools.

Btw, what makes the minimal Qbs version 1.4.2?

bjorn:
Hmm, I just tried running Tiled from Qt Creator 3.5 on OS X and it fails to run Tiled.app, even though it does manage to run all the other applications. The error message is:

_Executable /Users/bjorn/Projects/build-tiled-Desktop_Qt_5_4_1_clang_64bit-Debug/qtc_Desktop__cb18690c-debug/install-root/Tiled does not exist._

I am however able to run Tiled using `open /Users/bjorn/Projects/build-tiled-Desktop_Qt_5_4_1_clang_64bit-Debug/qtc_Desktop__cb18690c-debug/install-root/Tiled.app` from the command line. Maybe Qt Creator is confused about whether this product is a bundle or not?

jakepetroules:
That's a known issue - [QBS-858](https://bugreports.qt.io/browse/QBS-858), which [will be fixed](https://codereview.qt-project.org/#/c/123997/) in Qt Creator 3.5.1 and Qbs 1.4.3 (probably due around September - October time frame).

bjorn:
@jakepetroules Alright, thanks!

I just ran into another issue when compiling against Qt 5.5. Now it won't even start from the console either, and instead crashes on startup as it can't find the Qt libraries:

```
Dyld Error Message:
  Library not loaded: @rpath/QtCore.framework/Versions/5/QtCore
  Referenced from: /Users/USER/*/Tiled.app/Contents/MacOS/Tiled
  Reason: image not found
```

jakepetroules:
That's the "limitations of Qbs" I mentioned. For now you can add a Rule that rsync's over the Qt frameworks into your app bundle (or runs macdeployqt, but it currently doesn't support multiple subsequent runs very well).

bjorn:
Hmm, running `macdeployqt` does not seem to be working at least:
```
MacMini:tiled bjorn$ ~/Qt/5.5/clang_64/bin/macdeployqt /Users/bjorn/Projects/build-tiled-Desktop_Qt_5_5_0_clang_64bit2-Debug/qtc_Desktop__e049b762-debug/install-root/Tiled.app/
WARNING:
WARNING: Could not find any external Qt frameworks to deploy in "/Users/bjorn/Projects/build-tiled-Desktop_Qt_5_5_0_clang_64bit2-Debug/qtc_Desktop__e049b762-debug/install-root/Tiled.app"
WARNING: Perhaps macdeployqt was already used on "/Users/bjorn/Projects/build-tiled-Desktop_Qt_5_5_0_clang_64bit2-Debug/qtc_Desktop__e049b762-debug/install-root/Tiled.app" ?
WARNING: If so, you will need to rebuild "/Users/bjorn/Projects/build-tiled-Desktop_Qt_5_5_0_clang_64bit2-Debug/qtc_Desktop__e049b762-debug/install-root/Tiled.app" before trying again.
```
This was the first time I tried running `macdeployqt`.

bjorn:
Do you have the rsync rule lying around somewhere maybe?

jakepetroules:
No but add something like the following to the Tiled.app qbs file:

    ...
    type: base.concat(["macdeployqt.run"])
    ...
    Rule {
        inputs: ["installable"]
        Artifact {
            filePath: "macdeployqt.run"
            fileTags: ["macdeployqt.run"]
        }
        prepare: {
            var cmd = new Command(FileInfo.joinPaths(product.moduleProperty("Qt.core", "binPath"), "macdeployqt"), [FileInfo.joinPaths(qbs.installRoot, "Tiled.app")]);
            cmd.description = "macdeployqt Tiled.app";
            return cmd;
        }
    }
    ...

Note that this is kind of a hack.

I am on IRC (Freenode) as jakepetroules in #qt-qbs channel where you can ask any questions you may have.

jakepetroules:
I just noticed you said macdeployqt didn't work. I believe this is fixed in dev branch... dunno if there's a workaround. You could just try an rsync solution instead of macdeployqt though, and no I don't have a Rule for that sitting around but it shouldn't be too hard to figure out. :)

PR #1075, opened 2015-09-09
Updated German translation

Ablu:
I created this pullrequest for him since he is apparently struggling with the git tools :)

### DIFF
--- a/translations/tiled_de.ts
+++ b/translations/tiled_de.ts
@@ -38,7 +38,7 @@
     <message>
         <location line="+23"/>
         <source>Format not recognized (see --export-formats)</source>
-        <translation type="unfinished"></translation>
+        <translation>Unbekanntes Dateiformat (siehe --export-formats)</translation>
     </message>
     <message>
         <location line="+12"/>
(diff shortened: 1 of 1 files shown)


Ablu:
Will review it later.

bjorn:
Awesome, thanks guys! I'll wait till you guys got it reviewed. I think I forgot to mention in my announcement that I was planning to release in two weeks, so you got some time.

wayfu:
I have definitely stopped struggling and given up, to preserve my sanity. :)

translations/tiled_de.ts
```diff
@@ -593,17 +593,17 @@
...
         <location line="+8"/>
         <source>&amp;Never</source>
-        <translation type="unfinished"></translation>
+        <translation>&amp;Nie</translation>
     </message>
     <message>
         <location line="+11"/>
         <source>For &amp;Selected Objects</source>
-        <translation type="unfinished"></translation>
+        <translation>Für ausge&amp;wählte Objekte</translation>
```
Ablu: Hm. This changes the accessor key... @bjorn @wayfu wouldn't it make sense to put the `&` before the `s` in the German part too?

translations/tiled_de.ts
```diff
@@ -1348,22 +1348,22 @@ Zeile %1, Spalte %2</translation>
...
         <location line="+12"/>
         <source>Uncaught exception in script. Please check console.</source>
-        <translation type="unfinished"></translation>
+        <translation>Nicht abgefangener Ausnahmefehler im Skript. Bitte Konsole überprüfen.</translation>
     </message>
     <message>
         <location line="+5"/>
         <source>Script returned false. Please check console.</source>
-        <translation type="unfinished"></translation>
+        <translation>Skript liefert "falsch" zurück. Bitte Konsole überprüfen.</translation>
```
Ablu: @wayfu *`lieferte` würde ich sagen...

Ablu:
Looks fine otherwise

translations/tiled_de.ts
bjorn: Yes, as long as that doesn't conflict with another menu entry.

wayfu: I was under the impression that the accessor key was generated from the ampersand in the text and not hardcoded? @bjorn how do you handle it when the original key is not present in the translation? is there a real value in having the same accessor key in all languages, and does it trump the value of having memorizable/sensible accessors?

translations/tiled_de.ts
wayfu: jo, geht auch

translations/tiled_de.ts
bjorn: Right, for these accelerators, you should adjust them to the language and just avoid conflicts within the same menu. You guys decide whether **s** or **w** is best here. :-)

Ablu: I think sticking to the &S makes it easier for the people who jump 
between the english and german version. At least so far I tried to stick 
to the english accessor keys.

Ablu:
@wayfu are you planning to adapt some of the changes I suggested?

wayfu:
Oh, I thought you were going to do that. Misunderstanding :) Should I send you the file again with the two changes above?

Ablu:
Well I do not mind :) Best would be a git commit, but you can mail me 
the file too, then I can do the commit for you.

Ablu:
Added @wayfu's fixes. Looks good to me now.

bjorn:
Great, thanks!

PR #1078, opened 2015-09-12
Update Czech translation

encukou:


bjorn:
Looking good, thanks!

PR #1080, opened 2015-09-12
Updated fr translation + Removed remaining shortcuts translations.

Bertram25:
Cheers! :D

### DIFF
--- a/translations/tiled_fr.ts
+++ b/translations/tiled_fr.ts
@@ -41,7 +41,7 @@
     <message>
         <location line="+23"/>
         <source>Format not recognized (see --export-formats)</source>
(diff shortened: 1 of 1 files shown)


bjorn:
Thanks @Bertram25!

I did not notice any actual errors in the shortcut translations, btw, but it's alright to remove them entirely. Problems only occur when the names of the keys that are part of the shortcut are translated.

Bertram25:
> I did not notice any actual errors in the shortcut translations, btw, but it's alright to remove them entirely. Problems only occur when the names of the keys that are part of the shortcut are translated.

Ok, duly noted. Have a nice week-end! :)

PR #1087, opened 2015-09-19
Add "Split into tiles" option to Export As Image dialog

mitchcurtis:
This allows users to split their map into individual images, which
may be useful if they have static tiles (no animation) and don't
want to load lots of little tiles.

### DIFF
--- a/src/tiled/exportasimagedialog.cpp
+++ b/src/tiled/exportasimagedialog.cpp
@@ -41,6 +41,7 @@ static const char * const VISIBLE_ONLY_KEY = "SaveAsImage/VisibleLayersOnly";
 static const char * const CURRENT_SCALE_KEY = "SaveAsImage/CurrentScale";
 static const char * const DRAW_GRID_KEY = "SaveAsImage/DrawGrid";
 static const char * const INCLUDE_BACKGROUND_COLOR = "SaveAsImage/IncludeBackgroundColor";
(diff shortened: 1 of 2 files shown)


mitchcurtis:
Not sure how useful this is in practice. My thinking is that rather than have potentially thousands of 32x32 QQuickImages in memory (for a fully explored map), it might be better to have a handful of larger ones. Do you see this being a better alternative for static tiles?

bjorn:
I think the feature is useful, but I'm not sure about building it into the Ui like this. I think it would be nice if such a thing could be done with a user script instead, or if it was a feature of the `tmxrasterizer` tool. That way, the export action can be more easily reproduced, which I expect to be important for such a feature. What do you think?

mitchcurtis:
Are the user scripts/tmxrasterizer tool documented? My only concern with adding it elsewhere would be that users can't find it, as the documentation is quite sparse at the moment.

What do you mean by the export action being more easily reproduced? I'm a bit confused. :)

bjorn:
> Are the user scripts/tmxrasterizer tool documented? My only concern with adding it elsewhere would be that users can't find it, as the documentation is quite sparse at the moment.

Of course the documentation should be improved. Not finding the functionality is a valid concern, but I'd also like to keep the UI simple (I still have to check out the dialog with your changes).

> What do you mean by the export action being more easily reproduced? I'm a bit confused. :)

I just mean that it's harder to make sure the Export as Image settings are the same every time, than if you would put the right parameters in a batch file and run that each time you need to update the export.

PR #1095, opened 2015-09-23
Use tileset as the key in the map used to look up firstgids.

kiheru:
Fixes saving maps that have tilesets with duplicate names.


bjorn:
Hmm, what I'm wondering, why isn't this just an array of firstGids instead of a HashMap?

kiheru:
Probably for the lookup in getGid().

bjorn:
Hmm, right, otherwise you'd have to search for the index of the tileset all the time.

Well, thanks for the fix, looks better this way!

PR #1100, opened 2015-09-28
RandomPicker no longer accepts things that have probability zero or less

henu:
I'm developing [a game](https://youtu.be/mrLlPNSvDJ0) with my friend, that includes random level generator. The level generator works better, if I set terraintypes to as many tiles as possible. However, I don't want these tiles to appear when modifying terrain in Tiled, so I set their probability to zero. This causes problems, because things in QMap container in RandomPicker get overwritten.

In this pull request, I've modifier the RandomPicker class, so it ignores those things that have zero probability or less. It fixes the problem.

I tested it with stamp and bucket fill with random mode, and I also tested terrain tool. It didn't cause any problems, not even when I set everything to have zero or negative probability.

### DIFF
--- a/src/tiled/randompicker.h
+++ b/src/tiled/randompicker.h
@@ -40,8 +40,10 @@ public:
 
     void add(const T &value, qreal probability = 1.0)
     {
-        mSum += probability;
-        mThresholds.insert(mSum, value);
+        if (probability > 0) {
+            mSum += probability;
+            mThresholds.insert(mSum, value);
+        }
     }
 
     bool isEmpty() const


bjorn:
Ah, good catch! Indeed without this check it will overwrite valid choices with values that should be impossible...

bjorn:
Looks like a nice game, btw. :-)

bjorn:
Btw, I Just noticed your `tmx_sync` tool. Did you try using external tilesets instead or would that solve only part of the problem?

henu:
Damn, I didn't know there is such feature. I will try it as soon as I get to my machine again. It would be perfect, if I understood what you mean.

And thanks for merging! I was afraid this is considered too marginal case, but for us it is mandatory fix :)

bjorn:
It was a mandatory fix for Tiled too. :-)

You can export a tileset with the little Export button below the tileset view. You can then add that same tiles to other maps.

When you want to move your other maps over to refer to the same tileset, you could either choose to Export there too and then overwrite the external tileset file, or you could edit your maps in a text editor to change the tileset element to a reference to external tileset as you'll find in the first map file.

PR #1103, opened 2015-09-29
Update tiled_pt_BR.ts

laetemn:
My translation suggestion to the 'tiled_pt_BR.ts' file. (Brazilian Portuguese)

OBS: 
I wonder if the term "seal" present in several lines refers to a seal or is a reference to watermark?

### DIFF
--- a/translations/tiled_pt_BR.ts
+++ b/translations/tiled_pt_BR.ts
@@ -11,7 +11,7 @@
     <message>
         <location line="+83"/>
         <source>Donate</source>
-        <translation type="unfinished"></translation>
+        <translation type="unfinished">Doar</translation>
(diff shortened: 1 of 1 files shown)


bjorn:
You pushed some changes and then a revert of those changes, leaving the pull request with no changes at all. Could you fix that?

I don't see the term "seal" in any lines and don't know what it may be referring to. Where are you seeing this term?

I'll try to get in touch with the previous pt_BR translator for a review of the changes.

bjorn:
Next time, please update the existing pull request instead of opening a new one. Just force-push to the same branch you used for the pull request.

So for future reference, this moved to issue #1104.

PR #1114, opened 2015-10-07
Building on debian requires libqt5opengl5-dev.

technomancy:
Just an update to the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Compiling
 
 Make sure the Qt (>= 5.1) development libraries are installed:
 
-* In Ubuntu/Debian: `apt-get install qt5-default qttools5-dev-tools zlib1g-dev`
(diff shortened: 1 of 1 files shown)


bjorn:
Hmm, right, `qtbase5-dev` only "recommends" `libqt5opengl5-dev` but does not depend on it. Strange that I didn't run into this earlier, and for whatever reason it seems not necessary on Travis CI (using Ubuntu 14.04).

Thanks for the fix!

technomancy:
I think apt is configured to install recommended packages by default, but it's common to turn this off in order to not get a bunch of unnecessary stuff installed.

Thanks!

bjorn:
Alright, thanks for the explanation!

Now, for the next version of Tiled I will actually try to drop this dependency. :-)

PR #1126, opened 2015-10-16
Fix isometric rendering bug

### LINKED ISSUE #85: Isometric map issue with odd numbers
Hi! I noticed that the map moves out of the grid when you use odd tile widths and heights.
http://imageshack.us/photo/my-images/215/tiledbug.png/
I used 29 width and height.

And I have a suggestion: Why don´t you use normal tilesets, u´d use for orthogonal maps, and just rotate them by 45°? Itd be way easier to use and not so fault-prone..

RySchmitt:
Removes a previous issue where the render logic rounded odd tile heights.

Closes #85

### DIFF
--- a/src/libtiled/isometricrenderer.cpp
+++ b/src/libtiled/isometricrenderer.cpp
@@ -228,8 +228,8 @@ void IsometricRenderer::drawTileLayer(QPainter *painter,
 
     CellRenderer renderer(painter);
 
-    for (int y = startPos.y(); y - tileHeight < rect.bottom();
-         y += tileHeight / 2)
+    for (int y = startPos.y() * 2; y - tileHeight * 2 < rect.bottom() * 2;
+         y += tileHeight)
     {
         QPoint columnItr = rowItr;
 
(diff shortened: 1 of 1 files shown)


RyanSchmittDev:
Here is my test set: https://drive.google.com/file/d/0B6ONTdXiK6RLYk1nT0pkQ2htUlE/view?usp=sharing

Being unable to find a proper odd-numbered isometric tile-set, I just used improper parameters on an existing tile set.

bjorn:
Are you personally using an isometric grid with an odd tile height?

I'm wondering a little about your fix, since it basically means that it's trying to render every other row at half a pixel, right? This feels to me like it can't be the right thing to do.

Also, your test seems to use a tile height of 32, so how does it apply to this issue?

RySchmitt:
Sorry, I sent the wrong tmx, just replace the original tmx with this one: https://drive.google.com/file/d/0B6ONTdXiK6RLU0lUUklBUG9Ua00/view

I am not personally using an odd tile height, but the fix is equivalent to the original code for even numbered heights. The new version just doubles each parameter until the render step so that the odd height is not rounded off.

bjorn:
Thanks for the fix and sorry for the delay!

I still don't think it's very useful to draw every other row at half pixels, but it's definitely better than the repainting issues Tiled was having before!

Issue #489, opened 2013-08-04
Add support for nested (dictionary-type) properties


LearnCocos2D:
Currently it is not possible to create your own nested properties (like Position, Size). If a single property needs to encode multiple parameters, it needs to be done by embedding all parameters in the same string:

FollowPath | path=p1;speed=10;repeat=0

Or alternatively each property needs to have its own name:

FollowPath | p1
FollowPathSpeed | 10
FollowPathRepeat | 0

Both solutions are supoptimal for editing, and terrible for reading properties in a generic way. For example the above would require reading each property individually, while the example below could be mapped to a "createFollowPath" method that simply gets the properties dictionary passed in. With technologies like KVC (Key Value Coding, ie setting properties in the object hierarchy by key path like "FollowPath.speed") this would lend itself towards more automatisms.

So what I'd really like is to be able to add "dictionary" type of objects. Ideally with unlimited nesting, which once the nesting functionality is there should come natural anyway. The above would then look like this:

V  FollowPath
         path | p1
         speed | 10
         repeat | 0

To make this possible the user should be able to create a "dictionary" property. 

This could be done by offering another [+] icon next to it, perhaps [++] that creates a property of type dictionary. Then with the dictionary property selected, or one of its children, any click on [+] or [++] adds the new property to the selected dictionary property instead of the root properties.

theHacker:
I am missing such a feature for a long time.

For a first step (may be this could be an independent feature) I would like to see Tiled not removing custom XML data in a file. Currently if I write my custom tags inside a Tiled file and edit the file with Tiled and save it, Tiled would remove all tags it does not know.

If Tiled would preserve these tags it would be possible to easily save nested properties within the files. Tiled doesn't even have to understand these very game specific properties. The only thing Tiled would have to do is providing a possibility to view and edit these XML fragments.

Custom tags could be nested inside a special tag (for example `<custom-properties>`) to prevent clashes with existing tags.

With this solution the Tiled file would not be as bloated as with @Chaoseiro's XML and offers more flexibility because you can use tags and attributes as you seem fit.

    <properties>
       
    </properties>
    <custom-properies>
      <a-property>some value</a-property>
      <nested-properties-1 foo="some" bar="value" attributes-as-you="wish">
        <a-nested-property>some nested value</a-nested-property>
        <a-nested-property-2>some nested value 2</a-nested-property-2>
      </nested-properties-1>
      <nested-properties-2>
        <a-nested-property>some nested value</a-nested-property>
[...]

PR #1131, opened 2015-10-23
Added Types for Custom Properties

CaptainFrog:
It is now possible to add custom properties with a type instead of only Strings. 
The available types are
* int
* float
* bool
* String

One can add a new property with the brand new "Add Property Dialog" as such:
addproperty

And in the PropertyBrowser the new types can be edited using the property browser editors:
floatedit

A new attibute "type" was added in the XML so we can read the types back:

```xml
<object name="myObject" id="1" x="416" y="640" width="32" height="32">
 <properties>
  <property name="myBool" value="false" type="bool"/>
  <property name="myFloat" value="0.32000000000000001" type="double"/>
  <property name="myInt" value="2" type="int"/>
  <property name="myString" value="string value" type="QString"/>
 </properties>
</object>
```

Older maps not developed with this modification have their custom properties defaulted to String when read by tiled (for compatibility's sake)

Just a proposition if anybody needs this.


bjorn:
That's a great change that I'm sure many people will love! I'll write some feedback inline.

src/libtiled/mapreader.cpp
```diff
@@ -964,7 +965,12 @@ void MapReaderPrivate::readProperty(Properties *properties)
         }
     }
 
-    properties->insert(propertyName, propertyValue);
+    if(propertyType.isEmpty() || propertyType.isNull()){
```
bjorn: `QString::isNull` implies `QString::isEmpty` so you can drop the second condition.

src/libtiled/mapreader.cpp
```diff
@@ -964,7 +965,12 @@ void MapReaderPrivate::readProperty(Properties *properties)
...
+        propertyType = QLatin1String(QVariant::typeToName(QVariant::String)); // Default to String
```
bjorn: Please just use "string" for the string type instead of "QString".

src/libtiled/object.h
```diff
@@ -89,7 +90,11 @@ class TILEDSHARED_EXPORT Object
      * Returns the value of the object's \a name property.
      */
     QString property(const QString &name) const
-    { return mProperties.value(name); }
+    { return mProperties.value(name).toString(); }
```
bjorn: Wouldn't it make sense to change the return type to `QVariant` instead?

src/libtiled/object.h
```diff
@@ -100,8 +105,13 @@ class TILEDSHARED_EXPORT Object
     /**
      * Sets the value of the object's \a name property to \a value.
      */
-    void setProperty(const QString &name, const QString &value)
-    { mProperties.insert(name, value); }
+    void setProperty(const QString &name, const QString &value, QVariant::Type type = QVariant::String)
```
bjorn: I'd say this setter should just take a `const QVariant &value`.

src/libtiled/object.h
```diff
@@ -89,7 +90,11 @@ class TILEDSHARED_EXPORT Object
      * Returns the value of the object's \a name property.
      */
     QString property(const QString &name) const
-    { return mProperties.value(name); }
+    { return mProperties.value(name).toString(); }
+
+    QVariant::Type propertyType(const QString &name) const {
+        return mProperties.value(name).type();
+    }
```
bjorn: If `property` returns `QVariant`, this function becomes obsolete.

src/tiled/addpropertydialog.cpp
```diff
@@ -0,0 +1,57 @@
...
+    QString typeText = ui->typeBox->currentText();
+
+    // Returned the correct type according to the name
+    if(typeText == QLatin1String("float")){ return QVariant::Double;}
+    if(typeText == QLatin1String("String")){ return QVariant::String;}
+
+    if(typeText == QLatin1String("Size")){ return QVariant::Size;}
+    if(typeText == QLatin1String("Size Float")){ return QVariant::SizeF;}
+    if(typeText == QLatin1String("Point")){ return QVariant::Point;}
+    if(typeText == QLatin1String("Point Float")){ return QVariant::PointF;}
```
bjorn: Future plans? :-)

src/tiled/addpropertydialog.h
```diff
@@ -0,0 +1,32 @@
...
+{
+    Q_OBJECT
+
+public:
+    explicit AddPropertyDialog(QWidget *parent = 0);
+    ~AddPropertyDialog();
+
+
+    QString getPropertyName();
+    QVariant::Type getPropertyType();
```
bjorn: Please follow the existing coding style, in that getters do not have "get" in their name.

src/tiled/addpropertydialog.h
```diff
@@ -0,0 +1,32 @@
...
+
+
+
+
+private slots:
+    void on_name_textChanged(const QString &arg1);
```
bjorn: I prefer not to rely on the automagic connections.

src/tiled/changeproperties.h
```diff
@@ -88,6 +89,7 @@ class SetProperty : public QUndoCommand
     QList<Object*> mObjects;
     QString mName;
     QString mValue;
+    QVariant::Type mType;       // The type of the property
```
bjorn: Here again, I believe it will be easier to just change the type of `mValue` to `QVariant`.

src/tiled/propertiesdock.cpp
```diff
@@ -191,15 +192,17 @@ void PropertiesDock::tilesetFileNameChanged(Tileset *tileset)
...
-    dialog->setInputMode(QInputDialog::TextInput);
-    dialog->setLabelText(tr("Name:"));
-    dialog->setWindowTitle(tr("Add Property"));
-    dialog->open(this, SLOT(addProperty(QString)));
+    AddPropertyDialog *dialog = new AddPropertyDialog(mPropertyBrowser);
+    int result = dialog->exec();
+    if(result == AddPropertyDialog::Accepted){
+        addProperty(dialog->getPropertyName(),dialog->getPropertyType());
+    }
+    delete dialog;
```
bjorn: Instead of using `new/delete`, please just allocate it on the stack.

bjorn:
@CaptainFrog Do you expect to get back to this pull request sometime, or should I continue your work once I get around to it?

CaptainFrog:
@bjorn Sorry for the long delay, I have been really busy recently, I think it would be better if you'd take over as I don't know when I'll have time to complete this. If I ever find some time to finish it, I'll reach out to you. :)

bjorn:
@CaptainFrog Alright, thanks for the heads up! I'll see when I get around to this. I think it would be nice to still do this before Tiled 1.0.

bjorn:
Merged and [finished](https://github.com/bjorn/tiled/commit/434643a953ea01eda3403b04523c01430234cd15)! :-)

CaptainFrog:
Awesome!
Next step "dictionary type"

bjorn:
> Awesome!
> Next step "dictionary type"

Right, that's covered by issue #489. There's a lot of next steps though. :-)

PR #1139, opened 2015-11-22
Cell properties WIP

JoeDupuis:


bjorn:
Alright, first of all thanks for opening this pull request. Looking through the changes, it's clear that it will probably work fine, but there are some things I'm uneasy about.

By deriving Cell from Object, there is likely a big increase in tile layer memory consumption as well as layer construction time. Of course, it won't show on small maps, but it will affect big maps. To be sure this impact would have to be measured, but I think switching to one additional `QMap<QPoint,Properties>` to store this meta-information would be better. But of course, that'll make all the rest of the patch rather more complicated...

The map reading and saving routines can put the properties in a compressed JSON string and even implement custom parsing of that data. I think this overcomplicates things and in the TMX format the properties should only be stored in a plain XML format, best just using the `properties` elements already used elsewhere instead of using attributes.

In a way, I'm wondering whether this whole properties tool couldn't be done as something that just makes it easier to place tile-aligned objects. But then again, tile-aligned objects only exist in orthogonal and isometric maps, and not in staggered or hexagonal maps. Also, it is more convenient to have the properties more directly connected to the tiles...

peterschade:
Please support cell/tile properties per placed tile. The object Layer is not an replacement for this. So this could be really useful for level creation without extra markup files/hacks.

twistedjoe:
Object property doesn’t cut it on all the use cases. In my game for example each slot on a layer can only have one “block” (tile). 
Object allows you to put more than one object at the same place. 
Sure I could check for this at run time or make a script that check the map validity before running it. 
Plus I have to handle the conversion between position to “tile index” since I store everything in an array. Also let’s assume a level designer put two objects at the same place, he might not find that before testing the validity of the map, then some property might be lost or split between the two superposed object.

Working with objects for tile-based game is asking for trouble.

We should test the performance of the current implementation on big maps before “pre-optimizing”. 
Maybe it’s not even a problem. 
On my maps, it was imperceptible, but my maps are on the small side, so it doesn’t means much.

peterschade:
@twistedjoe could you provide a mac binary build?

twistedjoe:
There is no garantee it will work on your machine, but here it is:
https://dl.dropboxusercontent.com/u/11477187/Tiled.app.zip
Not all the map formats are supported though. We work in lua so this one works and I believe the default one too. 
The others I am not sure, I haven't tested.  I have not work on it for quite a while now so I don't recall much.

peterschade:
@twistedjoe Thanks for uploading. It fail with loading the qt4 lib
```
Dyld Error Message:
  Library not loaded: /usr/local/lib/QtGui.framework/Versions/4/QtGui
  Referenced from: /Users/USER/Desktop/*/Tiled.app/Contents/MacOS/Tiled
  Reason: image not found
```
It seems i must download the whole qt4 environment to compile this for testing :(

twistedjoe:
I remember that I was quite surprise by the speed it took to install with brew. 
Just Install brew and then 
Brew install qt
Be sure that it install version 4 and not 5

peterschade:
Good Work!

This worked:

     brew instal qt4

No need to recompile your package, because it finds the libs ;)

By the way this is exactly what i need to script my custom animations and behaviors for tiles/cells. Yet i need to extend my parser... night work

TheDerek:
Any indication as to when this will be merged? Considering making my own fork for my game if this takes too long.

bjorn:
@TheDerek I have no plans to simply press the merge button on this one. I have explained my reservations in my previous comment and I think it needs careful consideration and probably some major changes before being included. So if you really need this, don't hesitate to compile a version for yourself.

theKidOfArcrania:
This might be a little late, but may I propose a partial solution? I have observed how the java port has an intriguing, unused (i think), feature that might work. The Java version does something like have a `tiledproperties` xml tag to specify tile-specific properties. Within it, there are child `tile` tags and you can specify the `x` and `y` as attributes to avoid empty property entries within the file. I have a [version](https://github.com/theKidOfArcrania/tiled) of this implementation (although it doesn't address the internal memory issues, but I think it might be closer to solving it?)

bjorn:
@theKidOfArcrania I forgot that Tiled Java already had code for this. It was actually not by code (see [this change](https://github.com/bjorn/tiled-java/commit/d7260b4d557458c889eea1a8ccf67e0f5e27ba61)). Note that it is implemented as a `HashMap<Object, Properties>` (actually should be `HashMap<Point, Properties>`), rather than having a dictionary set up for each tile location. Also you'll find that the orthogonal map renderer was visualizing tiles that had properties set on them, but indeed in the editor UI the functionality is commented out, and in the TODO file there is:
```
* Add support for tile instance properties (by Christian Henz) (WIP)
  - Implement an effective way of displaying the presence of these properties
```
I think the functionality was never finished because Tiled Java 0.7 added support for placing rectangle objects to which properties could be assigned, which had a big overlap in functionality.

PR #1163, opened 2015-12-29
Pike TMX parser moved from gitorious to gitlab :)

endstation:


bjorn:
Thanks for the update!

Unfortunately the library seems to no longer get updated, but better to have a working link. :-)

PR #1166, opened 2016-01-04
cygwin build workaround

okami-1:
"-z" flag is unsupported on cygwin


bjorn:
It seems a bit strange to me that none of the options available with -z would be available on cygwin, but at least the RPATH stuff is probably not useful on Windows anyway so this patch should be fine.

Thanks!

PR #1174, opened 2016-01-10
Consider the RPATH variable while building the terraingenerator

Ablu:


Ablu:
well... this would be for the `0.15.0` branch. But there is none yet ;)

bjorn:
Thanks. :-)

PR #1177, opened 2016-01-17
fix tmx doc to no longer say width/height are ignored

zachreizner:


bjorn:
Good catch and thanks for the update!

Issue #953, opened 2015-05-18
Command line arguments aren't working


ghost:
Hi,

build Tiled from Git, now when I try to run Tiled with an command line argument, nothing happens. I run Windows 8.1 64-bit, Qt 4.8.6 - MinGW.
Thats what I get:

```
user@Main-PC ~/Documents/Remote Repositories/tiled (master)
$ tiled.exe --version

user@Main-PC ~/Documents/Remote Repositories/tiled (master)
$ tiled.exe --show-formats
```

Nothing happens, only when I start Tiled using ``--disable-opengl``.

bjorn:
It seems that for some reason the output is not written to the console. I'm not sure whether Windows applications can do that which are graphical applications.

lefticus:
For a Qt project I used to work on, we ended up popping up a message box on Windows to display results from command line switches. It wasn't ideal, but it was the only solution we found. IIRC, Windows applications either do have a console (and it opens every time the application is launched, which is not ideal) or they do *not* have a console, and displaying output to the user is difficult.

There's some other application specific cruft in here, but you can see I just did a QMessageBox

https://github.com/NREL/OpenStudio/blob/v1.1.1/openstudiocore/src/runmanager/app/main.cpp#L32-L45

-Jason

buckle2000:
FYI: http://msdn.microsoft.com/en-us/library/ms681952(VS.85).aspx

bjorn:
@buckle2000 Thanks, that is worth looking into sometime. Did you actually use it yourself or do you have a more elaborate example of how this is supposed to be used?

buckle2000:
```cpp
	bool has_console = AttachConsole(-1);
	HANDLE handle = GetStdHandle(-11); // stdout
	auto s = u"你好";
	WriteConsole(handle, s, sizeof(s), NULL, NULL);
```

bjorn:
Thanks @buckle2000. I've just tried it out and I was able to output to the console using `WriteConsole` using the following snippet (the `u` thing didn't work with VS 2013, or may not be part of C++11):
```c++
    bool has_console = AttachConsole(ATTACH_PARENT_PROCESS);
    if (has_console) {
        HANDLE handle = GetStdHandle(STD_OUTPUT_HANDLE);
        QString s(QStringLiteral("foo bar baz\n"));
        WriteConsole(handle, s.constData(), s.length(), NULL, NULL);
    }
```
[...]

PR #1182, opened 2016-01-20
Command-line map format conversion utility

tcoxon:
It is possible to convert a map from one format to another by opening it in Tiled and saving it to the new format. I recently found the need to do this to tens of thousands of map files, which would take a lot of clicking!

My solution was to create a command line utility, that could be scripted to automate the conversion.


bjorn:
Were you aware that Tiled itself already has `--export-map` and `--export-formats` options to help with converting maps to the various supported formats? It does not support converting of tilesets yet though, since that only makes sense since recently.

I do think a separate command line tool makes sense, especially on Windows where a graphical application can't also do console output (issue #953).

Thanks for opening the pull request! It does seem like the commits could be cleaned up a little though. I see a commit of myself in there and a huge merge, so I think it they be rebased onto the latest master. Also ignoring of vim and ctags related files is better done globally (so you can add those to your git settings instead).

tcoxon:
There seems to be no way to change the branch of a pull request so I'm closing this one. My rewritten commits are on #1187.

bjorn:
Instead of changing the name of the source branch, GitHub expects you to force-push to the branch you already opened a pull request for. Something for next time :)

PR #1187, opened 2016-01-28
Command-line format conversion utility (cleaned up version)

tcoxon:
I created a pull request in #1182 that contained some extraneous commits. Here, I've provided a cleaned up version of history containing nothing that it shouldn't.

This command line utility converts files between Tiled's various formats (including tilesets), and should work on Windows where graphical utilities reportedly can't do console output (#953).

### DIFF
--- a/src/formatconvert/main.cpp
+++ b/src/formatconvert/main.cpp
@@ -31,28 +31,42 @@ Map *readMap(QString const &fileName) {
   }
 }
 
-void writeMap(Map *map, QString const &fileName) {
+bool writeMap(Map *map, QString const &fileName) {
   MapFormat *format = findFormat<MapFormat>(fileName);
   if (format) {
     qDebug() << "Writing " << format->nameFilter();
-    format->write(map, fileName);
+    if (!format->write(map, fileName)) {
+      qFatal("%s", qPrintable(format->errorString()));
+      return false;
+    }
   } else {
     qDebug() << "Writing TMX format";
     MapWriter writer;
-    writer.writeMap(map, fileName);
+    if (!writer.writeMap(map, fileName)) {
+      qFatal("%s", qPrintable(writer.errorString()));
+      return false;
+    }
   }
+  return true;
 }
 
-void writeTileset(SharedTileset tileset, QString const &fileName) {
+bool writeTileset(SharedTileset tileset, QString const &fileName) {
   TilesetFormat *format = findFormat<TilesetFormat>(fileName);
   if (format) {
     qDebug() << "Writing " << format->nameFilter();
-    format->write(*tileset, fileName);
+    if (!format->write(*tileset, fileName)) {
+      qFatal("%s", qPrintable(format->errorString()));
+      return false;
+    }
   } else {
     qDebug() << "Writing TSX format";
     MapWriter writer;
-    writer.writeTileset(*tileset, fileName);
+    if (!writer.writeTileset(*tileset, fileName)) {
+      qFatal("%s", qPrintable(writer.errorString()));
+      return false;
+    }
   }
+  return true;
 }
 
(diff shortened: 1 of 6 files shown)


src/formatconvert/main.cpp
```diff
@@ -0,0 +1,113 @@
...
+#include "mapreader.h"
+#include "mapwriter.h"
+
+using namespace Tiled;
+
+template <typename Format>
+Format *findFormat(QString const &fileName) {
+  FormatHelper<Format> formatHelper(FileFormat::ReadWrite, "");
+  foreach (Format *format, formatHelper.formats()) {
+    if (format->supportsFile(fileName))
```
bjorn: I realize that I should probably rename `supportsFile` to `canRead`. It is only meant to find out whether a certain plugin can read an existing file, and not for finding a plugin to write a format with. So when I tested this tool, it failed for me because I was trying to convert a map to a .lua file.

bjorn:
I think this is a good start, but there are a number of things to improve:

* File format detection by extension is nice, but won't always work. There are for example two plugins that export to a "lua" file. Hence, please check how `tiled --export-map [format[ <source file> <target file>` is implemented. This conversion tool will need to support the same explicit `[format]` option to resolve such conflicts I think,  unless you have a better idea.

* The coding style is not matching with this project's style in many ways. Are you interested in making sure it does? Examples:
    * Indentation should be with 4 spaces, not 2 spaces.
    * `const` comes first in for example `const QString &string`.
    * For functions, the `{` starting the body should be on its own line.

* The name `formatconvert` is too generic for a command-line tool, at least on Linux. You've only just managed to avoid a collision with image conversion tool `convertformat` from the [leptonica](http://www.leptonica.com/) package. Unfortunately, I don't have a good alternative since for example "convertmap" is too specific given that it also deals with tilesets. Since actually I think format conversion isn't the only thing we'll want to do from the command-line (it already isn't with `tmxrasterizer` also being there), I'm somewhat inclined to suggest `tiledcmd`, and have various actions as second parameter, for now just `tiledcmd convert ...`, but then we can also integrate `tmxrasterizer` as `tiledcmd render ...` for example. And when there is scripting added eventually, this tool could also provide access to running scripts like `tilecmd run <scriptfile> ...`.

src/formatconvert/main.cpp
```diff
@@ -0,0 +1,113 @@
...
+  }
+  return true;
+}
+
+int main(int argc, char *argv[]) {
+  QGuiApplication app(argc, argv);
+  QGuiApplication::setApplicationName("formatconvert");
+  QGuiApplication::setApplicationVersion("0.1");
+
+  QCommandLineParser parser;
```
bjorn: This will raise the Qt dependency from 5.1 to 5.2, but I have no problems with that. Actually Travis CI builds are done with Qt 5.2 so there should be no issues. But please do raise the minimum Qt version to 5.2.0 in `tiled.pro`.

tcoxon:
Thanks for the feedback. I'm still intending to fix these issues - just need to find time to sit down and work on it.

bjorn:
@tcoxon No problem, thanks for the heads up! So much to do, so little time!

bjorn:
Just a heads up that Tiled now already depends on Qt 5.2 (09bb4c3a6a6cbe39546fc291d0eaa0e4b4061d6f), so at least there's no need to change anything related to that anymore. :-)

bjorn:
With issue #953 resolved as of Tiled 0.18 and the ability to write TMX or read other formats added with change f55ea17c40f4369ac89aac1494ec6c81cabbfbb5 for Tiled 1.1, the remaining functionality of this PR is the ability to convert tilesets. Since no separate tool is required anymore for the command-line output, this functionality is probably better added as done in PR #1872, so I'll close this one.

Issue #1127, opened 2015-10-16
Use an eye icon instead of a checkbox for layer visibility


bjorn:
Would look nicer and be more intuitive. It's what all other graphical applications are doing!

Also this would allow extending it with a lock icon later on, when implementing #734.

CaptainFrog:
+1 for this!

IMMZ:
I'll put it in my 'Tiled Contribution Pack'.
Some moments:
1. Where to get icons?
2. What about implementation? I offer to create a new class 'LayerCheckBox' with properties 'locked' and 'visible'. If i correct icon could be shown by reimplementing paint (...) func.

PR #1196, opened 2016-02-08
Added eye icon as an indicator of layer visibility.

IMMZ:
Worked on #1127 . Now layers visibility looks like in GIMP. Icon was taken from GIMP sources.
eye


bjorn:
That seems like a nice and short implementation. One thing I'm worrying about is whether it will still have a native look-and-feel on all platforms, since you're using `QItemDelegate` rather than `QStyledItemDelegate`. But I see the latter does not have the convenient `drawCheck` virtual function. So in any case this will need some testing on different platforms.

Visually, please try to center the eye vertically. At the moment it appears aligned to the top.

src/tiled/layerdock.h
```diff
@@ -77,6 +80,22 @@ private slots:
 };
 
 /**
+ * Delegate for drawing an eye icon in LayerView when the layer is visible.
+ */
+class LayerVisibilityDelegate: public QItemDelegate
```
bjorn: Since this class is not used anywhere else than in `layerdock.cpp`, please move it entirely into the cpp file.

src/tiled/layerdock.h
```diff
@@ -77,6 +80,22 @@ private slots:
...
+{
+public:
+    explicit LayerVisibilityDelegate(QObject *parent = 0);
```
bjorn: I would prefer if you typed `nullptr` instead of `0` here. Tiled already depends on C++11.

PR #1197, opened 2016-02-08
Mini code improvements. STL math functions replaced by Qt ones.

IMMZ:
I'm not sure if this improvement is really useful. I think if the project is written with Qt it would be good to use Qt on maximum and I've tried to make a small step closer to purity.


bjorn:
I hope you don't mind if I reject this patch. I just think there is no reason to rely on Qt APIs instead of standard C++ APIs for reasons of purity.

When standard C++ APIs are available, for example for containers or shared pointers, I often still prefer to use Qt versions because its APIs are more comfortable and often more readable as well. But this advantage does not apply to basic math functions.

IMMZ:
Ok.

PR #1198, opened 2016-02-10
Update OSX install instructions for QT

briancain:
Since Tiled requires QT to be above version 5, you need to install qt5
with homebrew, not qt (which is version 4). Also since qt5 is keg-only,
it must be linked with the --force flag to be added to the path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,9 @@ Make sure the Qt (>= 5.1) development libraries are installed:
 * In Ubuntu/Debian: `apt-get install qt5-default qttools5-dev-tools zlib1g-dev libqt5opengl5-dev`
 * In Fedora:        `yum install qt-devel`
 * In Arch Linux:    `pacman -S qt`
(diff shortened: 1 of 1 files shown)


bjorn:
Thanks!

I'm not personally using Brew to install Qt for compiling Tiled, but rather take the binaries from qt.io, so I wasn't aware of this.

Instead of linking with --force, maybe the instructions could also be changed to instruct how to run qmake without adding it to the path?

briancain:
@bjorn - Yeah, that's not a bad idea. I did it that way just to keep the instructions within the homebrew ecosystem, but you could also just add it to your .bash_profile or symlink it, but this seemed like the most straight forward way to document how to get it working to me.

bjorn:
@briancain Alright, we'll leave it at this then and assume that people who want things differently will know how to do that. Thanks!

briancain:
@bjorn - awesome! sounds good! Thanks for taking a look at the pull request.

Issue #1199, opened 2016-02-12
Add XML-item to new map dialog


GarryDvaraza:

bjorn:
See comment at #1200.

PR #1200, opened 2016-02-12
Issue #1199

GarryDvaraza:


bjorn:
I have removed this option in Tiled 0.15 on purpose since this format is deprecated. You should not use it because it is very space-inefficient so your map file will easily be huge and take a long time to load. The XML format has no advantages compared to the CSV based layer data format, which is also easy to parse but much more space-efficient (though for best performance you'll want to choose zlib-compressed binary layer data).

GarryDvaraza:
Oh, I'm sorry bjorn. I use xml for my network shooter. There are don't have a big map.
Please tell me what bugs are critical? :)
Best regards.

bjorn:
Alright. Of course for small maps it doesn't matter much, but you'd still be better off using CSV since it makes the file a little more readable in a text editor or a diff. I mainly decided to deprecate the "XML" option since it was a bit of a trap and some people ended up with maps of several megabytes that were slow to load.

There should not be any critical bugs, but feel free to look at [all open reported bugs](https://github.com/bjorn/tiled/issues?q=is%3Aissue+is%3Aopen+label%3ABug) or the [issues I assigned to Tiled 1.0](https://github.com/bjorn/tiled/milestones/Tiled%201.0) if you'd like to help our with something. The latter is a collection of things which I believe are important before I can release Tiled 1.0. In the list of open bugs you'll find a lot of things with little information or problems outside of Tiled's source code, though it would be nice if we could reduce this list of course.

GarryDvaraza:
Ok. Thank you :)

PR #1203, opened 2016-02-14
Update tmx-map-format.md

bitJericho:
add hexsidelength attribute

### DIFF
--- a/docs/reference/tmx-map-format.md
+++ b/docs/reference/tmx-map-format.md
@@ -19,6 +19,7 @@ Have a look at the [changelog](tmx-changelog.md) when you're interested in what 
 * <b>height:</b> The map height in tiles.
 * <b>tilewidth:</b> The width of a tile.
 * <b>tileheight:</b> The height of a tile.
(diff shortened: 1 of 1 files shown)


docs/reference/tmx-map-format.md
```diff
@@ -19,6 +19,7 @@ Have a look at the [changelog](tmx-changelog.md) when you're interested in what
 * <b>height:</b> The map height in tiles.
 * <b>tilewidth:</b> The width of a tile.
 * <b>tileheight:</b> The height of a tile.
+* <b>hexsidelength:</b> For staggered and hexagonal maps, determines the amount of width or height (depending on the staggered axis) of the tile's edge, in pixels.
```
bjorn: Thanks for adding this! But I think the "staggered and" should be removed here. I'm just wondering why you wrote that there? In my testing and as is intended, only hexagonal maps use this property.

bitJericho: Oh yes it looks like you are correct.

bjorn:
Alright, merged and fixed bit in 3e73c025b1090fb4bf6e8d179c7eeee5c540c02d. :-)

PR #1207, opened 2016-02-19
Fixed compilation on Windows with MinGW and qmake.

IMMZ:
I have a resource compilation error when tried to build Tiled on Windows 7 with MinGW (Qt 5.4.2) with qmake. The error occured because of wrong resource filename in *.pro file. In *.qbs file name is written correctly. After this fix 'Tiled' was built OK.


bjorn:
Ideally, it should of course do the replacement of `@VERSION@`, like the Qbs build system is doing, but for now it's good that the qmake builds for Windows are at least working.

PR #1210, opened 2016-02-21
Prevent segfault when adding new property to many tiles.

henu:
Found a bug: If you select multiple tiles, and try to add a new property to them, the editor segfaults. This does not happen if you select the tile with lowest ID last.

I think the reason is, that mObject doesn't have the property (yet?), since the value.type() will be 0. When I selected the lowest ID tile last, the value.type() was 10 and there was no segfault.

When you then call mVariantManager->addProperty(value.type(), name); and value.type() is 0, then it returns property that is NULL, and things start going wrong.

I made a little fix, but I have to admit that **I have no idea if this is the correct way to fix this**. Anyway, it seems to work with tiles now. I also tested with map properties and I also played with undo/redo and I didn't found any problems after this fix.

### DIFF
--- a/src/tiled/propertybrowser.cpp
+++ b/src/tiled/propertybrowser.cpp
@@ -294,7 +294,7 @@ void PropertyBrowser::propertyAdded(Object *object, const QString &name)
         QtProperty *precedingProperty = (index > 0) ? properties.at(index - 1) : nullptr;
 
         mUpdating = true;
-        QVariant value = mObject->property(name);
+        QVariant value = object->property(name);
         QtVariantProperty *property = mVariantManager->addProperty(value.type(), name);
         property->setValue(value);
         mCustomPropertiesGroup->insertSubProperty(property, precedingProperty);


bjorn:
Good catch! The fix is totally fine, though I just noticed it is incomplete. While it will avoid a crash, the insertion index will likely be off because of the line above:
```c++
const int index = mObject->properties().keys().indexOf(name);
```
But fixing this correctly, that line should actually be changed to:
```c++
const int index = mCombinedProperties.keys().indexOf(name);
```
I'll push a fix for that tomorrow and update the snapshots.

Thanks!

bjorn:
Actually that `indexOf` line is a little tricker to fix, since I realized `mCombinedProperties` is not being kept up-to-date as properties are added/removed. So for now i just pushed this crash fix to the `snapshot` branch, along with the Terrain Brush improvements regarding erasing of terrain, so that people using the snapshots will get an update.

PR #1215, opened 2016-02-28
It is now possible to set probability of multiple tiles.

henu:
I tested undo and redo and they seem to work.

There is one case that doesn't work:

1. Have tiles with probabilities 0, 0, 0, 1.
2. Select them from first to last, so that last selected has probability 1.
3. Now try to set probability to 1 for all of them and nothing happens.

The reason is, that the whole changing event is not triggered, since the value of the last tile wasn't really changed (1 changes to 1).

### DIFF
--- a/src/tiled/changetileprobability.cpp
+++ b/src/tiled/changetileprobability.cpp
@@ -29,22 +29,28 @@ namespace Tiled {
 namespace Internal {
 
 ChangeTileProbability::ChangeTileProbability(MapDocument *mapDocument,
-                                             Tile *tile,
+                                             const QList<Tile*>& tiles,
                                              float probability)
     : mMapDocument(mapDocument)
-    , mTile(tile)
-    , mProbability(probability)
+    , mTiles(tiles)
 {
+    mProbabilities.reserve(tiles.size());
+    for (int i = 0; i < tiles.size(); ++ i) {
+        mProbabilities.append(probability);
+    }
     setText(QCoreApplication::translate("Undo Commands",
                                         "Change Tile Probability"));
 }
 
 void ChangeTileProbability::swap()
 {
-    float probability = mTile->probability();
-    mTile->setProbability(mProbability);
-    mProbability = probability;
-    mMapDocument->emitTileProbabilityChanged(mTile);
+    for (int i = 0; i < mTiles.size(); ++ i) {
+        Tile* tile = mTiles[i];
+        float probability = tile->probability();
+        tile->setProbability(mProbabilities[i]);
+        mProbabilities[i] = probability;
+        mMapDocument->emitTileProbabilityChanged(tile);
+    }
 }
 
 } // namespace Internal
--- a/src/tiled/propertybrowser.cpp
+++ b/src/tiled/propertybrowser.cpp
@@ -937,7 +937,8 @@ void PropertyBrowser::applyTileValue(PropertyId id, const QVariant &val)
     switch (id) {
     case TileProbabilityProperty:
         undoStack->push(new ChangeTileProbability(mMapDocument,
-                                                  tile, val.toFloat()));
+                                                  mMapDocument->selectedTiles(),
+                                                  val.toFloat()));
         break;
     case ImageSourceProperty:
         undoStack->push(new ChangeTileImageSource(mMapDocument,
--- a/src/tiled/changetileprobability.h
+++ b/src/tiled/changetileprobability.h
@@ -35,7 +35,7 @@ class ChangeTileProbability : public QUndoCommand
 {
 public:
     ChangeTileProbability(MapDocument *mapDocument,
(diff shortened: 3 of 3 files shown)


src/tiled/propertybrowser.cpp
```diff
@@ -936,9 +936,17 @@ void PropertyBrowser::applyTileValue(PropertyId id, const QVariant &val)
 
     switch (id) {
     case TileProbabilityProperty:
+    {
+        QList<Tile*> tiles;
+        QList<float> probabilities;
+        for (Object *obj : mMapDocument->currentObjects()) {
```
bjorn: Please use `MapDocument::selectedTiles` here.

src/tiled/changetileprobability.cpp
```diff
@@ -29,11 +29,11 @@ namespace Tiled {
 namespace Internal {
 
 ChangeTileProbability::ChangeTileProbability(MapDocument *mapDocument,
-                                             Tile *tile,
-                                             float probability)
+                                             QList<Tile*> tiles,
+                                             QList<float> probabilities)
```
bjorn: These containers should be passed as `const ... &`. It's only saving some reference counting, but still.

src/tiled/changetileprobability.cpp
```diff
@@ -41,10 +41,14 @@ ChangeTileProbability::ChangeTileProbability(MapDocument *mapDocument,
...
-    mTile->setProbability(mProbability);
-    mProbability = probability;
-    mMapDocument->emitTileProbabilityChanged(mTile);
+    QList<float>::iterator itProbability = mProbabilities.begin();
+    for (Tile* tile : mTiles) {
+        float probability = tile->probability();
+        tile->setProbability(*itProbability);
+        *itProbability = probability;
+        mMapDocument->emitTileProbabilityChanged(tile);
+        ++ itProbability;
```
bjorn: I think I'd find this loop more readable as an index-based loop that indexes both arrays, rather than combining a range-based for loop with an iterator.

src/tiled/changetileprobability.cpp
bjorn: Btw, the reason you need a list of probabilities rather than just one is only because it is used to store the previous values. Hence, please just take a single value in the constructor and make the list within this class rather than having the `PropertyBrowser` do that.

bjorn:
Nice catch, I had not realized or remembered that this wasn't possible before! I've made some inline comments which you could address before I merge this. Thanks!

bjorn:
> The reason is, that the whole changing event is not triggered, since the value of the last tile wasn't really changed (1 changes to 1).

Yep, this is an issue that affects all properties. Not sure how this could be solved. For example, we can't connect to the closing of the editor, because no change should happen when the user just focused it and then unfocused it, without at least pressing Enter or something.

henu:
@bjorn Okay thanks for the feedback. I will fix those :)

I will also change the ChangeTileProbability, so that it doesn't take a collection of probabilities, but only one. Because I think such case will never happen, where you would like to set probability of multiple tiles, but have different probability for each. Of course inside ChangeTileProbability, there will still be a container of floats for undo/redo. Please comment if you think this is bad thinking.

bjorn:
> Of course inside ChangeTileProbability, there will still be a container of floats for undo/redo. Please comment if you think this is bad thinking.

No, that was exactly what I meant. :-)

src/tiled/propertybrowser.cpp
```diff
@@ -936,9 +936,12 @@ void PropertyBrowser::applyTileValue(PropertyId id, const QVariant &val)
 
     switch (id) {
     case TileProbabilityProperty:
+    {
```
bjorn: Note that the addition of this scope is no longer necessary.

henu:
Done.

Hope you like it now :) I didn't knew QList can be indexed, as I thought the name means its linked list.

bjorn:
I think the change looks fine now. Can you squash it into a single commit?

> I didn't knew QList can be indexed, as I thought the name means its linked list.

Yeah, this is a little confusing, but QList is a pretty smart class. Insertion is efficient by storing pointers to objects whenever the size of its objects is larger than a pointer. So it's a generic list class that is usually more efficient than an actual linked list (which is QLinkedList), and not much different from a QVector for small data types.

henu:
>Can you squash it into a single commit?

I have never done this before, but I guess everything went fine..? I did this:

```
git reset --soft HEAD~4
git commit
git push -f
```

bjorn:
> I have never done this before, but I guess everything went fine..? I did this:

Yep, you did fine! Personally I normally take the following approach:

```
git rebase -i HEAD~4
git push -f
```

Interactive rebase has a good amount of convenience for various scenarios, so if you're unfamiliar with it you may want to try that out sometime. :-)

PR #1220, opened 2016-03-09
correct the index of csvTileIds

ZacharyJia:
When the map is not square, the tiles will be disordered because of the wrong index. fix it by changing it from `ml.getHeight()` to `ml.getWidth() `


bjorn:
Nice catch, thanks for the fix!

PR #1236, opened 2016-03-22
Updated German translation

Ablu:
@wayfu Want to review again? :)

### DIFF
--- a/translations/tiled_de.ts
+++ b/translations/tiled_de.ts
@@ -33,12 +33,12 @@
     <message>
         <location filename="../src/tiled/addpropertydialog.ui" line="+14"/>
         <source>Add Property</source>
-        <translation type="unfinished">Eigenschaft hinzufügen</translation>
+        <translation>Eigenschaft hinzufügen</translation>
     </message>
     <message>
         <location line="+12"/>
         <source>Property name</source>
-        <translation type="unfinished"></translation>
+        <translation>Eigenschaftsname</translation>
     </message>
 </context>
 <context>
@@ -117,7 +117,7 @@
     <message>
         <location line="+5"/>
         <source>Start a new instance, even if an instance is already running</source>
-        <translation type="unfinished"></translation>
+        <translation>Eine neue Instanz starten, selbst wenn bereits eine Instanz läuft</translation>
     </message>
     <message>
         <location line="+32"/>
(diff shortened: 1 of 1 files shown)


translations/tiled_de.ts
```diff
@@ -3899,7 +3900,7 @@ Bitte Format angeben.</translation>
     <message>
         <location line="+1"/>
         <source>Erase Terrain</source>
-        <translation type="unfinished"></translation>
+        <translation>Terrain löschen</translation>
```
bjorn: In this case, it's "erase" as in using an eraser. Are you sure "löschen" is the right word?

Ablu: Ohh.... I thought the button would deleted an entire terrain entry... Probably because it is placed on the list and not in the toolbar. Why is it there while all other mapping tools are in the toolbar?

bjorn: Hmm, because it still uses the Terrain Brush rather than being a separate tool. That way you can also capture "no terrain" from the map and erase with it. The button is below the list in a way that I thought it would be suggesting a similar mode of operation, rather than to delete the selected terrain. But, I can see how it could be somewhat confusing. Anyway, not going to change for Tiled 0.16 anymore so maybe you have a more appropriate German word here?

bjorn:
I just tweaked that one translation a bit:

https://github.com/bjorn/tiled/commit/01205a4e5ea61daca9d626e560b53c219f1c8e5d

Thanks for the update!

Ablu:
Ah sorry... My Mail-Filter seems to be broken, did not got the github 
mails with me in CC into a different folder...

bjorn:
Ah, no problem!
 
 
On Mon, Mar 28, 2016, at 19:05, Erik Schilling wrote:
> Ah sorry... My Mail-Filter seems to be broken, did not got the github
>  mails with me in CC into a different folder...
>
>  On 03/28/2016 12:49 PM, Thorbjørn Lindeijer wrote:
>  > I just tweaked that one translation a bit:
>  >
>  > 01205a4
>  > <https://github.com/bjorn/tiled/commit/01205a4e5ea61daca9d626e560b53c219f1c8e5d>
>  >
>  > Thanks for the update!
>  >
>  > —
>  > You are receiving this because you authored the thread.
>  > Reply to this email directly or view it on GitHub
>  > <https://github.com/bjorn/tiled/pull/1236#issuecomment-202339516>
>  >
>
> —
> You are receiving this because you modified the open/close state.
> Reply to this email directly or view it on GitHub[1]
>
 

Links:

  1. https://github.com/bjorn/tiled/pull/1236#issuecomment-202487804

PR #1254, opened 2016-04-17
Turkish translations updated.

nuriu:
Just updated Turkish translations with minor fixes/replacements and new translations.


bjorn:
Looks good, thanks a lot!

PR #1261, opened 2016-04-29
Norwegian (bokmål) translation

peterjohansen:
A few things I took note of while translating:

- I did not translate Automap/Automapping because it's a named feature.
- Accelerators might not be optimally placed, but Qt Linguist didn't complain.
- I translated strings that were inside <>-symbols. Not entirely sure if I should've.
- "Norwegian Bokmål" turns into "NorwegianBokmal" in the language picker.
- The string "Paste in place" is not localized (noticed this randomly while running the program).

Some words might not be found in the dictionary as there are lots of technical words that don't exist in Norwegian (e.g. tileset), so I compounded words to explain the concepts as best as I could. Also there are probably a few errors and inconsistencies that I missed, but everything should be understandable and mostly correct.

Please let me know if there is anything that should be changed.


bjorn:
Hey, thanks for the new translation! And respect! It's not a particular small project to add a new translation from scratch.

Now, before I merge it I'd like to do at least one review, but I personally know very little Norwegian. Would you know anybody who could have a look?

> I translated strings that were inside <>-symbols. Not entirely sure if I should've.

I guess those were in command-line instructions. Should be fine.

> "Norwegian Bokmål" turns into "NorwegianBokmal" in the language picker.

I'll have a look at this, but this comes from Qt so maybe it should be fixed there.

> The string "Paste in place" is not localized (noticed this randomly while running the program).

That's because this is a feature I added recently and wasn't included in Tiled 0.16. I do a "string freeze" shortly before each new feature release. I'll include this new translation in Tiled 0.16.1, which will not have the "Paste in Place" action yet.

peterjohansen:
I doubt I know someone who would willingly go through the entire thing, but I can run some of the key words by some people to see if they agree with the translation. Is it enough that I talk to them, or do you need them to comment here?

I see the Travis CI build failed. I looked at the output, but I'm not very familiar with C++, so I wasn't able to decipher the error.

bjorn:
> I doubt I know someone who would willingly go through the entire thing, but I can run some of the key words by some people to see if they agree with the translation. Is it enough that I talk to them, or do you need them to comment here?

Ah, just showing the parts you're not sure about to someone else will do fine I think.

> I see the Travis CI build failed. I looked at the output, but I'm not very familiar with C++, so I wasn't able to decipher the error.

It's the OS X build, which has started failing since Qt 5.6. I do not know why yet either...

peterjohansen:
Based on feedback from a few people I made one major change: the word for tile was "flis" but is now "rute". Both work, but "rute" sounds a bit better.

bjorn:
"Rute" sounds better to me as well. It's quite funny to read the Norwegian. I know a little bit since I often went there on holiday with my parents, and there are many similarities with Dutch / German. :-)

Thanks for doing the feedback round and the improvements! I've cherry-picked it to the `0.16` branch for inclusion with Tiled 0.16.1, which I'll probably publish soon. I've also merged it to `master` and forwarded the `snapshot` branch, so the daily builds will see it soon.

PR #1266, opened 2016-05-04
Wip/tilesetdocument

theshadowx:
- Adding missing headers in some cpp files and adding missing files in .pro file in order that the compilation succeed. 
- Replacing scrollingPerPixel by scollingPerModel in TerrainView class to prevent the row from extending thus showing horizontal scrollbar.

Note: I think there is a bug generated by the mode scrollingPerPixel.


bjorn:
Hey @theshadowx, thanks for these fixes! However, the last commit on this branch was labeled "WIP" and I was already aware of the issues you fixed in your first two commits, and I'd like to just include those changes in that commit once I push the final version. I had pushed it already because I hoped to get around to continuing that work from home, but so far I didn't have the time.

It's interesting that you found that the horizontal scroll bar issue relates to scrolling per pixel. Rather than working this around, it would of course be nice to see if we can build a small example that reproduces the issue which we can then use in a bug report to Qt.

theshadowx:
After investigating, I think there is no bug.

The problem comes from the instantiation of the TerrainModel in TerrainDock : 

here https://github.com/bjorn/tiled/blob/wip/tilesetdocument/src/tiled/terraindock.cpp#L107
```
 mTerrainView(new TerrainView),
```
So the view is initialized with its default width, which is 640, that's why the horizontal scrollbar is shown.
To prevent this, the parent should be specified in the instantiation. 

```
 mTerrainView(new TerrainView(this)),
```

Is there anything connected to adding "this" in the instantiation as you added "this" in the other objects instantiation but not for mTerrainView : 

```
mToolBar(new QToolBar(this)),
mAddTerrainType(new QAction(this)),
mRemoveTerrainType(new QAction(this)),
mDocument(nullptr),
mTerrainView(new TerrainView),  //  < ------------------
mCurrentTerrain(nullptr),
mProxyModel(new TerrainFilterModel(this))
```

Because otherwise there will be a memory leak.

bjorn:
> So the view is initialized with its default width, which is 640, that's why the horizontal scrollbar is shown. To prevent this, the parent should be specified in the instantiation.

I tried this workaround but it didn't work, at least not completely. In testing, I noticed that on startup the scrollbar is actually not there, but it always appears as soon as you reduce the width of the terrains list, regardless of the width of the contents. So it does have something to do with the initial size, but there is a bug in that it fails to adjust when the size changes.

> Because otherwise there will be a memory leak.

There won't be a memory leak because the `mTerrainView` gets the parent `w` at the following line:
```c++
horizontal->addWidget(mTerrainView);
```
It may still be a good idea to set the parent directly, but then it should be set to `w` and not `this`.

theshadowx:
Weird that the scrollbar shows up. I've tried to reduce the width to the minimum possible but still the scrollbar doesn't show up.

isometric_grass_and_water tmx tiled_197

tiled_198

theshadowx:
here is the case where the scrollbar shows up : (I remove the text of the remove button to be able to reduce more)

tiled_199

tiled_200

bjorn:
Hmm, strange, you're absolutely right that it solves the problem on the `wip/tilesetdocument` branch. I had tested it on `master` and I just checked again while compiling it to the same Qt version, but on `master` this change does not do the job for some reason.

theshadowx:
Ok, I'll have a look on the master branch

bjorn:
I've spent some time today to finalize this commit (see 9802a4d91ebc5e69579e1977790dad089d3ebd03) and included also the fixes you had to do. Thanks for your help!

Issue #1269, opened 2016-05-06
Make snapshot builds for Fedora work again


bjorn:
The Fedora daily builds hosted at https://copr.fedorainfracloud.org/coprs/ablu/tiled-dailies/ seem to be stale. The last build was triggered 5 months ago and failed.

Hopefully @Ablu can help getting them to work again, as well as switching them from the `master` to the `snapshot` branch.

Ablu:
argh.... I need to setup some tracking for that...

Ablu:
I have a nearly working tito setup now. This would allow you to register a webhook to trigger rebuilds in copr automatically.

Issues in tito which prevent this at the moment:
 * dgoodwin/tito#222
 * dgoodwin/tito#227

bjorn:
Cool! The second one seems like the hardest one on their part, but we could put a file in the top level for now once dgoodwin/tito#222 is fixed?

Ablu:
Yeah... that should work... though we still need to wait until the first bug is fixed, in a release and the release is on the copr servers. Well I will still prepare the pullrequest. But it will only start working once the other steps are done.

bjorn:
@Ablu What's the status on this? It seems daily builds are up to date now, so is there still something to do before closing this issue?

Ablu:
Yeah the builds work again. I still do not like that the package has the same name as the official package...  The current procedure forced me to an ugly epoch trick. I will likely rename it in the future somewhen so switching back to the normal tiled package is easier. But that should work with the current infrastructure.

bjorn:
Alright, closing this issue then. Thanks!

PR #1282, opened 2016-05-22
Added scripts for building fedora dailies

Ablu:
The script needs to be invoked with a copr api token as the parameters.

After merging, can you push this to the snapshot branch too please? Then I can do final testing


Ablu:
Fixes #1269.

bjorn:
Thanks @Ablu. I'll sync `snapshots` now so you can test that part.

PR #1283, opened 2016-05-22
Bumped epoch of tiled.spec

Ablu:
I apparently pushed a build with a version number of 9999 while testing
tito. So I need to bump epoch in order to allow a clean upgrade path.

No need to merge to the snapshot branch. I triggered a manual build.

### DIFF
--- a/dist/copr/tiled.spec
+++ b/dist/copr/tiled.spec
@@ -1,5 +1,6 @@
 Name:           tiled
 Version:        0.16.0
+Epoch:          1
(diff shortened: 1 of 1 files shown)


bjorn:
So we have entered a new epoch!

Btw, you did notice I pushed a 0.16.1 release a while back, right? :-)

Ablu:
Yes i did :P. But I need to do that on my real pc... That one builds a 
lot faster.

PR #1307, opened 2016-06-17
Fix the export choice parameter from CLI

Pedroagaponto:
Filter match with the right format even if it comes prepended with the
absolute path. As all the options are read as filesToOpen.
(which get the values from QFileInfo(arg).absoluteFilePath())


bjorn:
Hrm, this is a regression in Tiled 0.16. I think a better fix would be to check if the file exists before turning it into an absolute path in `CommandLineParser::parse`.

bjorn:
Or, to wait with turning the files into absolute paths until it's actually necessary, in `main.cpp`.

Pedroagaponto:
much better solution, thanks

PR #1323, opened 2016-07-26
add tile xy-offset to Tiled

nobody-special666:
This merge adds the ability to add an XY drawing offset to individual tiles in a tileset.
So if somebody wants to align/offset how tiles are drawn against cell boundary they can do so.

you can edit+undo the offset in the propertybrowser. changing values forces a map redraw.
It only does load+save to the xml format, since I didn't know if the change would be accepted in current form. It adds an "offsetx" and "offsety" attribute to the tile tag for the tileset. It also works for tsx file.

TODO BUG: when using the 1-cell stamp brush it leaves artifacts on screen
when moving the mouse over the grey-area outside the map (seen in isometric maps).
I could not understand the brush code in order to fix this.
hopefully somebody can correct this small error.


bjorn:
Nice! I will try to review and test this change before the Tiled 0.17 string freeze. It could take a while until I get around to it, because I'm on holiday upcoming weekend.

Btw, AppVeyor test failed because the new files also need to be added to `src/tiled/tiled.qbs` (you can open the top-level `tiled.qbs` project file in Qt Creator to do this).

bjorn:
Btw, I noticed you're not committing with a valid e-mail address. Is that on purpose? It's not even associating your commits with your GitHub account.

nobody-special666:
no it wasn't intentional, although i'd prefer not to have my email in the commits.
tiled.qbs is fixed now

src/libtiled/tile.h
```diff
@@ -116,6 +116,9 @@ class TILEDSHARED_EXPORT Tile : public Object
 
     QPoint offset() const;
 
+    QPoint tileOffsetXY() const;
+    void setTileOffsetXY(QPoint offset);
```
bjorn: Please rename `Tile::offset` to `Tile::totalOffset` and then rename these functions to `Tile::offset` and `Tile::setOffset`. Then, if you make `Tile::totalOffset` return the sum of the local tile offset and the tileset tile offset, you would no longer need to change `maprenderer.cpp` and I think it may resolve some issues with dealing with the offset for tile objects.

src/libtiled/tile.h
```diff
@@ -152,6 +155,7 @@ class TILEDSHARED_EXPORT Tile : public Object
     QVector<Frame> mFrames;
     int mCurrentFrameIndex;
     int mUnusedTime;
+    QPoint mTileOffsetXY;
```
bjorn: This would also just become `mOffset`.

src/tiled/mapdocument.h
```diff
@@ -492,6 +494,15 @@ inline void MapDocument::emitTileLayerDrawMarginsChanged(TileLayer *layer)
 }
 
 /**
+ * Emits the signal notifying about the xy-offset of a tile changing.
+ */
+inline void MapDocument::emitTileOffsetXYChanged(Tile *tile)
+{
+    emit tileOffsetXYChanged(tile);
+    emit mapChanged();
```
bjorn: Please leave this second signal out if you can. Instead, you should connect to `tileOffsetXYChanged` (to be renamed to `tileOffsetChanged` in relevant places.

src/tiled/propertybrowser.h
```diff
@@ -115,6 +115,7 @@ private slots:
         YProperty,
         WidthProperty,
         HeightProperty,
+        TileOffsetXYProperty,
```
bjorn: Before renaming this to `TileOffsetProperty`, please rename the existing `TileOffsetProperty` to `TilesetTileOffsetProperty`.

bjorn:
> TODO BUG: when using the 1-cell stamp brush it leaves artifacts on screen
> when moving the mouse over the grey-area outside the map (seen in isometric maps).
> I could not understand the brush code in order to fix this.
> hopefully somebody can correct this small error.

Well, this is actually somewhat tricky to fix. The artifacts are due to Tiled trying to keep its painting area as small as possible when drawing things in the map view. Normally it's quite easy to figure out what parts of the map need to be painted, but it gets tricky when things are shifted by arbitrary amounts.

So far, tilesets could have an offset and layers could have an offset. I think Tiled was dealing with both of them quite well. But to deal with the tileset tile offset for example, the map is keeping track of which tilesets it is using, and then calculating its "draw margins" in `Map::recomputeDrawMargins`. This code will somehow need to be adjusted to take into account offsets in individual tiles.

Since it's undoable to take into account each individual tile (which would destroy the whole point of reducing the repaint area due to computational overhead), I would propose we calculate the maximum "offset margins" for each Tileset, and then combine that in `Map::recomputeDrawMargins`. It's alright if you'd prefer to leave this part to me.

nobody-special666:
I applied all your renames.
I assumed you also wanted ChangeTileOffsetXY class renamed as well, so I did it.
 
Your recommendation that removed the changes to maprenderer.cpp also fixed the artifacts for the stamp brush. copy-pasting the tile regions also applies the tile offsets correctly.

I did leave the "emit MapChanged()" hack in emitTileOffsetChanged() because recomputeDrawMargins did not force a map redraw. Please replace it with the "proper" way.

src/tiled/changetileoffset.cpp
```diff
@@ -0,0 +1,72 @@
+/*
+ * changetileoffsetXY.cpp
```
bjorn: File name in comment still needs to be renamed.

src/libtiled/tile.cpp
```diff
@@ -41,6 +41,7 @@ Tile::Tile(int id, Tileset *tileset):
     mProbability(1.f),
     mObjectGroup(nullptr),
     mCurrentFrameIndex(0),
+    mOffset(0,0),
```
bjorn: Please leave these out since the default constructor will also create a 0,0 point.

src/libtiled/mapreader.cpp
```diff
@@ -401,6 +401,14 @@ void MapReaderPrivate::readTilesetTile(Tileset &tileset)
     if (!probability.isEmpty())
         tile->setProbability(probability.toFloat());
 
+    // Read tile xy-offset
+    QStringRef tileoffsetx = atts.value(QLatin1String("offsetx"));
+    QStringRef tileoffsety = atts.value(QLatin1String("offsety"));
+    if (!tileoffsetx.isEmpty())
+        tile->setOffset(QPoint(tileoffsetx.toInt(), tile->offset().y()));
+    if (!tileoffsety.isEmpty())
+        tile->setOffset(QPoint(tile->offset().x(), tileoffsety.toInt()));
```
bjorn: Since `toInt` will return 0 when the string is empty, please simplify this to:
```c++
    QPoint tileOffset(atts.value(QLatin1String("offsetx")).toInt(),
                      atts.value(QLatin1String("offsety")).toInt());
    tile->setOffset(tileOffset);
```

src/libtiled/mapwriter.cpp
```diff
@@ -341,6 +342,10 @@ void MapWriterPrivate::writeTileset(QXmlStreamWriter &w, const Tileset &tileset,
                 w.writeAttribute(QLatin1String("terrain"), makeTerrainAttribute(tile));
             if (probability != 1.f)
                 w.writeAttribute(QLatin1String("probability"), QString::number(probability));
+            if (tileoffsetxy.x() != 0 || tileoffsetxy.y() != 0) {
```
bjorn: This is `if (!tileoffset.isNull()) {`

src/tiled/mapdocument.h
```diff
@@ -346,6 +347,7 @@ class MapDocument : public QObject
 
     void tileImageSourceChanged(Tile *tile);
     void tileTerrainChanged(const QList<Tile*> &tiles);
+    void tileOffsetXYChanged(Tile *tile);
```
bjorn: This signal should still be renamed to remove the XY.

src/tiled/propertybrowser.h
```diff
@@ -134,7 +135,7 @@ private slots:
         FlippingProperty,
         DrawOrderProperty,
         FileNameProperty,
-        TileOffsetProperty,
+        TilesetTileOffsetProperty,
```
bjorn: On second thought you don't need this one, you can us `TileOffsetProperty` for both the tile and the tileset, since they are mutually exclusive in the properties view.

bjorn:
Thanks for all the updates! I've found a few more things to change.

Also, it's great that this resolved the artifacts for the stamp brush, though I'm sure there are still cases where we'll need to do additional margin computation like mentioned in my previous comment.

> I did leave the "emit MapChanged()" hack in mapdocument::emitTileOffsetChanged() because recomputeDrawMargins did not force a map redraw. Please replace it with the "proper" way.

Ok I can have a look at this after merging. Essentially, the relevant places affected by tile offset changes would need to connect to the new `tileOffsetChanged` signal (which apparently overlaps with the places connected to `mapChanged`, but I think it may not be all of them).

nobody-special666:
Changes committed. Please review.
AppVeyor is complaining, but I don't know why.

bjorn:
> AppVeyor is complaining, but I don't know why.

Something broke with the Visual Studio build with the latest AppVeyor software update. I have no idea why yet, but it isn't related to your changes at least.

src/tiled/mapdocument.h
bjorn: Please still do this rename.

src/libtiled/tile.h
```diff
@@ -252,6 +256,23 @@ inline unsigned Tile::terrain() const
 }
 
 /**
+ * Returns the offset that is applied when drawing the tiles in this
+ * tileset.
```
bjorn: This comment should be updated to "Returns the offset that is applied when drawing the tile."

nobody-special666:
updated. and i put a proper email for the copyright notice.
please accept.

bjorn:
You still haven't renamed `tileOffsetXYChanged`, can you do this please?

nobody-special666:
fixed.

bjorn:
Ok, I just looked at testing this, but unfortunately it currently makes the tileset tile offset completely non-functional. All the usages of `Tile::offset` in `maprenderer.cpp`, `isometricrenderer.cpp` and `orthogonalrenderer.cpp` should still be renamed to `Tile::totalOffset`.

In addition there are still the repaint issues remaining. I will aim to look into a solution for those soon.

bjorn:
This functionality was requested [on the forum](https://discourse.mapeditor.org/t/does-tiled-allow-different-sized-tiles/2115/4) a long time ago as well. So far has been buried beneath other work, but remains a relevant feature to add.

bjorn:
Unfortunately these changes are now conflicting with other changes done on `master`. In the meantime, @Chase-san has started a new PR for this change at #3488, so I'll close this one.

I'm sorry for letting this change sit there for so long. I can't believe it's been over 6 years. So much has happened in this time, but no time was found for per-tile offsets. :-(

Issue #1173, opened 2016-01-09
Add color picker for setting transparent color of a tileset


bjorn:
It would be helpful, if the transparent color of a tileset could be chosen by using a color picker tool on the tileset image.

mrjnumber1:
i know you said this wasn't high priority, but i hope you'll reconsider. sorry i have nothing to add, just want to say it's a convenient feature to have :)

bjorn:
Thanks @Alturos for starting to work on this. Please feel free to ask _any_ questions you may have.

The way I imagined to implement this, would be to add a color picking button, which would cause a pop-up widget to open where the image is loaded and displayed. Left-clicking on it should pick the color of clicked pixel and clicking outside or right-clicking should close it again. Of course, if you have other ideas feel free to suggest or try them.

Alturos:
That's actually very close to what my plan was. I was effectively going to take a similar route as the QColourDialog. 
-Add a button next to the pick a colour button
-Pop up a window with the image in it
-Mouseover updates a little area with the colour preview, clicking seals it.
-Hit Ok to accept, or cancel to not.

That feels a little more consistent with the general UI theme, but happy to slant it more toward the quicker-to-done path you outline above.

bjorn:
I like the idea of a little preview area (and possibly a display of the hex color value next to it). But indeed, I think a single click should be enough rather than adding OK / Cancel buttons. I would try the `Qt::Popup` window flag to avoid creating a draggable window.

Alturos:
Alright, I can go with that. One question next is on window size, do we allow unlimited window size, add a scrollable area, resizable window?

My other question is on styling and formatting, do you have any guidelines published on that? So far I'm just sort of making things looks the same as they do, but there are things I'm curious about, like `auto` for example.

bjorn:
> One question next is on window size, do we allow unlimited window size, add a scrollable area, resizable window?

I think just having a reasonable maximum (maybe based on desktop size) would be enough. Beyond that the image would be rendered scaled down. Of course one could get fancy and allow zooming/panning, but I think it's rarely needed.

> My other question is on styling and formatting, do you have any guidelines published on that? So far I'm just sort of making things looks the same as they do, but there are things I'm curious about, like auto for example.

I'm pretty much using the Qt coding style, but there are some differences. Making things look like the existing code should be fine.

[...]

PR #1327, opened 2016-07-29
#1173: Colour picker

Alturos:
Alright then! Here we have it, the first pass at adding Colour Picker support to Tilesets.

This is my first time out with QT in well over a decade (and I didn't go deep then either), and my first contribution to Tiled since it moved off of Java. 

I tried to match your stylings, but I'm sure I missed something here or there (I'm rather used to just using VS with C# styles, and some of your rules are different than I'm used to), so feel free to point out anywhere that I'm being silly.

Hope it does the trick!


Alturos:
Heh, it's the innocuous changes that always get ya. Have that fixed in a jiffy.

Alturos:
Alright, I've fixed the errors and warnings I CAN fix. It looks like I picked a version of QT (5.7, which is >= 5.1, as the readme says!) that comes with a method that this PR sort of depends on (QImage.pixelColor).

bjorn:
I don't really understand what the new `QImage::pixelColor` is adding compared to `QImage::pixel`, which was available since Qt 4. At least for this use-case, the older function should be fine, right?

Alturos:
Alright, so, this is a clear cut case of "I don't really know the SDK that well yet." XD I'm not sure how I missed Pixel completely, but that change totally worked fine. Sorry for the confusion/delay there :)

DeadlyRedCube:
bjorn: For what it's worth, I believe that pixelColor, in natively returning a QColor instead of a QRgb, will give you other color spaces (HSV, CMYK) with no loss of precision if you happen to be working with images in those spaces.

Otherwise, pixel would convert to RGB and then you'd have to convert back, with all the precision loss that implies.

bjorn:
> For what it's worth, I believe that pixelColor, in natively returning a QColor instead of a QRgb, will give you other color spaces (HSV, CMYK) with no loss of precision if you happen to be working with images in those spaces.

I think you're partly right, but QImage does not support HSV or CMYK color spaces. It does however support some RGB modes where 10 bits per color are used, for which precision would get lost when converting to QRgb.

Also, the QColor(QRgb) constructor ignores the alpha channel, but this is not relevant for our use-case (if the image had an alpha channel, it wouldn't be using a magic color for transparency).

@Alturos I'll do a detailed review of the code as soon as possible (was away the past 4 days).

src/tiled/clickablelabel.cpp
```diff
@@ -0,0 +1,30 @@
+#include "clickablelabel.h"
```
bjorn: Please add the copyright header to this file.

src/tiled/imagecolorpickerwidget.cpp
```diff
@@ -0,0 +1,106 @@
...
+ImageColorPickerWidget::~ImageColorPickerWidget()
+{
+    delete mUi;
+}
+
+bool ImageColorPickerWidget::selectColor(const QString &image)
+{
+    QPixmap pix(image);
+    if (!pix.isNull())
+    {
```
bjorn: Except when defining functions, I always place the `{` on the same line. Please adjust the code accordingly.

src/tiled/imagecolorpickerwidget.cpp
```diff
@@ -0,0 +1,106 @@
...
+    }
+    return false;
+}
+
+void ImageColorPickerWidget::onMouseMove(QMouseEvent* event)
+{
+    if (!mImage.isNull())
+    {
+        mPreviewColor = mImage.pixel(event->pos().x() / mScaleX, event->pos().y() / mScaleY);
+        if(!mPreviewColor.isValid())
```
bjorn: Please put a space between `if` and `(` (affects also other places).

src/tiled/imagecolorpickerwidget.cpp
```diff
@@ -0,0 +1,106 @@
...
+    {
+        this->close();
+    }
+}
+
+void ImageColorPickerWidget::resizeEvent(QResizeEvent *)
+{
+    move(
+           findScreen().center() - rect().center()
+        );
```
bjorn: This should just be on one line.

src/tiled/imagecolorpickerwidget.cpp
```diff
@@ -0,0 +1,106 @@
...
+}
+
+QRect ImageColorPickerWidget::findScreen() const
```
bjorn: It seems this convenience function should just be called `screenGeometry`.

src/tiled/imagecolorpickerwidget.cpp
```diff
@@ -0,0 +1,106 @@
...
+{
+    QDesktopWidget wind;
+    return wind.availableGeometry(wind.screenNumber(this));
```
bjorn: Actually this should be:
```
return QApplication::desktop()->availableGeometry(this);`
```
And maybe it's not worth putting that in a member function.

src/tiled/imagecolorpickerwidget.h
```diff
@@ -0,0 +1,73 @@
...
+class ImageColorPickerWidget : public QWidget
+{
+    Q_OBJECT
+
+public:
+    ImageColorPickerWidget(QWidget *parent = 0);
+    ~ImageColorPickerWidget();
+
+    bool selectColor(const QString &image);
+    QColor mSelectedColor;
```
bjorn: Please make this member private.

src/tiled/newtilesetdialog.cpp
```diff
@@ -62,6 +63,8 @@ NewTilesetDialog::NewTilesetDialog(QWidget *parent) :
     mNameWasEdited(false)
 {
     mUi->setupUi(this);
+    mPopup = new ImageColorPickerWidget(mUi->dropperButton);
+    mPopup->setWindowFlags(Qt::Popup);
```
bjorn: Since the `ImageColorPickerWidget` is always meant to be used as a popup, please move the setting of this window flag there. Also, please construct it on-demand in the `pickColorFromImage` function.

src/tiled/imagecolorpickerwidget.h
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * A popup widget for selecting a colour from an image.
+ */
+class ImageColorPickerWidget : public QWidget
+{
+    Q_OBJECT
+
+public:
+    ImageColorPickerWidget(QWidget *parent = 0);
```
bjorn: Please write `nullptr` rather than `0`.

src/tiled/newtilesetdialog.ui
```diff
@@ -275,9 +275,23 @@
...
+       <widget class="QToolButton" name="dropperButton">
+        <property name="enabled">
+         <bool>false</bool>
+        </property>
+        <property name="text">
+         <string/>
+        </property>
+        <property name="icon">
+         <iconset resource="tiled.qrc">
+          <normaloff>:/images/22x22/stock-tool-by-color-select.png</normaloff>:/images/22x22/stock-tool-by-color-select.png</iconset>
```
bjorn: I've just pushed `:/images/16x16/stock-tool-color-picker-16.png`, please use that icon instead.

src/tiled/imagecolorpickerwidget.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+void ImageColorPickerWidget::onMouseRelease(QMouseEvent * event)
+{
+    if(event->button() == Qt::MouseButton::LeftButton)
+    {
+        if(!mImage.isNull())
+        {
+            mSelectedColor = mPreviewColor;
+            emit colorSelected(mSelectedColor);
+            this->close();
```
bjorn: There should be no need for `this->`, right?

src/tiled/imagecolorpickerwidget.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+        mPreviewIcon.fill(mPreviewColor);
+        mUi->preview->setPixmap(mPreviewIcon);
+    }
+    else
+    {
+        mPreviewColor = mSelectedColor;
+    }
+
+    event->accept();
```
bjorn: You don't need this, since events are accepted by default.

bjorn:
Alright, I'm sorry for all the small comments, and that I don't have my coding style well defined somewhere.

There some other changes I would make, like deriving the ImageColorPickedWidget from QFrame rather than QWidget, so that it can have a border (I'd probably use Panel, Plain).

I'd also try to avoid actually creating a scaled down version of the image, and instead rather than using a QLabel, implement a custom QWidget that paints the image scaled down when needed, which can then also take care of the mouse handling.

But unless you want to, there is no need to make these changes. The popup is already functional after all.

If you can still make the other requested changes by tomorrow I could merge it before the string freeze for Tiled 0.17.

Alturos:
Hey, thanks, it's cool. I should be able to get this in today. :)

Alturos:
Alright, that should be everything outside functionally rewriting the presentation. My QT know-how isn't at the level that I'm confident I could get that done this afternoon, so if it's cool, I'll leave it as is.

src/tiled/newtilesetdialog.cpp
```diff
@@ -90,15 +88,15 @@ NewTilesetDialog::NewTilesetDialog(QWidget *parent) :
...
     mUi->imageGroupBox->setVisible(tilesetType == 0);
     updateOkButton();
 }
 
 NewTilesetDialog::~NewTilesetDialog()
 {
     delete mUi;
-    delete mPopup;
+    if (mPopup != nullptr)
+        delete mPopup;
```
bjorn: Just a note, you never need to check a pointer before deleting it. The `delete` operator is by design safe to use on a null pointer.

src/tiled/newtilesetdialog.cpp
```diff
@@ -284,6 +282,11 @@ void NewTilesetDialog::updateOkButton()
  */
 void NewTilesetDialog::pickColorFromImage()
 {
+    if (mPopup != nullptr) {
```
bjorn: I think you didn't test this, because this check is the wrong way around. Also, it should have been written shorter as `if (!mPopup) {`.

bjorn: On second thought, you probably did test it, but it worked by chance because `mPopup` isn't being initialized to `nullptr`.

src/tiled/clickablelabel.cpp
```diff
@@ -0,0 +1,50 @@
+/*
+ * clickablelabel.cpp
+ * Copyright 2009-2016, Thorbjørn Lindeijer <thorbjorn@lindeijer.nl>
```
bjorn: Here should have been your name of course. :-)

bjorn:
Thanks for your contribution and for all the corrections! Of course, you can now claim [the small reward](https://www.bountysource.com/issues/29732420-add-color-picker-for-setting-transparent-color-of-a-tileset) on Bountysource.

I hope you'll be looking for more things to improve! :-)

src/tiled/clickablelabel.cpp
Alturos: I wasn't sure on that!  Now I know for next time :smile:

Alturos:
Awesome! I wanted this one to go all the way through before I picked up any others, just in case.

bjorn:
Now I've also made the layout improvements I was proposing, along with some other fixes (change a70307838605bfc6bafe92a1d85fbbe7857a6078).

Issue #1286, opened 2016-05-28
Cannot open json files


Coconaut11:
I save my map as JSON file, I close tiled, and if I want to open the map again, I can't!!!

Tiled Says:

Error, It's not a map file.
Line 1, Column 1

[The File](https://bitbucket.org/Coconaut_/indie-game/src/8e3a6faa9ebb792f2b09a2ecb930101ae0d5d9f0/Survival%20Zombie/res/maps/testMap.tmx?at=master)

bjorn:
If you link into your repository, please make sure it is publicly viewable. Right now I get:

> Access denied
>
> Use the links at the top to get back.

Coconaut11:
ok Here is it:

{"height":16,
 "layers":[
        {
[...]

PR #1328, opened 2016-08-03
Added extension filter mismatch warning to 'Save As' action

IMMZ:
This is first-try solution for #1286 . Now when filter doesn't support specified extension user will see the message box with question.

### DIFF
--- a/src/tiled/mainwindow.cpp
+++ b/src/tiled/mainwindow.cpp
@@ -916,6 +916,16 @@ bool MainWindow::saveFileAs()
         return false;
 
     MapFormat *format = helper.formatByNameFilter(selectedFilter);
+
+    if ((!format && !TmxMapFormat().supportsFile(fileName)) || (format && !format->supportsFile(fileName))) {
+        QMessageBox::StandardButton answer = QMessageBox::question(this, tr("Wrong extension"),
+            tr("The file extension does not match the chosen file type. "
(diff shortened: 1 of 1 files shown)


bjorn:
Hmm, the problem is that we can't use `supportsFile` for this, which is expected to be used only with files that exist (since some plugins may inspect the contents of the file rather than the file extension). I think we have to add a `supportsFileName` function.

bjorn:
Alternatively, we could parse the extensions out of the `nameFilter`.

IMMZ:
@bjorn I'd prefer the 1st solution. Please provide me some help: what extensions (or maybe full filenames) should be correct for next formats?

- TMWPlugin
- TEnginePlugin
- DefoldPlugin

bjorn:
I had already added the `supportsFileName` virtual function to the `FileFormat` interface, but I didn't like that it affected so many plugins and it felt redundant since there was already this `nameFilters` function. In the end I implemented a `fileNameMatchesNameFilter` utility function. I've also changed the dialog a bit to use the "informative text" property. Pushed as change 8c312b4dfb8b58134c019b2187feb8023a49fc8c.

Thanks anyway for helping with this! I'm just in a bit of a rush because I want to announce the string freeze for the next release, so I took it over.

IMMZ:
Ok, good.

PR #1332, opened 2016-08-10
Update: Turkish translations.

nuriu:
Just updated new entries.

Thanks for keep working.


bjorn:
Looks good, thanks for the update!

PR #1333, opened 2016-08-11
Updated Russian translation.

IMMZ:
All new strings were translated.

### DIFF
--- a/translations/tiled_ru.ts
+++ b/translations/tiled_ru.ts
@@ -653,7 +653,7 @@
     <message>
         <location line="+8"/>
         <source>Paste &amp;in Place</source>
-        <translation>Вставить &amp;в месте</translation>
+        <translation>Вставить на &amp;место</translation>
     </message>
     <message>
         <location line="+3"/>
@@ -1374,7 +1374,7 @@ Line %1, column %2</source>
         <location line="+27"/>
         <location line="+6"/>
         <source>Theme</source>
-        <translation>Оформление</translation>
+        <translation>Тема</translation>
     </message>
     <message>
         <location line="+20"/>
@@ -1384,7 +1384,7 @@ Line %1, column %2</source>
     <message>
         <location line="+5"/>
         <source>Fusion</source>
-        <translation type="unfinished"></translation>
+        <translation>Fusion</translation>
     </message>
     <message>
         <location line="+5"/>
@@ -1394,7 +1394,7 @@ Line %1, column %2</source>
     <message>
         <location line="+8"/>
         <source>Selection color:</source>
-        <translation>Выбор цвета:</translation>
+        <translation>Цвет выделения:</translation>
     </message>
     <message>
         <location line="+10"/>
@@ -2636,7 +2636,7 @@ Line %1, column %2</source>
     <message>
         <location line="+6"/>
         <source>Open Containing Folder...</source>
-        <translation>Открыть содержащую папку...</translation>
+        <translation>Открыть папку, содержащую карту...</translation>
     </message>
     <message>
         <location line="+139"/>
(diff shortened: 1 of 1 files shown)


IMMZ:
Oops, sorry, extension mismatch has incomplete translation. DO NOT MERGE! - I'll fix it now.

IMMZ:
Fixed and re-checked. Now all OK and ready for merge.

translations/tiled_ru.ts
```diff
@@ -4733,8 +4733,8 @@ Please select specific format.</source>
     <message numerus="yes">
         <location filename="../src/tiled/flipmapobjects.cpp" line="+39"/>
         <source>Flip %n Object(s)</source>
-        <translation type="unfinished">
-            <numerusform></numerusform>
+        <translation>
+            <numerusform>Повернуть %n объект(ов)</numerusform>
```
bjorn: It seems Russian has space for three plural forms, but here you're only using one. Probably you could split it up and avoid the "(ов)" part, right?

The "(s)" is only part of the source text out of necessity, because there is only one version of source string.

bjorn:
Thanks for updating the Russian translation! Normally @semtiko did this, but it seems you've beaten him to it. Maybe @semtiko would like to do a quick review as well?

I did find one issue with a plural-enabled string, please have a look at it.

translations/tiled_ru.ts
IMMZ: Don't sure if I understand correctly but I try to reply. In this case it could be 2 ways of Russian word endings depending of actual amount of objects. Example:
- 4 объект(а)
- 5 объект(ов)

In such case what type of fix you expect from me?
- Merge both endings: Повернуть %n объект(ов/а)
- Choose one default: Повернуть %n объектов
- Use no ending: Повернуть %n объект (this sounds strange because such form is only correct for singular forms.

bjorn: What I [read here](https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_and_Plurals#Plural_rule_.237_.283_forms.29) is that Russian has three plural forms concerning numbers:

1. ends in 1, excluding 11
1. ends in 2-4, excluding 12-14
1. everything else

I don't know what application you used for the translation, but if it is Qt Linguist you should see three fields for the translation. You should use these fields to put in the appropriate translation for each of the above cases. This way, there should be no need for things like "(ов/а)" in the string. Judging by your examples, the translations should be:

1. Повернуть %n объект
1. Повернуть %n объекта
1. Повернуть %n объектов

IMMZ:
@bjorn, I've fixed it. Really, some old examples from @semtiko already use three forms (' ', 'а' and 'ов'). Also I tried to use more appropriate translation for flip. I'll look by this PR (rebase + push) while we wait for @semtiko review.

bjorn:
I noticed @semtiko already pushed some translation updates to Weblate on Sunday. Since he was first, I've pushed those updates to the repository. @IMMZ, if you rebase your changes you could resolve conflicts in the translation file, and judge which version is better if there are any differences. If you find anything to improve please update this pull request. If not you can close it of course.

IMMZ:
@bjorn I left a best variants of our commits in this PR + a few improvements:
- Translated Theme as word-to-word 'Theme'  in Russian. I think this would be better because the users often see this word in Windows' or KDE's/XFCE's/etc or other apps' settings.
- Fixed incorrect translation of 'Selection color' in 'Theme'. It was as 'Choose color' in Russian and it is hard to understand what color you have to choose.
- Made second words in 'Flip %n Objects' lowercase. There are already other similar cases where the all words but 1st are lowercase. For example 'Duplicate %n Object(s)' in AbstractObjectTool. For unified GUI here we should use such rule.

bjorn:
@IMMZ Alright, thanks for the additional improvements and for explaining them! Sounds reasonable to me, so I'll just merge this.

IMMZ:
@bjorn Good :+1:

Issue #1329, opened 2016-08-05
Problem with snapping to pixels in the collision editor


RareBanjo:
I am expecting this to be a problem that is happening because I am an idiot and new to the program. But when I edit the collisions for my tiled I holding down CTRL and making a collision either doesn't select anything or select the whole thing. Then when I start at the bottom and drag up to make one it just makes a flat line. I am pretty new and would appreciate some help.

bjorn:
This is because Control snaps to tiles, not to pixels. I realize this isn't very useful, so it's something I should consider improving eventually.

For now, you can enable "Snap to Fine Grid" in the "View" menu to get more finer-grained snapping. If you want it to snap to pixels, you need to set the number fine grid divisions in the Preferences to match the number of pixels on your tiles.

PR #1334, opened 2016-08-12
Implemented 'Snap to pixels'.

IMMZ:
Tried to implement #1329 . @bjorn I'm waiting for your critics/notes. These are my notes about this PR:
- Don't sure if I correctly modified SnapHelper::toggleSnap(). Give some advices how it should work by the logic please if I done it wrong.
- Should I also update all *.ts and include them into this commit?

### DIFF
--- a/src/tiled/snaphelper.cpp
+++ b/src/tiled/snaphelper.cpp
@@ -1,6 +1,7 @@
 /*
  * snaphelper.cpp
  * Copyright 2015, Thorbjørn Lindeijer <thorbjorn@lindeijer.nl>
+ * Copyright 2016, Mamed Ibrahimov <ibramlab@gmail.com>
  *
  * This file is part of Tiled.
  *
@@ -32,6 +33,7 @@ SnapHelper::SnapHelper(const MapRenderer *renderer,
     Preferences *preferences = Preferences::instance();
     mSnapToGrid = preferences->snapToGrid();
     mSnapToFineGrid = preferences->snapToFineGrid();
+    mSnapToPixels = preferences->snapToPixels();
 
     if (modifiers & Qt::ControlModifier)
         toggleSnap();
@@ -55,6 +57,9 @@ void SnapHelper::snap(QPointF &pixelPos) const
             tileCoords = tileCoords.toPoint();
         }
         pixelPos = mRenderer->tileToPixelCoords(tileCoords);
+    } else if (mSnapToPixels) {
+        QPointF screenPos = mRenderer->pixelToScreenCoords(pixelPos);
+        pixelPos = mRenderer->screenToPixelCoords(screenPos.toPoint());
     }
 }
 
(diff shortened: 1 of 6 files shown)


bjorn:
> Should I also update all *.ts and include them into this commit?

No. I update the .ts files only before each new feature release, after which there is a "string freeze" until the release. After the release, strings are again changed/added but the .ts files are not updated until we are again close to a new release. People installing snapshot builds will be used to translations not being entirely complete.

IMMZ:
Ok.

src/tiled/snaphelper.cpp
```diff
@@ -55,6 +58,8 @@ void SnapHelper::snap(QPointF &pixelPos) const
             tileCoords = tileCoords.toPoint();
         }
         pixelPos = mRenderer->tileToPixelCoords(tileCoords);
+    } else if (mSnapToPixels) {
+        pixelPos = pixelPos.toPoint();
```
bjorn: I think the problem with this is, that even though internally this is called the "pixel position", it is not actually the "screen pixels" that the user would expect "snap to pixel" to work on. I think you'll need to do:
```c++
QPointF screenPos = mRenderer->pixelToScreenCoords(pixelPos);
pixelPos = mRenderer->screenToPixelCoords(screenPos.toPoint());
```

src/tiled/snaphelper.cpp
```diff
@@ -41,6 +43,7 @@ void SnapHelper::toggleSnap()
 {
     mSnapToGrid = !mSnapToGrid;
     mSnapToFineGrid = false;
+    mSnapToPixels = false;
```
bjorn: I think we shouldn't turn off pixel snapping with the modifier. More likely pixel snapping is something that will stay enabled or disabled for any given project, so turning it off because somebody temporarily wants to suppress tile snapping is probably not expected in general.

bjorn:
Seems all fine apart from my two comments.

IMMZ:
Code updated + rebased.

bjorn:
Awesome, thanks!

IMMZ:
Glad to help! :+1:

eevee:
I just noticed this and started using it — _thank you_ :)

One observation: if both "snap to grid" and "snap to pixels" are turned on, I seem to only snap to the grid.  This is backwards from "snap to grid" and "snap to fine grid", where the fine grid takes precedence.

Or maybe the three modes should be exclusive.  I suppose it makes sense to snap to _both_ the fine grid and pixels, but offhand it doesn't seem particularly useful.

Also, "To" is capitalized in the pixels label but not the other two.

bjorn:
@eevee Thanks for the feedback! I think you're right that these snapping options should be mutually exclusive. I'll also make sure to fix the capitalization.

bjorn:
@eevee Changed in 3409e330b7d7feb8a6746d7d3e21dad901ca40ec.

fosskers:
For those wandering in here from the future, you can turn this feature on from the top-level menus:

> View -> Snapping -> Snap to Pixels

PR #1335, opened 2016-08-15
Removed script creation part

kekdck:
Removed from defold plugin script creation part. Since it was needed only to me.


bjorn:
Good idea, thanks! Could you update it to also remove the .script file from the list of output files in `DefoldPlugin::outputFiles`?

bjorn:
Ok I've amended the change with that and pushed it as 2465cdc.

Issue #1341, opened 2016-08-21
New file dialog box should have Base64 (gzip compressed) and XML tile layer formats


AlexBezuska:
Linux version 2016-05-25

![https://i.imgur.com/3b45AOi.jpg](https://i.imgur.com/3b45AOi.jpg)
![https://i.imgur.com/p4p832N.jpg](https://i.imgur.com/p4p832N.jpg)

benwiley4000:
Addressed by https://github.com/bjorn/tiled/pull/1342

bjorn:
I intensionally removed the other options from the New Map dialog in Tiled 0.16. The [TMX changelog](http://doc.mapeditor.org/reference/tmx-changelog/) notes:

> The pure XML and Gzip-compressed tile layer data formats were deprecated, since they didn't have any advantage over other formats. Remaining formats are CSV, base64 and Zlib-compressed layer data.

Of course, I could not remove them from the Map Properties because already created maps may still be using those formats. However, I'd like to remove them there eventually. If you need these formats for compatibility reasons, please tell me why or which software you're using so we can help them upgrade.

Thanks @benwiley4000 for so quickly addressing this issue, but I'll close that pull request.

benwiley4000:
@bjorn this makes plenty of sense. I didn't realized zlib compression would outperform gzip, but it seems it does.

bjorn:
@benwiley4000 Right, while gzip and zlib use the same compression format, zlib is slightly lighter because it avoids some overhead of gzip that is meant for dealing with file data (http://www.gzip.org/zlib/zlib_faq.html#faq19). Through the zlib C API it is also more straight-forward to work with zlib when dealing with data in memory.

The reason Tiled supports both is because it first supported gzip and zlib was added later. Gzip remained for compatibility reasons.

I'm still curious why @AlexBezuska was missing these options, but I'll close this issue now.

PR #1342, opened 2016-08-21
All layer formats now available in new tileset dialog.

benwiley4000:
I added Base64 gzip and XML format options to the new tileset dialog window. Previously they were only available after initial setup via map properties.


bjorn:
See comment at #1341.

PR #1347, opened 2016-09-02
Replace %mappath with absolute path to map file's directory

zcabjro:
Useful for team projects where map and/or script locations differ


bjorn:
Thanks! :-)

Issue #1304, opened 2016-06-14
Changing property types without deleting them


karai17:
Currently, to change a property, you must delete it and re-add it with the new type and value. While this generally makes sense, it should be possible to convert between types to produce sensible values:

* int 0 -> bool false
* int 0 -> string "0"
* bool false -> string "false"
* bool false -> int 0
* string "false" -> bool false
* string "false" -> int 0

or other such things.

Alturos:
I'm curious, for this ticket, in the case of data that doesn't convert, would nuking the data be acceptable?

bjorn:
I think it should either leave out or disable menu items showing types to which no conversion is possible.

When opening the menu, Tiled could use `QVariant::convert` to attempt a conversion to each supported type and check the return value.

PR #1357, opened 2016-09-12
Changing property type context menu #1304

DeXP:
This actually does not changes the type, but deletes and adds the property

### DIFF
--- a/src/tiled/propertiesdock.cpp
+++ b/src/tiled/propertiesdock.cpp
@@ -91,6 +91,9 @@ PropertiesDock::PropertiesDock(QWidget *parent)
 
     setWidget(widget);
 
+    mPropertyBrowser->setContextMenuPolicy(Qt::CustomContextMenu);
+    connect(mPropertyBrowser, SIGNAL(customContextMenuRequested(const QPoint&)),
+        this, SLOT(showContextMenu(const QPoint&)));
     connect(mPropertyBrowser, SIGNAL(currentItemChanged(QtBrowserItem*)),
             SLOT(currentItemChanged(QtBrowserItem*)));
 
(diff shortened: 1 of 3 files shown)


bjorn:
Thanks for working on this!

> This actually does not changes the type, but deletes and adds the property

Was there a reason why you did not use the `SetProperty` command? I'm guessing maybe there is a problem with the property browser when a property changes type? If so I'd prefer to fix that instead of doing a remove/add.

When using `SetProperty` instead of `removeProperty` and `addProperty`, I think the implementation could move into the `PropertyBrowser` class.

DeXP:
> Was there a reason why you did not use the SetProperty command? I'm guessing maybe there is a problem with the property browser when a property changes type?

Yes. The GUI was not updating for new type. 

Now `removeProperty` and `addProperty` are in `PropertiesDock` class. I think, code for changing property type must be in that class too. But I can't call `updateCustomProperties` in my code since it's a `PropertyBrowser` private method...

bjorn:
> But I can't call updateCustomProperties in my code since it's a PropertyBrowser private method...

Yes, I think it should stay like that. But the `PropertyBrowser` should change the type when necessary in `PropertyBrowser::propertyChanged`. Would you like to give that a try?

src/tiled/propertiesdock.cpp
```diff
@@ -293,6 +296,42 @@ void PropertiesDock::renameProperty(const QString &name)
...
+    QAction* selectedItem = contextMenu->exec(globalPos);
+    if (selectedItem) {
+        unsigned int selectedType = selectedItem->data().toInt();
+        if (value.canConvert(selectedType)) {
+            value.convert(selectedType);
+            removeProperty();
+            addProperty(name, value);
+        }
+    }
+    delete contextMenu;
```
bjorn: Please allocate the `QMenu contextMenu` on the stack instead of on the heap, then you can drop this line.

src/tiled/propertiesdock.cpp
```diff
@@ -91,6 +91,9 @@ PropertiesDock::PropertiesDock(QWidget *parent)
 
     setWidget(widget);
 
+    mPropertyBrowser->setContextMenuPolicy(Qt::CustomContextMenu);
+    connect(mPropertyBrowser, SIGNAL(customContextMenuRequested(const QPoint&)),
+        this, SLOT(showContextMenu(const QPoint&)));
```
bjorn: For new code I prefer the new connect style:
```c++
    connect(mPropertyBrowser, &PropertyBrowser::customContextMenuRequested,
            this, &PropertiesDock::showContextMenu);
```

src/tiled/propertiesdock.cpp
```diff
@@ -293,6 +296,42 @@ void PropertiesDock::renameProperty(const QString &name)
...
+        if (value.canConvert(toType) && (value.type() != toType)) {
+            QAction* action = contextMenu->addAction(typeToName(toType));
+            action->setData(toType);
+        }
+    }
+
+    QAction* selectedItem = contextMenu->exec(globalPos);
+    if (selectedItem) {
+        unsigned int selectedType = selectedItem->data().toInt();
+        if (value.canConvert(selectedType)) {
```
bjorn: This check no longer makes sense since it already happened while adding items to the menu. Actually, rather than storing the selected type, what about storing the converted value directly with the action?

src/tiled/propertiesdock.cpp
```diff
@@ -293,6 +296,42 @@ void PropertiesDock::renameProperty(const QString &name)
...
+    QVariant value = object->property(name);
+
+    QMenu *contextMenu;
+    contextMenu = new QMenu(mPropertyBrowser);
+    contextMenu->addSection(tr("Convert to:"));
+
+    QList<unsigned int> convertTo;
+    convertTo << QVariant::Bool << QVariant::Color << QVariant::Double << filePathTypeId() << QVariant::Int << QVariant::String;
+    foreach (const unsigned int &toType, convertTo){
+        if (value.canConvert(toType) && (value.type() != toType)) {
```
bjorn: Could be:
```c++
    QVariant copy = value;
    if (value.type() != toType and copy.convert(toType)) {
        QAction *action = contextMenu->addAction(typeToName(toType));
        action->setData(copy);
    }
```

src/tiled/propertiesdock.cpp
```diff
@@ -91,6 +91,11 @@ PropertiesDock::PropertiesDock(QWidget *parent)
 
     setWidget(widget);
 
+    contextMenu = new QMenu(mPropertyBrowser);
```
bjorn: Please allocate this on the stack only where needed (in `showContextMenu`). Just put it like that:
```c++
    QMenu contextMenu(mPropertyBrowser);
```

bjorn:
Works good, thanks! I'll reward you the small bounty on BountySource. :-)

bjorn:
I made some small changes and added also Rename and Remove to this context menu in change 503eb6c.

PR #1360, opened 2016-09-15
Fix some Chinese translations

buckle2000:


bjorn:
Thanks for the fixes! Can you change the target branch for this to the 0.17 branch?

buckle2000:
Sure

bjorn:
Ok, changing the target branch is not so easy, because you would also need to change the source branch to the `0.17` branch in your repository. Since changing the source branch is not possible on GitHub you'd need to open a new pull request for that.

But, this isn't needed anymore because I've cherry-pick the change to the `0.17` branch now as commit 441fc4d.

Thanks for the fixes!

PR #1362, opened 2016-09-18
add drag-and-drop toggle mode to reorder tiles in the tilesetdock

nobody-special666:
This adds the ability to reorder tiles in Tiled.
it adds an "order" attribute to the tiled.xml file or external tileset tsx file.
It updates the TilesetView class and TilesetModel to add drag-n-drop support.
Reorder mode must be toggled by clicking on the "Reorder Tiles" button.
The reorder button is disabled until the external tileset is imported.
Reorder mode is turned off when a new map is selected.
It only supports moving 1 tile at a time. Undo/Redo is also supported.
Please consider for review and addition.


nobody-special666:
anybody home?

bjorn:
Sorry for the wait. I was on holiday for a week as I wrote in my last development update, and when I got back I was rather swamped and still haven't really caught up.

In addition, this change is relatively large and I wonder if it's really the right implementation. For sure there are people who want more control over the position of their tiles, but I'm not sure this is really the right approach. I'm more leaning to a way to define arbitrary "tile views", which are really just maps with a tile layer, where you can organize tiles from various tilesets more easily in order to select them from there.

So right now I'm unsure about adding this reordering option. What is your use-case that makes this solution particularly useful?

nobody-special666:
you may import tiles from a tilesheet or import a directory of tiles into a collection in whole, but you might not like the order and want to rearrange them. your tileviews sounds like a better idea. i hope we see it soon. disregard this then. thanks.

bjorn:
This feature actually got implemented somewhat differently in #2983.

The "tile views" feature is still on the table...

PR #1366, opened 2016-09-25
Always enable the "Object Properties" entry in the context menu

Ablu:
It is now possible to edit multiple properties at once.


bjorn:
Thanks! Submitted as 441b7a2a9e82fb1a887636debabce8bd6364041a.

Issue #397, opened 2013-03-26
libtiled-java and tmxviewer-java not in maven repository


jjYBdx4IL:
Hello!

Could you please put those two libraries into the maven repository? Would make handling licensing and dependencies much easier :-).

If you need me to write the pom.xml file, I'd be glad to help out.

Regards,
Mark

tomcashman:
@bjorn or @mikepthomas I saw Maven Central publishing was merged in but seems like it's still not published with each release?

bjorn:
@tomcashman I'm actually not sure what's required for this. I hope @mikepthomas can help.

mikepthomas:
No releases have yet been made to maven central repository as of yet...
You are able to get a snapshot of release 0.17-SNAPSHOT from when I implemented on Oct 17 by adding the following to your maven pom: 
```
    <dependencies>
        <dependency>
            <groupId>org.mapeditor</groupId>
            <artifactId>libtiled</artifactId>
            <version>0.17-SNAPSHOT</version>
            <type>jar</type>
        </dependency>
    </dependencies>
    <repositories>
        <repository>
            <id>sonatype-oss</id>
            <url>https://oss.sonatype.org/content/repositories/snapshots/</url>
            <snapshots>
                <enabled>true</enabled>
            </snapshots>
        </repository>
    </repositories>
```
Ideally the java libs should be built out as part of the ci process using maven but but am unfamilliar with the current process (using Travis I believe...) also the assets deployed to the oss snapshots repository are currently signed using my personal pgp key, I'm not too sure if @bjorn would be happy with this as I'd expect that he would want them to be signed with his own...

tomcashman:
What mvn commands are needed to deploy to Maven Central?

bjorn:
Hmm, I'd have to set up PGP again. I did this several times now, but I keep locking myself out of my own keys by forgetting their passphrase, since I'm inclined to pick a long one and then never use it.

Regarding the CI system, it could be integrated with either Travis CI (Linux or macOS) or AppVeyor (Windows), depending on what would be the easiest. Both Travis CI and AppVeyor seem to have Maven preinstalled on their build image.

mikepthomas:
Deploying to snapshots can easilly be done manually, I would have to look into how it integrate with the CI server (whichever is easiest) as for releasing to central this can be done with any of the deployed snapshots [using this guide](http://central.sonatype.org/pages/releasing-the-deployment.html).

bjorn:
[...]

PR #1372, opened 2016-10-01
Convert Java projects to Maven

mikepthomas:
Converted libtiled-java and tmxviewer to maven projects
Upgraded to Java 7
Deprecated use of old classes and methods
Add support for multi linked images in Tilesets
Auto generate licenses and missing JavaDocs


Ablu:
Apparently the CI checks broke. I think the Java parts are supposed to be build by Qbs now...

It seems to have some issues to find some libraries:
~~~
ERROR: Error: Process 'C:\Progra~1\Java\jdk1.8.0/bin/java' finished with exit code 1. The standard error output was:
C:\projects\tiled\util\java\libtiled-java\src\test\java\tiled\io\MapReaderTest.java:35: error: package junit.framework does not exist
import junit.framework.TestCase;
                      ^
C:\projects\tiled\util\java\libtiled-java\src\test\java\tiled\io\MapReaderTest.java:40: error: cannot find symbol
public class MapReaderTest extends TestCase {
                                   ^
  symbol: class TestCase
C:\projects\tiled\util\java\libtiled-java\src\test\java\tiled\io\MapReaderTest.java:46: error: cannot find symbol
        assertTrue(reader.accept(new File("example.tmx")));
        ^
~~~

I do not think that Qbs supports Maven at the moment...

Was there any specific reason for the switch to Maven?

mikepthomas:
Hi Eric, I have been using libtiled-java for a personal project written in Java using Maven for build and dependency management. I noticed you had an open issue that the library is not in the Maven central repository (https://github.com/bjorn/tiled/issues/397). The Java project would need to be converted to Maven to facilitate this.

bjorn:
Does libtiled-java now depend on junit? Could this dependency be avoided, for example by moving MapReaderTest.java to a separate directory?

mikepthomas:
JUnit is just used for the unit tests, I've declared it as a dependency in test scope in the Maven POM... If Maven is used for build, it will download it from Maven central. If Qbs cannot use Maven to build the JARs I can try re-instate the Ant build scripts to work alongside the new project structure if that would help?

mikepthomas:
Unfortunately I don't know why the CI build is failing... the JARs seem to build ok when using both Maven and the Ant scripts now. I've never used Qbs before, but it looks like the junit JAR is not in the classpath when trying to build.

bjorn:
> If Qbs cannot use Maven to build the JARs I can try re-instate the Ant build scripts to work alongside the new project structure if that would help?

Qbs is building the .jar files directly, it does not depend on the Maven or Ant build scripts. Feel free to remove the Ant script if it isn't useful to anybody anymore now that a Maven script is provided.

util/java/libtiled-java/src/main/java/tiled/core/Map.java
```diff
@@ -304,8 +402,19 @@ public String getFilename() {
      * Returns a vector with the currently loaded tileSets.
      *
      * @return Vector
+     * @deprecated Use getMapTileSets instead.
```
bjorn: Do we really need to do this? So far there aren't any compatibility guarantees and it should be a small thing to fix for projects upgrading to a new libtiled-java if we just change the return type, right?

mikepthomas: I wasn't too sure what level of API compatability you were after, If there's no compatibility guarantees, i've removed the deprecated methods and just changed the return type as suggested, I've also removed the commits that added the Ant Scripts and changes to junit.

As for the comment for adding support for multi linked images in tilesets, I have added a new method in TMXMapWriter, "writeImage" and a new property in Tile (with getter and setter), "source" that allows to write individual tiles to a tilemap as links to external files. This creates an identical format to using Type "Collection of Images" when creating a new Tileset in the Tiled application v0.17. e.g:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<tileset name="TileSet" tilewidth="32" tileheight="32" tilecount="2123" columns="0">
 <tile id="0">
  <image width="32" height="32" source="tiles/0.png"/>
 </tile>
</tileset>
```

bjorn:
> I've also removed the commits that added the Ant Scripts and changes to junit.

Something seems to have gone wrong here since now the history looks even stranger. There's a duplicated commit for "Update Qbs build scripts to point at new java root package" and a commit to remove the deprecated methods that is at the same time a merge commit.

Are you familiar enough with git to see about fixing this up or shall I try to nudge things in shape before merging it?

mikepthomas:
Sorry I only rebase removed them yesterday, rewriting history of pushed commits us usually frowned upon... I've now removed the commits from the history too

bjorn:
Thanks a lot for all the changes! Will you be able to close #397 now?

mikepthomas:
Hi @bjorn, I should be able to get it as far as the development version repository (Snapshots) A few minor changes will need to be made to get it into central. I'm going to apply for the namespace so that it can be deployed, if you could register an account from https://issues.sonatype.org/secure/Signup!default.jspa and let me know your username so you can also have admin access, I'll get the ball rolling

bjorn:
@mikepthomas Alright, I've signed up as `bjorn`. Thanks!

Issue #1377, opened 2016-10-16
Misaligned grid and selection after offset change


nykm:
Hello, I did not find any open issues relating exactly to this one.

Problem:

1. Open _sewers.tmx_ example
2. Select tiles (i.e. _Magic Wand_ tool)
3. Select _Offset Layers_ tool
4. Drag the Bottom layer with left mouse button to change the offset
5. Tile selection seem to be misaligned
6. Switch to Top layer
7. Both grid and selection seem to be misaligned

However, this misalignment is corrected when the Bottom layer is selected again which tells me that this is a bug (see the gif).

![offsetissue](https://cloud.githubusercontent.com/assets/9414577/19416191/b51b5bac-9392-11e6-8970-a65083dabd81.gif)

So I'm wondering, is this a bug or a feature?

Versions:
- Tiled 2016.05.25 for Linux (64-bit) 
- Compiled version from Github sources ~week 41 / 2016

bjorn:
The grid seems to be doing the right thing. It is displaying the grid for the currently selected layer, so it is shifted when you select the layer that has an offset.

But there's a bug in that the selection is not immediately moving along with the layer while you apply an offset to it.

PR #1378, opened 2016-10-16
Fix out-of-sync tile selection during offset change

nykm:
Here's my proposal for fixing the misalignment bug and it seems to be working for me. Issue: #1377

### DIFF
--- a/src/tiled/tileselectionitem.cpp
+++ b/src/tiled/tileselectionitem.cpp
@@ -39,6 +39,8 @@ TileSelectionItem::TileSelectionItem(MapDocument *mapDocument)
 
     connect(mMapDocument, &MapDocument::selectedAreaChanged,
             this, &TileSelectionItem::selectionChanged);
+    connect(mapDocument, &MapDocument::layerChanged,
+            this, &TileSelectionItem::layerChanged);
     connect(mMapDocument, &MapDocument::currentLayerIndexChanged,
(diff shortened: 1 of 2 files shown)


src/tiled/tileselectionitem.cpp
```diff
@@ -74,6 +76,12 @@ void TileSelectionItem::selectionChanged(const QRegion &newSelection,
     update(mMapDocument->renderer()->boundingRect(changedArea));
 }
 
+void TileSelectionItem::layerChanged(int index)
+{
+    const Layer *layer = mMapDocument->map()->layerAt(index);
```
bjorn: To avoid synchronizing the position to the wrong layer (mostly when using undo), we should add this check here:
```c++
if (layer == mMapDocument->currentLayer())
```

nykm:
Is it now ok?

bjorn:
Well it works, but I really meant to add that line to the code you had:
```c++
if (layer == mMapDocument->currentLayer())
```
Because right now, you're always setting the position regardless of which layer changed, whereas with the check like above, the position is only synchronized when needed (when the current layer changed).

nykm:
Oh I see, now it should be finally fixed.

bjorn:
Alright, I've squished the commits together and applied them to the `0.17` branch as change 86a189a6a2cd16eec1c61aa08826b82d82d064b4. That way it will be included in the upcoming 0.17.1 release. Thanks!

Issue #124, opened 2011-12-15
Update libtiled-java to support map files saved with 0.8.0


davifo:
Please update the java core in the v0.8.0.

Unsupported attribute 'color' on <objectgroup> tag
Unsupported attribute 'height' on <objectgroup> tag
Unsupported attribute 'width' on <objectgroup> tag

Tank You.

PR #1383, opened 2016-10-23
Suppress unsupported attribute warnings in libtiled-java

mikepthomas:
I have added color to the ObjectGroup class to suppress unsupported attribute warning to suppress unsupported attribute warning when loading map to fix issue #124 

When saving the sewers.tmx file using Tiled v0.17.0 the width and height are no longer saved to the ObjectGroup tag so I have not added these... should they be added too to support backwards compatibility with older formats?

### DIFF
--- a/src/tiled/magicwandtool.cpp
+++ b/src/tiled/magicwandtool.cpp
@@ -58,23 +58,29 @@ void MagicWandTool::tilePositionChanged(const QPoint &tilePos)
 
 void MagicWandTool::mousePressed(QGraphicsSceneMouseEvent *event)
 {
-    if (event->button() != Qt::LeftButton)
-        return;
-
+    const Qt::MouseButton button = event->button();
     const Qt::KeyboardModifiers modifiers = event->modifiers();
+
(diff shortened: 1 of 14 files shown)


bjorn:
I find it a bit strange to have a `setColor` taking a `String` and a `getColor` returning a `Color`. Would there be any way around this? Probably the fact that the map reader is based on reflection is causing problems here? (I'd like to rip out the reflection code)

bjorn:
> When saving the sewers.tmx file using Tiled v0.17.0 the width and height are no longer saved to the ObjectGroup tag so I have not added these... should they be added too to support backwards compatibility with older formats?

No. It would again be needed because of the reflection, right? When width/height are still present in a TMX file for objectgroup, they should simply be ignored.

mikepthomas:
Sorry @bjorn Ive made a bit of a mess of this pull request, Ive closed and will raise another

bjorn:
@mikepthomas No problem, but just note that you can always force-push (`git push -f ...`) to the branch that you opened the pull request for, after doing the cleanup. I see you even re-forked Tiled, but there is really no need for that.

PR #1384, opened 2016-10-25
Up/down buttons in objects dock + pixel coordinates for objects in status bar

iskolbin:
This pull request adds up/down buttons in objects dock for manually changing object index (undo is not implemented, sorry). It also adds object coordinates in status bar in pixels.


src/tiled/mapdocument.cpp
```diff
@@ -996,6 +996,39 @@ void MapDocument::moveObjectsToGroup(const QList<MapObject *> &objects,
     mUndoStack->endMacro();
 }
 
+// TODO add undo code
```
bjorn: If you want to look into adding this, note that there is already a `ChangeMapObjectsOrder` undo command that you can use for this purpose. You can see how it is used in the `RaiseLowerHelper`.

src/tiled/mapdocument.cpp
```diff
@@ -996,6 +996,39 @@ void MapDocument::moveObjectsToGroup(const QList<MapObject *> &objects,
...
+void MapDocument::moveObjectIndex(const MapObject *object, int count)
+{
+    int index = 0;
+    ObjectGroup *group = object->objectGroup();
+
+    while (index < group->objectCount() && group->objectAt( index ) != object)
+        index++;
```
bjorn: Or just do:
```c++
int index = group->objects().indexOf(object);
```

src/tiled/mapdocument.cpp
```diff
@@ -996,6 +996,39 @@ void MapDocument::moveObjectsToGroup(const QList<MapObject *> &objects,
...
+
+    if ( count < 0 && index > 0 ) {
+        mMapObjectModel->moveObjects( group, index, index - 1, 1 );
```
bjorn: Coding style: please don't put spaces on the insides of `(` and `)`.

bjorn:
@iskolbin Do you plan to continue working on this?

iskolbin:
@bjorn Sorry lots of work recently, didn't have time to response. Thank you very much for the comments, I'll try to fix noted issues next week

bjorn:
@iskolbin Alright, no problem and good to hear!

iskolbin:
@bjorn Hello! Sorry for long delay, I've just fixed noted issues and added undo feature as you suggested (It was so easy actually, thank you for the clean code). Sorry for newbie question, but how to do "rebase the relevant commits onto my latest master and force-push to update this pull request" -- could you please explain the details. Thanks in advance:)

bjorn:
> I've just fixed noted issues and added undo feature as you suggested (It was so easy actually, thank you for the clean code).

That's great to hear! The code could be optimized I think (I'll try to have a short look at this), but overal it looks good so I'll merge it.

> how to do "rebase the relevant commits onto my latest master and force-push to update this pull request" -- could you please explain the details. Thanks in advance:)

It won't be needed since I'll "Squash and merge" the changes, but you can read up on rebasing here:

https://git-scm.com/book/en/v2/Git-Branching-Rebasing
https://help.github.com/articles/about-git-rebase/

And more specifically about pull requests:

https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request

bjorn:
Ok, while looking into optimizing the implementation I found a number of other things to improve as well, pushed as change 9ee3cf465126b5bd64ad0f629d2f81d4e00a37b1:

    * Made sure "up" and "down" refer to the stacking order, so it makes sense
      for object layers using Manual rendering order. This does mean it no
      longer matches with the Objects dock display order, but I think that the
      display order needs to be reversed.
    
    * Used the icons from the theme when available.
    
    * Set tool tips and disable the actions when no objects are selected.
    
    * Fixed strange reordering happening when some set of objects is already
      at the bottom or top. Just do nothing instead.
    
    * Used RangeSet<int> to replace the sorting and reduce the number of
      ChangeMapObjectsOrder instances.
    
    * Replaced a bunch of 'foreach' loops with C++11 for loops.

Thanks for getting this change started! Unfortunately, especially with the now necessary reversal of the object list, I think it was only the beginning. :-)

bjorn:
> Thanks for getting this change started! Unfortunately, especially with the now necessary reversal of the object list, I think it was only the beginning. :-)

Well, it was indeed just the beginning. I ended up spending almost two full days working on a `ReversingProxyModel` (d82d9e67b28038695667097945065c4eaac6bfaf) and using it to reverse the objects in the Objects view (7c11cdc8d4f0b3b0064a7f0e83f5648f150aeeb8).

I spent some of those hours trying to do it without a reversing proxy model, but that made the code so ugly once I had it finally working that I wanted to try a different approach. I'm glad I did, even though it took way longer than I had anticipated.

PR #1385, opened 2016-10-25
Updated schema to follow the TMX format page

assofohdz:
It's a beginning towards an updated format. Many things in the XSD depends on the how you use it inside Tiled, which is ambigious in the TMX format page.

### DIFF
--- a/docs/map.xsd
+++ b/docs/map.xsd
@@ -71,9 +71,7 @@
 	</xs:simpleType>
 
 	<xs:simpleType name="degreesT">
-		<xs:restriction base="xs:nonNegativeInteger">
-			<xs:maxInclusive value="360"/>
-			<xs:minInclusive value="0"/>
+		<xs:restriction base="xs:float">
 		</xs:restriction>
 	</xs:simpleType>
 
@@ -153,7 +151,7 @@
 			<xs:documentation>A hexcolor</xs:documentation>
 		</xs:annotation>
 		<xs:restriction base="xs:string">
-			<xs:pattern value="([a-fA-F0-9][a-fA-F0-9]){3}"/>
+			<xs:pattern value="#?([a-fA-F0-9][a-fA-F0-9]){3}"/>
 		</xs:restriction>
 	</xs:simpleType>
 
@@ -169,7 +167,7 @@
 			<xs:documentation>A hexcolor with alpha</xs:documentation>
 		</xs:annotation>
 		<xs:restriction base="xs:string">
-			<xs:pattern value="#([a-fA-F0-9][a-fA-F0-9]){3,4}"/>
+			<xs:pattern value="#?([a-fA-F0-9][a-fA-F0-9]){3,4}"/>
 		</xs:restriction>
 	</xs:simpleType>
 
@@ -533,8 +531,8 @@
 
 	<xs:attributeGroup name="imagelayer">
 		<xs:attribute name="name" type="nameT"/>
-		<xs:attribute name="offsetx" type="xs:integer" default="0" use="optional"/>
-		<xs:attribute name="offsety" type="xs:integer" default="0" use="optional"/>
+		<xs:attribute name="offsetx" type="xs:double" default="0" use="optional"/>
+		<xs:attribute name="offsety" type="xs:double" default="0" use="optional"/>
 		<xs:attribute name="opacity" type="opacityT" default="1" use="optional"/>
 		<xs:attribute name="visible" type="xs:boolean" default="1" use="optional"/>
 	</xs:attributeGroup>
@@ -589,8 +587,8 @@
 		<xs:attribute name="height" type="xs:nonNegativeInteger" use="required"/>
 		<xs:attribute name="opacity" type="opacityT" default="1"/>
 		<xs:attribute name="visible" type="xs:boolean" default="1"/>
-		<xs:attribute name="offsetx" type="xs:integer" default="0"/>
-		<xs:attribute name="offsety" type="xs:integer" default="0"/>
+		<xs:attribute name="offsetx" type="xs:double" default="0"/>
+		<xs:attribute name="offsety" type="xs:double" default="0"/>
 	</xs:attributeGroup>
 
 	<xs:attributeGroup name="data.layer">
@@ -614,8 +612,8 @@
 		<xs:attribute name="color" type="hexColorT" use ="required"/>
 		<xs:attribute name="opacity" type="opacityT" use="optional" default="1"/>
 		<xs:attribute name="visible" type="xs:boolean" use="optional" default="1"/>
-		<xs:attribute name="offsetx" type="xs:integer" use="optional" default="0"/>
-		<xs:attribute name="offsety" type="xs:integer" use="optional" default="0"/>
+		<xs:attribute name="offsetx" type="xs:double" use="optional" default="0"/>
+		<xs:attribute name="offsety" type="xs:double" use="optional" default="0"/>
 		<xs:attribute name="draworder" type="draworderT" use="optional" default="topdown"/>
 	</xs:attributeGroup>
 
@@ -623,10 +621,10 @@
 		<xs:attribute name="id" type="xs:nonNegativeInteger" use="optional"/>
 		<xs:attribute name="name" type="nameT" use="required"/>
 		<xs:attribute name="type" type="xs:string" use="required"/>
-		<xs:attribute name="x" type="xs:integer" use="required"/>
-		<xs:attribute name="y" type="xs:integer" use="required"/>
-		<xs:attribute name="width" type="pixelT" use="optional" default="0"/>
-		<xs:attribute name="height" type="pixelT" use="optional" default="0"/>
+		<xs:attribute name="x" type="xs:double" use="required"/>
+		<xs:attribute name="y" type="xs:double" use="required"/>
+		<xs:attribute name="width" type="xs:double" use="optional" default="0"/>
+		<xs:attribute name="height" type="xs:double" use="optional" default="0"/>
 		<xs:attribute name="rotation" type="degreesT" use="optional" default="0"/>
 		<xs:attribute name="gid" type="tileId" use="optional" default="0"/>
 		<xs:attribute name="visible" type="xs:boolean" use="optional" default="1"/>
--- a/docs/map.xsd
+++ b/docs/map.xsd
@@ -24,12 +24,21 @@
 <xs:simpleType name="encodingT">
   <xs:restriction base="xs:NMTOKEN">
     <xs:enumeration value="base64" />
+    <xs:enumeration value="csv" />
   </xs:restriction>
 </xs:simpleType>
+
+  <xs:simpleType name="formatorsourceT">
+    <xs:restriction base="xs:NMTOKEN">
+      <xs:enumeration value="format" />
+      <xs:enumeration value="source" />
+    </xs:restriction>
+  </xs:simpleType>
 
 <xs:simpleType name="compressionT">
   <xs:restriction base="xs:NMTOKEN">
     <xs:enumeration value="gzip" />
+    <xs:enumeration value="zlip" />
   </xs:restriction>
 </xs:simpleType>
   
@@ -129,9 +138,9 @@
     </xs:restriction>
   </xs:simpleType>
 
-<xs:simpleType name="hexColourT">
+<xs:simpleType name="hexColorT">
   <xs:restriction base="xs:string">
-    <xs:pattern value="[a-fA-F0-9]{6}"/>
+    <xs:pattern value="([a-fA-F0-9][a-fA-F0-9]){3}"/>
   </xs:restriction>
 </xs:simpleType>
 
@@ -142,9 +151,9 @@
     </xs:restriction>
   </xs:simpleType>
   
-  <xs:simpleType name="hexColourAlphaT">
+  <xs:simpleType name="hexColorAlphaT">
     <xs:restriction base="xs:string">
-      <xs:pattern value="#[a-fA-F0-9]{8}"/>
+      <xs:pattern value="#([a-fA-F0-9][a-fA-F0-9]){3,4}"/>
     </xs:restriction>
   </xs:simpleType>
   
@@ -199,7 +208,6 @@
         <xs:attributeGroup ref="imagelayer"/>
       </xs:complexType>
   </xs:element>
-
     
 <xs:element name="map">
   <xs:complexType>
@@ -369,10 +377,10 @@
   <xs:attribute name="height" type="xs:nonNegativeInteger" use="required"/>
   <xs:attribute name="tilewidth" type="pixelT" use="required"/>
   <xs:attribute name="tileheight" type="pixelT" use="required"/>
-  <xs:attribute name="hexsidelength" type="pixelT" use="optional"/>
+  <xs:attribute name="hexsidelength" type="xs:integer" use="optional"/>
   <xs:attribute name="staggeraxis" type="staggeraxisT" use="optional"/>
   <xs:attribute name="staggerindex" type="staggerindexT" use="optional"/>
-  <xs:attribute name="backgroundcolor" type="hexColourAlphaT" use="optional"/>
+  <xs:attribute name="backgroundcolor" type="hexColorAlphaT" use="optional"/>
   <xs:attribute name="nextobjectid" type="xs:nonNegativeInteger" use="required"/>
 </xs:attributeGroup>
 
@@ -408,7 +416,7 @@
 
   <xs:attributeGroup name="data">
     <xs:attribute name="encoding" type="encodingT" use="required"/>
-    <xs:attribute name="compression" type="compressionT" use="required"/>
+    <xs:attribute name="compression" type="compressionT" use="optional"/>
   </xs:attributeGroup>
 
   <xs:attributeGroup name="frame">
@@ -417,16 +425,16 @@
   </xs:attributeGroup>
   
   <xs:attributeGroup name="image">
-    <xs:attribute name="format" type="imageformatT" use="required"/>
-    <xs:attribute name="source" type="xs:anyURI" use="required"/>
-    <xs:attribute name="trans" type="hexColourT" use="required"/>
-    <xs:attribute name="width" type="pixelT" use="required"/>
+    <xs:attribute name="format" type="imageformatT" use="optional"/>
+    <xs:attribute name="source" type="xs:anyURI" use="optional"/>
+    <xs:attribute name="trans" type="hexColorT" use="optional"/>
+    <xs:attribute name="width" type="pixelT" use="optional"/>
     <xs:attribute name="height" type="pixelT" use="optional"/>
   </xs:attributeGroup>
 
   <xs:attributeGroup name="image.tileset">
   <xs:attribute name="source" type="xs:anyURI"/>
-  <xs:attribute name="trans" type="hexColourT"/>
+  <xs:attribute name="trans" type="hexColorT"/>
 </xs:attributeGroup>
 
 <xs:attributeGroup name="tile.tileset">
@@ -458,9 +466,11 @@
 <xs:attributeGroup name="objectgroup">
   <!--Not sure if missing x,y which are defaulted to 0 and unchangable-->
   <xs:attribute name="name" type="nameT" use="required"/>
+  <!-- Deprecated:
   <xs:attribute name="width" type="pixelT" use="required"/>
   <xs:attribute name="height" type="pixelT" use="required"/>
-  <xs:attribute name="color" type="hexColourT" use ="required"/>
+  -->
+  <xs:attribute name="color" type="hexColorT" use ="required"/>
   <xs:attribute name="opacity" type="opacityT" use="optional" default="1"/>
   <xs:attribute name="visible" type="xs:boolean" use="optional" default="1"/>
   <xs:attribute name="offsetx" type="xs:integer" use="optional" default="0"/>
@@ -469,7 +479,7 @@
 </xs:attributeGroup>
 
 <xs:attributeGroup name="object">
-  <xs:attribute name="id" type="xs:nonNegativeInteger" use="required"/>
+  <xs:attribute name="id" type="xs:nonNegativeInteger" use="optional"/>
   <xs:attribute name="name" type="nameT" use="required"/>
   <xs:attribute name="type" type="xs:string" use="required"/>
   <xs:attribute name="x" type="xs:integer" use="required"/>
@@ -477,7 +487,7 @@
   <xs:attribute name="width" type="pixelT" use="optional" default="0"/>
   <xs:attribute name="height" type="pixelT" use="optional" default="0"/>
   <xs:attribute name="rotation" type="degreesT" use="optional" default="0"/>
-  <xs:attribute name="gid" type="tileId" use="optional"/>
+  <xs:attribute name="gid" type="tileId" use="optional" default="0"/>
   <xs:attribute name="visible" type="xs:boolean" use="optional" default="1"/>
 </xs:attributeGroup>
 
--- a/docs/map.xsd
+++ b/docs/map.xsd
@@ -1,472 +1,635 @@
 <?xml version="1.0" encoding="utf-8"?>
 
 <!--
-  format 0.7.0 by
-    Tiled Developers (mapeditor.org)
-  documented in schema form originally by
-    Olivier.Beaton@quadir.net
-    Creative Commons Attribution 3.0
-    http://creativecommons.org/licenses/by/3.0/
-  last updated on
-    2008-07-06
+	format 0.7.0 by
+		Tiled Developers (mapeditor.org)
+	documented in schema form originally by
+		Olivier.Beaton@quadir.net
+		Creative Commons Attribution 3.0
+		http://creativecommons.org/licenses/by/3.0/
+	last updated on
+		2008-07-06
 -->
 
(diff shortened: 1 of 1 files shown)


docs/map.xsd
```diff
@@ -204,41 +365,92 @@
...
   <xs:attribute name="version" type="xs:string" use="required"/>
   <xs:attribute name="orientation" type="orientationT" use="required"/>
-  <xs:attribute name="renderorder" type="renderorderT"/>
-  <xs:attribute name="width" type="xs:nonNegativeInteger" use="required"/>
-  <xs:attribute name="height" type="xs:nonNegativeInteger" use="required"/>
-  <xs:attribute name="tilewidth" type="xs:nonNegativeInteger" use="required"/>
-  <xs:attribute name="tileheight" type="xs:nonNegativeInteger" use="required"/>
+  <xs:attribute name="renderorder" type="renderorderT" default="right-down"/>
+  <xs:attribute name="width" type="pixelT" use="required"/>
+  <xs:attribute name="height" type="pixelT" use="required"/>
```
bjorn: The map width and height is not in pixels, so maybe we should just keep using `xs:nonNegativeInteger` here?

docs/map.xsd
```diff
@@ -204,41 +365,92 @@
...
+  <xs:attribute name="margin" type="pixelT"/>
+  <xs:attribute name="tilecount" type="xs:nonNegativeInteger"/>
+  <xs:attribute name="columns" type="xs:nonNegativeInteger"/>
 </xs:attributeGroup>
 
-<xs:attributeGroup name="image.tileset">
+  <xs:attributeGroup name="imagelayer">
+    <xs:attribute name="name" type="nameT"/>
+    <xs:attribute name="offsetx" type="pixelT" default="0" use="optional"/>
+    <xs:attribute name="offsety" type="pixelT" default="0" use="optional"/>
```
bjorn: The offset can actually be negative, so the non-negative `pixelT` would be the wrong type here. Maybe just use `xs:integer`?

docs/map.xsd
```diff
@@ -204,41 +365,92 @@
...
+  </xs:attributeGroup>
+
+  <xs:attributeGroup name="terrain">
+    <xs:attribute name="name" type="nameT"/>
+    <xs:attribute name="tile" type="tileId"/>
+  </xs:attributeGroup>
+
+  <xs:attributeGroup name="tileoffset">
+    <xs:attribute name="x" type="pixelT" use="required"/>
+    <xs:attribute name="y" type="pixelT" use="required"/>
```
bjorn: These can be negative as well.

docs/map.xsd
```diff
@@ -247,22 +459,33 @@
...
 <xs:attributeGroup name="object">
+  <xs:attribute name="id" type="xs:nonNegativeInteger" use="required"/>
   <xs:attribute name="name" type="nameT" use="required"/>
   <xs:attribute name="type" type="xs:string" use="required"/>
-  <xs:attribute name="x" type="xs:nonNegativeInteger" use="required"/>
-  <xs:attribute name="y" type="xs:nonNegativeInteger" use="required"/>
-  <xs:attribute name="width" type="xs:nonNegativeInteger"/>
-  <xs:attribute name="height" type="xs:nonNegativeInteger"/>
+  <xs:attribute name="x" type="pixelT" use="required"/>
+  <xs:attribute name="y" type="pixelT" use="required"/>
```
bjorn: These can be negative.

docs/map.xsd
```diff
@@ -247,22 +459,33 @@
...
   <xs:attribute name="name" type="nameT" use="required"/>
-  <xs:attribute name="width" type="xs:nonNegativeInteger" use="required"/>
-  <xs:attribute name="height" type="xs:nonNegativeInteger" use="required"/>
+  <xs:attribute name="width" type="pixelT" use="required"/>
+  <xs:attribute name="height" type="pixelT" use="required"/>
+  <xs:attribute name="color" type="hexColourT" use ="required"/>
+  <xs:attribute name="opacity" type="opacityT" use="optional" default="1"/>
+  <xs:attribute name="visible" type="xs:boolean" use="optional" default="1"/>
+  <xs:attribute name="offsetx" type="pixelT" use="optional" default="0"/>
+  <xs:attribute name="offsety" type="pixelT" use="optional" default="0"/>
```
bjorn: The `offsetx` and `offsety` can be negative.

docs/map.xsd
```diff
@@ -204,41 +365,92 @@
...
   <xs:attribute name="name" type="nameT" use="required"/>
-  <xs:attribute name="width" type="xs:nonNegativeInteger" use="required"/>
-  <xs:attribute name="height" type="xs:nonNegativeInteger" use="required"/>
-  <xs:attribute name="opacity" type="xs:decimal"/>
+  <xs:attribute name="width" type="pixelT" use="required"/>
+  <xs:attribute name="height" type="pixelT" use="required"/>
+  <xs:attribute name="opacity" type="opacityT" default="1"/>
+  <xs:attribute name="visible" type="xs:boolean" default="1"/>
+  <xs:attribute name="offsetx" type="pixelT" default="0"/>
+  <xs:attribute name="offsety" type="pixelT" default="0"/>
```
bjorn: The `offsetx` and `offsety` can be negative.

docs/map.xsd
```diff
@@ -204,41 +365,92 @@
...
 </xs:attributeGroup>
 
 <xs:attributeGroup name="layer">
+  <!--Not sure if missing x,y which are defaulted to 0 and unchangable-->
   <xs:attribute name="name" type="nameT" use="required"/>
-  <xs:attribute name="width" type="xs:nonNegativeInteger" use="required"/>
-  <xs:attribute name="height" type="xs:nonNegativeInteger" use="required"/>
-  <xs:attribute name="opacity" type="xs:decimal"/>
+  <xs:attribute name="width" type="pixelT" use="required"/>
+  <xs:attribute name="height" type="pixelT" use="required"/>
```
bjorn: These attributes are in tiles, not pixels. Maybe just use `xs:nonNegativeInteger`.

docs/map.xsd
```diff
@@ -204,41 +365,92 @@
...
+  <xs:attributeGroup name="image.tileset">
   <xs:attribute name="source" type="xs:anyURI"/>
   <xs:attribute name="trans" type="hexColourT"/>
 </xs:attributeGroup>
 
 <xs:attributeGroup name="tile.tileset">
-  <xs:attribute name="id" type="xs:nonNegativeInteger" use="required"/>
+  <xs:attribute name="id" type="tileId" use="required"/>
+  <xs:attribute name="terrain" type="cornerterrainT" use="optional"/>
+  <xs:attribute name="probability" type="percentageT" use="optional"/>
```
bjorn: I noticed `percentageT` is bound between 0 and 1, but the tile probability takes any float value.

docs/map.xsd
```diff
@@ -59,13 +129,33 @@
   </xs:restriction>
 </xs:simpleType>
 
+  <xs:simpleType name="cornerterrainT">
+    <xs:restriction base="xs:string">
+      <xs:pattern value="[0-9],[0-9],[0-9],[0-9]"/>
```
bjorn: Does this pattern allow for up to 255 terrains, or just up to 9?

docs/map.xsd
```diff
@@ -29,12 +33,78 @@
...
+  </xs:simpleType>
+
+  <xs:simpleType name="staggerindexT">
+    <xs:restriction base="xs:NMTOKEN">
+      <xs:enumeration value="even" />
+      <xs:enumeration value="odd" />
+    </xs:restriction>
+  </xs:simpleType>
+
+  <xs:simpleType name="propertyT">
```
bjorn: Should this be called `propertyTypeT`?

assofohdz:
Good feedback. Maybe ambigous isn't the right word, but vague. The types, bounds ands hierarchy are missing in the format page. I'll rework tonight and do another PR 📦

docs/map.xsd
assofohdz: No, all the custom simple types are simply named <name>T (T for Type). At least that's how it was when I started. So I just followed suit on that one.

docs/map.xsd
assofohdz: I'll do a double check and adjust

assofohdz: [0-9]{1,3},[0-9]{1,3},[0-9]{1,3},[0-9]{1,3} this would. But it would also allow up to 999,999,999,999 - we shouldn't do value validation in regex, only string restrictions.

docs/map.xsd
assofohdz: Done

docs/map.xsd
assofohdz: Done

docs/map.xsd
assofohdz: Done

docs/map.xsd
assofohdz: Done

docs/map.xsd
assofohdz: Done as `xs:integer`

docs/map.xsd
bjorn: I understand the T is for type, but this is not a type of property, but a type of property type. Hence I thought duplicating the type was appropriate.

docs/map.xsd
assofohdz: Ah, ok - so its just a float. I thought a probability would be some chance between 0 and 1 that something happens. I'll correct that.

docs/map.xsd
```diff
@@ -204,41 +358,92 @@
...
   <xs:attribute name="orientation" type="orientationT" use="required"/>
-  <xs:attribute name="renderorder" type="renderorderT"/>
+  <xs:attribute name="renderorder" type="renderorderT" default="right-down"/>
   <xs:attribute name="width" type="xs:nonNegativeInteger" use="required"/>
   <xs:attribute name="height" type="xs:nonNegativeInteger" use="required"/>
-  <xs:attribute name="tilewidth" type="xs:nonNegativeInteger" use="required"/>
-  <xs:attribute name="tileheight" type="xs:nonNegativeInteger" use="required"/>
+  <xs:attribute name="tilewidth" type="pixelT" use="required"/>
+  <xs:attribute name="tileheight" type="pixelT" use="required"/>
+  <xs:attribute name="hexsidelength" type="pixelT" use="optional"/>
```
bjorn: It just occurred to me that actually Tiled does not require the `hexsidelength` to be positive, so this should just be `xs:integer`.

docs/map.xsd
```diff
@@ -204,41 +358,92 @@
...
+  <xs:attribute name="staggeraxis" type="staggeraxisT" use="optional"/>
+  <xs:attribute name="staggerindex" type="staggerindexT" use="optional"/>
+  <xs:attribute name="backgroundcolor" type="hexColourAlphaT" use="optional"/>
```
bjorn: Actually, in any place where a `hexColourAlphaT` may be used, a `hexColourT` would also be valid (the alpha is then assumed to be 255). Could that be represented? Maybe as `#([a-fA-F0-9][a-fA-F0-9]){3-4}`

Also, maybe we should rename these to use US English `Color`?

docs/map.xsd
```diff
@@ -247,22 +452,33 @@
...
+  <xs:attribute name="gid" type="tileId" use="required"/>
 </xs:attributeGroup>
 
 <xs:attributeGroup name="objectgroup">
+  <!--Not sure if missing x,y which are defaulted to 0 and unchangable-->
   <xs:attribute name="name" type="nameT" use="required"/>
-  <xs:attribute name="width" type="xs:nonNegativeInteger" use="required"/>
-  <xs:attribute name="height" type="xs:nonNegativeInteger" use="required"/>
+  <xs:attribute name="width" type="pixelT" use="required"/>
+  <xs:attribute name="height" type="pixelT" use="required"/>
```
bjorn: `objectgroup` does not require `width` and `height`. In fact these properties are deprecated, not written out and ignored by Tiled. I guess we should just remove them here?

docs/map.xsd
```diff
@@ -247,22 +452,33 @@
 </xs:attributeGroup>
 
 <xs:attributeGroup name="tile.data.layer">
-  <xs:attribute name="gid" type="xs:nonNegativeInteger" use="required"/>
+  <xs:attribute name="gid" type="tileId" use="required"/>
 </xs:attributeGroup>
 
 <xs:attributeGroup name="objectgroup">
+  <!--Not sure if missing x,y which are defaulted to 0 and unchangable-->
```
bjorn: In the code, these x/y Layer attributes are an offset in tile units, but this is not supported for object groups by the current version of Tiled. I may rearrange things in libtiled to get rid of them and I think they should just be left out.

docs/map.xsd
```diff
@@ -247,22 +452,33 @@
...
+  <xs:attribute name="color" type="hexColourT" use ="required"/>
+  <xs:attribute name="opacity" type="opacityT" use="optional" default="1"/>
+  <xs:attribute name="visible" type="xs:boolean" use="optional" default="1"/>
+  <xs:attribute name="offsetx" type="xs:integer" use="optional" default="0"/>
+  <xs:attribute name="offsety" type="xs:integer" use="optional" default="0"/>
+  <xs:attribute name="draworder" type="draworderT" use="optional" default="topdown"/>
 </xs:attributeGroup>
 
 <xs:attributeGroup name="object">
+  <xs:attribute name="id" type="xs:nonNegativeInteger" use="required"/>
```
bjorn: While Tiled will always write out the id, it doesn't require them to be present when loading a map, so maybe it should be marked optional?

docs/map.xsd
```diff
@@ -247,22 +452,33 @@
...
-  <xs:attribute name="x" type="xs:nonNegativeInteger" use="required"/>
-  <xs:attribute name="y" type="xs:nonNegativeInteger" use="required"/>
-  <xs:attribute name="width" type="xs:nonNegativeInteger"/>
-  <xs:attribute name="height" type="xs:nonNegativeInteger"/>
+  <xs:attribute name="x" type="xs:integer" use="required"/>
+  <xs:attribute name="y" type="xs:integer" use="required"/>
+  <xs:attribute name="width" type="pixelT" use="optional" default="0"/>
+  <xs:attribute name="height" type="pixelT" use="optional" default="0"/>
+  <xs:attribute name="rotation" type="degreesT" use="optional" default="0"/>
+  <xs:attribute name="gid" type="tileId" use="optional"/>
```
bjorn: You can add a default of 0 here.

docs/map.xsd
```diff
@@ -28,13 +32,72 @@
     <xs:enumeration value="gzip" />
```
bjorn: Please add `zlib` compression here.

docs/map.xsd
```diff
@@ -17,6 +17,10 @@
   types
 -->
 
+  <xs:simpleType name="pixelT">
+    <xs:restriction base="xs:nonNegativeInteger"></xs:restriction>
+  </xs:simpleType>
+
 <xs:simpleType name="encodingT">
   <xs:restriction base="xs:NMTOKEN">
     <xs:enumeration value="base64" />
```
bjorn: Please add `csv` encoding here.

docs/map.xsd
```diff
@@ -204,41 +358,92 @@
...
+  <xs:attributeGroup name="frame">
+    <xs:attribute name="tileid" type="tileId" use="required"/>
+    <xs:attribute name="duration" type="xs:nonNegativeInteger" use="required"/>
+  </xs:attributeGroup>
+  
+  <xs:attributeGroup name="image">
+    <xs:attribute name="format" type="imageformatT" use="required"/>
+    <xs:attribute name="source" type="xs:anyURI" use="required"/>
+    <xs:attribute name="trans" type="hexColourT" use="required"/>
+    <xs:attribute name="width" type="pixelT" use="required"/>
```
bjorn: None of these attributes are required. Though, you would need to have at least `format` (for embedded data) or `source` (for external data), otherwise there is no image. Anyway I would mark both as optional if there is no easy way to represent that.

docs/map.xsd
```diff
@@ -204,41 +358,92 @@
...
+  </xs:attributeGroup>
+
+  <xs:attributeGroup name="tileoffset">
+    <xs:attribute name="x" type="xs:integer" use="required"/>
+    <xs:attribute name="y" type="xs:integer" use="required"/>
+  </xs:attributeGroup>
+
+  <xs:attributeGroup name="data">
+    <xs:attribute name="encoding" type="encodingT" use="required"/>
+    <xs:attribute name="compression" type="compressionT" use="required"/>
```
bjorn: `compression` is not required. It can be left out for uncompressed `base64` or `csv` data.

docs/map.xsd
assofohdz: Done

docs/map.xsd
assofohdz: Done

docs/map.xsd
assofohdz: Done

docs/map.xsd
assofohdz: Done

docs/map.xsd
assofohdz: Done

docs/map.xsd
assofohdz: Marked all as optional

docs/map.xsd
assofohdz: Yes, lets get the format as up-to-date as possible. I'll remove these.

docs/map.xsd
assofohdz: Ok. Though it seems weird to have an id (which is usually a unique key for some data) to be optional. But I'll mark it as optional :)

docs/map.xsd
assofohdz: I'll remove comment.

docs/map.xsd
```diff
@@ -17,24 +17,96 @@
...
+  </xs:simpleType>
+
 <xs:simpleType name="encodingT">
   <xs:restriction base="xs:NMTOKEN">
     <xs:enumeration value="base64" />
+    <xs:enumeration value="csv" />
   </xs:restriction>
 </xs:simpleType>
 
+  <xs:simpleType name="formatorsourceT">
```
bjorn: How does this work? It doesn't seem to be used anywhere?

docs/map.xsd
```diff
@@ -59,13 +131,31 @@
...
   <xs:restriction base="xs:string">
-    <xs:pattern value="[a-fA-F0-9]{6}"/>
+    <xs:pattern value="([a-fA-F0-9][a-fA-F0-9]){3}"/>
   </xs:restriction>
 </xs:simpleType>
 
-<xs:complexType name="simpleDataT">
+  <xs:simpleType name="pointsT">
+    <xs:restriction base="xs:string">
+      <xs:pattern value="(\[([0-9])+,([0-9])+\])+"/>
```
bjorn: Does this pattern actually work? Doesn't it need to account for the spaces separating the coordinate pairs? Or is that what the `[` and `]` are doing? An example value is actually `0,0 10,30 30,10`.

docs/map.xsd
assofohdz: Ah, a type I tried to add to create a choice between format or source. I'll remove it. Ended up choosing 'optional' as you suggested

docs/map.xsd
assofohdz: Sure. I didn't know about the format for points, so I assumed :) 

What is the max of the individual x's and y's?

bjorn: Hmm, actually I just realized they are floating point (double) values, and they can also be negative. So this positive integer pattern is not going to cut it.

I wondered if SVG, which uses the same list of points format, had anything useful in [its XSD file](https://www.w3.org/TR/2002/WD-SVG11-20020108/SVG.xsd), but they just do:
```xml
<simpleType name="PointsType">
  <annotation>
    <documentation>a list of points</documentation>
  </annotation>
  <restriction base="string"/>
</simpleType>
```

bjorn: I'm not sure if this is a good idea, because the floating point number could in theory also contain an exponent, unless you can include that in the pattern as well. In terms of documenting the format, I think something similar to what SVG did is more useful anyway.

assofohdz: Allright, I just thought we were talking about points in a grid-space. But I guess its more complex. I'll restrict it to string and let the consumer deal with the pattern.

docs/map.xsd
```diff
@@ -17,24 +17,89 @@
...
+      <xs:enumeration value="gif" />
+      <xs:enumeration value="jpg" />
+      <xs:enumeration value="bmp" />
+    </xs:restriction>
+  </xs:simpleType>
+
+  <xs:simpleType name="degreesT">
+    <xs:restriction base="xs:nonNegativeInteger">
+      <xs:maxInclusive value="360"/>
+      <xs:minInclusive value="0"/>
```
bjorn: I just noticed this. There's nothing requiring the degrees to be normalized between 0 and 360. Also, they aren't required to be integer.

docs/map.xsd
```diff
@@ -59,13 +124,31 @@
...
+  <xs:simpleType name="cornerterrainT">
+    <xs:restriction base="xs:string">
+      <xs:pattern value="[0-9]{1,3},[0-9]{1,3},[0-9]{1,3},[0-9]{1,3}"/>
+    </xs:restriction>
+  </xs:simpleType>
+
+<xs:simpleType name="hexColorT">
   <xs:restriction base="xs:string">
-    <xs:pattern value="[a-fA-F0-9]{6}"/>
+    <xs:pattern value="([a-fA-F0-9][a-fA-F0-9]){3}"/>
```
bjorn: Please do allow an (optional) `#` at the start here. I'd like to always have it there, but for compatibility reasons Tiled doesn't write the `#` on the `trans` attribute.

docs/map.xsd
```diff
@@ -108,12 +209,13 @@
...
         <xs:element ref="objectgroup"/>
       </xs:choice>
+      <xs:element ref="imagelayer" minOccurs="0"/>
     </xs:sequence>
     <xs:attributeGroup ref="map"/>
   </xs:complexType>
 </xs:element>
 
-<xs:element name="tileset">
+  <xs:element name="tileset">
```
bjorn: Small indentation inconsistency here.

bjorn:
Sorry for all the separate comments, I should have started another review. I thought we were almost done. :-)

docs/map.xsd
assofohdz: I changed it

docs/map.xsd
assofohdz: Added to both hexcolor and hexcoloralpha

docs/map.xsd
assofohdz: I did a general tabify on the whole document, should be ok now

bjorn:
Ah, but please bring the indentation back to 2 spaces. When you use tabs, it shows like 8 spaces in the browser and it becomes very hard to read, especially in side-by-side diff.

assofohdz:
In my editor its only 2 spaces per indent - let me check in the repository

bjorn:
> In my editor its only 2 spaces per indent - let me check in the repository

Well, that is rather on-standard and it's why using tabs for indentation is evil. :-)

assofohdz:
I can't seem to find anything wrong with indentation - let me know if there's something you see that should be changed.

bjorn:
I'm not really a fan of duplicating the documentation in this file, which has its main purpose in automatic checking. I especially don't like duplicating information that is readily apparent from the XSD, like which elements some element can contain.

assofohdz:
Granted. I'll remove it again.

bjorn:
> I can't seem to find anything wrong with indentation - let me know if there's something you see that should be changed.

Well, it looks like this to me (and to most other people who have not reconfigured their tab sizes):

screenshot_2016-10-30_21-18-35

Because this is unreadable, please change it back to use 2 spaces per indent, rather than using tabs.

bjorn:
Alright, I think it's time to update the comment at the start of the file. The current version should just be 1.0 and you should add your name and update the "last updated" date. :-)

assofohdz:
Added my name, updated version number and last update information. Its been fun working on this - it has given me some ideas for my own implementation of maps in my own game :)

bjorn:
A very nice update, thanks a lot for your persistence!

Please don't hesitate to show me your own implementation when it's ready. :-)

assofohdz:
No problem. It's been enlightening. I'll drop by the forum once I've gotten progress on my own format. It'll probably lean towards the TMX, but I want to create mine to suit my needs (and to make sure I fully understand every bit of it) - then afterwards, I can analyze how big the gap is towards TMX.

bjorn:
I'll just point out that the TMX format is of course not my ideal map format. It grew over time and breaking compatibility would hurt many more than it is worth, because it works as it is. But at some point I'd like to make a clean cut and make things more generic / modular / extendable. Maybe even without XML. :-P

So in any case I'm curious to see what you'll come up with.

assofohdz:
Yea, agreed. Unless there's a very firm strategy from the start, thats how things go. They grow and so does the complexity. My project is java, so I was thinking JSON to store the data. But the data model would not be all that different from TMX:

Map has List<TileSet> + List<Tiles>
Tiles has TileId + List<Property>
TileSet is an Image
Property describe events that can occur on this tile. But this is very game specific and makes sense to me - and probably wont to everyone else.

Still in progress. Not sure how it will end up

docs/map.xsd
johelegp: It says zli**p**.

bjorn: [Fixed](https://github.com/bjorn/tiled/commit/ed1a74f8488625126fab7ef89b3f379f7d482e59).

PR #1390, opened 2016-11-02
Copy the mVisible member member when cloning an object.

ranjak:
When an object was cloned, the clone's visibility was always set to default, which is `true`.
This caused hidden objects to become visible when using the map offsetting tool.

### DIFF
--- a/src/libtiled/mapobject.cpp
+++ b/src/libtiled/mapobject.cpp
@@ -131,5 +131,6 @@ MapObject *MapObject::clone() const
     o->setShape(mShape);
     o->setCell(mCell);
     o->setRotation(mRotation);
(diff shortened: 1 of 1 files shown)


bjorn:
Nice catch and thanks for the fix! If you'd like, please re-target the pull request to the `0.17` branch so that I can include it in the upcoming 0.17.1 release. Otherwise, I'll cherry-pick it so that branch when I get home.

bjorn:
Cherry-picked to the `0.17` branch as 7f1a3174fd61c73bb32dba2fcaaca29fe5c586db. Thanks again!

ranjak:
Oops, sorry, I was away and didn't see your comment :) You're welcome.

Issue #1237, opened 2016-03-23
Java tmx loader shifted / staggered issue


chriscamacho:
with reference to http://forum.mapeditor.org/t/java-tmx-loader-staggered-or-shifted/1188

I would have sent a pull request but I know too little about the internals of tiled, to go blundering around "fixing" something...

bjorn:
@mikepthomas Is this something you'd like to resolve?

PR #1391, opened 2016-11-06
fix libtiled-java shifted / staggered issue

mikepthomas:
Fixed shifted / staggered issue and added stagger axis, stagger index and hex side length properties to fix issue  #1237

### DIFF
--- a/util/java/libtiled-java/src/main/java/tiled/core/Map.java
+++ b/util/java/libtiled-java/src/main/java/tiled/core/Map.java
@@ -4,6 +4,7 @@
  * %%
  * Copyright (C) 2004 - 2016 Thorbjørn Lindeijer <thorbjorn@lindeijer.nl>
  * Copyright (C) 2004 - 2016 Adam Turk <aturk@biggeruniverse.com>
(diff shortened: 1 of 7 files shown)


util/java/libtiled-java/src/main/java/tiled/core/Map.java
```diff
@@ -44,38 +45,63 @@
...
+        /**
+        * Hexagonal.
+        *
+        * @since 0.11
+        */
+        HEXAGONAL,
+        /**
+        * Shifted/Staggered (used for iso and hex).
+        */
+        SHIFTED,
```
bjorn: `SHIFTED` was never used, and by the time I got around to implementing it, I called it `STAGGERED` instead, so you can remove this entry.

bjorn:
@mikepthomas The changes look fine. Just the same remarks about the commit message. Please squash it into a single commit, make sure the first line is no longer than 76 characters and leave a blank line followed by a longer description, then add `Closes #1237` at the bottom.

bjorn:
Thanks!

PR #1392, opened 2016-11-06
Suppress unsupported attribute warnings in libtiled-java

mikepthomas:
Update test tmx files to v0.17.0 format and add color to ObjectGroup to suppress unsupported attribute warning when loading map to fix issue #124

### DIFF
--- a/util/java/libtiled-java/src/main/java/tiled/core/ObjectGroup.java
+++ b/util/java/libtiled-java/src/main/java/tiled/core/ObjectGroup.java
@@ -29,6 +29,7 @@
  */
 package tiled.core;
 
+import java.awt.Color;
 import java.awt.Rectangle;
 import java.awt.Shape;
 import java.awt.geom.Area;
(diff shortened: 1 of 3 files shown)


mikepthomas:
@bjorn, as you mentioned in #1383, the getters and setters for color have different method signatures due to the way the map reader is using reflection. I can have a look at removing the reflection code if you like?

bjorn:
Hey @mikepthomas, I see this change includes only a few of the changes in #1383. I guess you plan on submitting other changes separately?

Just a few requests about your commit message:
* Please keep the first line short (like max `76` characters or so). If you need more space to describe the commit, leave a blank line and then write a longer message.
* Please add `Fixes #124` and the bottom of the commit message.

> I can have a look at removing the reflection code if you like?

Sure, that would be great!

bjorn:
> Hey @mikepthomas, I see this change includes only a few of the changes in #1383. I guess you plan on submitting other changes separately?

Sorry, I just realized that those changes are now part of #1391.

bjorn:
Yay! :-)

PR #1395, opened 2016-11-14
auto-adjust the column count for tileset collections in TilesetView

nobody-special666:
auto-adjust the column count for tileset collections when TilesetView is resized or scaled.
Currently the column count is hard-coded to 5 tiles which is a usability hindrance when working. 

I tested that this works when the TilesetView is resized or scaled and when the collection is empty.
Please consider for addition.

### DIFF
--- a/src/tiled/tilesetmodel.cpp
+++ b/src/tiled/tilesetmodel.cpp
@@ -33,6 +33,7 @@ using namespace Tiled::Internal;
 
 TilesetModel::TilesetModel(Tileset *tileset, QObject *parent):
     QAbstractListModel(parent),
+    mColumnCount(5),
     mTileset(tileset)
 {
     refreshTileIds();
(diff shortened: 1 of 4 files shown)


bjorn:
> Currently the column count is hard-coded to 5 tiles which is a usability hindrance when working.

Well, that is not true since you can configure how many columns your tileset should have in the tileset properties. However, that still does not provide dynamic adjustment like this patch here of course.

I'm not sure if I like the implementation though. It feels a bit hacky to have the view and the model communicate like that. Did you try using a `QListView` instead of a `QTableView`? Though of course that is tricky since the `TilesetView` derives from `QTableView`...

nobody-special666:
Making it a QListView is not a small change, that should come from a trusted contributor.
This is a very small patch and a reasonable compromise for an unreasonable restriction.
Easy to reverse or replace, when you're ready to implement your ideal solution.

bjorn:
I ended up using this approach after all, because indeed introducing a `QListView`-based tileset view would be quite a large patch at this point. Pushed as change 8c0ad00b3b643ecc3e850401d8c421ae3842b109.

PR #1398, opened 2016-11-15
Remove Reflection from TMXMapReader

mikepthomas:
Replace reflection with Java XML Binding (JAXB)


bjorn:
Hmm, quite nice that you could replace this with `@XmlAttribute`. Do you know if there are any performance improvements here?

I'm a little confused why `@XmlAttribute` was only added to the `setName` function in `MapLayer` for example. What about all the other attributes? How does it work?

mikepthomas:
>Hmm, quite nice that you could replace this with @XmlAttribute. Do you know if there are any performance improvements here?

Not too sure whether there will be any performance improvements, i haven't profilled it, however if there are than they are likely to be negligible as reflection seemed to be just used to load the tiles when they are explicitly defined and the object groups

>I'm a little confused why @XmlAttribute was only added to the setName function in MapLayer for example. What about all the other attributes? How does it work?

the @XmlAttribute annotation is just a hint to the JAX-B interpreter to tell it what the node is called, this is required here as the JAXBContext is mapping from the ObjectGroup class but the name property is in the parent class MapLayer so needs to be told where to map it to...

likewise the @XmlRootElement in ObjectGroup needs a hint to map the class to the objectgroup node or else it will try to map it from objectGroup as the nomclature of the class is in camel case not all lower case...

I have also added a few of the example maps from the root of the project to ensure they can be opened with tmxviewer, which only the orthogonal maps do however there is a bit of a problem with orthogonal_outside.tmx, it's throwing a NumberFormatException when loading the gid's of some objects as the gids seem to be larger than the size of an int.

bjorn:
> as the gids seem to be larger than the size of an int.

A gid is actually an unsigned int, which can be large than an int. Is there a way to support this in Java? The ones that are larger than an int must be flipped tile graphics.

Thanks for the additional explanation! I'd merge this, but now GitHub is telling me that there are conflicts. Could you look into rebasing this on top of the latest `master`?

UniteGroupServices:
There isn't such a thing as an unsigned int in Java all primitives are signed... you can use a long, however, the issue arises when TMXMapReader tries to get Tiles from the TilesSet. the getTile method can only get tiles from the ArrayList using an int as all Lists in Java only have a method signature "get(int index)" there is no long version.

mikepthomas:
There isn't such a thing as an unsigned int in Java all primitives are signed... you can use a long, however, the issue arises when TMXMapReader tries to get Tiles from the TilesSet. the getTile method can only get tiles from the ArrayList using an int as all Lists in Java only have a method signature "get(int index)" there is no long version.

bjorn:
> however, the issue arises when TMXMapReader tries to get Tiles from the TilesSet. the getTile method can only get tiles from the ArrayList using an int as all Lists in Java only have a method signature "get(int index)" there is no long version.

But there should never be a raw GID being used as index to an ArrayList. First the flags have to be cleared, and that puts the resulting value within range of an int because it leaves only 29 bits.

If a long is 64-bit in Java then we can use long to parse the GID and then cast it to an int after clearing the flags.

mikepthomas:
Good to know :) I will have a look at doing that tonight and merge the latest master as suggested

mikepthomas:
Was a little more difficult than I was expecting, but I think I've managed to extract and clear the flags before looking up the tile as you suggested (see TMXMapReader line 398-410) It no longer crashes when trying to load orthogonal_outside.tmx without the try/catch for the NumberFormatException that was previously added.

I've also rebased on latest upstream master branch so you should now be able to merge

bjorn:
Thanks!

PR #1399, opened 2016-11-17
docs: Add missing property parents

johelegp:


bjorn:
Whoops, maybe it's a bad idea to put random back-references in the docs. Quite hard to make sure it all stays up to date. :-)

Anyway, thanks for the fix!

PR #1401, opened 2016-11-30
Add a basic isometric renderer to libtiled-java

mikepthomas:
created a basic isometric map renderer to libtiled-java, it is able to load isometric maps such as isometric_grass_and_water.tmx in TMX Viewer e.g:
isometric

### DIFF
--- a/util/java/libtiled-java/src/main/java/tiled/view/IsometricRenderer.java
+++ b/util/java/libtiled-java/src/main/java/tiled/view/IsometricRenderer.java
@@ -1,0 +1,137 @@
+/*-
+ * #%L
+ * This file is part of libtiled-java.
+ * %%
+ * Copyright (C) 2004 - 2016 Thorbjørn Lindeijer <thorbjorn@lindeijer.nl>
+ * %%
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * 
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ * 
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDERS OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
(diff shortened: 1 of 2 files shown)


util/java/libtiled-java/src/main/java/tiled/view/IsometricRenderer.java
```diff
@@ -0,0 +1,94 @@
...
+                final Image image = tile.getImage();
+                if (image == null) {
+                    continue;
+                }
+
+                g.drawImage(
+                        image,
+                        ((x * tileWidth / 2) - (y * tileWidth / 2)) + hOffset,
+                        (y * tileHeight / 2) + (x * tileHeight / 2),
+                        null);
```
bjorn: Please simplify this to:
```java
               g.drawImage(
                        image,
                        (x - y) * tileWidth / 2 + hOffset,
                        (x + y) * tileHeight / 2,
                        null);
```

util/java/libtiled-java/src/main/java/tiled/view/IsometricRenderer.java
```diff
@@ -0,0 +1,94 @@
...
+    /** {@inheritDoc} */
+    @Override
+    public void paintTileLayer(Graphics2D g, TileLayer layer) {
+        final int tileWidth = map.getTileWidth();
+        final int tileHeight = map.getTileHeight();
+
+        final int hOffset = (map.getWidth() * tileWidth) / 2;
+
+        for (int x = 0; x < layer.getWidth(); x++) {
+            for (int y = 0; y < layer.getHeight(); y++) {
```
bjorn: Tiled renders row by row, iterating diagonally over the 2D tiles array, but I think in most cases there will be no visual difference.

util/java/libtiled-java/src/main/java/tiled/view/IsometricRenderer.java
mikepthomas: > Did you know isometric maps were supported by the Java version of Tiled, and it actually has an isometric map renderer

I didn't no... will have to take a look at that repo :) I think it will be better to port that code over to keep the render order consistent and it would be nice to keep the clip area optimization.

Also would it be worth me looking at porting the hex and shifted code over? Is there anything else in the repo you mentioned worth including too?

bjorn: I'm not sure if the HexMapView in that repository is really compatible with the one in Tiled Qt. I only wrote the one in Tiled Qt and it supports several options for customization that the Java version didn't. Also I remember some painting bugs with that renderer.

The ShiftedMapView is just a stub that was never finished, so if you want to add support for staggered maps to libtiled-java, you should look at the C++ code in libtiled instead. Actually in libtiled it is deriving from the hexagonal renderer, just with the "hexsidelength" property set to 0.

libtiled-java started as a stripped down version from that repository. Many things got stripped because they were too editor-specific. The renderers generally didn't make it, because the Java code was deriving from `JPanel` and `Scrollable` and included support for zooming (and caching scaled tile graphics at the current zoom level). I did not consider most of the functionality useful for games using Tiled maps.

What would probably be useful is the coordinate conversion. I just wasn't sure if that really belonged in the renderer.

mikepthomas:
I've integrated the code at https://github.com/bjorn/tiled-java/blob/master/src/tiled/view/IsoMapView.java... there probably some more optimization that can be done here in relation to calculating the start point to render the map but it is currently working as-is. If there's anything you'd like me to refactor more please let me know 👍

bjorn:
Works really well here. I can smoothly scroll the isometric example map at 4K even after making it 250x250 tiles. Nice! :-)

PR #1403, opened 2016-12-05
Update support-for-tmx-maps.md

sausagejohnson:
Adding Orx Converter to the list of supported tools for Tiled.

### DIFF
--- a/docs/reference/support-for-tmx-maps.md
+++ b/docs/reference/support-for-tmx-maps.md
@@ -142,6 +142,9 @@ Following entries are integrated solutions for specific game engines. They are t
 ### Node.js
 * [node-tmx-parser](https://github.com/andrewrk/node-tmx-parser) - loads the TMX file into a JavaScript object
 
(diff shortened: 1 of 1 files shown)


bjorn:
Looks fine, thanks!

sausagejohnson:
Thanks, Bjorn. I noticed the change hasn't flowed to http://doc.mapeditor.org/reference/support-for-tmx-maps. Is there a periodical build for the documentation?

bjorn:
> Is there a periodical build for the documentation?

Unfortunately there isn't at the moment. I've just pushed an update so now the change is live.

bjorn:
@sausagejohnson Btw, did you consider adding an export plugin to Tiled instead? That should be easier to use than an external tool.

sausagejohnson:
Thanks, Bjorn. I'd definitely consider doing an exporter if we get enough people wanting it. I'll certainly keep it in mind.

PR #1405, opened 2016-12-07
Fix macOS build

jakepetroules:
This raises the minimumQbsVersion to 1.5.2. In Qbs 1.5.2, osx was
renamed to macos. This is a forward compatible change, as Qbs guarantees
that macos is in the targetOS list if osx is.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,7 +12,8 @@ environment:
 configuration: Release
 
 install:
-  - set PATH=%PATH%;%QTDIR%\bin;%MINGW%\bin;C:\Qt\Tools\QtCreator\bin
+  - choco install -y qbs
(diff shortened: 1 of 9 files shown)


bjorn:
Hmm, I need to ask AppVeyor to update their version of Qbs first. :-)

I think we can actually drop this `-Wno-unknown-pragmas` flag now. I added it as a workaround for some huge amount of warnings, which I expect would be fixed by now.

Many thanks in any case, for working towards fixing the Qbs build on macOS!

jakepetroules:
Note that you can actually install [Qbs from Chocolatey](https://chocolatey.org/packages/qbs) now, and we (read: me) maintain it ourselves so it's always up to date. Then you don't have to rely on whatever version is in AppVeyor's Qt Creator install. ;)

bjorn:
Awesome, thanks!

PR #1406, opened 2016-12-07
Remove workaround for old bug with MinGW in qbs setup-toolchains

jakepetroules:
This might need to wait for #1405 since that updates Qbs to 1.6.1. Let's see...


bjorn:
Merged since apparently this bug is gone.

I have no idea what caused the Travis CI failure, but it isn't because of this change.

Thanks!

Issue #707, opened 2014-07-29
Connecting Objects


IcemarkUK:
Currently looking through Tiled to see if I can use it for a couple of my projects. I can get round most of my issues ( default properties on objects, easily creating objects of type ), but the one feature that would find very useful is the ability to connect two objects.

Sometimes you want to have a relationship between two objects, and you would like that relationship to by sticky. So it would be handy if you could select a connecting tool ( some combination of shift/alt/ctrl and select of object ) and drag a connection from one object to another.

You could have multiple connections from and to objects.

A connecting line would be visible on screen.

A connecting line could have name and properties to allow definition of the relationship.

Deleting an object would remove all connections on other objects.

For now workarounds are to place the name of an object into the properties of another object. But a visual process would be much nicer. I have looked at using the Polyline but not being able to resize is a pain, and not sticking to an object when you move it is more the real problem. Being able to meld the Polyline to objects would be a solution.

![screen-shot-2014-07-29-at-17 58 27](https://cloud.githubusercontent.com/assets/8301416/3738618/bcefd016-1742-11e4-80d8-7f604233f127.png)

![screen-shot-2014-07-29-at-17 54](https://cloud.githubusercontent.com/assets/8301416/3738613/b814916c-1742-11e4-9b9c-8f4e333aa361.png)

naiello:
I'd potentially be interested in helping implement something like this.  I used Tiled in a project at University last year and I had to work around this by referencing the name of an object in a custom property, so a connecting objects feature would've been really useful.

Since you mentioned using object connections in order to model physical constraints between items, I was thinking each connection would need its own property or set of properties in order to specify the type of constraint (since a physical constraint could be rigid, elastic, rotary, slider, etc).  What are your thoughts about the best way to show this in the Object Property dock?

bjorn:
@naiello First of all, thanks for offering to help with this!

Thinking about your question, I considered that actually a physical constraint between two objects would really just need to be its own object, since it would need its own set of custom properties for defining the parameters. And two of these parameters would be the actual "connection" to other objects, or most commonly, the bodyA and bodyB involved in the constraint.

So I think the way to do this will be to add types to custom properties, and one of those types would be "Object reference" / "Link". That way, the custom property system can be used to add any number of named connections to other objects. Tiled now already has a simple system by which each object gets its own unique ID, so these IDs can be used as the values of these properties.

[...]

PR #1408, opened 2016-12-08
Add object reference property type

simlrh:
Adds an "object" property type; really just another integer but with some UI for easily setting this as the ID of an object in the map.

Similar to the text editor, it's a spinbox for if you already know the ID, or click the button to open a dialog where you can do a text search over all the objects.

Addresses the use case in #707, although only the object ref, not the fancy connecting lines etc!


simlrh:
Hmm. I think there is propbably a simple solution to these build errors, but could someone familiar with Qt point me toward the problem?

bjorn:
> Hmm. I think there is probably a simple solution to these build errors, but could someone familiar with Qt point me toward the problem?

The Travis CI build started failing for some other reason, I'm not sure yet what to do about it. The AppVeyor build is failing because you added the new files only to the qmake project, but AppVeyor does the builds using Qbs.

bjorn:
Just having looked quickly over the code, I think this is a really great start! I'll be sure to try it out soon (probably Monday or Tuesday) and do a more detailed review of the code.

src/tiled/objectiddialog.cpp
```diff
@@ -53,8 +53,8 @@ ObjectIdDialog::ObjectIdDialog(QWidget *parent)
 
     Utils::restoreGeometry(this);
 
-    if (document = DocumentManager::instance()->currentDocument()) {
-        if (map = document->map()) {
+    if ((document = DocumentManager::instance()->currentDocument())) {
+        if ((map = document->map())) {
```
bjorn: I'd prefer if you fixed these warnings by instead localizing the scope of those variables. There won't be a warning when you write it like:
```c++
if (MapDocument *document = DocumentManager::instance()->currentDocument()) {
    if (Map *map = document->map()) {
```
In addition, note that there is never a need for that second check. A `MapDocument` _always_ has a `Map` instance.

src/libtiled/properties.h
```diff
@@ -41,6 +41,10 @@ struct FilePath {
     QString absolutePath;
 };
 
+struct ObjectId {
+    int id;
```
bjorn: This could use an initializing constructor, setting `id` to 0.

src/tiled/objectiddialog.cpp
```diff
@@ -0,0 +1,176 @@
...
+    , mUi(new Ui::ObjectIdDialog)
+    , mId(0)
+{
+    mUi->setupUi(this);
+    setWindowFlags(windowFlags() & ~Qt::WindowContextHelpButtonHint);
+
+    Utils::restoreGeometry(this);
+
+    if (MapDocument *document = DocumentManager::instance()->currentDocument()) {
+        QTableWidget *tableWidget = mUi->tableWidget;
```
bjorn: Rather than using a `QTabeWidget` and copying all objects into it, did you try using a `QTreeView` and re-using the `MapObjectModel`? When combined with a `QSortFilterProxyModel`, you can still have sorting and filtering work. Also, you could add an ID column to `MapObjectModel`, but hide it in the `ObjectsView` (in `objectsdock.cpp`).

src/tiled/objectiddialog.cpp
```diff
@@ -0,0 +1,176 @@
...
+    tableWidget->clearSelection();
+
+    QList<QTableWidgetItem *> items =
+            tableWidget->findItems(QString::number(mId), Qt::MatchExactly);
+
+    if (items.count() > 0) {
+        foreach (QTableWidgetItem *item, items) {
+            if (item->column() == 0) {
+                tableWidget->selectRow(item->row());
+                lineEdit->setText(QStringLiteral(""));
```
bjorn: Please use `QString()` to make an empty string. However, in this case, you should use `lineEdit->clear()`.

src/tiled/objectiddialog.cpp
```diff
@@ -0,0 +1,176 @@
...
+                tableWidget->setItem(tableWidget->rowCount() - 1, 1, new QTableWidgetItem(object->name()));
+                tableWidget->setItem(tableWidget->rowCount() - 1, 2, new QTableWidgetItem(object->type()));
+            }
+        }
+    }
+
+    connect(mUi->lineEdit, &QLineEdit::textChanged, this, &ObjectIdDialog::onTextChanged);
+    connect(mUi->tableWidget, &QTableWidget::itemSelectionChanged, this, &ObjectIdDialog::onItemSelectionChanged);
+    connect(mUi->tableWidget, &QTableWidget::itemDoubleClicked, this, &ObjectIdDialog::onItemDoubleClicked);
+    connect(mUi->pushButton, &QPushButton::clicked, this, &ObjectIdDialog::onButtonClicked);
```
bjorn: Rather than having a separate "Clear" button, please use the functionality built into `QLineEdit`: http://doc.qt.io/qt-5/qlineedit.html#clearButtonEnabled-prop. I think it's not needed that clicking this button also clears the selection in the list.

src/tiled/tiled.pro
```diff
@@ -221,7 +221,9 @@ SOURCES += aboutdialog.cpp \
     zoomable.cpp \
     clickablelabel.cpp \
     imagecolorpickerwidget.cpp \
-    reversingproxymodel.cpp
+    reversingproxymodel.cpp \
+    objectiddialog.cpp \
+    objectidedit.cpp
```
bjorn: I know it's annoying that it doesn't do it automatically, but please keep these lists sorted. It helps avoid merge conflicts.

src/tiled/variantpropertymanager.cpp
```diff
@@ -126,6 +135,32 @@ QString VariantPropertyManager::valueText(const QtProperty *property) const
         QVariant value = mValues[property].value;
         int typeId = propertyType(property);
 
+        if (typeId == objectIdTypeId()) {
+            ObjectId id = value.value<ObjectId>();
+
+            if (MapDocument *document = DocumentManager::instance()->currentDocument()) {
```
bjorn: Hmm, this may be a little problematic. There is no guarantee that the current document is the document to which the given `ObjectId` value is referring. Most of the time it will be true, but for example in the Tile Collision Editor, the objects have their own ID space (though it's a little hacky). I think the only real solution there would be to make the `ObjectId` contain also a `Map*`, but actually that won't work for the tile collision editor case since it uses a bare `ObjectGroup` instance. And if we add an `ObjectGroup*` to the `ObjectId` then we'd have to go and keep that up to date, which would be annoying as well.

What are you thoughts on this?

src/tiled/variantpropertymanager.cpp
```diff
@@ -126,6 +135,32 @@ QString VariantPropertyManager::valueText(const QtProperty *property) const
...
+
+                foreach(const ObjectGroup *group, objectGroups) {
+                    foreach (const MapObject *object, group->objects())
+                    {
+                        if (object->id() == id.id) {
+                            QString label = QStringLiteral("%1: ").arg(QString::number(object->id()));
+                            if (object->name().length())
+                                label.append(object->name());
+                            else
+                                label.append(QStringLiteral("Unnamed object"));
```
bjorn: Please use `tr("Unnamed object")`, so that it can be translated.

src/tiled/variantpropertymanager.cpp
```diff
@@ -126,6 +135,32 @@ QString VariantPropertyManager::valueText(const QtProperty *property) const
...
+
+                            if (object->type().length())
+                                label.append(QStringLiteral(" (%1)").arg(object->type()));
+                            return label;
+                        }
+                    }
+                }
+            }
+            return QStringLiteral("%1: Object not found").arg(QString::number(id.id));
```
bjorn: This should be `tr("%1: Object not found").arg...`

Also, I think a special text would be good for when the `id` is 0, which can never refer to a valid object. Maybe just "Unset".

src/tiled/objectiddialog.cpp
simlrh: Thanks for this suggestion, this was the first Qt project I've looked at so this was just the first way I figured out to display the items. I'll change this and the QLineEdit.

src/tiled/variantpropertymanager.cpp
simlrh: How about a `Map*` and an `int` referring to tile id for tile collision objects, or `0` for other objects? Does that uniquely identify all objects to your knowledge?

bjorn: Hmm, no, you'd need a `Tileset*` as well. I think adding a `Map*` and a `Tile*` would work.

bjorn: In the TMX file there should always just be the ID. The `Map*` and `Tile*` are only needed for the `VariantPropertyManager` to have enough context information to display the name of the object. In the TMX file, this context is clear.

bjorn:
Closing this now that it is superseded by #2712.

PR #1411, opened 2016-12-11
UPDATE - Translation to the Brazilian Portuguese language.

laetemn:
New update of the translation to the Brazilian Portuguese language.


bjorn:
Thanks a lot for the update!

I've pushed it to the `0.17` branch as 1d0d885, though at the moment I wasn't planning to do a `0.17.3` release. It's in any case also merged to `master` so that it will soon be available in snapshot builds.

I'm considering to branch off a Tiled 0.18, and if so I'll call for another update soon. :-)

Issue #427, opened 2013-05-02
Terrain sets don't work on staggered isometric maps


jrz:
The terrain tool does not work correctly with a saved terrain tile set from a diamond-iso metric map on a staggered iso map

PR #1430, opened 2017-01-11
terrain brush for isometric staggered map

luyungeng:
make terrain brush for isometric staggered available and add a sample
of it.

### DIFF
--- a/src/tiled/terrainbrush.cpp
+++ b/src/tiled/terrainbrush.cpp
@@ -372,11 +372,16 @@ void TerrainBrush::updateBrush(QPoint cursorPos, const QVector<QPoint> *list)
         int x = p.x(), y = p.y();
         int i = y*layerWidth + x;
 
+        // if we have already considered this point, skip to the next
+        // TODO: we might want to allow re-consideration if prior tiles... but not for now, this would risk infinite loops
+        if (checked[i])
+            continue;
+
         // to support isometric staggered, make edges into variables
-        QPoint upPoint(p.x(),p.y()-1);
-        QPoint bottomPoint(p.x(),p.y()+1);
-        QPoint leftPoint(p.x()-1,p.y());
-        QPoint rightPoint(p.x()+1,p.y());
+        QPoint upPoint(x, y-1);
+        QPoint bottomPoint(x, y+1);
+        QPoint leftPoint(x-1, y);
+        QPoint rightPoint(x+1, y);
 
         StaggeredRenderer* renderer = dynamic_cast<StaggeredRenderer*>(mapDocument()->renderer());
         if (renderer) {
@@ -390,12 +395,6 @@ void TerrainBrush::updateBrush(QPoint cursorPos, const QVector<QPoint> *list)
         int bottomIndex = bottomPoint.y()*layerWidth + bottomPoint.x();
         int leftIndex = leftPoint.y()*layerWidth + leftPoint.x();
         int rightIndex = rightPoint.y()*layerWidth + rightPoint.x();
-
-
-        // if we have already considered this point, skip to the next
-        // TODO: we might want to allow re-consideration if prior tiles... but not for now, this would risk infinite loops
-        if (checked[i])
-            continue;
 
         const Tile *tile = currentLayer->cellAt(p).tile();
         const unsigned currentTerrain = ::terrain(tile);
--- a/src/tiled/terrainbrush.cpp
+++ b/src/tiled/terrainbrush.cpp
@@ -52,6 +52,7 @@ TerrainBrush::TerrainBrush(QObject *parent)
     , mBrushBehavior(Free)
     , mLineReferenceX(0)
     , mLineReferenceY(0)
+    , mMapOrientation(-1)
 {
     setBrushMode(PaintTile);
 }
@@ -64,6 +65,13 @@ void TerrainBrush::activate(MapScene *scene)
 {
     AbstractTileTool::activate(scene);
     mIsActive = true;
+
+    if(scene && scene->mapDocument()){
+        const Map* map = scene->mapDocument()->map();
+        if(map){
+            mMapOrientation = (int)map->orientation();
+        }
+    }
 }
 
 void TerrainBrush::deactivate(MapScene *scene)
@@ -371,6 +379,25 @@ void TerrainBrush::updateBrush(QPoint cursorPos, const QVector<QPoint> *list)
         int x = p.x(), y = p.y();
         int i = y*layerWidth + x;
 
+        // to support isometric staggered, make edges into variables
+        int upperID = i - layerWidth;
+        int bottomID = i + layerWidth;
+        int leftID = i - 1;
+        int rightID = i + 1;
+
+        if((Map::Orientation)mMapOrientation==Map::Staggered){
+            if(y%2==0){
+                bottomID -= 1;
+                leftID -= layerWidth;
+                rightID = rightID + (layerWidth - 1);
+            }
+            else{
+                upperID += 1;
+                leftID = leftID - (layerWidth -1);
+                rightID += layerWidth;
+            }
+        }
+
         // if we have already considered this point, skip to the next
         // TODO: we might want to allow re-consideration if prior tiles... but not for now, this would risk infinite loops
         if (checked[i])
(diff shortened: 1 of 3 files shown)


bjorn:
Thanks for your patch, this is a very good start! The only real problem with it is that it only works for one type of staggered maps, because it does not take into account the "Stagger Axis" and "Stagger Index" of the map. To fix this you could cast the `MapRenderer` (accessible through the `MapDocument`) to a `StaggeredRenderer` so that you can use its `topLeft`, `topRight`, `bottomLeft` and `bottomRight` member functions (inherited from `HexagonalRenderer`) to get at the neighbors.

A smaller comment I have on this is that you do not need the `mMapOrientation` member, you could rather just query this from the map in `updateBrush`. This is even necessary because the user may have changed the map orientation in the meantime.

bjorn:
This patch closes issue #427.

luyungeng:
Thanks for your comment! I fixed the code again. It is the first time of me pulling request to github. Hope this helps :)

src/tiled/terrainbrush.cpp
```diff
@@ -385,16 +378,25 @@ void TerrainBrush::updateBrush(QPoint cursorPos, const QVector<QPoint> *list)
...
-            else{
-                upperID += 1;
-                leftID = leftID - (layerWidth -1);
-                rightID += layerWidth;
+        QPoint upPoint(p.x(),p.y()-1);
+        QPoint bottomPoint(p.x(),p.y()+1);
+        QPoint leftPoint(p.x()-1,p.y()-1);
+        QPoint rightPoint(p.x()+1,p.y());
+
+        if(mapDocument() && mapDocument()->map() && mapDocument()->map()->orientation() == Map::Staggered){
```
bjorn: You don't need the `mapDocument() && mapDocument()->map()` part, because `updateBrush` should never be called when there is no current map document, and a `MapDocument` instance _always_ has a map.

Just a minor thing about the code style, please put spaces in `if (` and `) {` (this applies to other parts of the patch as well).

src/tiled/terrainbrush.cpp
```diff
@@ -385,16 +378,25 @@ void TerrainBrush::updateBrush(QPoint cursorPos, const QVector<QPoint> *list)
...
+            StaggeredRenderer* renderer = dynamic_cast<StaggeredRenderer*>(mapDocument()->renderer());
+            if(renderer){
```
bjorn: Since you're relying on a `dynamic_cast` here, you actually don't need the orientation check above.

src/tiled/terrainbrush.cpp
```diff
@@ -371,6 +372,34 @@ void TerrainBrush::updateBrush(QPoint cursorPos, const QVector<QPoint> *list)
         int x = p.x(), y = p.y();
         int i = y*layerWidth + x;
 
+        // to support isometric staggered, make edges into variables
+        int upperID = i - layerWidth;
+        int bottomID = i + layerWidth;
+        int leftID = i - 1;
+        int rightID = i + 1;
```
bjorn: Please rename these to `upperIndex`, etc. because they're indexes and not IDs.

src/tiled/terrainbrush.cpp
```diff
@@ -371,6 +372,34 @@ void TerrainBrush::updateBrush(QPoint cursorPos, const QVector<QPoint> *list)
...
+                upperID = upPoint.y()*layerWidth + upPoint.x();
+
+                bottomPoint = renderer->bottomLeft(p.x(),p.y());
+                bottomID = bottomPoint.y()*layerWidth + bottomPoint.x();
+
+                leftPoint = renderer->topLeft(p.x(),p.y());
+                leftID = leftPoint.y()*layerWidth + leftPoint.x();
+
+                rightPoint = renderer->bottomRight(p.x(),p.y());
+                rightID = rightPoint.y()*layerWidth + rightPoint.x();
```
bjorn: You can move the assignments to `rightID` etc. out of the condition, since the same calculation works for orthogonal maps. 4 lines less code. :-)

src/tiled/terrainbrush.cpp
```diff
@@ -429,20 +458,20 @@ void TerrainBrush::updateBrush(QPoint cursorPos, const QVector<QPoint> *list)
             mask = 0;
 
             // depending which connections have been set, we update the preferred terrain of the tile accordingly
-            if (y > 0 && checked[i - layerWidth]) {
-                preferredTerrain = (::terrain(newTerrain[i - layerWidth]) << 16) | (preferredTerrain & 0x0000FFFF);
+            if (currentLayer->contains(upPoint) && upperID >= 0 && upperID < numTiles && checked[upperID]) {
```
bjorn: After the `currentLayer->contains` check, you don't need to check `upperID >= 0 && upperID < numTiles` as well, since you already know it will be a valid index. Same applies to the conditions below.

luyungeng:
I have changed it and also found a bug I wrote T_T. The terrain brush for orthogonal map don't work correctly for I defined  "leftPoint(p.x()-1, p.y()-1)" (by mistake)
This bug have been fixed and thanks for encouraging :-)

src/tiled/terrainbrush.cpp
```diff
@@ -371,6 +372,26 @@ void TerrainBrush::updateBrush(QPoint cursorPos, const QVector<QPoint> *list)
         int x = p.x(), y = p.y();
         int i = y*layerWidth + x;
 
+        // to support isometric staggered, make edges into variables
+        QPoint upPoint(p.x(),p.y()-1);
+        QPoint bottomPoint(p.x(),p.y()+1);
+        QPoint leftPoint(p.x()-1,p.y());
+        QPoint rightPoint(p.x()+1,p.y());
```
bjorn: I just realized there are already local variables `x` and `y` for the current point, so please replace all `p.x()` and `p.y()` with that for brevity.

src/tiled/terrainbrush.cpp
```diff
@@ -371,6 +372,26 @@ void TerrainBrush::updateBrush(QPoint cursorPos, const QVector<QPoint> *list)
...
+
+        int upperIndex = upPoint.y()*layerWidth + upPoint.x();
+        int bottomIndex = bottomPoint.y()*layerWidth + bottomPoint.x();
+        int leftIndex = leftPoint.y()*layerWidth + leftPoint.x();
+        int rightIndex = rightPoint.y()*layerWidth + rightPoint.x();
+
+
         // if we have already considered this point, skip to the next
         // TODO: we might want to allow re-consideration if prior tiles... but not for now, this would risk infinite loops
         if (checked[i])
```
bjorn: Please move up this check since we don't need to find out the neighbors if we have already checked `i` (you can also inline the calculation of `i` since I don't believe the variable is used anywhere else anymore).

luyungeng:
:-)

src/tiled/terrainbrush.cpp
```diff
@@ -376,6 +377,25 @@ void TerrainBrush::updateBrush(QPoint cursorPos, const QVector<QPoint> *list)
...
+        QPoint bottomPoint(x, y+1);
+        QPoint leftPoint(x-1, y);
+        QPoint rightPoint(x+1, y);
+
+        StaggeredRenderer* renderer = dynamic_cast<StaggeredRenderer*>(mapDocument()->renderer());
+        if (renderer) {
+            upPoint = renderer->topRight(p.x(),p.y());
+            bottomPoint = renderer->bottomLeft(p.x(),p.y());
+            leftPoint = renderer->topLeft(p.x(),p.y());
+            rightPoint = renderer->bottomRight(p.x(),p.y());
```
bjorn: Don't forget to replace `p.x()` also with `x` here. :-)

src/tiled/terrainbrush.cpp
```diff
@@ -376,6 +377,25 @@ void TerrainBrush::updateBrush(QPoint cursorPos, const QVector<QPoint> *list)
...
             continue;
 
+        // to support isometric staggered, make edges into variables
+        QPoint upPoint(x, y-1);
+        QPoint bottomPoint(x, y+1);
+        QPoint leftPoint(x-1, y);
+        QPoint rightPoint(x+1, y);
+
+        StaggeredRenderer* renderer = dynamic_cast<StaggeredRenderer*>(mapDocument()->renderer());
+        if (renderer) {
```
bjorn: One more thing here, since you're not going to use `renderer` outside of the `if` body, you should put it in its scope:
```c++
if (auto renderer = dynamic_cast<StaggeredRenderer*>(mapDocument()->renderer())) {
```

luyungeng:
I learned a lot from this. Thank you again.

luyungeng:
I have no idea what break the build. Need I change something and sync again?

bjorn:
> I have no idea what break the build. Need I change something and sync again?

It probably has something to do with it being a build of your repository, since encrypted values are involved. If it allowed using those from your repository then you could potentially expose them in the build log. I should find a way to suppress that part of the build for other repositories, so it's not your problem.

bjorn:
I've squashed the changes and pushed it to the `0.18` branch as change 18d633be5689eeba27a798e4865bc6cc026d1eea, since I think the fix would be nice to include in the upcoming 0.18.1 release. Thanks a lot for your contribution!

luyungeng:
Thank you a lot for everything :-)

PR #1431, opened 2017-01-12
Improved rendering quality of the minimap

gogoalex:


bjorn:
While this definitely looks much better, it's not an acceptable change. You're allocating an image the size of the map, which could easily take a huge amount of memory for large maps.

Maybe the quality could be improved by rendering the image at say twice the size of the minimap, and scaling it down from there.

gogoalex:
You're right.
Then we have to dig into the algorithm of the current map renderer. It's buggy when rendering into a small size image.
before : When the minimap window is small enough, the display can be terribly wrong as this.

kheftel:
Looks like it's possibly using nearest neighbor when drawing the minimap - as the minimap shrinks, each tile is represented by not an average of the pixels in the tile, but a couple discrete pixels picked from the tile.  If the tile has high contrast, you can easily end up with the selected pixels not looking very much like the tile as a whole. A solution could be to use a resampling mode like bilinear interpolation in the drawing of the minimap.  

(disclaimer: I have not looked at the pull request, nor do I know very well how tiled's code works, I'm just going off of the images and my own experience with the minimap. If I'm off base feel free to disregard).

Issue #1367, opened 2016-09-25
Jump to location by tile coordinates


bjorn:
It would be nice if you could jump to a location on the map by entering the tile coordinates.

JFonS:
Somewhat related or not, in some cases it would be nice to have a "tile ruler" to both axis of the map similar to the rulers in GIMP or other image editors.

bjorn:
@JFonS In terms of navigation it's related, but of course it's an entirely different feature. A ruler is also mentioned at issue #368.

leonidax:
I might take a look on this, have you done any progress on this?

bjorn:
@leonidax Not yet, feel free to have a go at it!

OsamahAnsari:
Hey, @Jchinjm and I are students who have to contribute to an open source project for an assignment. We were wondering if we could contribute to this feature request. Thanks!

bjorn:
@OsamahAnsari Sure, you're welcome to work on this feature!

You may take #1446 as a base, which was never entirely finished. However, it might be better to start from scratch. Tiled now has a "just to file" popup on <kbd>Ctrl+P</kbd>, which could feature a mode to jump to a tile location as well. So, <kbd>Ctrl+Shift+G</kbd> could then pop up this filter in the "go to tile" mode (similar to how in Sublime Text, this popup can jump to a line if you type a `:` at the start). Of course that'll not be entirely trivial since currently the `locatorwidget.cpp` is entirely written for project file search.

Dartho12:
I understand I am late to contributing, but I see this problem is still open. This will be my first time contributing to a GitHub project. Do you mind if I help on this feature? Should I also use 1446 as a base?

sirajahmadzai:
Is this issue still up for grabs?

PoonamMehan:
@bjorn I have tried to implement it. A review would be appreciated. Thank you very much

PR #1446, opened 2017-02-15
GoTo Tile

leonidax:
Hey I added the goto tile functionality [#1367](https://github.com/bjorn/tiled/issues/1367). I'm sure I might need to reorganize or refactor some part of the code but we can start from here with the review.


bjorn:
Hey @leeonix, why close it? If you want to update it with cleanups, please force-push to your `feature/gototile` branch to update this pull request.

leonidax:
Ok, I'll do it that way.

src/tiled/mapdocumentactionhandler.cpp
```diff
@@ -57,6 +58,9 @@ MapDocumentActionHandler::MapDocumentActionHandler(QObject *parent)
     Q_ASSERT(!mInstance);
     mInstance = this;
 
+    mActionGotoTile = new QAction(this);
+    mActionGotoTile->setShortcut(tr("Ctrl+G"));
```
bjorn: The Ctrl+G shortcut is taken for toggling the grid. However, I think Ctrl+G is a better suit for "go to" and instead the grid toggle could be changed. Maybe `#` works for that, which is what Inkscape uses.

src/tiled/mapdocumentactionhandler.cpp
```diff
@@ -175,6 +180,7 @@ MapDocumentActionHandler::~MapDocumentActionHandler()
 
 void MapDocumentActionHandler::retranslateUi()
 {
+    mActionGotoTile->setText(tr("&Go to"));
```
bjorn: At least as long as this action can only go to tiles, I would change the text to `&Go to Tile...` (note the added `...` because it spawns a dialog).

src/tiled/gotodialog.cpp
```diff
@@ -0,0 +1,94 @@
...
+        MapRenderer *renderer = mapDocument->renderer();
+        QPointF point = renderer->tileToScreenCoords(textX.toDouble(),textY.toDouble());
+        QSize size = mapView->viewport()->geometry().size();
+        QPoint origin(point.x() - size.width()/2.0, point.y() - size.height()/2.0f);
+
+        QRectF rect(origin,size);
+
+        if (!mapView->sceneRect().contains(rect)) {
+            QRectF newRect = mapView->sceneRect().united(rect);
+            mapView->setSceneRect(newRect);
```
bjorn: I'm not sure if it is a good idea to grow the scene rect. Did you try using `MapView::forceCenterOn`? That is what the minimap uses to scroll past the bounds.

src/tiled/mapdocumentactionhandler.cpp
```diff
@@ -321,6 +327,11 @@ void MapDocumentActionHandler::delete_()
     undoStack->endMacro();
 }
 
+void MapDocumentActionHandler::gotoTile()
+{
+    GotoDialog *gotoDialog = GotoDialog::showDialog();
```
bjorn: Variable not used, so just remove it.

src/tiled/tiled.qbs
```diff
@@ -189,6 +189,8 @@ QtGuiApplication {
         "flipmapobjects.h",
         "geometry.cpp",
         "geometry.h",
+        "gotodialog.cpp",
+        "gotodialog.h",
```
bjorn: Please add the new files to the `tiled.pro` file as well. The builds for Linux and macOS packages unfortunately still rely on qmake.

src/tiled/gotodialog.cpp
```diff
@@ -0,0 +1,94 @@
...
+#include "mapreader.h"
+#include "mapdocument.h"
+#include "maprenderer.h"
+
+using namespace Tiled::Internal;
+using namespace Tiled;
+
+GotoDialog *GotoDialog::mInstance;
+
+GotoDialog::GotoDialog(QWidget *parent, Qt::WindowFlags f )
```
bjorn: It should be called the `GotoTileDialog` unless we have bigger plans for it. Actually, I think it would be great if it could also jump to named objects, so we can keep this name for now.

ketanhwr:
@leonidax are you still working on this? If not, can I help in getting this one merged?

leonidax:
Hi @ketanhwr thanks for asking but I think I would rather finish it myself. I might need some help I have been pushing this for later since the build seems to be fail in Linux and I have not had time to setup a Linux machine, but I suspect it's something related to the namespaces.

ketanhwr:
The build for Linux is failing since you've not added the new files in `tiled.pro`. Therefore they aren't getting compiled. Anyway, thanks for your reply! :)

leonidax:
@ketanhwr you are right this might be the problem thanks for point that out, I just checked the job configuration and it uses qmake and the .pro file instead of the .qbs file.

bjorn:
> I just checked the job configuration and it uses qmake and the .pro file instead of the .qbs file.

Yep, it's annoying so I've just opened #1484 for fixing that. Until that is closed though, we need to keep the file list in the qmake projects up to date as well.

src/tiled/tiled.qbs
leonidax: Shame on me, I think I never read this commentary.

src/tiled/gotodialog.cpp
leonidax: I did try only by using forceCenterOn, the only problem is that if the scene rect is not big enough it would only make an approximation, so if the tile was in the upper left cornern it would scroll to the limit of the view without centering the tile.

I'm not sure how exaclty the view works but my impresion is that the scene rect does get bigger if you scroll to the left or to the top, for now I didn't see any side effect of doing this and I don't make the scene rect bigger the size required to center the corners of the map.

If this is no real issue, I think this fuctionality could be moved into the mapView, but I wanted you to review this part of the code first.

Beside this a nice addition would be to add a temporaly highlight to the tile since even with the tile exactly in the center it's hard to identify it.

src/tiled/gotodialog.cpp
leonidax: Yeah, I think if we add the goto object it might be better to call it FindDialog

src/tiled/gotodialog.cpp
```diff
@@ -0,0 +1,94 @@
+#include "gotodialog.h"
```
leonidax: Done

src/tiled/gotodialog.cpp
```diff
@@ -0,0 +1,107 @@
+/*
+ * %gotodialog.cpp%
```
bjorn: You are supposed to remove the `%` characters. :-)

The `%FILENAME%` is a placeholder that makes Qt Creator automatically fill in the file name in the license template (configured in Qt Creator > Options > C++ > File Naming).

src/tiled/gotodialog.cpp
```diff
@@ -0,0 +1,107 @@
...
+ * Copyright 2017, Your Name <leon.moctezuma@gmail.com>
```
bjorn: Don't forget to replace "Your Name" with your name. :-)

src/tiled/gotodialog.cpp
bjorn: The issue with changing the scene rect is that it is permanent (until the scene rect gets refreshed, for example when switching maps). Whereas when scrolling past the boundaries, the scroll bar range gets temporarily extended and shrinks again as you scroll back to the map.

> I did try only by using forceCenterOn, the only problem is that if the scene rect is not big enough it would only make an approximation, so if the tile was in the upper left cornern it would scroll to the limit of the view without centering the tile.

There might be an issue with `forceCenterOn`, though it seems to be working fine for the minimap.

> Beside this a nice addition would be to add a temporaly highlight to the tile since even with the tile exactly in the center it's hard to identify it.

Nice idea! Feel free to open an issue about it or make it your next pull request. :-)

src/tiled/gotodialog.h
```diff
@@ -0,0 +1,42 @@
+#ifndef GOTODIALOG_H
```
bjorn: Of course, this file should get the license header as well.

Also, please use `#pragma once` instead of header guards (I've replaced them all a while ago: 1ecd50adb793dddc35ca3e161e396dc95a89f1a7).

src/tiled/tiled.pro
```diff
@@ -141,6 +141,7 @@ SOURCES += aboutdialog.cpp \
     flexiblescrollbar.cpp \
     flipmapobjects.cpp \
     geometry.cpp \
+	gotodialog.cpp \
```
bjorn: You've used a tab here while the other lines are using 4 spaces. Please use 4 space as well (also applies to header file entry below).

src/tiled/gotodialog.h
```diff
@@ -0,0 +1,42 @@
...
+
+private:
+    void goToCoordinates();
+    QLineEdit *lineEditX;
+    QLineEdit *lineEditY;
+
+    static GotoDialog *mInstance;
+};
+
+}
```
bjorn: Coding style: because it can be hard to see what these closing brackets are for, I have a habit of annotating them as follows:
```c++
} // namespace Internal
} // namespace Tiled
```

src/tiled/gotodialog.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+using namespace Tiled::Internal;
+using namespace Tiled;
+
+GotoDialog *GotoDialog::mInstance;
+
+GotoDialog::GotoDialog(QWidget *parent, Qt::WindowFlags f )
+    : QDialog(parent, f)
+{
+    setWindowTitle(tr("Go to"));
```
bjorn: Please make the title "Go to Tile" to match its current functionality.

src/tiled/gotodialog.cpp
```diff
@@ -0,0 +1,107 @@
...
+ * You should have received a copy of the GNU General Public License along with
+ * this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "gotodialog.h"
+
+#include "documentmanager.h"
+#include "mapview.h"
+#include "mapscene.h"
+#include "mapreader.h"
```
bjorn: I'm pretty sure you don't need the `mapreader.h` include.

src/tiled/gotodialog.cpp
leonidax: facepalm

src/tiled/gotodialog.cpp
leonidax: double facepalm

src/tiled/gotodialog.cpp
leonidax: ok

src/tiled/gotodialog.cpp
leonidax: Ok

src/tiled/gotodialog.cpp
ketanhwr: `double facepalm` :joy:

bjorn:
Closing this old PR in favor of more recent ones that implement a `LocatorSource`. :-)

PR #1447, opened 2017-02-16
Support rotate hexagonal tiles 60 degrees

nicolaichuk:
http://discourse.mapeditor.org/t/rotate-hexagonal-tiles-60-degrees/1546/5

Please review  and merge to main branch.


src/libtiled/tilelayer.cpp
```diff
@@ -199,57 +199,132 @@ void TileLayer::flip(FlipDirection direction)
...
-                dest.setFlippedHorizontally(!source.flippedHorizontally());
-            } else if (direction == FlipVertically) {
-                const Cell &source = cellAt(x, mHeight - y - 1);
-                dest = source;
-                dest.setFlippedVertically(!source.flippedVertically());
-            }
-        }
-    }
+    const Map* const globalMap = map();
+    const Map::Orientation currentOrientation = (globalMap ? globalMap->orientation() : Map::Unknown);
```
bjorn: Since we can't rely on the tile layer being part of a map in general, and because of the duplicated implementation, I think it would make sense to split this up into `flip` and `flipHexagonal`, `rotate` and `rotateHexagonal`. We can decide which version to call in `tilestamp.cpp`, where a null check on the map is not needed.

src/libtiled/gidmapper.cpp
```diff
@@ -38,6 +38,7 @@ using namespace Tiled;
 const int FlippedHorizontallyFlag   = 0x80000000;
 const int FlippedVerticallyFlag     = 0x40000000;
 const int FlippedAntiDiagonallyFlag = 0x20000000;
+const int FlippedLeftHexAntiDiagonallyFlag = 0x10000000;
```
bjorn: I don't entirely understand the name or behavior of this flag.

`FlippedAntiDiagonally` meant basically swapping the x/y axis, which was equivalent to flipping vertically and then rotating by 90 degrees. In hexagonal mode, this has changed to mean flipping vertically and then rotating by 60 degrees clockwise (which can no longer be described as an anti-diagonal flip).

`FlippedLeftHexAntiDiagonallyFlag` seems to mean to flip vertically and then rotating by 60 degrees counter-clockwise (which I guess is where the `Left` comes from).

What I'm wondering about is since these flags are not exactly diagonal flips, whether it is useful in any way to still implement them like that. Wouldn't it be simpler if we instead had:
```
const int Rotated60Flag = 0x20000000;   // a duplicate of FlippedAntiDiagonallyFlag for readability
const int Rotated120Flag = 0x10000000;
```
This way things should be easier to understand:

* 60: Rotated60
* 120: Rotated120
* 180: FlippedHorizontally | FlippedVertically
* 240: Rotated60 | FlippedHorizontally | FlippedVertically
* 300: Rotated120 | FlippedHorizontally | FlippedVertically

I hope you don't mind to adjust the patch accordingly.

src/libtiled/maprenderer.cpp
```diff
@@ -106,10 +106,11 @@ static bool hasOpenGLEngine(const QPainter *painter)
             type == QPaintEngine::OpenGL2);
 }
 
-CellRenderer::CellRenderer(QPainter *painter)
+CellRenderer::CellRenderer(QPainter *painter, const bool isHexagonal)
```
bjorn: Please make this `isHexagonal` into an enum like `CellType` with default value `OrthogonalCells` and alternative value `HexagonalCells`.

src/libtiled/maprenderer.cpp
```diff
@@ -164,8 +165,15 @@ void CellRenderer::render(const Cell &cell, const QPointF &pos, const QSizeF &si
...
-    if (cell.flippedAntiDiagonally()) {
-        fragment.rotation = 90;
+    qreal rotation = 0;
+    if (mIsHexagonal) {
+        rotation = (cell.flippedLeftHexAntiDiagonally() ? 300 : (cell.flippedAntiDiagonally() ? 60 : 0));
+    } else {
+        rotation = (cell.flippedAntiDiagonally() ? 90 : 0);
+    }
+
+    if (rotation != 0) {
```
bjorn: Here, please keep the `if (cell.flippedAntiDiagonally()) {` check and simplify the hexagonal case first by having something like:
```c++
if (mCellType == HexagonalCells) {
    if (cell.rotated60())
        fragment.rotation += 60;
    if (cell.rotated120())
        fragment.rotation += 120;
} else if (cell.flippedAntiDiagonally()) {
    ...
}
```

src/libtiled/gidmapper.cpp
bjorn: Just noting that this `Rotated60Flag` does not need to exist in this file. Since it is only for readability, it can be an alternative getter/setter as `bool Cell::rotated60()`.

nicolaichuk:
@bjorn:
Please review and merge to main branch.

src/libtiled/tilelayer.h
```diff
@@ -238,13 +255,27 @@ class TILEDSHARED_EXPORT TileLayer : public Layer
...
      * Rotate this tile layer by 90 degrees left or right. The tile positions
      * are rotated within the layer, and the tiles themselves are rotated. The
      * dimensions of the tile layer are swapped.
      */
     void rotate(RotateDirection direction);
 
     /**
+     * Hexagonal rotate this tile layer by 90 degrees left or right. The tile positions
+     * are rotated within the layer, and the tiles themselves are rotated. The
+     * dimensions of the tile layer are swapped.
```
bjorn: I just realized that swapping the dimensions of the layer really makes no sense at all, and indeed the rotation of multiple tiles by 60 degrees doesn't really do anything useful. Essentially, the layer is rotated by 90 degrees while the tile graphics are getting rotated by 60 degrees. Is this something you have considered, and do you see a good solution to this? Of course, it doesn't help that trying to paint with a bigger stamp on a hex map doesn't work properly half the time due to the staggering.

nicolaichuk: May be this documentation helps http://www.redblobgames.com/grids/hexagons/#rotation

A bird in the hand is worth two in the bush )

Rotate group tile is good feature, but now it rather rotate a single tile for purposes in my application.

bjorn: Ok, so we can finish the basic feature first for single-tiles, and I can keep an issue open for making the rotation of groups work properly. That documentation definitely helps.

nicolaichuk:
@bjorn:
Please review and merge to main branch.

bjorn:
Merged as fae6ebda331037b3aa2a2dfe4fbc53984fe338b7, thanks!

I've left out the Windows related changes to README.md. While I agree Qbs instructions should be added there, I didn't want to include them in the patch.

nicolaichuk:
@bjorn, thanks.

bjorn:
@nicolaichuk That would be Tiled 1.0, which I hope will be out in April or May. In the meantime though, this feature will be in the development snapshots later today, which are also easy to install on any platform.

nicolaichuk:
@bjorn, Ok, thanks.

bjorn:
@nicolaichuk If you happen to find any time to improve hex rotation of multiple tiles that would be appreciated. If not, it'll be on my list of things to do before the release.

nicolaichuk:
I will now switch to other tasks for my project.
If I have free time, I'll see how I can help.

Issue #1382, opened 2016-10-22
Need short and consistent --export-format


buckle2000:
Currently `--export-format`'s output varies from machine to machine.

Format like `lua` is better than `"Lua files (*.lua)"` because it can be used on a different machine without changing build script.

bjorn:
That's right. Formats should probably have some explicit short name for this, since the file extension is not guaranteed to be unique.

Note that in this particular case, you can just leave out the `--export-format` parameter, because the Lua format will be used automatically based on the file extension.

buckle2000:
In some cases, maps would be exported to lua files but have extension other than `.lua` (maybe `.map`). I know I can rename files but with specified format it is just redundant.

Also, I cannot export lua files with `tengine` plugin turned on.

bjorn:
Good points, and that's also why I agree there should be some kind of short string identifying each format. Unfortunately, the plugin name can't be used either because a single plugin could add multiple formats (currently done by the Python and JSON plugins).

josefnpat:
Originally I came here to mention that `"Lua files (*.lua)"` as an argument is odd and unwieldy, but honestly, I think it's good. It takes the guesswork out of what it's exporting as, especially since the there might be multiple object structures for future Lua exports (e.g. `"Tiled Lua File (*.lua)"` or `"<Possibly very specific framework that's popular but refuses to change> Lua File (*.lua)"`.

The real problem here is that when one runs `tiled --export-formats` it doesn't specifically mention that the string should be what is the list.

For example, 

```bash
seppi@seppi7:~♠ tiled --export-formats
Export formats:
  "CSV files (*.csv)"
  "Lua files (*.lua)"
  "Json map files (*.json)"
  "JavaScript map files (*.js)"
```

I was confused at first why `--export-map lua input.tmx` didn't work. Lua is in the list! (My first guess was the fact that I didn't capitalize "Lua"). On a whim, I tried the string, and was happy to see data popping up from my pipeline.

Maybe it would be nice to explain that the strings shown are the input args, or show a quick example to squash any doubt:

```bash
seppi@seppi7:~♠ tiled --export-formats
Export format arguments:
  "CSV files (*.csv)"
  "Lua files (*.lua)"
  "Json map files (*.json)"
  "JavaScript map files (*.js)"

Example:
tiled --export-map "CSV files (*.csv)" foo.tmx
```

bjorn:
[...]

PR #1454, opened 2017-02-28
Issue #1382: Added short and consistent map format names to use with --export-map

MarceColl:
For Issue #1382 
Added a new virtual method to MapFormat called `shortName` to use whenever a short descriptive name is needed. Particularly for CLI input/output.

Added short names for all map format plugins.

`--export-formats` now uses the `shortName()`
`--export-map` now uses the `shortName()`
`--export-map` with no filter argument still uses `nameFilter()`

I've used the names of the modules as reference, if any of them are not appropriate I will change them.

### DIFF
--- a/src/plugins/python/pythonplugin.cpp
+++ b/src/plugins/python/pythonplugin.cpp
@@ -363,6 +363,32 @@ QString PythonMapFormat::nameFilter() const
     return ret;
 }
 
+QString PythonMapFormat::shortName() const
+{
+    QString ret;
+
+    // find fun
+    PyObject *pfun = PyObject_GetAttrString(mClass, "shortName");
+    if (!pfun || !PyCallable_Check(pfun)) {
+        PySys_WriteStderr("Plugin extension doesn't define \"shortName\"\n");
+        return ret;
+    }
+
+    // have fun
+    PyObject *pinst = PyEval_CallFunction(pfun, "()");
+    if (!pinst) {
+        PySys_WriteStderr("** Uncaught exception in script **\n");
+    } else {
+        ret = PyString_AsString(pinst);
+        Py_DECREF(pinst);
+    }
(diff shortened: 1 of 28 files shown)


src/plugins/python/pythonplugin.cpp
```diff
@@ -363,6 +363,32 @@ QString PythonMapFormat::nameFilter() const
...
 
+QString PythonMapFormat::shortName() const
+{
+    QString ret;
+
+    // find fun
+    PyObject *pfun = PyObject_GetAttrString(mClass, "shortName");
+    if (!pfun || !PyCallable_Check(pfun)) {
+        PySys_WriteStderr("Plugin extension doesn't define \"shortName\"\n");
+        return ret;
```
bjorn: Maybe it would be helpful to fall back to returning the result from `nameFilter` here, for backwards compatibility?

MarceColl:
Added requested changes.
Sorry, there were some tr() functions that I forgot. Changed them to QLatin1String now. Made all shortName methods public and the python shortName fallbacks to nameFilter when shortName doesn't exist. I'm not sure if we should keep the warning that shortName is not implemented tho.

bjorn:
> I'm not sure if we should keep the warning that shortName is not implemented tho. I think we should if shortName is the expected implementation. Something like: Plugin extension doesn't define shortName. Fallbacking to nameFilter?

I agree. Though please use "Falling back" instead of "Fallbacking". :-P

MarceColl:
Okay, done!

bjorn:
Awesome, thanks a lot!

MarceColl:
Thank you for Tiled! I will start working on the y-offset tiles this afternoon.

bjorn:
@MarceColl You're welcome! I'm looking forward to your next PR. :-)

PR #1457, opened 2017-02-28
Fixes #1127

### LINKED ISSUE #1127: Use an eye icon instead of a checkbox for layer visibility

ketanhwr:
I've added the eye feature as mentioned in #1127 
I took reference from what @IMMZ had previously done and fixed some additional requests than @bjorn made. I've attached a screenshot of what I've done and the eye is symmetric vertically.

Screenshot


src/tiled/layervisibilitydelegate.h
```diff
@@ -0,0 +1,57 @@
...
+ * Delegate for drawing an eye icon in LayerView when the layer is visible.
+ */
+class LayerVisibilityDelegate: public QItemDelegate
+{
+public:
+    explicit LayerVisibilityDelegate(QObject *parent = nullptr);
+
+protected:
+    void drawCheck(QPainter *painter, const QStyleOptionViewItem &option,
+        const QRect &rect, Qt::CheckState state) const;
```
bjorn: Please mark this virtual function override with the C++11 `override` keyword.

ketanhwr:
@bjorn, the image is not exactly visible. Can you update the link? And I'll try to make the changes requested :)

bjorn:
> the image is not exactly visible. Can you update the link? And I'll try to make the changes requested :)

Hmm, the image is uploaded to GitHub and is displaying fine for me. What do you mean with "not exactly visible"?

ketanhwr:
It's now visible, I don't know it wasn't visible to me few hours back. And by the way, if we're going to implement the 'lock layer' option then there's no point of implementing this eye feature, right?

ketanhwr:
@bjorn 've marked the function with the `override` keyword and I've added another icon to make the invisible layers. Here is a screenshot of the same.

Screenshot

ketanhwr:
And instead of drawing a larger icon, won't it be alright if we could just shift the origin of the icon a bit down? That'll look pretty good and won't seem odd.

bjorn:
> And instead of drawing a larger icon, won't it be alright if we could just shift the origin of the icon a bit down? That'll look pretty good and won't seem odd.

Yes, centering the icon will be fine for now as well, though it'll need to be scaled properly before the next release. But I can do this later.

Can you look into also using this eye icon in the Objects view? It also features checkboxes to toggle visibility.

src/tiled/layervisibilitydelegate.h
```diff
@@ -0,0 +1,57 @@
...
+{
+public:
+    explicit LayerVisibilityDelegate(QObject *parent = nullptr);
+
+protected:
+    void drawCheck(QPainter *painter, const QStyleOptionViewItem &option,
+        const QRect &rect, Qt::CheckState state) const override;
+
+private:
+    QPixmap visiblePixmap, invisiblePixmap;
```
bjorn: Please use:
```c++
    QPixmap mVisiblePixmap;
    QPixmap mInvisiblePixmap;
```

ketanhwr:
Okay, I'll make the changes :+1:

src/tiled/layervisibilitydelegate.h
ketanhwr: Oh sorry about that, I totally forgot!

src/tiled/eyevisibilitydelegate.cpp
```diff
@@ -0,0 +1,68 @@
...
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "layerdock.h"
+#include "eyevisibilitydelegate.h"
```
bjorn: I forgot to mention, but a cpp file should always include its own header first. This is mainly to make sure all headers include their own dependencies.

src/tiled/eyevisibilitydelegate.cpp
```diff
@@ -0,0 +1,68 @@
...
+    mInvisiblePixmap(QLatin1String(":/images/12x12/layer-invisible.png"))
+{
+}
+
+void EyeVisibilityDelegate::drawCheck(QPainter *painter, const QStyleOptionViewItem &option,
+    const QRect &rect, Qt::CheckState state) const
+{
+    Q_UNUSED(option)
+    if (state == Qt::Checked)
+    {
```
bjorn: Coding style: Please don't use `{` and `}` for single-line bodies (also for multi-line bodies, they should not be on their own line, except for functions).

ketanhwr:
I've added the eye icon in object dock as well. I'll try to fix the origin thing, but I'm not sure how to reproduce it so it'll take some time. Apart from this, is everything else fine?

src/tiled/eyevisibilitydelegate.h
```diff
@@ -0,0 +1,58 @@
...
+#include <QDockWidget>
+#include <QItemDelegate>
+#include <QPixmap>
+#include <QTreeView>
+#include <QToolButton>
+
+class QAbstractProxyModel;
+class QLabel;
+class QModelIndex;
+class QUndoStack;
```
bjorn: Please clean up the includes and forward declarations. You do not need `mapdocument.h`, `QDockWidget`, etc.

ketanhwr:
@bjorn can you look into the icon problem now? I think changing the size of icon from `12x12` to `16x16` fixes the issue. I've also made the changes in the code that you reviewed :)

bjorn:
I'll give the patch another try in the evening.

Sorry to be nitpicking, but I think you can also remove includes for `QTreeView` and `QApplication`.

ketanhwr:
If it's a learning experience for me, then it's not nitpicking :smile:

ketanhwr:
@bjorn did you take a look into the patch?

bjorn:
Sorry, a lot of things came up yesterday evening, I'll try it again this evening!

src/tiled/eyevisibilitydelegate.cpp
```diff
@@ -0,0 +1,48 @@
+/*
+ * eyevisibilitydelegate.cpp
+ * Copyright 2008-2017, Thorbjørn Lindeijer <thorbjorn@lindeijer.nl>
+ * Copyright 2010, Andrew G. Crowell <overkill9999@gmail.com>
+ * Copyright 2010, Jeff Bland <jksb@member.fsf.org>
+ * Copyright 2011, Stefan Beller <stefanbeller@googlemail.com>
```
bjorn: Please have a clean list of authors here, this list makes no sense.

src/tiled/eyevisibilitydelegate.h
```diff
@@ -0,0 +1,48 @@
+/*
+ * eyevisibilitydelegate.h
+ * Copyright 2008-2013, Thorbjørn Lindeijer <thorbjorn@lindeijer.nl>
+ * Copyright 2010, Andrew G. Crowell <overkill9999@gmail.com>
```
bjorn: Here too.

src/tiled/eyevisibilitydelegate.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+void EyeVisibilityDelegate::drawCheck(QPainter *painter, const QStyleOptionViewItem &option,
+    const QRect &rect, Qt::CheckState state) const
+{
+    Q_UNUSED(option)
+
+    if (state == Qt::Checked)
+        painter->drawPixmap(rect, mVisiblePixmap, QRect(0, 0, rect.width(), rect.height()));
+    else if (state == Qt::Unchecked)
+        painter->drawPixmap(rect, mInvisiblePixmap, QRect(0, 0, rect.width(), rect.height()));
```
bjorn: For `drawPixmap`, the second rectangle you pass in is the _source_ rectangle. This has nothing to do with `rect` and should always just be the full image (which means you can just remove it, and rely on the `drawImage` overload the only specifies the target rectangle instead). The icon should automatically render larger at a higher DPI then, and it would also fix this "cut off" problem, meaning that you can probably keep the 16x16px icon you have now (though it may look blurry when drawn scaled to 14x14px).

src/tiled/tiled.pro
```diff
@@ -236,7 +236,8 @@ SOURCES += aboutdialog.cpp \
     utils.cpp \
     varianteditorfactory.cpp \
     variantpropertymanager.cpp \
-    zoomable.cpp
+    zoomable.cpp \
+    eyevisibilitydelegate.cpp
```
bjorn: Please keep this list of files sorted alphabetically.

src/tiled/eyevisibilitydelegate.cpp
ketanhwr: Oh that was a stupid mistake :/

src/tiled/eyevisibilitydelegate.cpp
ketanhwr: :+1:

src/tiled/eyevisibilitydelegate.h
ketanhwr: :+1:

src/tiled/tiled.pro
ketanhwr: :+1:

ketanhwr:
I took the icon from the Noun Project, they've a great bunch of icons! I didn't realise my mistake in `drawPixmap` one, but I've fixed it now. The icon looks pretty good now.

bjorn:
Alright, I've merged this, but I did spend some time looking for an alternative icon because I wasn't very  happy with this one. Eventually I went with the Tango icon used by Inkscape. I also added a higher resolution one, because a 14x14px scaled one didn't look so great on my screen. See change 6db4285f10872c747b8ce99493f893f30827038a.

ketanhwr:
The icon looks better, and I didn't know about `QIcon`, thanks! :smile:

PR #1460, opened 2017-03-02
Fixes #943

### LINKED ISSUE #943: Add Execute Command action also to menu
Adding the Execute Command action also to the menu should help with its discoverability and would be the only way to see the shortcut currently, apart from browsing the wiki.

Idea came up at issue #940.

ketanhwr:
This fixes #943. I've added the option of `Execute Command` in File Menu (I couldn't find a suitable category for this purpose). Here is a screenshot of the same.

Screenshot

### DIFF
--- a/src/tiled/mainwindow.cpp
+++ b/src/tiled/mainwindow.cpp
@@ -33,6 +33,8 @@
 #include "addremovetileset.h"
 #include "automappingmanager.h"
 #include "commandbutton.h"
+#include "commanddatamodel.h"
+#include "commanddialog.h"
 #include "consoledock.h"
 #include "documentmanager.h"
 #include "exportasimagedialog.h"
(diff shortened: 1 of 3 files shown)


src/tiled/mainwindow.cpp
```diff
@@ -1038,6 +1041,39 @@ void MainWindow::reload()
...
+            msgBox.setDefaultButton(QMessageBox::Ok);
+            msgBox.setEscapeButton(QMessageBox::Ok);
+
+            QAbstractButton *button = msgBox.buttons().last();
+            connect(button, SIGNAL(clicked()), SLOT(showCommandDialog()));
+
+            msgBox.exec();
+            return;
+        }
+    }
```
bjorn: It's not nice to duplicate all this code from the `CommandButton`. I think a `CommandManager` singleton class should be introduced, which will own a `CommandDataModel` instance and which can hold this function, to be used by the `CommandButton` and from here.

ketanhwr: I didn't think of it this way, nice!

ketanhwr:
Alright, I'll close this one and make another pull request with the required changes!

Issue #1227, opened 2016-03-19
Should Not Add Duplicate Tilesets (or atleast warn and ask)


reapazor:
When adding images to a tileset, if the file is identical to an already existing item it should warn per item (add anyway, ignore, cancel) 

There may be use cases for duplicates of the same tiles, however at times when you are dumping out parsed tilesets and adding in large number of images, having duplicates be created is frustrated.

lazybotter:
Good shout! There should be a warning to overwrite the file.

bjorn:
@lazybotter As far as I understand this isn't about overwriting anything (Tiled should already warn about overwriting in all relevant cases, so please let me know if it isn't somewhere).

This issue is rather about Image Collection tilesets, where it does rarely make sense to add the same image twice to the collection. Since this is almost never what you want, it does indeed make sense to check against this case. I wonder if we can't just silently ignore this, even.

reapazor:
:) Indeed i get no error/warning when adding image.png to an existing collection with image.png in it already.

PR #1461, opened 2017-03-02
Check for duplicates when adding tiles

simiaosimis:
This pull request closes https://github.com/bjorn/tiled/issues/1227 

When adding an image to a Tileset, it checks if the image already exists in the collection and warn the user if so.

add

The approach was: try to match for each added image, its image source with the images already in the collection (O(n^2))

### DIFF
--- a/src/tiled/tileseteditor.cpp
+++ b/src/tiled/tileseteditor.cpp
@@ -46,6 +46,7 @@
 #include "zoomable.h"
 
 #include <QAction>
+#include <QCheckBox>
 #include <QComboBox>
 #include <QCoreApplication>
 #include <QFileDialog>
@@ -435,21 +436,47 @@ void TilesetEditor::addTiles()
         QPixmap image;
     };
     QVector<LoadedFile> loadedFiles;
-
+    bool addImage = true, dontAskAgain = false;
     for (const QString &file : files) {
-        const QPixmap image(file);
-        if (!image.isNull()) {
-            loadedFiles.append(LoadedFile { file, image });
-        } else {
-            QMessageBox warning(QMessageBox::Warning,
-                                tr("Add Tiles"),
-                                tr("Could not load \"%1\"!").arg(file),
-                                QMessageBox::Ignore | QMessageBox::Cancel,
-                                mMainWindow->window());
-            warning.setDefaultButton(QMessageBox::Ignore);
+        // If the tile is already in the tileset, warn user and confirm addition
+        const QMap<int, Tile*> addedTiles = mCurrentTilesetDocument->tileset().data()->tiles();
+        for(auto tile : addedTiles) {
+            if(dontAskAgain)
+                break;
+            if(tile->imageSource() == file) {
+                QCheckBox *checkBox = new QCheckBox(tr("Apply this action to all tiles"));
+                QMessageBox warning(QMessageBox::Warning,
+                            tr("Add Tiles"),
+                            tr("Tile \"%1\" already exists in the Tileset!").arg(file),
+                            QMessageBox::Yes | QMessageBox::No,
+                            mMainWindow->window());
+                warning.setDefaultButton(QMessageBox::Yes);
+                warning.setInformativeText(tr("Add anyway?"));
+                warning.setCheckBox(checkBox);
+                if (warning.exec() != QMessageBox::Yes) {
+                    addImage = false;
+                }
+                if (checkBox->checkState() == Qt::Checked) {
+                    dontAskAgain = true;
+                }
+                break;
+            }
+        }
+        if(addImage){
+            const QPixmap image(file);
+            if (!image.isNull()) {
+                loadedFiles.append(LoadedFile { file, image });
+            } else {
+                QMessageBox warning(QMessageBox::Warning,
+                                    tr("Add Tiles"),
+                                    tr("Could not load \"%1\"!").arg(file),
+                                    QMessageBox::Ignore | QMessageBox::Cancel,
+                                    mMainWindow->window());
+                warning.setDefaultButton(QMessageBox::Ignore);
 
-            if (warning.exec() != QMessageBox::Ignore)
-                return;
+                if (warning.exec() != QMessageBox::Ignore)
+                    return;
+            }
         }
     }
 
@@ -465,28 +492,7 @@ void TilesetEditor::addTiles()
         Tile *newTile = new Tile(tileset->takeNextTileId(), tileset);
         newTile->setImage(loadedFile.image);
         newTile->setImageSource(loadedFile.imageSource);
-
-        // If the tile is already in the tileset, warn user and confirm addition
-        bool addImage = true;
-        const QMap<int, Tile*> addedTiles = mCurrentTilesetDocument->tileset().data()->tiles();
-        for(auto tile : addedTiles) {
-            if(tile->imageSource() == newTile->imageSource()) {
-                QMessageBox warning(QMessageBox::Warning,
-                            tr("Add Tiles"),
-                            tr("Tile already exists in the Tileset!"),
-                            QMessageBox::Yes | QMessageBox::No,
-                            mMainWindow->window());
-                warning.setDefaultButton(QMessageBox::Yes);
-                warning.setInformativeText(tr("Add anyway?"));
-                if (warning.exec() != QMessageBox::Yes) {
-                    addImage = false;
-                }
-                break;
-            }
-        }
-        if(addImage){
-            tiles.append(newTile);
-        }
+        tiles.append(newTile);
     }
 
     mCurrentTilesetDocument->undoStack()->push(new AddTiles(mCurrentTilesetDocument, tiles));
--- a/src/tiled/tileseteditor.cpp
+++ b/src/tiled/tileseteditor.cpp
@@ -445,6 +445,7 @@ void TilesetEditor::addTiles()
         QPixmap image;
     };
     QVector<LoadedFile> loadedFiles;
+
     // If the tile is already in the tileset, warn user and confirm addition
     bool dontAskAgain = false;
     bool rememberOption = true;
@@ -462,14 +463,10 @@ void TilesetEditor::addTiles()
             warning.setInformativeText(tr("Add anyway?"));
             warning.setCheckBox(checkBox);
             int warningBoxChoice = warning.exec();
-            if (checkBox->checkState() == Qt::Checked)
-                dontAskAgain = true;
-            if (warningBoxChoice != QMessageBox::Yes) {
-                rememberOption = false;
+            dontAskAgain = checkBox->checkState() == Qt::Checked;
+            rememberOption = warningBoxChoice == QMessageBox::Yes;
+            if (!rememberOption)
                 continue;
-            }
-            else
-                rememberOption = true;
         }
         const QPixmap image(file);
         if (!image.isNull()) {
--- a/src/tiled/tileseteditor.cpp
+++ b/src/tiled/tileseteditor.cpp
@@ -436,25 +436,31 @@ void TilesetEditor::addTiles()
         QPixmap image;
     };
     QVector<LoadedFile> loadedFiles;
-    bool addImage = true, dontAskAgain = false;
+    // If the tile is already in the tileset, warn user and confirm addition
+    bool dontAskAgain = false, rememberOption = true;
     for (const QString &file : files) {
-        // If the tile is already in the tileset, warn user and confirm addition
+        bool addImage = true;
         const QMap<int, Tile*> addedTiles = mCurrentTilesetDocument->tileset().data()->tiles();
-        for(auto tile : addedTiles) {
-            if(dontAskAgain)
-                break;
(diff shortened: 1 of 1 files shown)


src/tiled/tileseteditor.cpp
```diff
@@ -465,7 +465,28 @@ void TilesetEditor::addTiles()
         Tile *newTile = new Tile(tileset->takeNextTileId(), tileset);
         newTile->setImage(loadedFile.image);
         newTile->setImageSource(loadedFile.imageSource);
-        tiles.append(newTile);
+
+        // If the tile is already in the tileset, warn user and confirm addition
+        bool addImage = true;
+        const QMap<int, Tile*> addedTiles = mCurrentTilesetDocument->tileset().data()->tiles();
+        for(auto tile : addedTiles) {
```
bjorn: At this point we have already loaded all images the user is trying to add, which is the slowest part of the operation. Given that one use-case for this feature is that you can pick a large collection of images and have Tiled figure out which ones still need to be added, it would be good to do this duplicate check in the previous loop that iterated the chosen files.

src/tiled/tileseteditor.cpp
```diff
@@ -465,7 +465,28 @@ void TilesetEditor::addTiles()
...
+            if(tile->imageSource() == newTile->imageSource()) {
+                QMessageBox warning(QMessageBox::Warning,
+                            tr("Add Tiles"),
+                            tr("Tile already exists in the Tileset!"),
```
bjorn: Which tile exists already? Since the user may have selected multiple files, you need to include the file name as done in the "Could not load" message box above.

src/tiled/tileseteditor.cpp
```diff
@@ -465,7 +465,28 @@ void TilesetEditor::addTiles()
...
+                            QMessageBox::Yes | QMessageBox::No,
```
bjorn: Issue #1227 mentioned a "Cancel" action as well, and I think we should have it. The reason for this is that when the user selected a large amount of tiles and changes his mind because of duplicates, he needs to be able to abort the operation without clicking No a lot of times.

Ideally, the user should have the options:

* Don't add this one (No)
* Don't add any duplicates (No to All)
* Add this one (Yes)
* Add all (Yes to All, skipping duplication checks)
* Cancel

The "No to All" and "Yes to All" actions are sometimes implemented with some kind of "Don't ask again" checkbox, which could be added using `QMessageBox::setCheckBox`. I know Windows Explorer has something like this when you move multiple files and it finds existing ones, maybe we can copy the text it is using (can't remember it now).

simiaosimis: I added the checkbox with "Apply this action to all tiles", similar to what shows on cinnamon, seemed more appropriate than "Don't show this again" from Windows Explorer.

A choice that I have made is that this checkbox is valid only for the images being added, not for all future actions, as the user does not have a place to configure this preferences and would not be able to revert this behaviour.

simiaosimis: revisao

vikasTmz:
This should be done in **O(1)** if you were to hash the existing images on uploading. 
Also, does this check `tile->imageSource() == file` compare the file names or does pixel wise comparison?

bjorn:
> This could be done in O(1) if you were to hash the existing images on uploading.

Well, it's at best O(n) since you have to process all tiles getting added, and only if you were to keep around a QHash in the tileset to map image sources to tiles, but I don't think that's worth it. You really need to have a lot of tiles to start noticing this.

> Also, does this check tile->imageSource() == file compare the file names or does pixel wise comparison?

The source is the file name. If it was a pixel wise comparison, hashing would indeed be mandatory for keeping things fast.

src/tiled/tileseteditor.cpp
bjorn: I think that's fine. Indeed the checkbox can only apply to the current operation.



> seemed more appropriate than "Don't show this again" from Windows Explorer.

I was referring to the Explorer because I assumed would have a better text than that. I just tried it, and actually it solves the problem without a checkbox:

2017-03-03 10_27_17-replace or skip files

That is getting really fancy though, so I won't ask you to go that far. I think it's no longer worth the time investment at that point.

src/tiled/tileseteditor.cpp
```diff
@@ -435,8 +436,41 @@ void TilesetEditor::addTiles()
...
+                if (warning.exec() != QMessageBox::Yes) {
+                    rememberOption = addImage = false;
+                }
+                else {
+                    rememberOption = true;
+                }
+                if (checkBox->checkState() == Qt::Checked) {
+                    dontAskAgain = true;
+                }
+                break;
```
bjorn: Since you're always going to break here, this code could be cleaner written when the function that searches for an existing tile with the same image source is separately implemented. Since that would also avoid the nested loop, you'd no longer need the `addImage` variable either.

src/tiled/tileseteditor.cpp
```diff
@@ -435,8 +436,41 @@ void TilesetEditor::addTiles()
         QPixmap image;
     };
     QVector<LoadedFile> loadedFiles;
-
+    // If the tile is already in the tileset, warn user and confirm addition
+    bool dontAskAgain = false, rememberOption = true;
```
bjorn: Coding style: please declare these on their own line.

src/tiled/tileseteditor.cpp
```diff
@@ -435,8 +436,41 @@ void TilesetEditor::addTiles()
...
+                            tr("Add Tiles"),
+                            tr("Tile \"%1\" already exists in the tileset!").arg(file),
+                            QMessageBox::Yes | QMessageBox::No,
+                            mMainWindow->window());
+                warning.setDefaultButton(QMessageBox::Yes);
+                warning.setInformativeText(tr("Add anyway?"));
+                warning.setCheckBox(checkBox);
+                if (warning.exec() != QMessageBox::Yes) {
+                    rememberOption = addImage = false;
+                }
```
bjorn: Coding style: please leave out the brackets when the body is only a single line.

src/tiled/tileseteditor.cpp
```diff
@@ -435,21 +436,47 @@ void TilesetEditor::addTiles()
...
-                                mMainWindow->window());
-            warning.setDefaultButton(QMessageBox::Ignore);
-
-            if (warning.exec() != QMessageBox::Ignore)
-                return;
+        // If the tile is already in the tileset, warn user and confirm addition
+        const QMap<int, Tile*> addedTiles = mCurrentTilesetDocument->tileset().data()->tiles();
+        for(auto tile : addedTiles) {
+            if(dontAskAgain)
+                break;
```
simiaosimis: Sorry about that, it was indeed with an abnormal behavior, trying to simplify the logic plus moving the code to a different place (due to your request) I ended up not checking all the case scenarios :(

bjorn: It's alright! Indeed it's not uncommon for bugs to creep in when making "last minute" reorganizations.

src/tiled/tileseteditor.cpp
simiaosimis: Although it would avoid the nested loop, it could also complicate the logic, since the use of the `dontAskAgain` variable would only be possible if as a reference parameter in the separated method, or if it become a class variable of tileseteditor. This happens because the `QMessageBox` solves two problems in the same place (should add tile and should remember this option). So for me the cleaner way of implementing this separately would be a boolean method that returns if the tile should be added plus a reference parameter to set if that action should be remembered, if that is what you meant I can make the changes, but I still think that as a matter of clarity, this code is easier to understand.

bjorn: I meant to only move the part that searches for an existing tile with the same image source into a separate function. The message box and the `dontAskAgain` variable only becomes relevant once that function returns a tile instead of `nullptr`.

simiaosimis: If I only move that part I would still need the `addImage` variable, and if I am not in a loop I can't break on the `dontAskAgain` if, so I would need an if-else, is that ok? I can commit so you can take a look and decide what is best.

bjorn: I think you're not understanding what I mean. Feel free to submit since it may help clear up the confusion.

src/tiled/tileseteditor.cpp
```diff
@@ -417,6 +418,14 @@ void TilesetEditor::retranslateUi()
     mEditTerrain->setText(tr("Edit &Terrain Information"));
 }
 
+bool hasTileInTileset(QString tileName, const QMap<int, Tile*> addedTiles){
```
bjorn: Please mark this function as `static` since it's not used from other modules.

Coding style: please put the `{` on a new line (for functions).

Naming: `tileName` is confusing, please call it `imageSource` instead.
Naming: `addedTiles` are not actually the added tiles, they're the tiles already in the tileset. Please pass in `const Tileset &tileset` instead.

src/tiled/tileseteditor.cpp
```diff
@@ -435,8 +444,35 @@ void TilesetEditor::addTiles()
         QPixmap image;
     };
     QVector<LoadedFile> loadedFiles;
-
+    // If the tile is already in the tileset, warn user and confirm addition
+    bool dontAskAgain = false;
+    bool rememberOption = true;
     for (const QString &file : files) {
+        bool addImage = true;
+        if (hasTileInTileset(file, mCurrentTilesetDocument->tileset().data()->tiles())) {
```
bjorn: You only need to do this search if `!(dontAskAgain && rememberOption)`, so:
```c++
if (!(dontAskAgain && rememberOption) && hasTileInTileset(file, *tileset)) {
```

src/tiled/tileseteditor.cpp
```diff
@@ -435,8 +444,35 @@ void TilesetEditor::addTiles()
...
+            if (dontAskAgain) 
+                addImage = rememberOption;
```
bjorn: Now you can just `continue;` here, because if `dontAskAgain` you know that `rememberOption == false`.

src/tiled/tileseteditor.cpp
```diff
@@ -435,8 +444,35 @@ void TilesetEditor::addTiles()
...
+                QMessageBox warning(QMessageBox::Warning,
+                            tr("Add Tiles"),
+                            tr("Tile \"%1\" already exists in the tileset!").arg(file),
+                            QMessageBox::Yes | QMessageBox::No,
+                            mMainWindow->window());
+                warning.setDefaultButton(QMessageBox::Yes);
+                warning.setInformativeText(tr("Add anyway?"));
+                warning.setCheckBox(checkBox);
+                if (warning.exec() != QMessageBox::Yes)
+                    rememberOption = addImage = false;
```
bjorn: Didn't say yes, so `continue;` instead of setting `addImage`.

src/tiled/tileseteditor.cpp
```diff
@@ -435,8 +444,35 @@ void TilesetEditor::addTiles()
...
+                else 
+                    rememberOption = true;
+                if (checkBox->checkState() == Qt::Checked)
+                    dontAskAgain = true;
+            }
+        }
+        if(!addImage){
+            continue;
+        }
```
bjorn: No longer needed.

src/tiled/tileseteditor.cpp
bjorn: Using `continue` also means you don't need to indent the rest into an `else`.

src/tiled/tileseteditor.cpp
```diff
@@ -435,8 +445,32 @@ void TilesetEditor::addTiles()
         QPixmap image;
     };
     QVector<LoadedFile> loadedFiles;
-
```
bjorn: Please don't remove the empty line.

src/tiled/tileseteditor.cpp
```diff
@@ -435,8 +445,32 @@ void TilesetEditor::addTiles()
...
+            warning.setCheckBox(checkBox);
+            int warningBoxChoice = warning.exec();
+            if (checkBox->checkState() == Qt::Checked)
+                dontAskAgain = true;
+            if (warningBoxChoice != QMessageBox::Yes) {
+                rememberOption = false;
+                continue;
+            }
+            else
+                rememberOption = true;
```
bjorn: The above lines could be simplified as:
```
dontAskAgain = checkBox->checkState() == Qt::Checked;
rememberOption = warningBoxChoice == QMessageBox::Yes;
if (!rememberOption)
    continue;
```

simiaosimis: Much prettier, but I didn't know if you liked this kind of code :)

bjorn: I think it's short and efficient, without compromising on readability. That's the kind of code I like. :-)

I don't like the name "rememberOption", but nothing better came to mind so far. For general readability, I considered whether using an `enum { Ask, Yes, No } addDuplicate;` (basically a tribool) could help. Feel free to explore that option if you want.

bjorn:
Thanks for your persistence! :-)

Issue #1458, opened 2017-03-01
Keyboard Shortcuts not working in Ubuntu 16.04


ketanhwr:
I'm using Tiled latest developer build (from current `master` branch) and I've noticed that only a few keyboard shortcuts are working for me. I'm using Ubuntu 16.04.

The keyboard shortcuts for `Stamp Brush`, `Terrain Brush`, `Bucket Fill Tool`, etc are working, but `Ctrl+Z` for `Undo`, `Zoom in` and `Zoom out` (`Ctrl++` `Ctrl+-`) aren't working. I've not tried every keyboard shortcut, but `Ctrl+E` for Export isn't working too.

Can anyone reproduce this?

bjorn:
They're working fine on Arch Linux / Xfce. I could try later on Xubuntu 16.10, but I fear it may be related to the customized menus in standard Ubuntu / Unity and I don't have an install ready for testing.

vadi2:
It's most likely due to @ketanhwr using Qt 5.7 or 5.8 and runnning Ubuntu - I believe the bug is localised to that combination.

bjorn:
@vadi2 Provided a link to [QTBUG-59557](https://bugreports.qt.io/browse/QTBUG-59557) in #1497, which covers this issue.

bjorn:
The Qt issue was closed as "cannot reproduce" against Qt 5.10. Since it's been over 4 years now and no additional reports of this kind have come in, I'll close this issue as well.

PR #1463, opened 2017-03-03
Option to run commands from Menu as well as edit them - Fixes #943

### LINKED ISSUE #943: Add Execute Command action also to menu

ketanhwr:
This fixes #943. I previously made a PR, but it lacked a lot of features. Here's a basic gist of what I've done now:

- Created a menu just for Commands
- Contains an action to `Edit Commands`
- Contains a list of all the enabled commands. Also has a shortcut for the first enabled command. This list gets refreshed each time the menu is opened. This is to ensure that the command list is updated and the disabled/enabled commands are updated.

Here's a screenshot of the same:

Screenshot

I made a `CommandManager` class which is instantiated once in `MainWindow`. This manages the menu (updates the commands) and the signals/slots for them.

### DIFF
--- a/src/tiled/commandbutton.cpp
+++ b/src/tiled/commandbutton.cpp
@@ -44,10 +44,10 @@ CommandButton::CommandButton(QWidget *parent)
     setMenu(mMenu);
 
     // Populate the menu once in order to register shortcuts beforehand
-    CommandManager::instance()->populateMenu(mMenu);
+    CommandManager::instance()->populateMenu(mMenu, true);
 
     connect(mMenu, &QMenu::aboutToShow,  
-            [this]() { CommandManager::instance()->populateMenu(mMenu); });
+            [this]() { CommandManager::instance()->populateMenu(mMenu, true); });
     connect(this, SIGNAL(clicked()), SLOT(runCommand()));
 }
 
--- a/src/tiled/commandbutton.cpp
+++ b/src/tiled/commandbutton.cpp
@@ -43,6 +43,9 @@ CommandButton::CommandButton(QWidget *parent)
     setPopupMode(QToolButton::MenuButtonPopup);
     setMenu(mMenu);
 
+    // Populate the menu once in order to register shortcuts beforehand
+    CommandManager::instance()->populateMenu(mMenu);
+
     connect(mMenu, &QMenu::aboutToShow, 
             CommandManager::instance(), [this]() { CommandManager::instance()->populateMenu(mMenu); });
     connect(this, SIGNAL(clicked()), SLOT(runCommand()));
--- a/src/tiled/commandbutton.cpp
+++ b/src/tiled/commandbutton.cpp
@@ -127,5 +127,6 @@ void CommandButton::changeEvent(QEvent *event)
 
 void CommandButton::retranslateUi()
 {
+    setShortcut(QKeySequence(tr("F5")));
     setToolTip(tr("Execute Command"));
 }
--- a/src/tiled/commandbutton.cpp
+++ b/src/tiled/commandbutton.cpp
@@ -35,7 +35,6 @@ using namespace Tiled::Internal;
 CommandButton::CommandButton(QWidget *parent)
     : QToolButton(parent)
     , mMenu(new QMenu(this))
-    , mCommandManager(new CommandManager(this))
 {
     setIcon(QIcon(QLatin1String(":images/24x24/system-run.png")));
     setThemeIcon(this, "system-run");
@@ -45,7 +44,7 @@ CommandButton::CommandButton(QWidget *parent)
     setMenu(mMenu);
 
     connect(mMenu, &QMenu::aboutToShow, 
-            mCommandManager, [this]() { mCommandManager->populateMenu(mMenu); });
+            CommandManager::instance(), [this]() { CommandManager::instance()->populateMenu(mMenu); });
     connect(this, SIGNAL(clicked()), SLOT(runCommand()));
 }
 
--- a/src/tiled/commandbutton.cpp
+++ b/src/tiled/commandbutton.cpp
@@ -127,6 +127,6 @@ void CommandButton::changeEvent(QEvent *event)
 
 void CommandButton::retranslateUi()
 {
-    setShortcut(QKeySequence(tr("F5")));
     setToolTip(tr("Execute Command"));
+    setShortcut(QKeySequence(tr("F5")));
 }
--- a/src/tiled/commandbutton.cpp
+++ b/src/tiled/commandbutton.cpp
@@ -21,6 +21,7 @@
 #include "commandbutton.h"
 #include "commanddatamodel.h"
 #include "commanddialog.h"
+#include "commandmanager.h"
 #include "utils.h"
 
 #include <QEvent>
@@ -34,6 +35,7 @@ using namespace Tiled::Internal;
 CommandButton::CommandButton(QWidget *parent)
     : QToolButton(parent)
     , mMenu(new QMenu(this))
+    , mCommandManager(new CommandManager(this))
 {
     setIcon(QIcon(QLatin1String(":images/24x24/system-run.png")));
     setThemeIcon(this, "system-run");
@@ -42,7 +44,8 @@ CommandButton::CommandButton(QWidget *parent)
     setPopupMode(QToolButton::MenuButtonPopup);
     setMenu(mMenu);
 
-    connect(mMenu, SIGNAL(aboutToShow()), SLOT(populateMenu()));
+    connect(mMenu, &QMenu::aboutToShow, 
+            mCommandManager, [this]() { mCommandManager->populateMenu(mMenu); });
     connect(this, SIGNAL(clicked()), SLOT(runCommand()));
 }
 
@@ -85,33 +88,6 @@ void CommandButton::showDialog()
     dialog.exec();
 }
 
-void CommandButton::populateMenu()
-{
-    mMenu->clear();
-
-    // Use a data model for getting the command list to avoid having to
-    // manually parse the settings
-    const CommandDataModel model;
-    const QList<Command> &commands = model.allCommands();
-
-    foreach (const Command &command, commands) {
-        if (!command.isEnabled)
-            continue;
-
-        QAction *action = mMenu->addAction(command.name);
-        action->setStatusTip(command.command);
-        action->setData(command.toQVariant());
-        connect(action, SIGNAL(triggered()), SLOT(runCommand()));
-    }
-
-    if (!mMenu->isEmpty())
-        mMenu->addSeparator();
-
-    // Add "Edit Commands..." action
-    QAction *action = mMenu->addAction(tr("Edit Commands..."));
-    connect(action, SIGNAL(triggered()), SLOT(showDialog()));
-}
-
 void CommandButton::changeEvent(QEvent *event)
 {
     QToolButton::changeEvent(event);
@@ -128,5 +104,4 @@ void CommandButton::changeEvent(QEvent *event)
 void CommandButton::retranslateUi()
 {
     setToolTip(tr("Execute Command"));
-    setShortcut(QKeySequence(tr("F5")));
 }
--- a/src/tiled/commandbutton.cpp
+++ b/src/tiled/commandbutton.cpp
@@ -128,5 +128,4 @@ void CommandButton::changeEvent(QEvent *event)
 void CommandButton::retranslateUi()
 {
     setToolTip(tr("Execute Command"));
-    setShortcut(QKeySequence(tr("F5")));
 }
--- a/src/tiled/commandbutton.cpp
+++ b/src/tiled/commandbutton.cpp
@@ -46,8 +46,8 @@ CommandButton::CommandButton(QWidget *parent)
     // Populate the menu once in order to register shortcuts beforehand
     CommandManager::instance()->populateMenu(mMenu);
 
-    connect(mMenu, &QMenu::aboutToShow, 
-            CommandManager::instance(), [this]() { CommandManager::instance()->populateMenu(mMenu); });
+    connect(mMenu, &QMenu::aboutToShow,  
+            [this]() { CommandManager::instance()->populateMenu(mMenu); });
     connect(this, SIGNAL(clicked()), SLOT(runCommand()));
 }
 
--- a/src/tiled/commandmanager.cpp
+++ b/src/tiled/commandmanager.cpp
@@ -62,6 +62,7 @@ void CommandManager::populateMenu(QMenu *menu)
 {
 	menu->clear();
 
+	// Add Edit Commands first
 	QAction *mEditCommands = new QAction(this);
     mEditCommands->setIcon(
             QIcon(QLatin1String(":/images/24x24/system-run.png")));
@@ -70,4 +71,24 @@ void CommandManager::populateMenu(QMenu *menu)
     menu->addAction(mEditCommands);
 
     connect(mEditCommands, SIGNAL(triggered()), this, SLOT(showDialog()));
+
+    menu->addSeparator();
+
+    // Add all enabled commands now
+
+    bool firstEnabledCommand = true;
+
+    const CommandDataModel model;
+    const QList<Command> &commands = model.allCommands();
+
+    foreach (const Command &command, commands) {
+        if (!command.isEnabled)
+            continue;
+
+        QAction *mAction = menu->addAction(command.name);
+        if(firstEnabledCommand)
+        	mAction->setShortcut(QKeySequence(tr("F5")));
+        firstEnabledCommand = false;
+        
+    }
 }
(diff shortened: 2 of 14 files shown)


ketanhwr:
@bjorn status?

src/tiled/commandbutton.cpp
```diff
@@ -127,6 +127,6 @@ void CommandButton::changeEvent(QEvent *event)
 
 void CommandButton::retranslateUi()
 {
-    setToolTip(tr("Execute Command"));
     setShortcut(QKeySequence(tr("F5")));
+    setToolTip(tr("Execute Command"));
```
bjorn: Why would you change the order of these two lines?

src/tiled/commandmanager.cpp
```diff
@@ -0,0 +1,102 @@
+/*
+ * commandmanagar.cpp
```
bjorn: Typo: commandmanager.cpp (also affects header file)

src/tiled/commandmanager.cpp
```diff
@@ -0,0 +1,102 @@
...
+#include <QMenu>
+#include <QSignalMapper>
+#include <QWidget>
+
+using namespace Tiled;
+using namespace Tiled::Internal;
+
+CommandManager::CommandManager(QObject *parent, QWidget *window)
+	: QObject(parent)
+	, mMainWindow(window)
```
bjorn: Coding style: please use 4 spaces for indentation, no tabs.

src/tiled/commandmanager.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+    // Add all enabled commands now
+
+    bool firstEnabledCommand = true;
+    int counter = -1;
+
+    CommandDataModel *mModel = new CommandDataModel;
+    const QList<Command> &commands = mModel->allCommands();
+
+    foreach (const Command &command, commands) {
```
bjorn: Since you're actually using the index, please use a traditional for loop instead:
```c++
    for (int i = 0; i < commands.size(); ++i) {
        const Command &command = commands.at(i);
```

src/tiled/mainwindow.cpp
```diff
@@ -310,6 +314,8 @@ MainWindow::MainWindow(QWidget *parent, Qt::WindowFlags flags)
     connect(mUi->actionZoomNormal, SIGNAL(triggered()), SLOT(zoomNormal()));
     connect(mUi->actionFullScreen, &QAction::toggled, this, &MainWindow::setFullScreen);
 
+    connect(mUi->menuCommand, SIGNAL(aboutToShow()), mCommandManager, SLOT(populateMainWindowMenu()));
```
bjorn: You can do this instead, to avoid the need for `setMainWindowMenu` earlier:
```c++
connect(mUi->menuCommand, &QMenu::aboutToShow,
        mCommandManager, [this]() { mCommandManager->populateMenu(mUi->menuCommand); });
```

src/tiled/commandmanager.cpp
```diff
@@ -0,0 +1,102 @@
...
+}
+
+void CommandManager::populateMainWindowMenu()
+{
+	populateMenu(this->mMainWindowMenu);
+}
+
+void CommandManager::showDialog()
+{
+	CommandDialog dialog(mMainWindow);
```
bjorn: I think you can use the static `QApplication::activeWindow()` as parent, to avoid the need for passing around a pointer to the main window.

src/tiled/commandmanager.cpp
```diff
@@ -0,0 +1,102 @@
...
+CommandManager::CommandManager(QObject *parent, QWidget *window)
+	: QObject(parent)
+	, mMainWindow(window)
+{
+	
+}
+
+CommandManager::~CommandManager()
+{
+}
```
bjorn: Just leave out the destructor. Since it's not doing anything useful, the default destructor will do fine.

src/tiled/commandmanager.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+void CommandManager::populateMenu(QMenu *menu)
+{
+	menu->clear();
+
+	// Add Edit Commands first
+	QAction *mEditCommands = new QAction(this);
+    mEditCommands->setIcon(
+            QIcon(QLatin1String(":/images/24x24/system-run.png")));
+    mEditCommands->setText(tr("Edit Commands"));
```
bjorn: Please add this action as the last item in the list. Also, its text should end with `...` since it opens a dialog. Also, you should be able to re-use this `populateMenu` function in the `CommandButton` class. I think it will cause no problems to have the F5 shortcut there as well, but please verify that it won't lead to "ambiguous shortcut" problems.

src/tiled/commandmanager.cpp
```diff
@@ -0,0 +1,102 @@
...
+    const QList<Command> &commands = mModel->allCommands();
+
+    foreach (const Command &command, commands) {
+    	++counter;
+
+        if (!command.isEnabled)
+            continue;
+
+        QAction *mAction = menu->addAction(command.name);
+        if(firstEnabledCommand)
```
bjorn: Coding style: Please put a space after the `if`.

src/tiled/commandmanager.cpp
```diff
@@ -0,0 +1,102 @@
...
+        	mAction->setShortcut(QKeySequence(tr("F5")));
+        firstEnabledCommand = false;
+
+        QSignalMapper *mapper = new QSignalMapper(mAction);
```
bjorn: You do not need a `QSignalMapper` instance for each action. Instead you could use a single instance to map all these actions. However, I would prefer to use lambdas instead:
```c++
connect(mAction, &QAction::triggered, [this,counter]() { mModel->execute(i); });
```

src/tiled/commandmanager.cpp
```diff
@@ -0,0 +1,102 @@
...
+    connect(mEditCommands, SIGNAL(triggered()), this, SLOT(showDialog()));
+
+    menu->addSeparator();
+
+    // Add all enabled commands now
+
+    bool firstEnabledCommand = true;
+    int counter = -1;
+
+    CommandDataModel *mModel = new CommandDataModel;
```
bjorn: You shouldn't create a new `CommandDataModel` instance each time the menu is populated. You'll end up with a lot of instances after a while and they'll never be deleted. Instead, move this `mModel` to be a member of the `CommandManager` and set the manager as its parent (you'll need to add `QObject *parent = nullptr` to the constructor of the `CommandDataModel` and pass this on the `QObject` constructor).

src/tiled/commandmanager.cpp
```diff
@@ -0,0 +1,102 @@
...
+        if(firstEnabledCommand)
+        	mAction->setShortcut(QKeySequence(tr("F5")));
+        firstEnabledCommand = false;
+
+        QSignalMapper *mapper = new QSignalMapper(mAction);
+        mapper->setMapping(mAction, counter);
+        connect(mAction, SIGNAL(triggered()), mapper, SLOT(map()));
+        connect(mapper, SIGNAL(mapped(int)), mModel, SLOT(execute(int)));
+    }
+}
```
bjorn: Coding style: please make sure the file ends with a newline (most editors have an option for this).

src/tiled/commandbutton.cpp
ketanhwr: I didn't notice this, I'll revert this one.

src/tiled/commandmanager.cpp
ketanhwr: I previously did exactly this, but it wasn't working. Probably because, when we edit the commands, the `mModel` doesn't get updated. So I instantiated this each time while updating and it worked fine after that.

src/tiled/mainwindow.cpp
ketanhwr: Nice solution, thanks!

src/tiled/commandmanager.cpp
ketanhwr: It leads to "ambiguous shortcut" problem. So I've added a small modification to assign shortcut to only 1 of the menu.

ketanhwr:
@bjorn, I've made the requested changes. Although the problem with `CommandDataModel` still persists. I need to instantiate a `CommandDataModel` object every time I populate the menu, otherwise the list doesn't get updated.

ketanhwr:
Alright, so what should I do for this task? Remove the 'Ok' and 'Cancel' buttons and add a 'Close' button instead?

bjorn:
> Alright, so what should I do for this task? Remove the 'Ok' and 'Cancel' buttons and add a 'Close' button instead?

Yes, that's what I would suggest.

src/tiled/commandmanager.h
```diff
@@ -0,0 +1,56 @@
...
+     */
+    void populateMenu(QMenu *menu);
+
+    /**
+     * Displays the dialog to edit the commands
+     */
+    void showDialog();
+
+private:
+    QMenu *mMainWindowMenu;
```
ketanhwr: Didn't notice this, sorry!

src/tiled/commandbutton.cpp
```diff
@@ -34,6 +35,7 @@ using namespace Tiled::Internal;
 CommandButton::CommandButton(QWidget *parent)
     : QToolButton(parent)
     , mMenu(new QMenu(this))
+    , mCommandManager(new CommandManager(this))
```
ketanhwr: I created this instance so I could connect the signal `QMenu::aboutToShow` to the slot `CommandManager::populateMenu`. How can I do this without creating a new instance?

ketanhwr: I could pass the singleton instance, but that would lead to passing it via `TilesetEditor` and `MapEditor` to `MainToolbar` to `CommandButton` which won't be an effective solution, I think.

ketanhwr: @bjorn what should I do regarding this problem? ^

bjorn: You don't pass around singletons! Please look up the design pattern. See the `TilesetManager` for an example singleton.

ketanhwr: Alright, thanks :+1:

ketanhwr:
Should I shift `CommandManager` from namespace `Tiled\Internal` to namespace `Tiled` like `TilesetManager`?

bjorn:
> Should I shift CommandManager from namespace Tiled\Internal to namespace Tiled like TilesetManager?

No, the `TilesetManager` is outside of the `Internal` namespace because it lives in `src/libtiled`. The `Internal` basically means `Editor`.

ketanhwr:
@bjorn, I've made the changes. Please have a look.

src/tiled/commandmanager.cpp
```diff
@@ -0,0 +1,107 @@
...
+#include <QLatin1String>
+#include <QMenu>
+
+namespace Tiled {
+namespace Internal {
+
+CommandManager *CommandManager::mInstance;
+
+CommandManager::CommandManager()
+    : mModel(new CommandDataModel)
```
bjorn: Even though this is a singleton, for correctness sake please do still pass the manager as the parent of the CommandDataModel (you'll also still need to add that constructor parameter and pass it on to the superclass). This makes sure the CommandDataModel is deleted when the CommandManager is deleted.

src/tiled/commandmanager.cpp
```diff
@@ -0,0 +1,107 @@
...
+    return mInstance;
+}
+
+void CommandManager::deleteInstance()
+{
+    delete mInstance;
+    mInstance = nullptr;
+}
+
+CommandDataModel *CommandManager::getCommandDataModel()
```
bjorn: Naming style: getters do not have a "get" prefix, this should just be `commandDataModel`.

src/tiled/commandmanager.cpp
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    // Add Edit Commands action
+    if (!menu->isEmpty())
+        menu->addSeparator();
+
+    QAction *mEditCommands = new QAction(this);
+    mEditCommands->setIcon(
+            QIcon(QLatin1String(":/images/24x24/system-run.png")));
+    mEditCommands->setText(tr("Edit Commands..."));
```
bjorn: Please add:
```c++
Utils::setThemeIcon(mEditCommands, "system-run");
```

src/tiled/commandbutton.cpp
```diff
@@ -42,7 +43,11 @@ CommandButton::CommandButton(QWidget *parent)
     setPopupMode(QToolButton::MenuButtonPopup);
     setMenu(mMenu);
 
-    connect(mMenu, SIGNAL(aboutToShow()), SLOT(populateMenu()));
+    // Populate the menu once in order to register shortcuts beforehand
+    CommandManager::instance()->populateMenu(mMenu);
+
+    connect(mMenu, &QMenu::aboutToShow, 
+            CommandManager::instance(), [this]() { CommandManager::instance()->populateMenu(mMenu); });
```
bjorn: Since we know the `CommandButton` will not outlive the `CommandManager`, we can simplify this to:
```c++
    connect(mMenu, &QMenu::aboutToShow, 
            [this]() { CommandManager::instance()->populateMenu(mMenu); });
```
Same for the connection to `mUi->menuCommand` in the `mainwindow.cpp`.

ketanhwr: Nice!

ketanhwr:
@bjorn, done! :-)

bjorn:
Alright, I just tested it and unfortunately it's not working. When I press F5 it's outputting:
```
QAction::eventFilter: Ambiguous shortcut overload: F5
```
I think it is because there are still two actions bound to F5, it just so happens that they would execute the same command, but Qt doesn't know this. So you'll need to make sure the same QAction instances are used in each menu.

ketanhwr:
It's working fine on mine. Intact, I just realised, the shortcuts for the actions in the File menu are not getting registered. But they are being registered in your machine, apparently. Why is this happening?

ketanhwr:
I am thinking of converting `populateMenu(QMenu *)` to `populateMenu(QMenu *, bool)` and this depending on this bool variable, I'll register the shortcut on that particular menu. This will solve the above problem, but I still don't understand why the shortcuts are not getting registered in the File menu on my machine :/

ketanhwr:
@bjorn, can you test this again?

ketanhwr:
I guess this is just a part of #1458. I was facing the same problem before as well. Basically, the shortcuts registered on the menu aren't working.

bjorn:
You're probably right that you can't reproduce it because of #1458. I don't know what is causing that problem, though I think it must have something to do with Unity and the menu bar integration plugin they're using.

I would prefer if you re-used the same QAction instances across menus because this would allow the shortcuts to show in both locations. But I understand this will change the patch quite a bit again, so it could be done as a follow-up improvement instead.

So I'll merge this in the evening. If you have a bit of time, please still add a separator also below the Commands entry in the File menu.

ketanhwr:
I'll look into this issue once, and try to see what's causing this. I've added the separator in my last commit.

ketanhwr:
For using the same `QAction` instances, I was thinking that I could store pointer to both the menus in the `CommandManager` class and the slot `populateMenu` would populate both the menus at once. Would this be okay?

ketanhwr:
I've done the above changes in 49b9cd4a2e0a345a7a78a8d7c05c67a8543b0a42, this is working alright for me.

I defined a `QList<QMenu*>` which stores the pointers to all the menus. The function `addMenu(QMenu*)` appends a menu to the `QList`. Now whenever `populateMenu` is called, it creates actions in the first menu and adds the same actions to all the other menus. Although I'm assuming that no menu is ever destructed (is this assumption true?).

If this seems fine, I can add these changes to this pull request itself.

bjorn:
@ketanhwr I think it's almost fine. I would prefer if you did it in such a way that you first build up the list of actions and then just call `addActions` on each menu. Also I think `registerMenu` would be a better name than `addMenu`.

bjorn:
Note that you can use `QAction::setSeparator` to add a separator to the list, so that you don't have to special-case that and the Edit Commands action.

ketanhwr:
Can you merge this one? I'll make another pull request for this update.

bjorn:
> Can you merge this one? I'll make another pull request for this update.

Since this still introduces the ambiguous shortcut issue, I think you should fix that before I merge.

ketanhwr:
I fixed that with 5b378a58a2b80466b06a395ecda2866802c3d52a
I guess you didn't notice this commit in between the comments :-/

bjorn:
> I guess you didn't notice this commit in between the comments :-/

Sorry, I had noticed but I had forgotten it yesterday evening somehow. I had even said "But I understand this will change the patch quite a bit again, so it could be done as a follow-up improvement instead."

Anyway, now it's merged, and thanks a lot for your contribution! I'm looking forward to your follow-up patch and hopefully the customizable shortcuts later on. :-)

ketanhwr:
Thanks a lot!  :)

Issue #1190, opened 2016-01-28
Highlight hovered objects when using Object Selection Tool


bjorn:
To make it easier to see which object you'll be interacting with before clicking, the hovered object should be highlighted in some way.

Based on feedback at http://forum.mapeditor.org/t/art-study-castlevania-legends/1070

vikasTmz:
Hello @bjorn, I'd like to work on this feature. How should the highlight on the hovered object appear? A bounding box centered on the object or a translucent layer on the object?

Could you guide me on how the tile mousehover highlight was done or which relevant code should I go through.

PR #1465, opened 2017-03-05
Added #1190 feature : object highlight on mouseover

vikasTmz:
Highlight hovered objects when using Object Selection Tool #1190

### DIFF
--- a/src/tiled/objectselectiontool.cpp
+++ b/src/tiled/objectselectiontool.cpp
@@ -17,7 +17,6 @@
  * You should have received a copy of the GNU General Public License along with
  * this program. If not, see <http://www.gnu.org/licenses/>.
  */
-
 #include "objectselectiontool.h"
 
 #include "changepolygon.h"
@@ -686,6 +685,8 @@ static bool canResizeAbsolute(const MapObject *object)
 
     return false;
 }
+
+bool highlightState = false;
 
 /* This function returns the actual bounds of the object, as opposed to the
  * bounds of its visualization that the MapRenderer::boundingRect function
@@ -1431,7 +1432,7 @@ void ObjectSelectionTool::saveSelectionState()
 void ObjectSelectionTool::refreshCursor()
 {
     Qt::CursorShape cursorShape = Qt::ArrowCursor;
-
+    const MapObject *prevObject = nullptr;
     switch (mAction) {
     case NoAction: {
         const bool hasSelection = !mapScene()->selectedObjectItems().isEmpty();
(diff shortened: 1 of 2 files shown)


vikasTmz:
#1190

ketanhwr:
@vikasTmz why did you close this PR?

vikasTmz:
There was a bug in my code. It's reopened now.

src/tiled/objectselectiontool.cpp
```diff
@@ -1431,20 +1433,45 @@ void ObjectSelectionTool::saveSelectionState()
...
-
+    const MapObject *prevObject = nullptr;
     switch (mAction) {
     case NoAction: {
         const bool hasSelection = !mapScene()->selectedObjectItems().isEmpty();
 
         if ((mHoveredObjectItem || ((mModifiers & Qt::AltModifier) && hasSelection && !mHoveredHandle)) &&
                 !(mModifiers & Qt::ShiftModifier)) {
             cursorShape = Qt::SizeAllCursor;
+            MapObject *currentObject = mHoveredObjectItem->mapObject();
```
bjorn: This may crash since there is no guarantee `mHoveredObjectItem` isn't null at this point.

src/tiled/objectselectiontool.cpp
```diff
@@ -1431,20 +1433,45 @@ void ObjectSelectionTool::saveSelectionState()
...
+
+            if (currentObject != prevObject) {
```
bjorn: `prevObject` is always going to still be `nullptr` here. `prevObject` will go out of scope after it is being assigned, so all the values assigned to it will never be used.

src/tiled/objectselectiontool.cpp
```diff
@@ -687,6 +687,8 @@ static bool canResizeAbsolute(const MapObject *object)
     return false;
 }
 
+bool highlightState = false;
```
bjorn: This variable would really need to be a member of `ObjectSelectionTool`, otherwise you'll run into problems once two instances of the tool are being used (which is actually the case, since another instance of the tool is used in the tile collision editor).

src/tiled/objectselectiontool.cpp
```diff
@@ -1431,20 +1433,45 @@ void ObjectSelectionTool::saveSelectionState()
...
+            if (currentObject != prevObject) {
+                prevObject = currentObject;
+                if (highlightState)
+                    mapScene()->removeItem(mSelectionRectangle);
+
+                MapRenderer *renderer = mapDocument()->renderer();
+                QRectF boundingRect = objectBounds(currentObject, renderer,
+                                                   objectTransform(currentObject, renderer));
+
+                mSelectionRectangle->setRectangle(boundingRect);
```
bjorn: I think it's clever to reuse the selection rectangle, but it doesn't really do what I'd like to see. There is another class, the `MapObjectOutline` in `objectselectionitem.cpp` that I think would be more suitable because it matches the position as well as rotation of the object. You could define a new class `MapObjectHighlight` in `objectselectiontool.cpp` as a copy of `MapObjectOutline` and tweak it's `paint` override to look like the one of the `SelectionRectangle` class.

Later on, I think I'd like to change the look of the hover highlight, but that would suffice for now.

src/tiled/objectselectiontool.cpp
vikasTmz: where should I get the `mapObject` from instead ?

bjorn: You get it from the `mHoveredObjectItem`, but only if it isn't null.

vikasTmz:
@bjorn I've added the necessary changes.

src/tiled/objectselectiontool.cpp
```diff
@@ -1431,20 +1489,43 @@ void ObjectSelectionTool::saveSelectionState()
 void ObjectSelectionTool::refreshCursor()
 {
     Qt::CursorShape cursorShape = Qt::ArrowCursor;
-
+    QHash<MapObject*, MapObjectHighlight*> highlightItems;
```
vikasTmz: > I think especially for the second case, it could make sense to move the management of this hover highlight into ObjectSelectionItem, because it is already connected to all the relevant signals. 

Yes, I'll start working on this change. Ran into some class assignments lately, so I may need a day or two.

> would be necessary to communicate the hovered object from the ObjectSelectionTool to the ObjectSelectionItem. I think it would be best to do this through the MapDocument, adding an mHoveredObject pointer there.

How will this exactly solve dereferencing a roaming pointer? Or you mean to say since the second issue will be solved, this issue can not be reproduced?

bjorn: > How will this exactly solve dereferencing a roaming pointer? Or you mean to say since the second issue will be solved, this issue can not be reproduced?

Communicating the hovered object through the MapDocument is only a design change that makes the implementation easier (by moving the display to `ObjectSelectionItem`, where it is more easily kept up to date) and it would enable possible future uses like highlighting the hovered object in the Objects view at the same time. It has nothing to do with solving the roaming pointer issue.

bjorn:
Closing this since #1190 was implemented a while ago.

Issue #1424, opened 2017-01-03
Change current layer depending on selected objects


bjorn:
When the list of selected objects changes and all of them are on a different layer than the current one, it would be helpful to change the current layer. I've noticed this behavior in Inkscape.

gardenvarietyse:
Came here to suggest exactly this!

One issue I have occasionally is, I'll select an object and duplicate it, but oh no, it's on the wrong layer because the selection didn't change. I tend to give them different colors which helps avoid this, but if I didn't I'd probably end up not even noticing every time.

bjorn:
@gardenvarietyse If you duplicate an object, the duplicate ends up on the same layer as the original object, regardless of the currently selected layer. Still, I'm sure we can avoid some confusion by changing the current layer.

I just realized that `ObjectsView::selectionChanged` already implements this behavior, but it only works when objects are selected in the Objects view.

PR #1466, opened 2017-03-05
Change current layer depending on selected objects, usability issue #1424

Glavak:
I added little usability improvement from #1424. Now when selecting object not from current layer, the layer with selected object will become active.
I'm not sure should I write my code as static function or private method.

### DIFF
--- a/src/tiled/mapscene.cpp
+++ b/src/tiled/mapscene.cpp
@@ -166,33 +166,6 @@ void MapScene::setMapDocument(MapDocument *mapDocument)
     refreshScene();
 }
 
-// If none of selected objects is on current layer, sets current layer to one with selected objects
-static void updateCurrentLayerBySelection(MapDocument *mapDocument)
-{
-    QList<MapObject*> selectedObjects = mapDocument->selectedObjects();
-
-    if (selectedObjects.empty())
-        return;
-
-    ObjectGroup *currentObjectLayer = dynamic_cast<ObjectGroup *>(mapDocument->currentLayer());
-
-    // Return if any of selected objects is in selected layer
-    for (MapObject *object : selectedObjects) {
-        if (currentObjectLayer->objects().contains(object))
-            return;
-    }
-
-    // Find layer that contains first selected object, and set it as current
-    for (Layer *layer : mapDocument->map()->layers()) {
-        ObjectGroup *layerAsObjectGroup = dynamic_cast<ObjectGroup *>(layer);
-
-        if (layerAsObjectGroup && layerAsObjectGroup->objects().contains(selectedObjects.at(0))) {
-            mapDocument->setCurrentLayer(layer);
-            return;
-        }
-    }
-}
-
 void MapScene::setSelectedObjectItems(const QSet<MapObjectItem *> &items)
 {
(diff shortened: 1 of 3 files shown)


src/tiled/mapscene.cpp
```diff
@@ -166,6 +166,33 @@ void MapScene::setMapDocument(MapDocument *mapDocument)
...
+    ObjectGroup *currentObjectLayer = dynamic_cast<ObjectGroup *>(mapDocument->currentLayer());
+
+    // Return if any of selected objects is in selected layer
+    for (MapObject *object : selectedObjects) {
+        if (currentObjectLayer->objects().contains(object))
+            return;
+    }
+
+    // Find layer that contains first selected object, and set it as current
+    for (Layer *layer : mapDocument->map()->layers()) {
```
bjorn: This unfortunately no longer iterates all layers. You need to use the `LayerIterator` for that, because the layers for a hierarchy. However, you actually don't need to iterate all layers, because you're looking for `selectedObjects.first()->objectGroup()`.

Glavak:
Thank you for pointing at my mistakes.
I got a bit confused with layer vs object group, and didn't notice the `object->objectGroup()` getter first time. Also with a new approach code got simpler, so I put it directly to the `MapDocument::setSelectedObjects` method
And I just noticed that @WinterGrascph ended up with almost exactly the same code as mine

WinterGrascph:
@Glavak Hey :)

Sorry to have made a mess, you were first.

bjorn:
@Glavak Actually you've put it in `MapScene::setSelectedObjectItems`, so please still move it to `MapDocument::setSelectedObjects`. Also, when you do so you can clean up the related code in `ObjectsView::selectionChanged` because it will be no longer needed.

While I agree with @WinterGrascph that it's nice to squash the commits, this is no longer necessary on GitHub since there is a "Squash and merge" button, and I can look at the squashed version in "Files changed". So for small patches like this I prefer to squash when merging, since then it remains clear what changes were made based on feedback. Squashing fixups remains preferable in more complicated changes where it would be nice to preserve some of the development history and keep a merge commit.

Glavak:
It makes sense for code to be in MapDocument, so I moved it there.
I also noticed that when selecting multiple objects from other layer, this layers gets not only active, but also selected in a properties window, which is a bit unexpected behavior. I guess I should dig deeper in source codes to understand all the qt signals that leads to that behavior and fix it

bjorn:
@Glavak Yes, I think you're right that in this case the "current object" should be left alone. It is actually the `setCurrentObject` call in `setCurrentLayer` that causes this behavior. I think we could add a check there, so that it only sets the current object when the previous current object was already a layer (`typeId() == Object::LayerType`).

Glavak:
Thank you for pointing the right approach, now it seems to work fine

bjorn:
Thanks!

Unfortunately, I think I pressed the merge button a little too soon though... upon second review I realized that probably there is no guarantee that `currentObject()` won't return null, so it may crash in some cases. I'll add a check there.

bjorn:
Amended with change 11f1ef91c966bb9a0c61ed890fff303e81357be2.

PR #1468, opened 2017-03-05
Switch current layer when selecting objects from a single layer.

WinterGrascph:
Fix for #1424. When selecting objects from the editor, the current
layer is switched when the objects are from only a single layer.

### DIFF
--- a/src/tiled/mapdocument.cpp
+++ b/src/tiled/mapdocument.cpp
@@ -670,7 +670,7 @@ void MapDocument::setSelectedObjects(const QList<MapObject *> &selectedObjects)
 
     // Switch the current object layer if only one object layer (and/or its objects)
     // are included in the current selection.
-    if(singleObjectGroup)
+    if (singleObjectGroup)
         setCurrentLayer(singleObjectGroup);
 
     if (selectedObjects.size() == 1)
--- a/src/tiled/mapdocument.cpp
+++ b/src/tiled/mapdocument.cpp
@@ -659,6 +659,20 @@ void MapDocument::setSelectedObjects(const QList<MapObject *> &selectedObjects)
     mSelectedObjects = selectedObjects;
     emit selectedObjectsChanged();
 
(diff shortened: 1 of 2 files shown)


bjorn:
It's nice that you used the same approach as in `ObjectsView::selectionChanged`. However, rather than duplicating this functionality I think it can be moved to `MapDocument::setSelectedObjects` like I also just noted in #1466.

It's a bit unfortunate to have two people at the same time working to fix this issue. :-)

WinterGrascph:
I've moved the code to MapDocument::setSelectedObjects.

Sadly I only noticed too late, that @Glavak has already started working on the issue.

src/tiled/mapdocument.cpp
```diff
@@ -659,6 +659,20 @@ void MapDocument::setSelectedObjects(const QList<MapObject *> &selectedObjects)
...
+        if (singleObjectGroup && nextMapObject->objectGroup() != singleObjectGroup) {
+            singleObjectGroup = nullptr;
+            break;
+        }
+        singleObjectGroup = nextMapObject->objectGroup();
+    }
+
+    // Switch the current object layer if only one object layer (and/or its objects)
+    // are included in the current selection.
+    if(singleObjectGroup)
```
bjorn: Coding style: please put a space after `if`.

src/tiled/objectsdock.cpp
```diff
@@ -368,11 +368,6 @@ void ObjectsView::selectionChanged(const QItemSelection &selected,
             selectedObjects.append(o);
     }
 
-    // Switch the current object layer if only one object layer (and/or its objects)
-    // are included in the current selection.
-    if (singleObjectGroup)
-        mMapDocument->setCurrentLayer(singleObjectGroup);
```
bjorn: Since `singleObjectGroup` is no longer actually used now, you can clean up the code above.

bjorn:
Thanks, @WinterGrascph!

Issue #1462, opened 2017-03-02
In objects view being able to add columns


issy123:
### Request
It would be nice if we could add columns to the objects list. Like id, x,y.
When you right click on the columns you should see a list of columns you can add/remove.

### Example

![27u04nt](https://cloud.githubusercontent.com/assets/531764/23527189/404523f2-ff95-11e6-8d1d-200d70e3010b.png)

### Reason

Same as #1453

### Note

It would be nice to add custom attributes too! But maybe I ask too much.

bjorn:
> It would be nice if we could add columns to the objects list. Like id, x,y.
> When you right click on the columns you should see a list of columns you can add/remove.

Good idea! It could be implemented by using `QHeaderView::setSectionHidden` and `Qt::CustomContextMenu`.

> Same as #1453

What does this have to do with exporting to PNG? Wrong number?

> It would be nice to add custom attributes too! But maybe I ask too much.

That is definitely asking too much! :-P

Actually this would be entirely possible, but the implementation would look very different and could work relatively independent of the other feature, so it should just be done in its own patch. You would not use `setSectionHidden` but would instead talk to the `MapObjectModel` about adding custom columns.

issy123:
>>Same as #1453

>What does this have to do with exporting to PNG? Wrong number?

This:
> Some games may have user contributed levels/maps (take for example happy wheels). We want the users to make their own map with tiled and upload a .tmx file. 

It has to do with parsing contributed maps and giving the users some information to work with if errors occur.


If I tell the user the object with id x has an invalid value of property y
If he has to click on every single object he has made to get object x, this wouldn't be user-friendly.

You may ask why not refer to object name or type? Well it can be that this is also not defined, the only way to refer to an object with empty values is by it's ID or location (x,y).

Ps. i'm heavily using tiled so expect more feature requests 🏭 
Hope you won't mind ghehe

bjorn:
Alright, that makes sense.

> Ps. i'm heavily using tiled so expect more feature requests 
> Hope you won't mind ghehe

Not at all, it's great receiving your feedback!

bjorn:
An example implementation of this can be found at https://github.com/rsachetto/sir/blob/0f7d463e9be25c883e49bc734dfc136564c4fe93/src/widgets/treewidgetheader.cpp. Though I'd prefer an implementation that did not rely on subclassing the QHeaderView, but rather used `Qt::CustomContextMenu` to change the context menu of the existing header.

Issue #1467, opened 2017-03-05
Being able to search objects


issy123:
### Request
It would be nice if we could search objects by it's id,name or type.

### Example

![image](https://cloud.githubusercontent.com/assets/9143253/23588346/a3330bae-01bc-11e7-8d23-39eb7d22fc8f.png)

### Reason

Same as #1462

### Note

It would be nice to add custom attributes too! But maybe I ask too much(yeah yeah I repeat myself ik).

bjorn:
This would be a nice feature, though I would probably prefer the primary search function to take more the form of the Qt Creator "Locate..." (Ctrl+K) feature. The right side of the open documents tab bar seems like a suitable place for something like this.

Of course, that's a little bit more tricky since it involves a bit more complicated input handling. On the other hand, Qt Creator source code can be used as a reference.

bjorn:
It just occurred to me that another option would be to add a filter box on top of the objects view, like I did in the "Tile Stamps" dock. That shouldn't be very hard to implement. The downside is that it provides no real way of filtering on certain properties specifically, but I think you rarely need that.

thabetx:
I started working on this issue. I'm trying to change the base class for ReversingProxyModel from QAbstractProxyModel to QSortFilterProxyModel and implement the filtering function, would like to know if this would be the preferred approach.

bjorn:
@thabetx Nice that you started working on this! Actually turning the `ReversingProxyModel` into a subclass of `QSortFilterProxyModel` is likely not going to work or at least will be a lot more complicated than simply chaining these proxies together. Check out how `MapEditor::layerComboActivated` uses two chained proxy models

thabetx:
Alright, I thought chaining would be more complicated, I will check it out, thank you.

thabetx:
I'm not sure about group names should the search match them?

PR #1469, opened 2017-03-06
Added switchable columns to objects view

i-ka:
Implemented feature requested in #1462. Now objects view has switchable id, x, y columns. Columns can be switched by calling context menu on objects view header.

### DIFF
--- a/src/tiled/mapobjectmodel.cpp
+++ b/src/tiled/mapobjectmodel.cpp
@@ -445,7 +445,7 @@ void MapObjectModel::emitObjectsChanged(const QList<MapObject *> &objects, const
     }
 }
 
-void MapObjectModel::emitObjectsChanged(const QList<MapObject *> &objects, const MapObjectModel::Column column)
+void MapObjectModel::emitObjectsChanged(const QList<MapObject *> &objects, Column column)
 {
     emitObjectsChanged(objects,
                        QList<MapObjectModel::Column>() << column);
--- a/src/tiled/mapobjectmodel.cpp
+++ b/src/tiled/mapobjectmodel.cpp
@@ -43,6 +43,8 @@ MapObjectModel::MapObjectModel(QObject *parent):
     mObjectGroupIcon(QLatin1String(":/images/16x16/layer-object.png"))
 {
     mObjectGroupIcon.addFile(QLatin1String(":images/32x32/layer-object.png"));
+    connect(this, &MapObjectModel::objectsChanged,
+            this, &MapObjectModel::emitObjectsDataChanged);
 }
 
 QModelIndex MapObjectModel::index(int row, int column,
@@ -430,6 +432,13 @@ void MapObjectModel::tileTypeChanged(Tile *tile)
             }
         }
     }
+}
+
+void MapObjectModel::emitObjectsDataChanged(const QList<MapObject *> &objects)
+{
+    emit dataChanged(index(objects.first(), Name),
+                     index(objects.last(), ColumnCount - 1),
+                     QVector<int>() << Qt::EditRole);
 }
 
 QList<Layer *> &MapObjectModel::filteredChildLayers(GroupLayer *parentLayer) const
--- a/src/tiled/mapobjectmodel.cpp
+++ b/src/tiled/mapobjectmodel.cpp
@@ -108,7 +108,7 @@ int MapObjectModel::rowCount(const QModelIndex &parent) const
 int MapObjectModel::columnCount(const QModelIndex &parent) const
 {
     Q_UNUSED(parent)
-    return ColumnsCount;
+    return ColumnCount;
 }
 
 QVariant MapObjectModel::data(const QModelIndex &index, int role) const
@@ -241,10 +241,10 @@ Qt::ItemFlags MapObjectModel::flags(const QModelIndex &index) const
     Qt::ItemFlags rc = QAbstractItemModel::flags(index);
     if (index.column() == 0)
         rc |= Qt::ItemIsUserCheckable | Qt::ItemIsEditable;
-    else if (toMapObject(index))
-        if (index.column() == Type) { // allow to edit only type column
+    else if (toMapObject(index)) {
+        if (index.column() == Type)// allow to edit only type column
             rc |= Qt::ItemIsEditable;
-        }
+    }
     return rc;
 }
 
@@ -255,7 +255,7 @@ QVariant MapObjectModel::headerData(int section, Qt::Orientation orientation,
         switch (section) {
         case Name: return tr("Name");
         case Type: return tr("Type");
-        case Id: return tr("Id");
+        case Id: return tr("ID");
         case X: return tr("X");
         case Y: return tr("Y");
         }
--- a/src/tiled/mapobjectmodel.cpp
+++ b/src/tiled/mapobjectmodel.cpp
@@ -433,7 +433,7 @@ void MapObjectModel::tileTypeChanged(Tile *tile)
     }
 }
 
-void MapObjectModel::emitObjectsChanged(const QList<MapObject *> &objects, const QList<Columns> &columns)
+void MapObjectModel::emitObjectsChanged(const QList<MapObject *> &objects, const QList<Column> &columns)
 {
     emit objectsChanged(objects);
     if (columns.isEmpty())
@@ -443,6 +443,12 @@ void MapObjectModel::emitObjectsChanged(const QList<MapObject *> &objects, const
     for (auto object : objects) {
         emit dataChanged(index(object, *minMaxPair.first), index(object, *minMaxPair.second));
     }
+}
+
+void MapObjectModel::emitObjectsChanged(const QList<MapObject *> &objects, const MapObjectModel::Column column)
+{
+    emitObjectsChanged(objects,
+                       QList<MapObjectModel::Column>() << column);
 }
 
 QList<Layer *> &MapObjectModel::filteredChildLayers(GroupLayer *parentLayer) const
--- a/src/tiled/mapobjectmodel.cpp
+++ b/src/tiled/mapobjectmodel.cpp
@@ -108,7 +108,7 @@ int MapObjectModel::rowCount(const QModelIndex &parent) const
 int MapObjectModel::columnCount(const QModelIndex &parent) const
 {
     Q_UNUSED(parent)
-    return 2; // MapObject name|type
+    return ColumnsCount;
 }
 
 QVariant MapObjectModel::data(const QModelIndex &index, int role) const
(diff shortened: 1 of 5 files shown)


src/tiled/objectsdock.cpp
```diff
@@ -401,6 +405,45 @@ void ObjectsView::selectedObjectsChanged()
...
+    contextMenu.exec(QCursor::pos());
+}
+
+void ObjectsView::updateColumnVisibilityActions()
+{
+    hideExtraColumns();
+    mActions.clear();
+    QAbstractItemModel *model = mProxyModel->sourceModel();
+    for (int i = 0; i < model->columnCount(); i++) {
+        if (i == MapObjectModel::Name || i == MapObjectModel::Type)
```
bjorn: I think it's fine to allow hiding the Type column, it just needs to stay visible by default.

src/tiled/mapobjectmodel.cpp
```diff
@@ -108,7 +108,7 @@ int MapObjectModel::rowCount(const QModelIndex &parent) const
 int MapObjectModel::columnCount(const QModelIndex &parent) const
 {
     Q_UNUSED(parent)
-    return 2; // MapObject name|type
+    return ColumnsCount;
```
bjorn: Small nit: should be `ColumnCount`.

src/tiled/objectsdock.h
```diff
@@ -98,12 +98,18 @@ private slots:
...
+    void showCustomMenu(const QPoint &point);
 
 private:
+    void updateColumnVisibilityActions();
+    void hideExtraColumns();
     void synchronizeSelectedItems();
 
     MapDocument *mMapDocument;
     QAbstractProxyModel *mProxyModel;
+    QList<QAction*> mActions;
```
bjorn: Please call this `mColumnActions`, just to be a little more clear. However, why keep around the actions at all? You could also just build up the menu in `showCustomMenu` and remove `updateColumnVisibilityActions`.

src/tiled/objectsdock.cpp
```diff
@@ -401,6 +405,45 @@ void ObjectsView::selectedObjectsChanged()
...
+    Q_UNUSED(point)
+    QMenu contextMenu(this);
+    contextMenu.addActions(mActions);
+    contextMenu.exec(QCursor::pos());
+}
+
+void ObjectsView::updateColumnVisibilityActions()
+{
+    hideExtraColumns();
+    mActions.clear();
```
bjorn: This leaks all the QAction instances. However, as per another comment, I think you don't need `mActions` at all.

src/tiled/mapobjectmodel.cpp
```diff
@@ -233,7 +242,9 @@ Qt::ItemFlags MapObjectModel::flags(const QModelIndex &index) const
     if (index.column() == 0)
         rc |= Qt::ItemIsUserCheckable | Qt::ItemIsEditable;
     else if (toMapObject(index))
-        rc |= Qt::ItemIsEditable; // MapObject type
+        if (index.column() == Type) { // allow to edit only type column
+            rc |= Qt::ItemIsEditable;
+        }
```
bjorn: Coding style: Please wrap the outer if/else if with braces and remove them from the nested if.

src/tiled/mapobjectmodel.cpp
```diff
@@ -242,8 +253,11 @@ QVariant MapObjectModel::headerData(int section, Qt::Orientation orientation,
 {
     if (role == Qt::DisplayRole && orientation == Qt::Horizontal) {
         switch (section) {
-        case 0: return tr("Name");
-        case 1: return tr("Type");
+        case Name: return tr("Name");
+        case Type: return tr("Type");
+        case Id: return tr("Id");
```
bjorn: I've used "ID" in the Properties view for this field, so this should also use uppercase.

i-ka:
I think maybe unite X and Y columns in one Position column

src/tiled/mapobjectmodel.cpp
```diff
@@ -43,6 +43,8 @@ MapObjectModel::MapObjectModel(QObject *parent):
     mObjectGroupIcon(QLatin1String(":/images/16x16/layer-object.png"))
 {
     mObjectGroupIcon.addFile(QLatin1String(":images/32x32/layer-object.png"));
+    connect(this, &MapObjectModel::objectsChanged,
+            this, &MapObjectModel::emitObjectsDataChanged);
```
bjorn: Instead of connecting this signal, I'd prefer if `emitObjectsDataChanged` was only called when needed, and only with the actually changed columns. The `objectsChanged` signal gets emitted for many things that aren't actually exposed by this model, and `dataChanged` is already emitted for a few things explicitly as well.

src/tiled/mapobjectmodel.h
```diff
@@ -101,6 +110,7 @@ private slots:
     void layerChanged(Layer *layer);
     void layerAboutToBeRemoved(GroupLayer *groupLayer, int index);
     void tileTypeChanged(Tile *tile);
+    void emitObjectsDataChanged(const QList<MapObject *> &objects);
```
bjorn: I think this can be a private function, and I'd prefer the name to be simply `emitDataChanged`, since the parameter already makes clear that it is for objects.

src/tiled/objectsdock.cpp
```diff
@@ -401,6 +406,52 @@ void ObjectsView::selectedObjectsChanged()
...
+{
+    QAction *action = qobject_cast<QAction*>(sender());
+    if (!action)
+        return;
+
+    int column = action->data().toInt();
+    header()->setSectionHidden(column, !visible);
+
+    QSettings *settings = Preferences::instance()->settings();
+    QVariantList showedColumns;
```
bjorn: `visibleSections`

src/tiled/objectsdock.cpp
```diff
@@ -401,6 +406,52 @@ void ObjectsView::selectedObjectsChanged()
...
+        action->setCheckable(true);
+        action->setChecked(!header()->isSectionHidden(i));
+        action->setData(i);
+        connect(action, &QAction::triggered, this, &ObjectsView::setColumnVisibility);
+        contextMenu.addAction(action);
+    }
+    contextMenu.exec(QCursor::pos());
+}
+
+void ObjectsView::hideExtraColumns()
```
bjorn: This name isn't accurate, maybe `restoreVisibleSections` would be better.

src/tiled/mapobjectmodel.cpp
i-ka: I tried to emit dataChanged in ```MapObjectModel::setObjectPosition```, but seems ```setObjectPosition``` behaves not as expected when object was moved by mouse. It doesn't changes object position because it already changed directly by ```ObjectSelectionTool``` and  model doesn't know about changes. I think emit ```dataChanged``` in ```ObjectSelectionTool``` it's not a good idea. :-)

i-ka:
I tried to emit dataChanged in ```MapObjectModel::setObjectPosition```, but seems ```setObjectPosition``` behaves not as expected when object was moved by mouse. It doesn't changes object position because it already changed directly by ```ObjectSelectionTool``` and  model doesn't know about changes. I think emit ```dataChanged``` in ```ObjectSelectionTool``` it's not a good idea. :-)

bjorn:
Right, that doesn't work because `objectsChanged` is emitted after applying new positions to possibly multiple objects, as a performance optimization (compared to calling `setObjectPosition` on the model for each object and then each time emitting the signal for a single object).

A possible fix would be to add a `const QList<MapObject::Property> &properties` argument to the `emitObjectsChanged` function, and to use that to possibly emit `dataChanged` signals in addition to the `objectsChanged` signal. And then of course the `ObjectSelectionTool` should use `emitObjectsChanged` and pass in the right properties each time.

src/tiled/mapobjectmodel.cpp
```diff
@@ -434,11 +433,16 @@ void MapObjectModel::tileTypeChanged(Tile *tile)
...
+void MapObjectModel::emitObjectsChanged(const QList<MapObject *> &objects, const QList<Columns> &columns)
 {
-    emit dataChanged(index(objects.first(), Name),
-                     index(objects.last(), ColumnCount - 1),
-                     QVector<int>() << Qt::EditRole);
+    emit objectsChanged(objects);
+    if (columns.isEmpty())
+        return;
+
+    auto minMaxPair = std::minmax_element(columns.begin(), columns.end());
```
bjorn: That's really neat, I didn't even realize STL had such a helper function. :-)

src/tiled/mapobjectmodel.cpp
```diff
@@ -418,6 +433,18 @@ void MapObjectModel::tileTypeChanged(Tile *tile)
     }
 }
 
+void MapObjectModel::emitObjectsChanged(const QList<MapObject *> &objects, const QList<Columns> &columns)
```
bjorn: Thanks for realizing this should indeed by `Columns` rather than `MapObject::Property`! However, please rename the enum to `Column`.

src/tiled/objectselectiontool.cpp
```diff
@@ -1020,7 +1020,8 @@ void ObjectSelectionTool::updateMovingItems(const QPointF &pos,
         mapObject->setPosition(newPos);
     }
 
-    emit mapDocument()->mapObjectModel()->objectsChanged(changingObjects());
+    mapDocument()->mapObjectModel()->emitObjectsChanged(changingObjects(),
+                                                        QList<MapObjectModel::Columns>() << MapObjectModel::Position);
```
bjorn: So I see now that currently the only interesting column to be included here is the Position column. I think a helper override taking a single `MapObjectModel::Column` argument, which can in turn just call the list version, would be nice.

src/tiled/mapobjectmodel.h
```diff
@@ -97,7 +97,8 @@ class MapObjectModel : public QAbstractItemModel
     void setObjectRotation(MapObject *o, qreal rotation);
 
     void setObjectProperty(MapObject *o, MapObject::Property property, const QVariant &value);
-    void emitObjectsChanged(const QList<MapObject *> &objects, const QList<Columns> &columns = QList<Columns>());
+    void emitObjectsChanged(const QList<MapObject *> &objects, const QList<Column> &columns = QList<Column>());
+    void emitObjectsChanged(const QList<MapObject*> &objects, const Column column);
```
bjorn: Please leave out the `const` here, it makes no difference to the caller whether a parameter passed by value is declared const or not. Sorry to nitpick about it, but I don't want to do a commit just to remove it. :-)

src/tiled/mapobjectmodel.cpp
```diff
@@ -445,6 +445,12 @@ void MapObjectModel::emitObjectsChanged(const QList<MapObject *> &objects, const
     }
 }
 
+void MapObjectModel::emitObjectsChanged(const QList<MapObject *> &objects, const MapObjectModel::Column column)
```
bjorn: You don't need to prefix with `MapObjectModel::` there since you're already in the `MapObjectModel` context.

bjorn:
Really nice change, thanks for finishing it!

Do you have any plans of maybe also tackling #1467, or maybe adding support for adding custom property columns?

i-ka:
Thanks for feedback!

I think, i can tackling #1467. If it will be implemented with custom property support, then there will no needed to add custom property columns :)

PR #1479, opened 2017-03-09
Remember tileset zoom level in the Edit Terrain Information dialog

Sharganov:


src/tiled/preferences.h
```diff
@@ -37,7 +37,7 @@ namespace Internal {
  * This class holds user preferences and provides a convenient interface to
  * access them.
  */
-class Preferences : public QObject
+class  Preferences : public QObject
```
bjorn: Please don't add a space there.

src/tiled/preferences.h
```diff
@@ -88,6 +88,9 @@ class Preferences : public QObject
     Map::RenderOrder mapRenderOrder() const;
     void setMapRenderOrder(Map::RenderOrder mapRenderOrder);
 
+    qreal getTerrainEditorZoom(QString name);
```
bjorn: Naming style: getters do not have a "get" prefix, so this would be just `terrainEditorZoom`. Also, the getter should be defined `const`.

Efficiency: Please pass `QString` always as `const QString &`. Even if of minor impact, this avoids the reference counting so it's good to get used to always pass it like that.

src/tiled/tileseteditor.cpp
```diff
@@ -235,6 +241,7 @@ void TilesetEditor::removeDocument(Document *document)
     tilesetDocument->disconnect(this);
 
     TilesetView *view = mViewForTileset.take(tilesetDocument);
+    Preferences::instance()->setTerrainEditorZoom(tilesetDocument->fileName(),view->scale());
```
bjorn: Please call it `setTilesetScale`, since this isn't only for terrain and the property is already called `scale`.

Coding style: please put a space after `,`

src/tiled/preferences.cpp
```diff
@@ -350,6 +350,22 @@ void Preferences::setMapRenderOrder(Map::RenderOrder mapRenderOrder)
                         mMapRenderOrder);
 }
 
+void Preferences::setTerrainEditorZoom(QString name, qreal value)
+{
+    QString str;
+    str += QLatin1String("Zoom/");
```
bjorn: Please use a little more specific prefix, like `"TilesetEditor/TilesetScales/"`

src/tiled/preferences.h
ketanhwr: @Sharganov you haven't fixed this one.

Sharganov: oh, miss it. Thanks @ketanhwr

src/tiled/preferences.cpp
```diff
@@ -350,6 +350,22 @@ void Preferences::setMapRenderOrder(Map::RenderOrder mapRenderOrder)
                         mMapRenderOrder);
 }
 
+void Preferences::setTilesetScale(const QString & name, qreal value)
+{
+    QString str;
+    str += QLatin1String("TilesetEditor/TilesetScales/");
+    str += name;
```
bjorn: While testing this I noticed that for embedded tilesets this name is empty, which makes sense since they do not have their own file name. To avoid storing a shared zoom level for all embedded tilesets, I think it would make sense to fall back on the name of the tileset.

src/tiled/tileseteditor.cpp
```diff
@@ -196,7 +196,13 @@ void TilesetEditor::addDocument(Document *document)
 
     TilesetView *view = new TilesetView(mWidgetStack);
     view->setTilesetDocument(tilesetDocument);
-    view->setZoomable(new Zoomable(this));
+
+    auto zoomable = new Zoomable(this);
+    auto fileName = tilesetDocument->fileName();
+    qreal zoom = Preferences::instance()->TilesetScale(fileName);
```
bjorn: Actually, I think you can move this loading and saving of the zoom level into the `TilesetView`. That way it will work everywhere, including the Tilesets dock in the map editor and the tileset view in the animation editor.

That does mean that the Zoomable instance should be created and owned by the TilesetView. You can remove `TilesetView::setZoomable` and instead only have a getter. You'll need to use `Zoomable::setComboBox` in `TilesetDock::currentTilesetChanged` to make sure the combo box in the TilesetsDock displays the right zoom level.

Sharganov: Do you want that zoom in TilesetsDock window to be the same as in Edit Tileset window? I don't understand what "TilesetsDock displays the right zoom level" mean

bjorn: The Tilesets dock currently displays the zoom level of its own single Zoomable instance, which it sets on each created TilesetView. That way, all tileset views are at the same zoom level. However, you are now storing the zoom level for each tileset, and the Tilesets dock should probably use this per-tileset zoom level as well.

On the other hand, it now occurred to me that you may want to zoom out the tileset in the dock and not in the editor, so I'm now wondering whether it would then make sense to store two zoom levels for each tileset...

Sharganov: hmm, it is exactly the problem i try to solve now. 

If to store two different zooms, you need to add path parameter to `Preferences::tileSetScale` and call it from `TilesetDock` or `TilesetEditor`.

However, using the same zoom value not what the user expects, i think.

bjorn: > If to store two different zooms, you need to add path parameter to Preferences::tileSetScale and call it from TilesetDock or TilesetEditor, TileAnimationEditor

Let's keep the behavior more like it is already, and only add the per-tileset zoom level to the TilesetEditor. For the TileAnimationEditor and the TilesetDock, and actually also the TileCollisionEditor, it would already be a nice improvement if they'd simply store and restore their current zoom level.

bjorn: Please do this by interacting with QSettings directly, rather than adding more to the Preferences class.

Sharganov: Ok, just one more question, every tileset in docker should has the same zoom that was stored or only first opened?

Sharganov: The problem is that different tilesets had different zooms in docker, so open them with the same zoom not much than open with 1.0, i think.

bjorn:
> The problem is that different tilesets had different zooms in docker when you closed Tiled editor, so open them with the same zoom not much better than open with 1.0, i think.

The current behavior in Tiled is that all tilesets in the Tilesets dock share the same zoom level. The only thing I'd change, for now, is to simply make this zoom level persistent. Just forget about my request to move the storing and restoring of the zoom level into the TilesetView, because that raises questions that can be considered later.

Sharganov:
I've tried to implement some ideas which discussed, hope i understand what you mean.

src/tiled/tileanimationeditor.cpp
```diff
@@ -270,12 +270,10 @@ TileAnimationEditor::TileAnimationEditor(QWidget *parent)
     mUi->setupUi(this);
     resize(Utils::dpiScaled(size()));
 
-    Zoomable *zoomable = new Zoomable(this);
-    zoomable->setComboBox(mUi->zoomComboBox);
-
     mUi->frameList->setModel(mFrameListModel);
     mUi->tilesetView->setMarkAnimatedTiles(false);
-    mUi->tilesetView->setZoomable(zoomable);
+    //TODO: test
```
bjorn: Did you test it?

Sharganov: no, because i don't how to use tile animation editor :)

bjorn: Ah, you can open it using _Tileset -> Tile Animation Editor_ (when you have a tileset open).

Sharganov: hm, looks like, it works fine

src/tiled/tilesetdock.cpp
```diff
@@ -392,8 +390,12 @@ void TilesetDock::dropEvent(QDropEvent *e)
 void TilesetDock::currentTilesetChanged()
 {
     if (const TilesetView *view = currentTilesetView())
+    {
```
bjorn: Coding style: please place the `{` on the same line as the `if` (like for `for` and `while`, except for functions).

bjorn:
Thanks a lot for this improvement!

Sharganov:
Thanks for your help! :)

PR #1480, opened 2017-03-10
Same QAction to all QMenu. Follow up patch for #1463

ketanhwr:
This is the follow-up patch for #1463. I've also modified the code on the basis of comments mentioned by @bjorn in the same pull request.

I've added a function `registerMenu(QMenu *menu)` which adds `menu` to the list of all menus that the singleton `CommandManger` has stored in `QList<QMenu*> mMenus`. Another function `updateActions` is added, which deletes all the previously created `QAction` instances and updates the list on the basis of `CommandDataModel`. The slot `populateMenu` will just add all the `QAction` instances present in `mActions` to all the registered `QMenu`.

I've also fixed a small bug: whenever the `Edit Commands...` dialog is prompted, and the list is updated, the effect is not visible until the user opens a particular menu (the menus and the commands get updated only when some menu is opened). So what I've done now is that I fired the `populateMenu` function when the `Close` button is hit in the Dialog. This fixed the problem.

### DIFF
--- a/src/tiled/commanddialog.cpp
+++ b/src/tiled/commanddialog.cpp
@@ -46,8 +46,7 @@ CommandDialog::CommandDialog(QWidget *parent)
     setWindowTitle(tr("Edit Commands"));
     Utils::restoreGeometry(this);
 
-    connect(mUi->pushButton, &QPushButton::clicked, 
-            [this]() { CommandManager::instance()->populateMenu(); });
+    setWindowFlags(Qt::Window | Qt::WindowTitleHint | Qt::CustomizeWindowHint);
 }
 
 CommandDialog::~CommandDialog()
@@ -62,6 +61,8 @@ void CommandDialog::accept()
 
     mUi->treeView->model()->setSaveBeforeExecute(mUi->saveBox->isChecked());
     mUi->treeView->model()->commit();
+
+    CommandManager::instance()->updateActions();
 }
 
 CommandTreeView::CommandTreeView(QWidget *parent)
--- a/src/tiled/commanddialog.cpp
+++ b/src/tiled/commanddialog.cpp
@@ -45,8 +45,6 @@ CommandDialog::CommandDialog(QWidget *parent)
 
     setWindowTitle(tr("Edit Commands"));
     Utils::restoreGeometry(this);
-
-    setWindowFlags(Qt::Window | Qt::WindowTitleHint | Qt::CustomizeWindowHint);
 }
 
 CommandDialog::~CommandDialog()
@@ -55,9 +53,9 @@ CommandDialog::~CommandDialog()
     delete mUi;
 }
 
-void CommandDialog::accept()
+void CommandDialog::closeEvent(QCloseEvent *event)
 {
-    QDialog::accept();
+    QDialog::closeEvent(event);
 
     mUi->treeView->model()->setSaveBeforeExecute(mUi->saveBox->isChecked());
     mUi->treeView->model()->commit();
--- a/src/tiled/commanddialog.cpp
+++ b/src/tiled/commanddialog.cpp
@@ -45,6 +45,9 @@ CommandDialog::CommandDialog(QWidget *parent)
 
     setWindowTitle(tr("Edit Commands"));
     Utils::restoreGeometry(this);
+
+    connect(mUi->pushButton, &QPushButton::clicked, 
+            [this]() { CommandManager::instance()->populateMenu(); });
 }
 
 CommandDialog::~CommandDialog()
--- a/src/tiled/commandmanager.cpp
+++ b/src/tiled/commandmanager.cpp
@@ -59,17 +59,37 @@ CommandDataModel *CommandManager::commandDataModel()
     return mModel;
 }
 
+void CommandManager::registerMenu(QMenu *menu)
+{
+    mMenus.append(menu);
+}
+
 void CommandManager::showDialog()
 {
     CommandDialog dialog(QApplication::activeWindow());
     dialog.exec();
 }
 
-void CommandManager::populateMenu(QMenu *menu, bool flag)
+void CommandManager::populateMenu()
 {
-    menu->clear();
+    updateActions();
 
-    // Add all enabled commands
+    for (int i = 0; i < mMenus.size(); ++i) {
+        QMenu *menu = mMenus.at(i);
+        menu->clear();
+
+        for (int j = 0; j < mActions.size(); ++j) {
+            menu->addAction(mActions.at(j));
+        }
+    }
+
+}
+
+void CommandManager::updateActions()
+{
+    for (int i = 0; i < mActions.size(); ++i)
+        delete mActions.at(i);
+    mActions.clear();
 
     bool firstEnabledCommand = true;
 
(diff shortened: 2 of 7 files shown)


src/tiled/commandmanager.h
```diff
@@ -39,13 +40,22 @@ class CommandManager : public QObject
...
+     */
+    void registerMenu(QMenu* menu);
+
 public slots:
+
     /**
      * Populates the menu pointed by menu
      */
-    void populateMenu(QMenu *menu, bool flag = false);
+    void populateMenu();
```
bjorn: Since this populates all registered menus, it should be called `populateMenus` now. You also still need to update the documentation.

src/tiled/commandmanager.h
```diff
@@ -57,8 +67,15 @@ public slots:
 
     CommandManager();
 
+    /**
+     * Updates mActions QList
+     */
```
bjorn: Please don't write comments about things that are immediately obvious when reading the code.

src/tiled/commandmanager.cpp
```diff
@@ -59,17 +59,37 @@ CommandDataModel *CommandManager::commandDataModel()
...
-    menu->clear();
+    updateActions();
+
+    for (int i = 0; i < mMenus.size(); ++i) {
+        QMenu *menu = mMenus.at(i);
+        menu->clear();
+
+        for (int j = 0; j < mActions.size(); ++j) {
+            menu->addAction(mActions.at(j));
+        }
```
bjorn: Please use `menu->addActions(mActions)`.

src/tiled/commandmanager.cpp
```diff
@@ -59,17 +59,37 @@ CommandDataModel *CommandManager::commandDataModel()
     return mModel;
 }
 
+void CommandManager::registerMenu(QMenu *menu)
+{
+    mMenus.append(menu);
```
bjorn: I think it would be a good idea to add the actions to this menu here. That way, you don't need to call `populateMenu` from the outside at all and it can be made a private function.

src/tiled/commanddialog.cpp
```diff
@@ -45,6 +45,9 @@ CommandDialog::CommandDialog(QWidget *parent)
 
     setWindowTitle(tr("Edit Commands"));
     Utils::restoreGeometry(this);
+
+    connect(mUi->pushButton, &QPushButton::clicked, 
+            [this]() { CommandManager::instance()->populateMenu(); });
```
bjorn: You don't need a connection for this, since just calling it from the `accept` method should work fine.

src/tiled/commandmanager.cpp
```diff
@@ -59,17 +59,37 @@ CommandDataModel *CommandManager::commandDataModel()
...
+        }
+    }
+
+}
 
-    // Add all enabled commands
+void CommandManager::updateActions()
+{
+    for (int i = 0; i < mActions.size(); ++i)
+        delete mActions.at(i);
```
bjorn: Qt has `qDeleteAll(mActions)` for this.

src/tiled/commandmanager.cpp
```diff
@@ -81,28 +101,35 @@ void CommandManager::populateMenu(QMenu *menu, bool flag)
         if (!command.isEnabled)
             continue;
 
-        QAction *mAction = menu->addAction(command.name);
-        if (firstEnabledCommand && flag)
+        QAction *mAction = new QAction(this);
+
+        QString commandText = command.name;
```
bjorn: Please avoid the needless local variable. You can even pass this directly into the QAction constructor.

src/tiled/mainwindow.cpp
```diff
@@ -316,8 +316,10 @@ MainWindow::MainWindow(QWidget *parent, Qt::WindowFlags flags)
     connect(mUi->actionZoomNormal, SIGNAL(triggered()), SLOT(zoomNormal()));
     connect(mUi->actionFullScreen, &QAction::toggled, this, &MainWindow::setFullScreen);
 
+    CommandManager::instance()->registerMenu(mUi->menuCommand);
+
     connect(mUi->menuCommand, &QMenu::aboutToShow, 
-            [this]() { CommandManager::instance()->populateMenu(mUi->menuCommand); });
+            [this]() { CommandManager::instance()->populateMenu(); });
```
bjorn: You don't need this connection if you're keeping the menu contents updated already anyway.

src/tiled/commandmanager.cpp
ketanhwr: Making `populateMenu` private would require `updateActions` to be public. I would  then call `updateActions` each time the Command Dialog is opened.

src/tiled/commandmanager.cpp
ketanhwr: Passing it to the local constructor directly wasn't working for me. I'll try again.

ketanhwr: It works now, I must've been doing some stupid mistake :/

src/tiled/commanddialog.cpp
ketanhwr: I realised another bug, if the user clicks on the cross button of the dialog instead of clicking on the close button, the results don't get committed! I'll fix this in the next commit.

ketanhwr: Since there are no `Okay` and `Cancel` options, I think disabling the Close (X) button in the CommandDialog would be a simple solution for this.

ketanhwr:
@bjorn I've made `populateMenus` private and this function would be called by `updateActions` method. `updateActions` would be called everytime the `CommandDialog` is closed.

src/tiled/commanddialog.cpp
```diff
@@ -45,6 +45,8 @@ CommandDialog::CommandDialog(QWidget *parent)
 
     setWindowTitle(tr("Edit Commands"));
     Utils::restoreGeometry(this);
+
+    setWindowFlags(Qt::Window | Qt::WindowTitleHint | Qt::CustomizeWindowHint);
```
bjorn: Please instead of removing the close button, which I'm sure will confuse some people, change the `accept` method to a `closeEvent` override. It didn't actually make sense to keep it as `accept` when you can only just close the dialog anyway. You'll also need to connect the "Close" button at the bottom to the "close()" slot instead of the "accept()" slot.

src/tiled/commandmanager.cpp
```diff
@@ -59,17 +59,32 @@ CommandDataModel *CommandManager::commandDataModel()
...
     CommandDialog dialog(QApplication::activeWindow());
     dialog.exec();
 }
 
-void CommandManager::populateMenu(QMenu *menu, bool flag)
+void CommandManager::populateMenus()
 {
-    menu->clear();
+    for (int i = 0; i < mMenus.size(); ++i) {
+        QMenu *menu = mMenus.at(i);
```
bjorn: Since you don't actually care about the index of the menu, just use:
```c++
    for (QMenu *menu : mMenus) {
```

src/tiled/commanddialog.cpp
ketanhwr: I was searching for this thing only, except I didn't know about this. Thanks!

ketanhwr:
@bjorn done :)

bjorn:
Thanks! :-)

src/tiled/commandmanager.cpp
ketanhwr: @bjorn, I'm a bit confused here. In `updateActions` too, we didn't actually use the indices. Why didn't you suggest using this structure there as well?

bjorn: You were not using indexes? `updateActions` does use the index now at least, in the following line:
```c++
connect(mAction, &QAction::triggered, [this,i]() { mModel->execute(i); });
```

ketanhwr: Oh forgot about this!

Issue #1475, opened 2017-03-07
Flipping a group of objects doesn't behave like expected


bjorn:
When you select a group of objects and flip them, each object is flipped individually. Instead, the group should be flipped around its shared center point.

This also means that object rotation needs to be handled correctly at part of the flip.

See `FlipMapObjects::flip`.

Mentioned by @eevee in his [feedback post](http://discourse.mapeditor.org/t/feedback-after-a-couple-months-with-tiled/2090).

PR #1481, opened 2017-03-10
Added flipping a group of objects like expected

Vitek1425:
Implemented feature requested in #1475. Now group of objects flipping like expected

### DIFF
--- a/src/libtiled/mapobject.cpp
+++ b/src/libtiled/mapobject.cpp
@@ -200,7 +200,7 @@ void MapObject::setMapObjectProperty(Property property, const QVariant &value)
  * Flip this object in the given \a direction. This doesn't change the size
  * of the object.
  */
-void MapObject::flip(FlipDirection direction)
+void MapObject::flip(FlipDirection direction, const QPointF &flipCenter)
 {
     if (!mCell.isEmpty()) {
         if (direction == FlipHorizontally)
@@ -209,16 +209,40 @@ void MapObject::flip(FlipDirection direction)
             mCell.setFlippedVertically(!mCell.flippedVertically());
     }
 
-    if (!mPolygon.isEmpty()) {
-        const QPointF center2 = mPolygon.boundingRect().center() * 2;
+    //creating flipmatrix
+    QMatrix flipMatrix;
+    flipMatrix.translate(flipCenter.x(), flipCenter.y());
+    if (direction == FlipHorizontally)
+        flipMatrix.scale(1, -1);
+    else //direction == FlipVertically
+        flipMatrix.scale(-1, 1);
+    flipMatrix.translate(-flipCenter.x(), -flipCenter.y());
 
-        if (direction == FlipHorizontally) {
-            for (int i = 0; i < mPolygon.size(); ++i)
-                mPolygon[i].setX(center2.x() - mPolygon[i].x());
-        } else if (direction == FlipVertically) {
-            for (int i = 0; i < mPolygon.size(); ++i)
-                mPolygon[i].setY(center2.y() - mPolygon[i].y());
-        }
+    //computing position offset
+    QPointF positionOffset;
+    QSizeF objectSize = mPolygon.empty() ? size() : mPolygon.boundingRect().size();
+    if (direction == FlipHorizontally)
+        positionOffset = QPointF(0, objectSize.height());
+    else //direction == FlipVertically
+        positionOffset = QPointF(objectSize.width(), 0);
+
+    if (!mPolygon.isEmpty()) {  //fliping polygon
+        QPointF polygonCenter = mPolygon.boundingRect().center();
+
+        QMatrix flipPolygonMatrix;
+        flipPolygonMatrix.translate(polygonCenter.x(), polygonCenter.y());
+        if (direction == FlipHorizontally)
+            flipPolygonMatrix.scale(1, -1);
+        else //direction == FlipVertically
+            flipPolygonMatrix.scale(-1, 1);
+        flipPolygonMatrix.translate(-polygonCenter.x(), -polygonCenter.y());
+
+
+        mPolygon = flipPolygonMatrix.map(mPolygon);
+        setPosition(flipMatrix.map(position()) - positionOffset);
+    }
+    else {
+        setPosition(flipMatrix.map(position() + positionOffset));
     }
 }
 
--- a/src/libtiled/mapobject.cpp
+++ b/src/libtiled/mapobject.cpp
@@ -240,7 +240,7 @@ void MapObject::flipRectObject(FlipDirection direction, const QPointF &origin)
         flipTransform.scale(1, -1);
     flipTransform.translate(-origin.x(), -origin.y());
 
-    // point 1 is position
+    // point 0 is position
     // 0-----1
     // |     |
     // 3-----2
@@ -276,7 +276,7 @@ void MapObject::flipPolygonObject(FlipDirection direction, const QPointF &origin
         flipTransform.scale(1, -1);
     flipTransform.translate(-origin.x(), -origin.y());
 
-    // point 1 is position
+    // point 0 is position
     // 0-----1
     // |     |
     // 3-----2
@@ -333,7 +333,7 @@ void MapObject::flipTileObject(FlipDirection direction, const QPointF &origin)
         flipTransform.scale(1, -1);
     flipTransform.translate(-origin.x(), -origin.y());
 
-    // point 1 is position
+    // point 0 is position
     // 3-----2
     // |     |
     // 0-----1
--- a/src/libtiled/mapobject.cpp
+++ b/src/libtiled/mapobject.cpp
@@ -276,44 +276,46 @@ void MapObject::flipPolygonObject(FlipDirection direction, const QPointF &origin
         flipTransform.scale(1, -1);
     flipTransform.translate(-origin.x(), -origin.y());
 
-    // point 0 is position
-    // 0-----1
-    // |     |
-    // 3-----2
+    // PolygonBorder
+    // 0_________1
+    // | /‾‾‾‾‾‾\|
+    // |/  .pos /|
+    // |\      / |
+    // | \____/  |
+    // 3‾‾‾‾‾‾‾‾‾2
 
-    QRectF polygonRect = mPolygon.boundingRect();
-    qreal polyWidth = polygonRect.width();
-    qreal polyHeight = polygonRect.height();
+    QTransform polygonToMapTransform;
+    polygonToMapTransform.translate(position().x() + mPolygon.first().x(),
+                                    position().y() + mPolygon.first().y());
+    polygonToMapTransform.rotate(rotation());
+    polygonToMapTransform.translate(-(mPolygon.first().x()),
+                                    -(mPolygon.first().y()));
 
     QPointF points[4];
-    points[0] = position();
-    points[1] = QPointF(cos(qDegreesToRadians(rotation())) * polyWidth + x(),
-                        sin(qDegreesToRadians(rotation())) * polyWidth + y());
-
-    points[3] = QPointF(cos(qDegreesToRadians(rotation() + 90)) * polyHeight + x(),
-                            sin(qDegreesToRadians(rotation() + 90)) * polyHeight + y());
-
-    points[2] = QPointF(cos(qDegreesToRadians(rotation())) * polyWidth + points[3].x(),
-                        sin(qDegreesToRadians(rotation())) * polyWidth + points[3].y());
-
-
+    points[0] = polygonToMapTransform.map(mPolygon.boundingRect().normalized().topLeft());
+    points[1] = polygonToMapTransform.map(mPolygon.boundingRect().normalized().topRight());
+    points[2] = polygonToMapTransform.map(mPolygon.boundingRect().normalized().bottomRight());
+    points[3] = polygonToMapTransform.map(mPolygon.boundingRect().normalized().bottomLeft());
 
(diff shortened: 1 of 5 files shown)


src/libtiled/mapobject.cpp
```diff
@@ -209,16 +209,40 @@ void MapObject::flip(FlipDirection direction)
...
-        }
+    //computing position offset
+    QPointF positionOffset;
+    QSizeF objectSize = mPolygon.empty() ? size() : mPolygon.boundingRect().size();
+    if (direction == FlipHorizontally)
+        positionOffset = QPointF(0, objectSize.height());
+    else //direction == FlipVertically
+        positionOffset = QPointF(objectSize.width(), 0);
+
+    if (!mPolygon.isEmpty()) {  //fliping polygon
```
bjorn: Typo "fliping"

src/libtiled/mapobject.cpp
```diff
@@ -209,16 +209,40 @@ void MapObject::flip(FlipDirection direction)
...
+            flipPolygonMatrix.scale(1, -1);
+        else //direction == FlipVertically
+            flipPolygonMatrix.scale(-1, 1);
+        flipPolygonMatrix.translate(-polygonCenter.x(), -polygonCenter.y());
+
+
+        mPolygon = flipPolygonMatrix.map(mPolygon);
+        setPosition(flipMatrix.map(position()) - positionOffset);
+    }
+    else {
```
bjorn: Coding style: please put `} else {` on one line.

src/tiled/flipmapobjects.cpp
```diff
@@ -43,10 +43,23 @@ FlipMapObjects::FlipMapObjects(MapDocument *mapDocument,
 
 void FlipMapObjects::flip()
 {
-    // TODO: Flip them properly as a group
     const auto &objects = mMapObjects;
+
+    //computing objects center
+    QPainterPath boundaringPath;
+    for (MapObject *object : objects)
+    {
```
bjorn: Coding style: please put `{` on the same line.

src/libtiled/mapobject.cpp
```diff
@@ -209,16 +209,40 @@ void MapObject::flip(FlipDirection direction)
             mCell.setFlippedVertically(!mCell.flippedVertically());
     }
 
-    if (!mPolygon.isEmpty()) {
-        const QPointF center2 = mPolygon.boundingRect().center() * 2;
+    //creating flipmatrix
+    QMatrix flipMatrix;
```
bjorn: The documentation of QMatrix says "This class is obsolete. It is provided to keep old source code working. We strongly advise against using it in new code."

Please use QTransform instead.

src/libtiled/mapobject.cpp
```diff
@@ -200,7 +200,7 @@ void MapObject::setMapObjectProperty(Property property, const QVariant &value)
  * Flip this object in the given \a direction. This doesn't change the size
  * of the object.
  */
-void MapObject::flip(FlipDirection direction)
+void MapObject::flip(FlipDirection direction, const QPointF &flipCenter)
```
bjorn: Naming: I would call it "origin" instead of "flipCenter".

src/libtiled/mapobject.cpp
```diff
@@ -209,16 +209,40 @@ void MapObject::flip(FlipDirection direction)
             mCell.setFlippedVertically(!mCell.flippedVertically());
     }
 
-    if (!mPolygon.isEmpty()) {
-        const QPointF center2 = mPolygon.boundingRect().center() * 2;
+    //creating flipmatrix
+    QMatrix flipMatrix;
+    flipMatrix.translate(flipCenter.x(), flipCenter.y());
+    if (direction == FlipHorizontally)
+        flipMatrix.scale(1, -1);
```
bjorn: While possibly ambiguous, this changes the interpretation of what a horizontal flip is compared to what it was before. A horizontal flip is supposed to flip the left/right sides, whereas a vertical flip is supposed to flip top/bottom. This means horizontal flip should use a -1 scale on x rather than y.

src/libtiled/mapobject.cpp
```diff
@@ -209,16 +209,40 @@ void MapObject::flip(FlipDirection direction)
...
-            for (int i = 0; i < mPolygon.size(); ++i)
-                mPolygon[i].setY(center2.y() - mPolygon[i].y());
-        }
+    //computing position offset
+    QPointF positionOffset;
+    QSizeF objectSize = mPolygon.empty() ? size() : mPolygon.boundingRect().size();
+    if (direction == FlipHorizontally)
+        positionOffset = QPointF(0, objectSize.height());
+    else //direction == FlipVertically
+        positionOffset = QPointF(objectSize.width(), 0);
```
bjorn: Note that the object's origin is unfortunately different depending on whether we're dealing with a shape object or a tile object. Please make sure it works in both cases.

Vitek1425:
question
Flipping must be like number one or two?
If like two, then on number three axis angles must be computing like average rotation objects angle?

Vitek1425:
test24
I improved code again.
The flipping code is so complicated, because a position of an object is not a center of an object.
Why code MapObject::bounds() dont calculate real boundary? It is a bug?

bjorn:
> Why code MapObject::bounds() dont calculate real boundary? It is a bug?

The real boundary in what coordinate space? The object position is in "pixel coordinates" (which aren't always "screen coordinates", the MapRenderer maps between these). The object size is in "pixel coordinates" for rectangle and circle objects, but in "screen coordinates" for tile objects. So `bounds` is already inconsistent in this way and its return value needs to be interpreted depending on the context.

For polygons and polylines in particular, you usually want to know the bounds in "screen coordinates", at least for things like finding the visual center, but the `bounds` function would only be able to give you the bounds in "pixel coordinates". Though I do think this may be more useful than returning an empty rectangle at the polygon starting point...

In general, we have unfortunately the situation of a rather inconsistent state, where on an isometric map, all shape objects are projected but tile objects are not. It simply didn't make sense to project tile objects, but that did introduce an inconsistency. In addition, rotation is applied after projection, which is actually really strange if you think about it, and it's mainly because of the way it has been implemented (as a rotation of the QGraphicsItem) and because it was anyway mostly useful for tile objects only (which aren't projected anyway). Finally, tile objects have a different origin than rectangle and ellipse objects (and this even depends on the renderer, cause on isometric maps, their origin is bottom center whereas it is bottom left for all other maps).

So all in all, when we look at all usages of `MapObject::bounds`, we'll find many to be actually problematic and not handling certain cases outlined above correctly.

Also, I think this unfortunately causes many problems with implementing the flipping correctly. I have no idea for example, how one would flip an isometric projected polygon and a non-projected tile object together "correctly". The only thing that saves us, is that it is a very unlikely situation for a user to get himself into.

src/tiled/propertybrowser.cpp
```diff
@@ -947,15 +947,19 @@ QUndoCommand *PropertyBrowser::applyMapObjectValueTo(PropertyId id, const QVaria
         const bool flippedVertically = flippingFlags & 2;
 
         // You can only change one checkbox at a time
+        Cell newCell = mapObject->cell();
         if (mapObject->cell().flippedHorizontally() != flippedHorizontally) {
-            command = new FlipMapObjects(mMapDocument,
-                                         QList<MapObject*>() << mapObject,
-                                         FlipHorizontally);
+            newCell.setFlippedHorizontally(!mapObject->cell().flippedHorizontally());
```
bjorn: You don't need to toggle the existing value, please just do:
```c++
newCell.setFlippedHorizontally(flippedHorizontally);
```
You can even remove the condition and so you can assign the flipped state directly with `flippingFlags & 1` and `flippingFlags & 2`, since there is no longer a need for the additional variables regarding readability.

src/libtiled/mapobject.cpp
```diff
@@ -197,29 +199,17 @@ void MapObject::setMapObjectProperty(Property property, const QVariant &value)
 }
 
 /**
- * Flip this object in the given \a direction. This doesn't change the size
+ * Flip this object in the given \a direction . This doesn't change the size
```
bjorn: Please don't add a space there.

bjorn:
I think in general there is an issue here in that the rotation may or may not need to be changed, for the flip to work. In case of tile objects, the flip does not need to affect the rotation as long as the rotation is a multiple of 90 degrees. In case of a polygon, the flip could always be done internally, but it may be easier to follow the same logic as for a tile object. And finally the rectangle and circle objects are not affected at all by 180 degree rotations, apart from their location.

Personally I would expect the rotation to stay unaffected when it is not necessary, but of course I'd also want the flipping code to be as straight-forward as possible.

As it stands, I have two main issues with the patch:

* The approach for finding the new rotation of the object seems overly complicated. Flipping a rotation horizontally should be (180° - rotation) and flipping vertically should be simply negating the rotation.

* There is some bug with flipping polygon objects that can cause them to move. Check out the behavior when trying to flip the "guard" NPC path in the bottom-left of the `orthogonal-outside.tmx` map, it's actually shifting about and trying to undo it only makes it worse. I think you shouldn't be using the first point of the polygon when setting up `polygonToMapTransform` (and maybe you can remove this transform entirely).

Vitek1425:
I fixed polygon rotation and the approach for finding the new rotation, 
but i always change the rotation, because if i add condition for rotation, the code become overly complicated.
If you think it is necessary i can to do it.

src/libtiled/mapobject.cpp
```diff
@@ -240,4 +230,83 @@ MapObject *MapObject::clone() const
...
+    flipTransform.translate(origin.x(), origin.y());
+    qreal newRotation = 0;
+    if (direction == FlipHorizontally) {
+        newRotation = 180.0 - rotation();
+        flipTransform.scale(-1, 1);
+    } else { //direction == FlipVertically
+        flipTransform.scale(1, -1);
+        newRotation = -rotation();
+    }
+    flipTransform.translate(-origin.x(), -origin.y());
```
bjorn: The whole bit about determining the new rotation and the flip transform seems to be the same for each type of object. Please move that bit into `MapObject::flip` and pass the transform into the other functions instead of the direction and origin.

src/tiled/flipmapobjects.cpp
```diff
@@ -43,10 +44,37 @@ FlipMapObjects::FlipMapObjects(MapDocument *mapDocument,
...
+            QTransform polygonToMapTransform;
+            polygonToMapTransform.translate(object->x(),
+                                            object->y());
+            polygonToMapTransform.rotate(object->rotation());
+            boundaryObjectsRect = boundaryObjectsRect.united(polygonToMapTransform.mapRect(QRectF(objectPolygon.boundingRect())));
+        } else { //computing bound rect for other
+            boundaryObjectsRect = boundaryObjectsRect.united(objectTransform.mapRect(object->bounds()));
+        }
+    }
+    QPointF objectsCenter = boundaryObjectsRect.center();
```
bjorn: Since the list of objects is not going to change after the `FlipMapObjects` instance is created, please move the calculation of the center to the constructor.

Vitek1425:
I updated the code

src/tiled/flipmapobjects.cpp
```diff
@@ -40,13 +40,9 @@ FlipMapObjects::FlipMapObjects(MapDocument *mapDocument,
...
                                         "Flip %n Object(s)",
                                         nullptr, mapObjects.size()));
-}
-
-void FlipMapObjects::flip()
-{
-    const auto &objects = mMapObjects;
 
     //computing objects center
+    const auto &objects = mMapObjects;
```
bjorn: Please just use `mapObjects`, since it's the same list and already `const &`.

Vitek1425: I found two same lines.

src/tiled/flipmapobjects.cpp
```diff
@@ -39,14 +40,39 @@ FlipMapObjects::FlipMapObjects(MapDocument *mapDocument,
     setText(QCoreApplication::translate("Undo Commands",
                                         "Flip %n Object(s)",
                                         nullptr, mapObjects.size()));
+
+    //computing objects center
+    QRectF boundaryObjectsRect;
+    for (MapObject *object : mMapObjects) {
```
bjorn: It is important to use `mapObjects` here rather than `mMapObjects`, because the latter is not `const`. Iterating a non-const Qt container that has implicit copies (as is the case here) will cause it to detach (make a deep copy).

src/tiled/flipmapobjects.cpp
```diff
@@ -39,14 +40,39 @@ FlipMapObjects::FlipMapObjects(MapDocument *mapDocument,
...
+            boundaryObjectsRect = boundaryObjectsRect.united(objectTransform.mapRect(object->bounds()));
+        }
+    }
+    mObjectsCenter = boundaryObjectsRect.center();
 }
 
 void FlipMapObjects::flip()
 {
-    // TODO: Flip them properly as a group
-    const auto &objects = mMapObjects;
```
bjorn: Please keep this line since using the `const` alias can prevent a deep copy of the (implicitly shared) Qt container.

bjorn:
Alternatively, add `const` to the `mMapObjects` member.

Vitek1425:
I added const to the mMapObjects member.

bjorn:
Thanks a lot! I think it's a great improvement.

Now, I'll still have to resolve this annoying ambiguous shortcut "X", which since recently can also means "Insert Text"... :-S

Vitek1425:
What's the problem? 
I can't found it.

PR #1482, opened 2017-03-12
Custom shortcuts for commands - Fixes #1456

### LINKED ISSUE #1456: Custom shortcuts on commands
While customizable shortcuts in general would be nice (#215), there is a special case for commands. Users may set up an arbitrary number of them, but at the moment there is only one hardcoded shortcut (F5) for running the first enabled one.

It would be nice if custom shortcuts could be set in the Edit Commands dialog. For this, a separate (not in the table view) QKeySequenceEdit could be added. These shortcuts then need to be registered on the QMainWindow, probably by instantiating QShortcut instances that will map to the custom commands.

http://discourse.mapeditor.org/t/more-shortcuts-for-commands/2082

ketanhwr:
This fixes #1456. Here's a gist of what I've done:

- Added a `QKeySequenceEdit` and a `QLabel` in the `CommandDialog`.
- The value of `QKeySequenceEdit` gets updated whenever a new row in the `CommandTreeView` is selected.
- When the value of `QKeySequenceEdit` is changed, the shortcut for the respective command gets changed and updated in the menu as well.

Here are the screenshots:
Screenshot 1
Screenshot 2

Although here's a thing I'm confused about: If the user enters a key sequence which is already registered, then it will result in the ambiguous shortcut error. We can probably remove this by allowing these shortcuts to be the function keys.


src/tiled/commanddialog.h
```diff
@@ -78,8 +88,12 @@ private slots:
       */
     void removeSelectedCommands();
 
+    void currentChanged(const QModelIndex &current,
+                        const QModelIndex&) override;
+
 private:
     CommandDataModel *mModel;
+    CommandDialog *mDialog;
```
ketanhwr: Is there anyway of accessing some function of the parent QWidget instead of storing it in `mDialog` ?

src/tiled/command.cpp
```diff
@@ -105,6 +107,8 @@ Command Command::fromQVariant(const QVariant &variant)
         command.name = hash[namePref].toString();
     if (hash.contains(commandPref))
         command.command = hash[commandPref].toString();
+    if (hash.contains(shortcutPref))
+        command.shortcut = QKeySequence(hash[shortcutPref].toString());
```
bjorn: Since you didn't convert the key sequence to a string when you put it in the variant, a better way to get the key sequence back out should be `hash[shortcutPref].value<QKeySequence>()`. I didn't try it though.

src/tiled/commanddatamodel.cpp
```diff
@@ -433,6 +434,22 @@ bool CommandDataModel::dropMimeData(const QMimeData *data, Qt::DropAction, int,
...
 
+QKeySequence CommandDataModel::shortcut(const QModelIndex &index)
+{
+    const bool isNormalRow = index.row() < mCommands.size();
+
+    if (isNormalRow) {
+        return mCommands[index.row()].shortcut;
+    }
+    else
+        return QKeySequence();
```
bjorn: I'd prefer a rather shorter implementation:
```c++
if (index.isValid())
    return mCommands[index.row()].shortcut;
return QKeySequence();
```
Also this method should be declared `const`.

src/tiled/commanddialog.cpp
```diff
@@ -45,6 +45,13 @@ CommandDialog::CommandDialog(QWidget *parent)
 
     setWindowTitle(tr("Edit Commands"));
     Utils::restoreGeometry(this);
+
+    mKeySequenceEdit = mUi->keySequenceEdit;
```
bjorn: You do not need `mKeySequenceEdit`. Just use `mUi->keySequenceEdit` everywhere.

src/tiled/commanddialog.cpp
```diff
@@ -45,6 +45,13 @@ CommandDialog::CommandDialog(QWidget *parent)
...
+
+    connect(mKeySequenceEdit, &QKeySequenceEdit::keySequenceChanged, 
+            mUi->treeView, &CommandTreeView::setShortcut);
```
bjorn: Why does the view have a `setShortcut` method? Please move that to the `CommandDialog`.

src/tiled/commanddialog.cpp
```diff
@@ -118,3 +135,8 @@ void CommandTreeView::removeSelectedCommands()
     const QModelIndexList indices = selection->selectedRows();
     mModel->removeRows(indices);
 }
+
+void CommandTreeView::currentChanged(const QModelIndex &current, const QModelIndex&)
```
bjorn: Rather than having this in the view, please handle this in the `CommandDialog`. This way you can also remove `mDialog`. To do this you should connect to the `currentChanged` signal of the `treeView->selectionModel()`.

src/tiled/command.cpp
ketanhwr: In the `QVariant` documentation, it says that it casts `QKeySequence` to `QString` automatically. Is this still required?

src/tiled/commanddialog.cpp
ketanhwr: Thanks!

src/tiled/commanddatamodel.cpp
ketanhwr: This implementation leads to a segmentation fault, because `isValid()` returns `true` when the last row is chosen when the model doesn't actually exist.

ketanhwr: By the last row I mean, `<new command>`

src/tiled/commanddialog.cpp
```diff
@@ -63,6 +69,19 @@ void CommandDialog::closeEvent(QCloseEvent *event)
...
 }
 
+void CommandDialog::setShortcut(const QKeySequence &keySequence)
+{
+    mUi->treeView->model()->setShortcut(mUi->treeView->currentIndex(), keySequence);
+}
+
+void CommandDialog::updateKeySequenceEdit(const QModelIndex &current, const QModelIndex &)
+{
+    if (current.row() < mUi->treeView->model()->rowCount(current))
```
ketanhwr: This check is required otherwise a segmentation fault occurs.

ketanhwr:
@bjorn, I've made the changes :)

src/tiled/commanddialog.cpp
ketanhwr: `isValid()` returns true even when the last row `<new command>` is selected which results in this.

bjorn: Whoops, good point. I had totally forgotten about that dummy row.

ketanhwr:
Is there anything else left to do regarding this task?

src/tiled/commanddatamodel.cpp
```diff
@@ -433,6 +434,19 @@ bool CommandDataModel::dropMimeData(const QMimeData *data, Qt::DropAction, int,
     return false;
 }
 
+QKeySequence CommandDataModel::shortcut(const QModelIndex &index) const
+{
+    if (index.isValid())
+        return mCommands[index.row()].shortcut;
```
bjorn: Since you correctly pointed out this may crash, please keep the range check you had before.

src/tiled/commanddialog.cpp
bjorn: `rowCount(current)` will always return 0 unless current happens to be invalid, in which case it would be commandCount+1, so this check still seems wrong. I think you want:

```c++
if (current.row() < mUi->treeView->model()->rowCount() - 1)
```

ketanhwr: Yes, thanks! Didn't notice this.

ketanhwr: `mUi->keySequenceEdit->clear()` is giving segmentation fault now. Can you help me with this? I'm not able to debug this one.

ketanhwr: I think this line is not required anyway. So I'm removing `mUi->keySequenceEdit->clear()` line entirely. Is this fine?

bjorn: The segmentation fault would be because you're calling it when there is no valid current index, but clearing the shortcut changes it and then you're calling CommandDataModel::setShortcut with an invalid index, which then tries to set the shortcut of the command at -1.

ketanhwr: Oh, damn me :/

ketanhwr:
@bjorn I've tested everything now (I think). It works fine :)

ketanhwr:
Alright. I'll add this.

ketanhwr:
@bjorn please have a look :-)

src/tiled/commanddialog.cpp
```diff
@@ -63,6 +69,24 @@ void CommandDialog::closeEvent(QCloseEvent *event)
     CommandManager::instance()->updateActions();
 }
 
+void CommandDialog::setShortcut(const QKeySequence &keySequence)
+{
+    if (!mUi->keySequenceEdit->isEnabled())
+        mUi->keySequenceEdit->setEnabled(true);
```
ketanhwr: Is it okay to call public slots directly like functions? I mean, syntactically it's correct, but in Qt style, is it considered correct? I don't know why I get this feeling that there is some better way to do it rather than call a public slot directly :confused:

bjorn: A public slot is just a public method that can optionally be used in an old-style connect statement (since it can be dynamically called by its name, which also makes it callable from JavaScript for example). There is no reason not to also use it at a plain function. In fact, many of the Qt APIs are slots even though they are mostly called directly, just because it enables the convenience of connecting things sometimes.

bjorn: Why is this here, though? First of all there is no point in checking whether the sequence edit is disabled before setting it to enabled, but the main thing I'm wondering is why should it get enabled in this function? I think this should happen in `updateKeySequenceEdit`.

src/tiled/commanddialog.cpp
```diff
@@ -81,7 +84,7 @@ void CommandDialog::updateKeySequenceEdit(const QModelIndex &current, const QMod
     if (current.row() < mUi->treeView->model()->rowCount(QModelIndex()) - 1)
         mUi->keySequenceEdit->setKeySequence(mUi->treeView->model()->shortcut(current));
     else
-        mUi->keySequenceEdit->clear();
```
bjorn: It should still be cleared. It makes no sense to show a disabled widget with a shortcut from the previously selected command.

src/tiled/commanddialog.cpp
ketanhwr: Oh, thanks! :smile:

ketanhwr: I don't know what I was thinking when I did that :-/

src/tiled/commanddialog.cpp
```diff
@@ -71,20 +71,21 @@ void CommandDialog::closeEvent(QCloseEvent *event)
...
 
 void CommandDialog::updateKeySequenceEdit(const QModelIndex &current, const QModelIndex &)
 {
-    if (current.row() < mUi->treeView->model()->rowCount(QModelIndex()) - 1)
+    if (current.row() < mUi->treeView->model()->rowCount(QModelIndex()) - 1) {
+        mUi->keySequenceEdit->setEnabled(true);
         mUi->keySequenceEdit->setKeySequence(mUi->treeView->model()->shortcut(current));
-    else
+    }
+    else {
```
bjorn: Coding style: these should go on the same line `} else {`. I think it'll be my last comment. :-)

bjorn:
Ok, two more comment after checking the behavior:

* When I set a shortcut, but I close the dialog before it times out (no longer showing the ", ..."), the shortcut is simply not applied. Unfortunately, judging by the QKeySequenceEdit API, there is no way to fix this because `finishEditing` is in the `QKeySequenceEditPrivate` class. Maybe to get a better use feedback, we should add a Shortcut column to the view so that it's easier to see for the user when the shortcut change is applied.

* There seems to be no (proper) way to clear the shortcut. A workaround that I've noticed is that you could press the Windows key, which does not lead to a recorded key but it does clear any previous shortcut. So a clear button should be added. Strangely enough, the QKeySequenceEdit implementation has a "// TODO: add clear button", and I wonder why this hasn't simply been done already (maybe I'll find out).

ketanhwr:
So what should I do now? Add another column? And regarding clearing the shortcut, I also tried a lot and on Ubuntu 16.04 the control button clears it which is quite weird. Should I add a button clicking on which would clear the shortcut?

bjorn:
> So what should I do now? Add another column? And regarding clearing the shortcut, I also tried a lot and on Ubuntu 16.04 the control button clears it which is quite weird. Should I add a button clicking on which would clear the shortcut?

Right, those were the things I asked about:

* Add a Shortcut column to the CommandDataModel.
* Add a Clear button connected to the `clear` slot of the `QKeySequenceEdit`.

ketanhwr:
Alright.

ketanhwr:
I've added the column and the clear button. One thing that I think is still a problem is that when the shortcut is updated, the user needs to click somewhere in order to see the change in the TreeView. Is there a way this can be fixed? Like when the signal of `editingFinished` is emitted, the column automatically gets updated.

bjorn:
> One thing that I think is still a problem is that when the shortcut is updated, the user needs to click somewhere in order to see the change in the TreeView. Is there a way this can be fixed? Like when the signal of editingFinished is emitted, the column automatically gets updated.

Yes, to make the view update when you change the data, you'll need to emit the `dataChanged` signal with the range of affected model indexes. You should do this in `CommandDataModel::setShortcut`.

ketanhwr:
@bjorn I've made the changes. I also missed a small thing which I've fixed in this commit.

src/tiled/commanddialog.cpp
```diff
@@ -72,8 +72,9 @@ void CommandDialog::closeEvent(QCloseEvent *event)
 void CommandDialog::setShortcut(const QKeySequence &keySequence)
 {
     const QModelIndex &current = mUi->treeView->currentIndex();
-    if (current.row() < mUi->treeView->model()->rowCount(QModelIndex()))
+    if (current.row() < mUi->treeView->model()->rowCount(QModelIndex())) {
         mUi->treeView->model()->setShortcut(current, keySequence);
+    }
```
bjorn: Why adding brackets here? Coding style is actually to leave them out for single-line bodies.

src/tiled/commanddatamodel.cpp
```diff
@@ -451,8 +453,10 @@ QKeySequence CommandDataModel::shortcut(const QModelIndex &index) const
 
 void CommandDataModel::setShortcut(const QModelIndex &index, const QKeySequence &keySequence)
 {
-    if (index.row() < mCommands.size())
+    if (index.row() < mCommands.size()) {
         mCommands[index.row()].shortcut = keySequence;
+        emit dataChanged(index, index);
```
bjorn: Actually you can't necessarily re-use the given index here, since you need to make sure to tell the view that the `ShortcutColumn` changed. If you test it, you'll notice that it only works if you selected the command by clicking on the shortcut column.

So you need to do:
```c++
QModelIndex shortcutIndex = this->index(index.row(), ShortcutColumn);
emit dataChanged(shortcutIndex, shortcutIndex);
```

src/tiled/commanddialog.cpp
ketanhwr: Oh, I added a line then removed it but forgot to remove the brackets :|

src/tiled/commanddatamodel.cpp
ketanhwr: Oh, thanks!

src/tiled/commanddatamodel.cpp
```diff
@@ -455,7 +453,9 @@ void CommandDataModel::setShortcut(const QModelIndex &index, const QKeySequence
 {
     if (index.row() < mCommands.size()) {
         mCommands[index.row()].shortcut = keySequence;
-        emit dataChanged(index, index);
+
+        QModelIndex shortcutIndex = this->index(index.row(), ShortcutColumn);
+        emit dataChanged(index, shortcutIndex);
```
bjorn: Not sure why you're not using shortcutIndex for the first argument here? It should be.

ketanhwr: :man_facepalming:

src/tiled/commanddatamodel.h
```diff
@@ -25,6 +25,7 @@
 #include <QAbstractTableModel>
 #include <QSettings>
 
+class QKeySequence;
```
bjorn: Hmm, I didn't notice this earlier, but you can't actually rely on a forward-declaration for a class that is used as a return value. The reason it compiles anyway is because `command.h` is already including the QKeySequence header. I'd say just get rid of this line.

bjorn:
We made it! :-)

ketanhwr:
Thanks a lot for your help! :-)

Issue #1484, opened 2017-03-13
Use Qbs everywhere and remove the qmake project files


bjorn:
Since it is annoying to keep both the Qbs and the qmake projects up to date with new files, we should move forward with only one system. Qbs is already the only supported build method on Windows, however the following tasks are open before we can remove the qmake projects:

- [x] Build macOS packages using Qbs
- [x] Build Linux binary packages using Qbs
- [x] Update translation files `.ts` with Qbs or possibly using some script (currently, `make ts` is run in the `translations` folder)
- [x] Make sure Linux distributions can use Qbs for building their package
- [x] Build the tests using Qbs (currently only included in qmake project).
- [x] Build the snap using Qbs

This issue is here to document parts still depending on qmake and what the Qbs project files are possibly still lacking, which we'll find out as we try to move from qmake to Qbs in the above areas.

bjorn:
@Ablu If you have some time, you could help with the 4th item on the list by trying to change the Fedora package of Tiled to be built with Qbs. :-)

bjorn:
@svenstaro Maybe you'd like to help make sure Qbs is ready (or find out in what way it is not) by making the Arch Linux package build with Qbs. I'll try it myself eventually in any case, so feel free to wait as well.

jakepetroules:
FYI: The Arch Linux package recently moved from AUR to an official package in Extras.

svenstaro:
Can you provide some packaging-friendly qbs instructions to build tiled?

bjorn:
@svenstaro It seems to essentially work by replacing the `build()` and `package()` functions in your package as follows:
```sh
build() {
  cd $srcdir/$pkgname-$pkgver
  qbs build --no-install
}

package() {
  cd $pkgname-$pkgver
  qbs qbs.installRoot:$pkgdir/usr
}
```
And of course adding `qbs` to the `makedepends`.

However, looking at the file list it is clear that I've got some work to do, since the desktop file, icons, man pages, mime stuff and thumbnailer all still need to be added. Also I'll probably need to add an option to install the header files, and possibly also to disable RPATH (is that needed?).

[...]

PR #1485, opened 2017-03-13
Added Executable files in tests directory to gitignore

ketanhwr:
I ran the command `qmake` and then `make` inside the tests directory and it resulted in two untracked files. As far as I know, these two files should be in `.gitignore`.

I don't know if my assumption is true though. Please have a look!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,8 @@ moc_*.cpp
 ui_*.h
 qrc_*
 default/
+tests/mapreader/mapreader
+tests/staggeredrenderer/staggeredrenderer
 
 # Build artifacts on Windows
 tiled.exe
--- a/.gitignore
+++ b/.gitignore
@@ -11,8 +11,6 @@ moc_*.cpp
 ui_*.h
 qrc_*
 default/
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -11,6 +11,8 @@ moc_*.cpp
 ui_*.h
 qrc_*
 default/
+tests/mapreader/mapreader
+tests/staggeredrenderer/staggeredrenderer
```
bjorn: Note that an outdated entry `tests/test_mapreader` is already in this file below. Please put these lines there instead.

I think it's fine to include this for now, but we can clean up this stuff a lot once the qmake projects are out (#1484), since Qbs always puts build artifacts in a separate directory.

ketanhwr: Yes that's true. But I'm just curious, as to how no one noticed this in a long time.

bjorn: I'm personally not using qmake while developing, except for testing whether it works. And when I do, I rarely do an in-source build (Qt Creator builds outside of the source folder by default). And then still, I don't build the tests. So, there you have it. :-)

PR #1488, opened 2017-03-15
Added an executable picker - Fixes #942

### LINKED ISSUE #942: Add an executable picker for setting up commands
It would be convenient if a file picker would be available for finding the executable to execute for a command.

Requested at issue #940.

ketanhwr:
This fixes #942 

Command Dialog

Firstly, I've modified the layout of the Command Dialog. Also, I've made the option "Save Map before Execute" unique to each command. The user can now either enter the command, or he can browse an executable.

### DIFF
--- a/src/tiled/commanddatamodel.h
+++ b/src/tiled/commanddatamodel.h
@@ -36,7 +36,7 @@ class CommandDataModel : public QAbstractTableModel
 
 public:
 
-    enum { NameColumn, CommandColumn, ShortcutColumn, EnabledColumn };
+    enum { NameColumn, ShortcutColumn, EnabledColumn };
 
     /**
       * Constructs the object and parses the users settings to allow easy
--- a/src/tiled/commanddatamodel.h
+++ b/src/tiled/commanddatamodel.h
@@ -133,7 +133,7 @@ public:
 
     bool saveBeforeExecute(const QModelIndex &index) const;
 
-    void setSaveBeforeExecute(const QModelIndex &index, const bool &value);
+    void setSaveBeforeExecute(const QModelIndex &index, bool value);
 
     QString command(const QModelIndex &index) const;
 
--- a/src/tiled/commanddatamodel.h
+++ b/src/tiled/commanddatamodel.h
@@ -49,20 +49,6 @@ public:
       */
     void commit();
 
-    /**
-      * Returns whether saving before executing commands is enabled.
-      */
-    bool saveBeforeExecute() const { return mSaveBeforeExecute; }
-
-    /**
-      * Enables or disables saving before executing commands.
-      */
-    void setSaveBeforeExecute(bool enabled) { mSaveBeforeExecute = enabled; }
-
-    /**
-      * Returns the first enabled command in the list, or an empty
-      * disabled command if there are no enabled commands.
-      */
     Command firstEnabledCommand() const;
 
     /**
@@ -143,7 +129,15 @@ public:
 
     QKeySequence shortcut(const QModelIndex &index) const;
 
-    void setShortcut(const QModelIndex &index, const QKeySequence &keySequence);
+    void setShortcut(const QModelIndex &index, const QKeySequence &value);
+
+    bool saveBeforeExecute(const QModelIndex &index) const;
+
+    void setSaveBeforeExecute(const QModelIndex &index, const bool &value);
+
+    QString command(const QModelIndex &index) const;
+
+    void setCommand(const QModelIndex &index, const QString &value);
 
 public slots:
 
@@ -182,7 +176,6 @@ private:
 
     QSettings mSettings;
     QList<Command> mCommands;
-    bool mSaveBeforeExecute;
 };
 
 } // namespace Internal
--- a/src/tiled/commanddatamodel.h
+++ b/src/tiled/commanddatamodel.h
@@ -70,12 +70,12 @@ public:
     /**
      * Returns the number of rows (this includes the <New Command> row).
      */
-    int rowCount(const QModelIndex &) const override;
+    int rowCount(const QModelIndex &parent = QModelIndex()) const override;
 
     /**
      * Returns the number of columns.
      */
-    int columnCount(const QModelIndex &) const override;
+    int columnCount(const QModelIndex &parent = QModelIndex()) const override;
 
     /**
      * Returns the data at <i>index</i> for the given <i>role</i>.
--- a/src/tiled/commanddialog.cpp
+++ b/src/tiled/commanddialog.cpp
@@ -30,6 +30,7 @@
 #include <QContextMenuEvent>
 #include <QModelIndex>
 #include <QFileDialog>
+#include <QStandardPaths>
 
 using namespace Tiled;
 using namespace Tiled::Internal;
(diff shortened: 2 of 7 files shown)


bjorn:
Please don't center the Close button! :-)

ketanhwr:
@bjorn status?

bjorn:
> @bjorn status?

@ketanhwr No need for such messages unless something is inactive for weeks and it may seem I have forgotten. I only have two full days to work on Tiled per week, and any activity beyond that is limited and random. Of course I will commit to spending a bit of time every day during GSoC, but I can't do this all the time. Rest assured that each activity here will usually sit in my inbox already until I get around to look at it.

src/tiled/commanddatamodel.h
```diff
@@ -143,7 +129,15 @@ class CommandDataModel : public QAbstractTableModel
...
     QKeySequence shortcut(const QModelIndex &index) const;
 
-    void setShortcut(const QModelIndex &index, const QKeySequence &keySequence);
+    void setShortcut(const QModelIndex &index, const QKeySequence &value);
+
+    bool saveBeforeExecute(const QModelIndex &index) const;
+
+    void setSaveBeforeExecute(const QModelIndex &index, const bool &value);
+
+    QString command(const QModelIndex &index) const;
```
bjorn: What about making this function return the full Command, like `firstEnabledCommand` is doing, avoiding the need for all the other getters?

src/tiled/commanddialog.cpp
```diff
@@ -76,17 +83,56 @@ void CommandDialog::setShortcut(const QKeySequence &keySequence)
         mUi->treeView->model()->setShortcut(current, keySequence);
 }
 
-void CommandDialog::updateKeySequenceEdit(const QModelIndex &current, const QModelIndex &)
+void CommandDialog::setSaveBeforeExecute(int state)
+{
+    const QModelIndex &current = mUi->treeView->currentIndex();
+    if (current.row() < mUi->treeView->model()->rowCount(QModelIndex()))
```
bjorn: Please add `QModelIndex()` as the default parameter of the `parent` for `rowCount`. It is the default in the base class already, and then we can avoid being needlessly verbose here.

src/tiled/commanddialog.cpp
```diff
@@ -76,17 +83,56 @@ void CommandDialog::setShortcut(const QKeySequence &keySequence)
...
+{
+    QString caption = tr("Select executable");
+    QString dir = tr("/bin");
+    QString executableName = QFileDialog::getOpenFileName(this, caption, dir);
+
+    if (!executableName.isEmpty())
+        mUi->commandEdit->setText(executableName);
+}
+
+void CommandDialog::enableWidgets(const bool enable)
```
bjorn: Please merge this function into `updateWidgets`. I see no gain in having this as a separate method.

src/tiled/commanddialog.cpp
```diff
@@ -76,17 +83,56 @@ void CommandDialog::setShortcut(const QKeySequence &keySequence)
...
+        mUi->saveBox->setChecked(mUi->treeView->model()->saveBeforeExecute(current));
+        mUi->commandEdit->setText(mUi->treeView->model()->command(current));
+    }
+    else
+        enableWidgets(false);
+}
+
+void CommandDialog::openFileDialog()
+{
+    QString caption = tr("Select executable");
```
bjorn: Window captions should capitalize each word, so "Select Executable".

src/tiled/commanddialog.cpp
```diff
@@ -76,17 +83,56 @@ void CommandDialog::setShortcut(const QKeySequence &keySequence)
...
+    QString dir = tr("/bin");
```
bjorn: That is only useful on Linux, and even there it should probably be `/usr/bin`. I think we should try relying on `QStandardPaths::ApplicationsLocation`.

src/tiled/commanddialog.cpp
```diff
@@ -76,17 +83,56 @@ void CommandDialog::setShortcut(const QKeySequence &keySequence)
...
+    QString executableName = QFileDialog::getOpenFileName(this, caption, dir);
+
+    if (!executableName.isEmpty())
+        mUi->commandEdit->setText(executableName);
```
bjorn: Maybe it would be useful to automatically add `%mapfile`? It will be what the user wants most of the time.

src/tiled/commanddialog.cpp
```diff
@@ -101,7 +147,6 @@ CommandTreeView::CommandTreeView(QWidget *parent)
     QHeaderView *h = header();
     h->setStretchLastSection(false);
     h->setSectionResizeMode(CommandDataModel::NameColumn, QHeaderView::Interactive);
-    h->setSectionResizeMode(CommandDataModel::CommandColumn, QHeaderView::Stretch);
```
bjorn: Please change the `NameColumn` from `Interactive` to `Stretch`.

src/tiled/commanddatamodel.h
ketanhwr: Overlooked this, thanks!

ketanhwr:
I thought you might have overlooked this one, so I tagged you, sorry! I'll keep this in mind next time :smiley:

src/tiled/commanddialog.cpp
bjorn: You added the default parameter but didn't remove the now no longer necessary argument here. :-)

src/tiled/commanddialog.cpp
```diff
@@ -76,19 +84,55 @@ void CommandDialog::setShortcut(const QKeySequence &keySequence)
...
+    }
+    else {
+        mUi->commandEdit->clear();
         mUi->keySequenceEdit->clear();
+        mUi->saveBox->setChecked(false);
+        mUi->commandEdit->setEnabled(false);
+        mUi->browseButton->setEnabled(false);
         mUi->keySequenceEdit->setEnabled(false);
         mUi->clearButton->setEnabled(false);
+        mUi->saveBox->setEnabled(false);
```
bjorn: Please use a local variable to de-duplicate all these `setEnabled` calls.

src/tiled/commanddialog.cpp
```diff
@@ -76,19 +84,55 @@ void CommandDialog::setShortcut(const QKeySequence &keySequence)
...
 }
 
+void CommandDialog::openFileDialog()
+{
+    QString caption = tr("Select Executable");
+    QString dir = QStandardPaths::writableLocation(QStandardPaths::ApplicationsLocation);
+    QString executableName = QFileDialog::getOpenFileName(this, caption, dir);
+
+    if (!executableName.isEmpty())
+        mUi->commandEdit->setText(executableName);
```
bjorn: What about the idea to automatically append `" %mapfile"`?

src/tiled/commanddialog.cpp
ketanhwr: It's behaving quite strangely now. If I remove these parameters, a compilation error shows up :/
I'm trying to figure it out though

src/tiled/commanddialog.cpp
ketanhwr: I think we should append `%mapfile` only if the user creates a new command.

ketanhwr:
Regarding the alignment, it's working fine for me, I don't know why it's working that way for you. Can you attach a screenshot, perhaps? And also, how about adding a keyboard shortcut for `Edit Commands...` as well? I think it might be a bit useful for people who'll customise their commands when we add more options to the Command Dialog.

src/tiled/commanddialog.cpp
bjorn: I think if we don't always add it here, we need to introduce a separate "arguments" field, like in Qt Creator. That way the arguments can default to `%mapfile` and choosing the executable becomes independent.

bjorn:
> Regarding the alignment, it's working fine for me, I don't know why it's working that way for you. Can you attach a screenshot, perhaps?

selection_154

> And also, how about adding a keyboard shortcut for Edit Commands... as well? I think it might be a bit useful for people who'll customise their commands when we add more options to the Command Dialog.

I think the action is too rare to have a default shortcut for it.

ketanhwr:
Alright, seems weird, I'll look into this.

src/tiled/commanddialog.cpp
ketanhwr: Oh found the error. I'm stupid.

ketanhwr:
I've fixed the UI problem. Still not sure what to do about the `%mapfile` appending.

src/tiled/commanddialog.cpp
```diff
@@ -72,23 +80,63 @@ void CommandDialog::closeEvent(QCloseEvent *event)
...
+        mUi->commandEdit->clear();
         mUi->keySequenceEdit->clear();
-        mUi->keySequenceEdit->setEnabled(false);
-        mUi->clearButton->setEnabled(false);
+        mUi->saveBox->setChecked(enable);
+        mUi->commandEdit->setEnabled(enable);
+        mUi->browseButton->setEnabled(enable);
+        mUi->keySequenceEdit->setEnabled(enable);
+        mUi->clearButton->setEnabled(enable);
+        mUi->saveBox->setEnabled(enable);
```
bjorn: You know you still need to de-duplicate these lines, right? I mean now you're using a variable, but...

src/tiled/commanddialog.cpp
```diff
@@ -72,23 +80,63 @@ void CommandDialog::closeEvent(QCloseEvent *event)
...
+
         mUi->keySequenceEdit->setKeySequence(mUi->treeView->model()->shortcut(current));
-    } else {
+        mUi->saveBox->setChecked(mUi->treeView->model()->saveBeforeExecute(current));
+        mUi->commandEdit->setText(mUi->treeView->model()->command(current));
+    }
+    else {
```
bjorn: Coding style: `} else {`. Why did you even change it? It was correct before.

src/tiled/commanddialog.cpp
```diff
@@ -72,23 +80,63 @@ void CommandDialog::closeEvent(QCloseEvent *event)
...
+    if (current.row() < mUi->treeView->model()->rowCount())
+        mUi->treeView->model()->setCommand(current, text);
+}
+
+void CommandDialog::updateWidgets(const QModelIndex &current, const QModelIndex &)
 {
-    if (current.row() < mUi->treeView->model()->rowCount(QModelIndex()) - 1) {
-        mUi->keySequenceEdit->setEnabled(true);
-        mUi->clearButton->setEnabled(true);
+    bool enable = true;
```
bjorn: What about:
```c++
bool enabled = current.row() < mUi->treeView->model()->rowCount() - 1;
```

src/tiled/commanddialog.cpp
ketanhwr: Kill me already ;_;

ketanhwr:
:man_facepalming:

bjorn:
Great, thanks!

About the `%mapfile`, I think ideally we need to separate the executable from the arguments. If you look at the Qt Creator screenshot I showed you in some issue, you see they have done the same. It would allow the current file to be passed to the command by default, independently of the executable. It also makes the "Select Executable" action less destructive, since it won't affect the argument list.

bjorn:
I've pushed some follow-up tweaks in change 83b26c48238709d03d1acb54b2dc7a6749af7dd8.

ketanhwr:
Oh, I forgot about removing all the getters and setters. Thanks for your help :-)

ketanhwr:
Yes, I was pretty impressed by the Qt Creator's custom command menu. I'm planning to implement something pretty similar in Tiled as well, would be a nice thing to add to my proposal, I think!

bjorn:
> Yes, I was pretty impressed by the Qt Creator's custom command menu. I'm planning to implement something pretty similar in Tiled as well, would be a nice thing to add to my proposal, I think!

I could definitely imagine that being part of your proposal. :-)

Issue #1491, opened 2017-03-15
Be able to select objects behind other objects


bjorn:
Currently, the only way to select objects covered entirely by other objects is to either move those other objects out of the way temporarily, or to hide them or their object layer.

It should be possible to select objects below other layers without resorting to such actions.

One suggested solution is to cycle between all objects at a certain location when the user hold the Alt modifier (without moving the mouse, because that would drag the objects). It may also be helpful to be able to see a list of all objects at a certain location in a menu.

Based on [feedback](http://discourse.mapeditor.org/t/feedback-after-a-couple-months-with-tiled/2090) from @eevee and others.

issy123:
A way to do is is by searching an object in the objects view, or show the x and y property (and sort on it), so you know the overlapping objects

faulknermano:
Right now, the selection is based on the bounding box. But I think it would be more intuitive, if by picking the foremost pixel of an object (ignoring those pixels with full transparency or alpha=0) would yield that object selected.

Photoshop has that capability: by holding down the CTRL modifier, with the Move tool selected, it picks the object whose pixel is under the cursor. In addition, Photoshop has a checkbox which will activate this 'select under cursor' function without having to use the CTRL modifier.

The interesting bit is that the Move is separate from Selection, so that you don't nudge objects inadvertently. I often mistakenly nudge an object in Tiled while selecting; it is mitigated by use of grid snapping, though snapping is sometimes not wanted.

Though this issue is about selection, I think manipulation is linked because there are no current ways to select only or transform only. My suggestion would be to adopt a similar functionality to Photoshop because I find it the clearest way of manipulating objects in a 2d space.

bjorn:
> Though this issue is about selection, I think manipulation is linked because there are no current ways to select only or transform only.

Well, there is:

* Shift modifier to force the selection rectangle. No drag operation will be started then, even if you click on an object.
* Alt modifier to force the drag operation. The selection will not change when starting a drag.

Issue #1477 covers ignoring transparent parts of objects.

I'm not sure how much sense it makes to separate the selection and movement tools at this point. Tiled is a bit different from an image editor, and its interaction model in terms of object handling is more based on Inkscape.

faulknermano:
@bjorn, yes I agree; that's cool. Apparently, I did not know about the Shift modifier and I think those two modifiers alone, satisfies my comments. Thanks for this info.i

I'm also glad to know that the transparent pixel issue is at the table. Thanks for your work.

[...]

PR #1494, opened 2017-03-16
Selection of underlying objects

Acuion:
My approach on #1491 
User can iterate through objects by pressing Alt while hovering over them with cursor, also there is a new submenu in the context menu, which allows to select certain object in the stack.
I'm not sure about using new field 'mCurrMouseScenePosition' to get in-scene mouse coordinates in keyPressed method, is there any better way to do it?
And due to reselection of the topmost object while dragging, an underlying object can't be moved this way, only by arrows. Does it have to be fixed? If so, I need a clue how to do it right.

### DIFF
--- a/src/tiled/abstractobjecttool.cpp
+++ b/src/tiled/abstractobjecttool.cpp
@@ -116,8 +116,7 @@ void AbstractObjectTool::mouseMoved(const QPointF &pos,
 void AbstractObjectTool::mousePressed(QGraphicsSceneMouseEvent *event)
 {
     if (event->button() == Qt::RightButton) {
-        showContextMenu(event->scenePos(),
-                        event->screenPos());
+        showContextMenu(event->scenePos(), event->screenPos());
     }
 }
 
@@ -224,8 +223,7 @@ void AbstractObjectTool::lowerToBottom()
  * Shows the context menu for map objects. The menu allows you to duplicate and
  * remove the map objects, or to edit their properties.
  */
-void AbstractObjectTool::showContextMenu(QPointF scenePos,
-                                         QPoint screenPos)
+void AbstractObjectTool::showContextMenu(QPointF scenePos, QPoint screenPos)
 {
     MapObjectItem *clickedObjectItem = topMostObjectItemAt(scenePos);
 
--- a/src/tiled/abstractobjecttool.cpp
+++ b/src/tiled/abstractobjecttool.cpp
@@ -96,6 +96,11 @@ void AbstractObjectTool::keyPressed(QKeyEvent *event)
 
         mSelectedRotationIndex = (mSelectedRotationIndex + 1) % objectList.size();
         return;
+	}
+    case Qt::Key_D:
+        if (event->modifiers() & Qt::ControlModifier) {
+            duplicateObjects();
+            return;
         }
     }
 
--- a/src/tiled/abstractobjecttool.cpp
+++ b/src/tiled/abstractobjecttool.cpp
@@ -133,7 +133,7 @@ ObjectGroup *AbstractObjectTool::currentObjectGroup() const
     return dynamic_cast<ObjectGroup*>(mapDocument()->currentLayer());
 }
 
-QList<MapObjectItem*> AbstractObjectTool::listOfObjectItemsAt(QPointF pos) const
+QList<MapObjectItem*> AbstractObjectTool::objectItemsAt(QPointF pos) const
 {
     const QList<QGraphicsItem *> &items = mMapScene->items(pos);
 
@@ -290,24 +290,6 @@ void AbstractObjectTool::showContextMenu(QPointF scenePos, QPoint screenPos)
     Utils::setThemeIcon(removeAction, "edit-delete");
     Utils::setThemeIcon(propertiesAction, "document-properties");
 
-    QList<MapObjectItem*> underlyingObjects = listOfObjectItemsAt(scenePos);
-    if (underlyingObjects.size() > 1) {
-        menu.addSeparator();
-
-        QMenu *selectUnderlyingMenu = menu.addMenu(tr("Select an underlying object"));
-
-        for (int levelNum = 0; levelNum < underlyingObjects.size(); ++levelNum) {
-            const QString& objectName = underlyingObjects[levelNum]->mapObject()->name();
-            QString actionName = (objectName.isEmpty() ? tr("Object at level %n", "", levelNum + 1) : objectName)
-                    + tr(levelNum ? "" : " (topmost)");
-            QAction *action = selectUnderlyingMenu->addAction(actionName);
-            if (levelNum == 0)
-                action->setEnabled(false);//just to set a starting point
-            else
-                action->setData(QVariant::fromValue(levelNum));
-        }
-    }
-
     QAction *action = menu.exec(screenPos);
     if (!action)
         return;
@@ -322,12 +304,5 @@ void AbstractObjectTool::showContextMenu(QPointF scenePos, QPoint screenPos)
     if (ObjectGroup *objectGroup = action->data().value<ObjectGroup*>()) {
         mapDocument()->moveObjectsToGroup(mapDocument()->selectedObjects(),
                                           objectGroup);
-    }
-
-    if (int indexOfObjectToBeSelected = action->data().value<int>()) {
-        auto selection = mapScene()->selectedObjectItems();
-        selection.clear();
-        selection.insert(underlyingObjects[indexOfObjectToBeSelected]);
-        mapScene()->setSelectedObjectItems(selection);
     }
 }
--- a/src/tiled/abstractobjecttool.cpp
+++ b/src/tiled/abstractobjecttool.cpp
@@ -82,10 +82,8 @@ void AbstractObjectTool::keyPressed(QKeyEvent *event)
     case Qt::Key_Home:      raiseToTop(); return;
     case Qt::Key_End:       lowerToBottom(); return;
     case Qt::Key_D:
-        if (event->modifiers() & Qt::ControlModifier) {
+        if (event->modifiers() & Qt::ControlModifier)
             duplicateObjects();
-            return;
-        }
         return;
     }
 
--- a/src/tiled/abstractobjecttool.cpp
+++ b/src/tiled/abstractobjecttool.cpp
@@ -67,7 +67,6 @@ AbstractObjectTool::AbstractObjectTool(const QString &name,
 void AbstractObjectTool::activate(MapScene *scene)
 {
     mMapScene = scene;
-    mSelectedRotationIndex = 0;
 }
 
 void AbstractObjectTool::deactivate(MapScene *)
@@ -82,26 +81,12 @@ void AbstractObjectTool::keyPressed(QKeyEvent *event)
     case Qt::Key_PageDown:  lower(); return;
     case Qt::Key_Home:      raiseToTop(); return;
     case Qt::Key_End:       lowerToBottom(); return;
-    case Qt::Key_Alt: {
-        auto objectList = listOfObjectItemsAt(mCurrMouseScenePosition);
-
-        if (!objectList.size())
-            return;
-        if (mSelectedRotationIndex >= objectList.size())
-            mSelectedRotationIndex = 0;//definitely another stack
-        QSet<MapObjectItem*> selection = mapScene()->selectedObjectItems();
-        selection.clear();
-        selection.insert(objectList.at(mSelectedRotationIndex));
-        mapScene()->setSelectedObjectItems(selection);
-
-        mSelectedRotationIndex = (mSelectedRotationIndex + 1) % objectList.size();
-        return;
-	}
     case Qt::Key_D:
         if (event->modifiers() & Qt::ControlModifier) {
             duplicateObjects();
             return;
         }
+        return;
     }
 
     event->ignore();
--- a/src/tiled/abstractobjecttool.cpp
+++ b/src/tiled/abstractobjecttool.cpp
@@ -67,6 +67,7 @@ AbstractObjectTool::AbstractObjectTool(const QString &name,
 void AbstractObjectTool::activate(MapScene *scene)
 {
     mMapScene = scene;
+    mSelectedRotationIndex = 0;
 }
 
 void AbstractObjectTool::deactivate(MapScene *)
@@ -81,6 +82,19 @@ void AbstractObjectTool::keyPressed(QKeyEvent *event)
     case Qt::Key_PageDown:  lower(); return;
     case Qt::Key_Home:      raiseToTop(); return;
     case Qt::Key_End:       lowerToBottom(); return;
+    case Qt::Key_Alt: {
+        auto objectList = listOfObjectItemsAt(mCurrMouseScenePosition);
+
+        if (!objectList.size())
+            return;
+        QSet<MapObjectItem*> selection = mapScene()->selectedObjectItems();
+        selection.clear();
+        selection.insert(objectList.at(mSelectedRotationIndex));
+        mapScene()->setSelectedObjectItems(selection);
+
+        mSelectedRotationIndex = (mSelectedRotationIndex + 1) % objectList.size();
+        return;
+        }
     }
 
     event->ignore();
@@ -105,12 +119,15 @@ void AbstractObjectTool::mouseMoved(const QPointF &pos,
     const int x = (int) std::floor(tilePosF.x());
     const int y = (int) std::floor(tilePosF.y());
     setStatusInfo(QString(QLatin1String("%1, %2 (%3, %4)")).arg(x).arg(y).arg(pixelPos.x()).arg(pixelPos.y()));
+
+    mCurrMouseScenePosition = pos;
+    mSelectedRotationIndex = 0;
 }
 
 void AbstractObjectTool::mousePressed(QGraphicsSceneMouseEvent *event)
 {
     if (event->button() == Qt::RightButton) {
-        showContextMenu(topMostObjectItemAt(event->scenePos()),
+        showContextMenu(event->scenePos(),
                         event->screenPos());
     }
 }
@@ -128,9 +145,22 @@ ObjectGroup *AbstractObjectTool::currentObjectGroup() const
     return dynamic_cast<ObjectGroup*>(mapDocument()->currentLayer());
 }
 
+QList<MapObjectItem*> AbstractObjectTool::listOfObjectItemsAt(QPointF pos) const
+{
+    const QList<QGraphicsItem *> &items = mMapScene->items(pos);
+
+    QList<MapObjectItem*> objectList;
+    for (auto item : items) {
+        if (MapObjectItem *objectItem = qgraphicsitem_cast<MapObjectItem*>(item))
+            objectList.append(objectItem);
+    }
+    return objectList;
+}
+
 MapObjectItem *AbstractObjectTool::topMostObjectItemAt(QPointF pos) const
 {
     const QList<QGraphicsItem *> &items = mMapScene->items(pos);
+
     for (QGraphicsItem *item : items) {
         if (MapObjectItem *objectItem = qgraphicsitem_cast<MapObjectItem*>(item))
             return objectItem;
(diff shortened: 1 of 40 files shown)


bjorn:
Hey @Acuion, thanks for taking on this issue!

> I'm not sure about using new field 'mCurrMouseScenePosition' to get in-scene mouse coordinates in keyPressed method, is there any better way to do it?

Why did you decide to cycle on Alt presses, rather than cycling on mouse clicks while Alt is held? That's how Inkscape does it and it would avoid this issue.

> And due to reselection of the topmost object while dragging, an underlying object can't be moved this way, only by arrows. Does it have to be fixed? If so, I need a clue how to do it right.

You can drag an underlying selected objects by holding Alt, which will force the drag even if not hitting any object and prevents any changes to the selection. So no change should be needed there, apart from keeping that behavior working.

Acuion:
Oh, yes, it was a quite odd decision, now should work as expected

`if (modifiers & Qt::AltModifier) {...}` is located outside the `if (mClickedObjectItem) {...}` block because when I click fast enough, something happens and mClickedObjectItem becomes nullptr.

src/tiled/abstractobjecttool.cpp
```diff
@@ -133,9 +131,22 @@ ObjectGroup *AbstractObjectTool::currentObjectGroup() const
     return dynamic_cast<ObjectGroup*>(mapDocument()->currentLayer());
 }
 
+QList<MapObjectItem*> AbstractObjectTool::listOfObjectItemsAt(QPointF pos) const
```
bjorn: Can just be called `objectItemsAt`.

src/tiled/abstractobjecttool.cpp
```diff
@@ -276,6 +288,24 @@ void AbstractObjectTool::showContextMenu(MapObjectItem *clickedObjectItem,
...
+
+        for (int levelNum = 0; levelNum < underlyingObjects.size(); ++levelNum) {
+            const QString& objectName = underlyingObjects[levelNum]->mapObject()->name();
+            QString actionName = (objectName.isEmpty() ? tr("Object at level %n", "", levelNum + 1) : objectName)
+                    + tr(levelNum ? "" : " (topmost)");
+            QAction *action = selectUnderlyingMenu->addAction(actionName);
+            if (levelNum == 0)
+                action->setEnabled(false);//just to set a starting point
+            else
+                action->setData(QVariant::fromValue(levelNum));
```
bjorn: If you put `Q_DECLARE_METATYPE(Tiled::Internal::MapObjectItem*)` at the bottom of `mapobjectitem.h`, then you can do `action->setData(QVariant::fromValue(underlyingObjects[levelNum])` here and `action->data().value<MapObjectItem*>()` below.

src/tiled/objectselectiontool.cpp
```diff
@@ -551,8 +552,22 @@ void ObjectSelectionTool::mouseReleased(QGraphicsSceneMouseEvent *event)
...
         }
         const Qt::KeyboardModifiers modifiers = event->modifiers();
+        QSet<MapObjectItem*> selection = mapScene()->selectedObjectItems();
+        if (modifiers & Qt::AltModifier) {
+            auto underlyingObjects = listOfObjectItemsAt(event->scenePos());
+
+            if (!underlyingObjects.size())
+                break;
+            if (mSelectedRotationIndex >= underlyingObjects.size())
+                mSelectedRotationIndex = 0;//definitely another stack
```
bjorn: I think instead of keeping track of the current "selection rotation index", the next selection should simply be based on the current selection. So basically, find the last selected object among the objects at the mouse position, and then take the next one (or cycle back to the first one if there is no next one).

It will be a bit more processing but I think it will make the system more predictable. At the moment there are some problems with the click sometimes not responding the first time you try it, or a deeper object getting selected than you might expect after having already used the feature somewhere else.

src/tiled/objectselectiontool.cpp
```diff
@@ -551,8 +552,22 @@ void ObjectSelectionTool::mouseReleased(QGraphicsSceneMouseEvent *event)
...
+            selection.clear();
+            selection.insert(underlyingObjects.at(mSelectedRotationIndex));
```
bjorn: Note that Tiled adds to the selection when Shift or Control modifiers are held. The same should apply here, so you can use Shift or Control to add below objects to the selection.

I have a hunch, btw, that you can re-use the code below by just changing the `mClickedObjectItem` to the next underlying object.

src/tiled/abstractobjecttool.cpp
```diff
@@ -276,6 +288,24 @@ void AbstractObjectTool::showContextMenu(MapObjectItem *clickedObjectItem,
     Utils::setThemeIcon(removeAction, "edit-delete");
     Utils::setThemeIcon(propertiesAction, "document-properties");
 
+    QList<MapObjectItem*> underlyingObjects = listOfObjectItemsAt(scenePos);
+    if (underlyingObjects.size() > 1) {
+        menu.addSeparator();
+
+        QMenu *selectUnderlyingMenu = menu.addMenu(tr("Select an underlying object"));
```
bjorn: What do you think about rather than adding the items here, showing an entirely different menu with just the list of objects, when right-clicking while holding Alt?

Acuion: To separate object-related actions and a selection is a good idea. But where is it better to place it, in the abstract or in the selection tool? On the one hand, in the abstract it can be accessed from all tools, on the other, alt+click is working only with the selection one, so it will be more logical.

src/tiled/abstractobjecttool.cpp
Acuion: Yay, that's what I missed

src/tiled/abstractobjecttool.cpp
```diff
@@ -115,8 +116,7 @@ void AbstractObjectTool::mouseMoved(const QPointF &pos,
 void AbstractObjectTool::mousePressed(QGraphicsSceneMouseEvent *event)
 {
     if (event->button() == Qt::RightButton) {
-        showContextMenu(topMostObjectItemAt(event->scenePos()),
-                        event->screenPos());
+        showContextMenu(event->scenePos(), event->screenPos());
```
Acuion: Should I return it back? Btw, I think it is more flexible

Acuion:
@bjorn I don't know do I have to comment to indicate it, but a new version is ready
- a new context menu in the selection tool
- shift&ctrl support in both alt+click and the menu

src/tiled/objectselectiontool.cpp
```diff
@@ -533,6 +534,40 @@ void ObjectSelectionTool::mousePressed(QGraphicsSceneMouseEvent *event)
 
         break;
     }
+    case Qt::RightButton:
+        if (event->modifiers() & Qt::AltModifier) {
+            QList<MapObjectItem*> underlyingObjects = objectItemsAt(event->scenePos());
+            if (underlyingObjects.size() > 1) {
```
Acuion: Perhaps the menu should always be shown, if the first line is now active

src/tiled/abstractobjecttool.cpp
bjorn: My first thought is that indeed it would be more logical to only supports this alternative menu in the ObjectSelectionTool.

src/tiled/abstractobjecttool.cpp
bjorn: Return what back? The menu? That's not possible because the menu is destroyed before the function returns. If we need it, we should have a `createContextMenu`, but I think we don't need it.

src/tiled/objectselectiontool.cpp
bjorn: Yes, I think the check should be `if (!underlyingObjects.isEmpty()) {`.

src/tiled/objectselectiontool.cpp
```diff
@@ -533,6 +534,40 @@ void ObjectSelectionTool::mousePressed(QGraphicsSceneMouseEvent *event)
...
+                    const QString& objectName = underlyingObjects[levelNum]->mapObject()->name();
+                    QString actionName = (objectName.isEmpty() ? tr("Object at level %n", "", levelNum + 1) : objectName)
+                            + tr(levelNum ? "" : " (topmost)");
+                    QAction *action = selectUnderlyingMenu.addAction(actionName);
+                    action->setData(QVariant::fromValue(underlyingObjects[levelNum]));
+                }
+
+                QAction *action = selectUnderlyingMenu.exec(event->screenPos());
+
+                if (MapObjectItem* objectToBeSelected = action->data().value<MapObjectItem*>()) {
```
bjorn: This crashes if `action` is null (no action chosen / clicked outside of menu).

src/tiled/abstractobjecttool.cpp
Acuion: The line with showContextMenu. Because the change from `topMostObjectItemAt` to `event->scenePos() `is no longer required.

src/tiled/objectselectiontool.cpp
```diff
@@ -551,8 +586,25 @@ void ObjectSelectionTool::mouseReleased(QGraphicsSceneMouseEvent *event)
...
         }
         const Qt::KeyboardModifiers modifiers = event->modifiers();
+        QSet<MapObjectItem*> selection = mapScene()->selectedObjectItems();
+        if (modifiers & Qt::AltModifier) {
+            auto underlyingObjects = objectItemsAt(event->scenePos());
+
+            if (underlyingObjects.isEmpty())
+                break;
+            auto currRotation = std::find(underlyingObjects.begin(),
+                                          underlyingObjects.end(), mSelectedInUnderlyingRotation);
```
bjorn: Please prefer to use Qt methods when available. This should just be `underlyingObjects.indexOf(mSelectedInUnderlyingRotation)`.

However, I suggested to not use any other state than the current selection, to make this action completely predictable. There should not be a `mSelectedInUnderlyingRotation`, but rather you could search for the bottom-most selected object and then select the next one (cyling back to top-most object if it doesn't exist).

It seems reverse iterators have only been added to QList in Qt 5.6, and Tiled currently needs to compile with Qt 5.4, so I think there will be not such a short way of finding the bottom-most selected object. It would probably just be easiest to do something like:
```c++
MapObjectItem *nextItem = underlyingObjects.first();
for (int i = underlyingObjects.size() - 1; i > 0; --i) {
    MapObjectItem *underlyingObject = underlyingObjects.at(i);
    if (selection.contains(underlyingObject)
        break;
    nextItem = underlyingObject;
}
```

src/tiled/objectselectiontool.cpp
```diff
@@ -551,8 +586,25 @@ void ObjectSelectionTool::mouseReleased(QGraphicsSceneMouseEvent *event)
...
+            if (currRotation == underlyingObjects.end())
+                currRotation = underlyingObjects.begin();//new stack of objects
+            else {
+                if ((modifiers & (Qt::ShiftModifier | Qt::ControlModifier)) == 0)
+                    selection.remove(mSelectedInUnderlyingRotation);//deselect previous
```
bjorn: Instead of doing this, please suppress the "Clicking one of the selected items changes the edit mode" below when Alt is held. Unless I'm missing something, that should do the same in less code.

src/tiled/objectselectiontool.h
```diff
@@ -25,6 +25,7 @@
 #include <QList>
 #include <QSet>
 #include <QVector>
+#include <QMenu>
```
bjorn: Please move that include to the `objectselectiontool.cpp` file.

src/tiled/abstractobjecttool.cpp
bjorn: If you meant reverting this signature change, please do because it doesn't matter in the end so keeping the patch smaller is better.

src/tiled/objectselectiontool.cpp
Acuion: I tried this approach and realized that it is not compatible with a shift selection. If you start Shift+Alt+Click'ing in the middle of the stack, once bottom-most element is included to the selection, `nextItem` will always be the first element in the stack instead of continue iterating.
Also it is not compatible with a shift **un**selection - in this case we need to access the next selected object.

Acuion:
Fixed, but `mSelectedInUnderlyingRotation` is still here

Acuion:
So, do you think that multiple in-stack selection can be removed from the alt+click logic (a user will still can append an item from the stack to an existing selection)? A new option "select all" in the context menu can be added instead. Other selection combination can be done with a shift+context menu selecting.

src/tiled/objectselectiontool.cpp
```diff
@@ -533,6 +535,43 @@ void ObjectSelectionTool::mousePressed(QGraphicsSceneMouseEvent *event)
...
+        if (event->modifiers() & Qt::AltModifier) {
+            QList<MapObjectItem*> underlyingObjects = objectItemsAt(event->scenePos());
+            if (underlyingObjects.empty())
+                break;
+            QMenu selectUnderlyingMenu;
+
+            for (int levelNum = 0; levelNum < underlyingObjects.size(); ++levelNum) {
+                const QString& objectName = underlyingObjects[levelNum]->mapObject()->name();
+                QString actionName = (objectName.isEmpty() ? tr("Object at level %n", "", levelNum + 1) : objectName)
+                        + tr(levelNum ? "" : " (topmost)");
```
Acuion: Maybe add a numeration instead? Like
1) Name
Or something like
Floor 1) Name
To create some associations

bjorn: Sure, that would be fine. Then you can even do `"&1) "` for the first 9 entries to enable the number keys to be pressed for selecting one of the entries (`&` sets the mnemonic).

bjorn:
> So, do you think that multiple in-stack selection can be removed from the alt+click logic (a user will still can append an item from the stack to an existing selection)?

I'm not sure what you mean by this. I don't think I thought anything in addition to what I wrote.

> A new option "select all" in the context menu can be added instead. Other selection combination can be done with a shift+context menu selecting.

While you could add a "Select All" option, I'm not sure if it will be very useful. If you want to select all the items, then you can already just drag a regular selection rectangle (possibly while holding Shift to force selection in case you need to start dragging on top of one of the items).

Acuion:
- A numeration in the context menu 
- The rotation is now based on a selection
  - Shift unselection is not supported in this mode

src/tiled/objectselectiontool.cpp
```diff
@@ -533,6 +534,42 @@ void ObjectSelectionTool::mousePressed(QGraphicsSceneMouseEvent *event)
...
+    case Qt::RightButton:
+        if (event->modifiers() & Qt::AltModifier) {
+            QList<MapObjectItem*> underlyingObjects = objectItemsAt(event->scenePos());
+            if (underlyingObjects.empty())
+                break;
+            QMenu selectUnderlyingMenu;
+
+            for (int levelNum = 0; levelNum < underlyingObjects.size(); ++levelNum) {
+                const QString& objectName = underlyingObjects[levelNum]->mapObject()->name();
+                QString actionName = QLatin1String(levelNum < 9 ? "&" : "") + tr("%n) ", "", levelNum + 1)
```
Acuion: Is there any better way to construct a `"%d) "` string in this case? I think it doesn't require any translation, so can be made without `tr`.
And how to properly update a .ts file? I can provide a russian translation for strings in my code (now it is only "Unnamed object")

src/tiled/objectselectiontool.cpp
```diff
@@ -595,24 +593,25 @@ void ObjectSelectionTool::mouseReleased(QGraphicsSceneMouseEvent *event)
...
+            while (selection.contains(underlyingObjects.at(lastSelectedIndex))
+                   && lastSelectedIndex != underlyingObjects.size() - 1);
+            mClickedObjectItem = underlyingObjects.at(lastSelectedIndex);
         }
         if (mClickedObjectItem) {
             if (modifiers & (Qt::ShiftModifier | Qt::ControlModifier)) {
-                if (selection.contains(mClickedObjectItem))
-                    selection.remove(mClickedObjectItem);
+                if (!(modifiers & Qt::AltModifier) && selection.contains(mClickedObjectItem))
+                    selection.remove(mClickedObjectItem);// Removal is not supported in alt+click mode
```
bjorn: I don't understand why you would block removal. I don't think there is no negative side-effect to it.

src/tiled/objectselectiontool.cpp
```diff
@@ -595,24 +593,25 @@ void ObjectSelectionTool::mouseReleased(QGraphicsSceneMouseEvent *event)
...
-            mClickedObjectItem = mSelectedInUnderlyingRotation
-                    = underlyingObjects.at((currRotationIndex + 1) % underlyingObjects.size());
+
+            int lastSelectedIndex = -1;
+            for (auto selected : selection)
+                lastSelectedIndex = std::max(lastSelectedIndex, underlyingObjects.indexOf(selected));
+            do lastSelectedIndex = (lastSelectedIndex + 1) % underlyingObjects.size();
+            while (selection.contains(underlyingObjects.at(lastSelectedIndex))
+                   && lastSelectedIndex != underlyingObjects.size() - 1);
+            mClickedObjectItem = underlyingObjects.at(lastSelectedIndex);
```
bjorn: Was something wrong with [the loop I wrote earlier](https://github.com/bjorn/tiled/pull/1494#discussion_r106858622)? I think it was easier to understand and more efficient than this approach.

src/tiled/objectselectiontool.cpp
```diff
@@ -559,17 +558,16 @@ void ObjectSelectionTool::mousePressed(QGraphicsSceneMouseEvent *event)
...
+                        selection.insert(objectToBeSelected);
                 } else {
                     selection.clear();
                     selection.insert(objectToBeSelected);
                 }
                 mapScene()->setSelectedObjectItems(selection);
             }
-        }
-        else
+        } else
```
bjorn: Coding style: if some if/else chain has brackets somewhere, it should have them everywhere. Please add `{` here.

src/tiled/objectselectiontool.cpp
bjorn: I mostly just make it a translatable string anyway, because you never know (especially since some languages have different writing directions). But since this number does not relate to a plural form, you should not use the plural form override of `tr`.

I would probably just do:
```c++
if (objectName.isEmpty())
    objectName = tr("Unnamed object");

QString actionName;
if (levelNum < 9)
    actionName = tr("&%1) %2").arg(levelNum + 1).arg(objectName)
else
    actionName = tr("%1) %2").arg(levelNum + 1).arg(objectName)
```

src/tiled/objectselectiontool.cpp
Acuion: Otherwise it will select and unselect the same object in the cycle, I think it looks buggy

src/tiled/objectselectiontool.cpp
Acuion: It doesn't support a last->first iteration. If the bottom part of the stack is Shift-selected, your code always point at the topmost element.

src/tiled/objectselectiontool.cpp
Acuion: Ok, but what about a .ts file update? Can it be done manually or only via lupdate?

Acuion:
@bjorn any new comments?

bjorn:
@Acuion I can only commit my time fully to Tiled two days per week (currently Monday and Tuesday). On other days I may have no time at all, so while I will try to respond earlier, it may take until Monday.

About the select and unselect, I think it's better than no response at all, but it's also because I've seen the same behavior in Inkscape.

> If the bottom part of the stack is Shift-selected, your code always point at the topmost element.

Again, that was intentional and matches the behavior in Inkscape. I will need some time to try out your code again to see what really the difference is and what behavior I'd prefer.

Acuion:
Ok, thank you :)
I just want the patch to be finished asap. Waiting for your next feedback.

bjorn:
@Acuion After looking at your code in detail, I liked the fact that it looks for the first non-selected object when no unselected object was available at the bottom. I did however [rewrite that part a little](https://github.com/bjorn/tiled/commit/59abdf7e2f918be4a821a3689b6aa4ff2385f051), since I think it was rather hard to see what it was doing.

I did allow removal from the selection, so that the user doesn't have to be so conscious about releasing his Alt key when he wants to exclude some object. Indeed, it can only work for the first object and as long as there is no unselected object below, but I don't think it's a reason to block it entirely since I think this case isn't so rare.

Thanks a lot for your contribution!

Acuion:
Thanks for your guidance! Yes, I think that an unselection of the topmost element really has its application.
Btw, you didn't explained how to modify the translation. This patch is already meged, but it will be useful in the future :).
PS: I've shared my draft via the GSoC system, waiting for the feedback

Acuion:
And looks like your code have a little bug
`for (int i = 1; i < underlyingObjects.size() - 2; ++i) {`
probably should be
`for (int i = 1; i < underlyingObjects.size() - 1; ++i) {`
or <=

In stack of three select the last element, shift+alt+click twice and realize that the second one never get the selection.

bjorn:
> Btw, you didn't explained how to modify the translation. This patch is already meged, but it will be useful in the future, so I will be grateful for the explanation :).

That is because I don't update translation files until about 2 weeks before a new release. This 2 week period is known as a "string freeze", in which no changes to translated strings are allowed. This gives all translators a chance to update the translations (and you could help update yours).

> And seems there is a little bug in your code

Thanks for noticing! I only meant to exclude the last element, but subtracted one too much. :-/

PR #1495, opened 2017-03-16
Coding style. Add basic vera++ usage

Sharganov:
Propose to discuss here code style rules.

### DIFF
--- a/src/tiled/abstracttool.cpp
+++ b/src/tiled/abstracttool.cpp
@@ -28,8 +28,10 @@
 namespace Tiled {
 namespace Internal {
 
-AbstractTool::AbstractTool(const QString &name, const QIcon &icon,
-                           const QKeySequence &shortcut, QObject *parent)
+AbstractTool::AbstractTool(const QString &name,
+                           const QIcon &icon,
+                           const QKeySequence &shortcut,
+                           QObject *parent)
     : QObject(parent)
     , mName(name)
     , mIcon(icon)
@@ -81,10 +83,12 @@ void AbstractTool::setMapDocument(MapDocument *mapDocument)
         return;
 
     if (mMapDocument) {
-        disconnect(mMapDocument, &MapDocument::layerChanged,
-                   this, &AbstractTool::updateEnabledState);
-        disconnect(mMapDocument, &MapDocument::currentLayerChanged,
-                   this, &AbstractTool::updateEnabledState);
+        disconnect(
+            mMapDocument, &MapDocument::layerChanged, this, &AbstractTool::updateEnabledState);
+        disconnect(mMapDocument,
+                   &MapDocument::currentLayerChanged,
+                   this,
+                   &AbstractTool::updateEnabledState);
     }
 
     MapDocument *oldDocument = mMapDocument;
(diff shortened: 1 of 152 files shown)


bjorn:
Please start your `coding_style` branch from my `master`, because this way the pull request includes changes that have already been merged as part of #1479. When you have your branch updated locally, you can force-push it to your fork on GitHub to update this pull request (no need to close/reopen the pull request).

ketanhwr:
@Sharganov don't forget to add vera++ in the file `appveyor.yml` as well :)

bjorn:
@ketanhwr I think executing it at either Travis or AppVeyor is enough, since I don't expect the output to depend on the operating system.

@Sharganov It seems like it's checking .ui files for C++ style as well:

```
src/tiled/preferencesdialog.ui:63: keyword 'if' not followed by a single space
src/tiled/preferencesdialog.ui:307: keyword 'for' not followed by a single space
```

Besides being wrong, since they are followed by a single space in both cases, is there a way we can avoid processing .ui files?

Sharganov:
I think it will be easy to fix checking .ui files. 
There is another tool - clang-format, it supports some standart [options](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) I plan to use it too. What do you think?

*Sorry for low activity, university takes the most time.

Ablu:
There also is a clang-format integration for QtCreator. Not sure if that 
works with vera++.

Without having used vera++ it looks like clang-format is more actively 
developed... But not sure how it compares to vera++ otherwise.

Sharganov:
I think that the only reason to use vera++ is support of custom scripts (rules).

src/tiled/abstracttool.cpp
```diff
@@ -81,21 +83,24 @@ void AbstractTool::setMapDocument(MapDocument *mapDocument)
         return;
 
     if (mMapDocument) {
-        disconnect(mMapDocument, &MapDocument::layerChanged,
-                   this, &AbstractTool::updateEnabledState);
-        disconnect(mMapDocument, &MapDocument::currentLayerChanged,
-                   this, &AbstractTool::updateEnabledState);
+        disconnect(
+            mMapDocument, &MapDocument::layerChanged, this, &AbstractTool::updateEnabledState);
+        disconnect(mMapDocument,
```
Sharganov: it's what I mean under controversial, two very similar disconnects have absolutely different calls.

Sharganov:
Tried different ClangFormat options combination, seems like there isn't "silver bullet". I commit .cpp file changes example. 
There are many improvements, however, some changes make code difficult to read. I followed Qt Creator Coding Rules in case of brackets and etc.

1) The positive(stable) results with brackets, indentations, short expression, pointers alignment.
2) The functions arguments and parameters alignment is controversial.

src/tiled/abstracttool.cpp
bjorn: It seems to be applying the ColumnLimit of 100? I personally use a soft-limit of 80 (that's where I have Qt Creator display a margin line), but it's a soft limit because I always prefer readability to adhering to line length limitations.

Due to explicitly having to name the two classes involved, the connect statements often exceed the line length limit, but I think it's more readable to have it on the same line so I usually keep them that way. In many places you also see them wrapped, which is just because I was more strict about the line length in the past.

bjorn:
I am not interested in a fully automated reformatting of the source code, btw. Hence I would prefer vera++, or anything that gives us more control about what we want to complain about. Does Clang have a mode like vera++, where it doesn't reformat but will instead print out warnings when certain style rules are violated?

Sharganov:
Don't think so.

Ablu:
This looks possible: 
https://stackoverflow.com/questions/22866609/can-clang-format-tell-me-if-formatting-changes-are-necessary

bjorn:
Closing old PR. While code style checks might be interesting to include in CI, I'm not personally interested in setting this up at the moment. I'm also not a fan of reformatting all the existing code.

PR #1496, opened 2017-03-16
Stamp brush now works with the stagger in hex mode as expected

Bdtrotte:
Previously when using the stamp brush tool when in hexagonal mode, moving down rows would cause a stamp to behave strangely (not accounting for the stagger). This has been patched over by shifting every other row of the stamp over in the appropriate direction when needed.

This is my first contribution, and major edit of tiled, so feedback on how anything could be done better would be greatly appreciated!

Ben

### DIFF
--- a/src/tiled/stampbrush.cpp
+++ b/src/tiled/stampbrush.cpp
@@ -45,6 +45,7 @@ StampBrush::StampBrush(QObject *parent)
                        parent)
     , mBrushBehavior(Free)
     , mIsRandom(false)
+    , mCaptureTopCorner(-1)
 {
 }
 
@@ -250,6 +251,8 @@ void StampBrush::endCapture()
 
     // Intersect with the layer and translate to layer coordinates
     QRect captured = capturedArea();
+    mCaptureTopCorner = captured.top();
+
     captured &= QRect(tileLayer->x(), tileLayer->y(),
                       tileLayer->width(), tileLayer->height());
 
@@ -385,7 +388,53 @@ void StampBrush::drawPreviewLayer(const QVector<QPoint> &list)
             const TileStampVariation variation = mStamp.randomVariation();
             mapDocument()->unifyTilesets(variation.map, mMissingTilesets);
 
-            TileLayer *stamp = variation.tileLayer();
+            TileLayer *stamp = variation.tileLayer()->copy(variation.tileLayer()->bounds());
+
+            //If using a hexagonal map, this takes into acount staggering
+            if(mapDocument()->map()->orientation() == 4 && stamp->height()>1)
+            {
+                int shiftMode;//-1 do not shift, 0 shift even rows (relatived to the stamp), 1 shift odd rows
+                int rowOfTop = p.y()-stamp->height()/2;
+                if(rowOfTop<0)rowOfTop*=-1; //Only used to check if even or odd, and being negative can mess this up.
+
+                if(mCaptureTopCorner == -1)
+                    shiftMode = 1;
+                else
+                //if what the stamp captured starts by going left
+                if((mCaptureTopCorner+1)%2 == mapDocument()->map()->staggerIndex())
+                {
+                    //if where the top is now will go down the correct way
+                    if((rowOfTop+1)%2 != mapDocument()->map()->staggerIndex())
+                        shiftMode = 0;
+                    else
+                        shiftMode = -1;
+                }
+                else//if going right
+                {
+                    //if where the top is now will go down the correct way
+                    if((rowOfTop+1)%2 == mapDocument()->map()->staggerIndex())
+                        shiftMode = 1;
+                    else
+                        shiftMode = -1;
+                }
+
+                if(shiftMode != -1)
+                {
+                    stamp->resize(QSize(stamp->width()+1,stamp->height()),QPoint());
+
+                    QRect boundBox = stamp->bounds();
+                    //Actually does the shift of the tiles. Maybe this could be done as
+                    //a method of TileLayer?
+                    for(int i = shiftMode; i <= boundBox.bottom(); i+=2)
+                    {
+                        for(int j = boundBox.right()-1; j >= 0; --j)
+                        {
+                            stamp->setCell(j+1,i,stamp->cellAt(j,i));
+                            stamp->setCell(j,i,Cell());
+                        }
+                    }
+                }
+            }
 
             QRegion stampRegion;
             if (regionCache.contains(stamp)) {
--- a/src/tiled/stampbrush.cpp
+++ b/src/tiled/stampbrush.cpp
@@ -391,7 +391,7 @@ void StampBrush::drawPreviewLayer(const QVector<QPoint> &list)
         QRegion paintedRegion;
         QVector<PaintOperation> operations;
         QHash<TileLayer *, QRegion> regionCache;
-        QHash<TileLayer *,TileLayer *> shiftedCopies;
+        QHash<TileLayer *, TileLayer *> shiftedCopies;
 
         for (const QPoint &p : list) {
             const TileStampVariation variation = mStamp.randomVariation();
@@ -414,7 +414,7 @@ void StampBrush::drawPreviewLayer(const QVector<QPoint> &list)
                         TileLayer *shiftedStamp = shiftedCopies.value(stamp);
                         if (!shiftedStamp) {
                             shiftedStamp = static_cast<TileLayer *>(stamp->clone());
-                            shiftedCopies.insert(stamp,shiftedStamp);
+                            shiftedCopies.insert(stamp, shiftedStamp);
 
                             shiftedStamp->resize(QSize(shiftedStamp->width() + 1, shiftedStamp->height()), QPoint());
 
@@ -433,7 +433,7 @@ void StampBrush::drawPreviewLayer(const QVector<QPoint> &list)
                         TileLayer *shiftedStamp = shiftedCopies.value(stamp);
                         if (!shiftedStamp) {
                             shiftedStamp = static_cast<TileLayer *>(stamp->clone());
-                            shiftedCopies.insert(stamp,shiftedStamp);
+                            shiftedCopies.insert(stamp, shiftedStamp);
 
                             shiftedStamp->resize(QSize(shiftedStamp->width(), shiftedStamp->height() + 1), QPoint());
 
--- a/src/tiled/stampbrush.cpp
+++ b/src/tiled/stampbrush.cpp
@@ -255,14 +255,11 @@ void StampBrush::endCapture()
                       tileLayer->width(), tileLayer->height());
 
     //gets if the relative stagger should be the same as the base layer
-    int stagIndexOffSet;
+    int staggerIndexOffSet;
     if (tileLayer->map()->staggerAxis() == Map::StaggerX)
-        stagIndexOffSet = captured.x() % 2;
+        staggerIndexOffSet = captured.x() % 2;
     else
-        stagIndexOffSet = captured.y() % 2;
-
-    captured &= QRect(tileLayer->x(), tileLayer->y(),
-                      tileLayer->width(), tileLayer->height());
+        staggerIndexOffSet = captured.y() % 2;
 
     if (captured.isValid()) {
         captured.translate(-tileLayer->x(), -tileLayer->y());
@@ -275,7 +272,7 @@ void StampBrush::endCapture()
                              map->tileHeight());
 
         stamp->setStaggerAxis(map->staggerAxis());
-        stamp->setStaggerIndex((Map::StaggerIndex)((map->staggerIndex() + stagIndexOffSet) % 2));
+        stamp->setStaggerIndex((Map::StaggerIndex)((map->staggerIndex() + staggerIndexOffSet) % 2));
 
         // Add tileset references to map
         foreach (const SharedTileset &tileset, capture->usedTilesets())
@@ -350,7 +347,6 @@ QRegion StampBrush::doPaint(int flags)
 struct PaintOperation {
     QPoint pos;
     TileLayer *stamp;
-    bool isCopy;
 };
 
 /**
@@ -395,54 +391,51 @@ void StampBrush::drawPreviewLayer(const QVector<QPoint> &list)
         QRegion paintedRegion;
         QVector<PaintOperation> operations;
         QHash<TileLayer *, QRegion> regionCache;
+        QHash<TileLayer *, TileLayer *> shiftedCopies;
 
         for (const QPoint &p : list) {
             const TileStampVariation variation = mStamp.randomVariation();
             mapDocument()->unifyTilesets(variation.map, mMissingTilesets);
 
             TileLayer *stamp = variation.tileLayer();
-            bool copyFlag = false;
+
+            Map::StaggerAxis mapStaggerAxis = mapDocument()->map()->staggerAxis();
+            Map::StaggerIndex mapStaggerIndex = mapDocument()->map()->staggerIndex();
+            Map::StaggerIndex stampStaggerIndex = variation.map->staggerIndex();
+
 
             //if staggered map, makes sure stamp stays the same
-            if ((mapDocument()->map()->orientation() == Map::Hexagonal
-                    || mapDocument()->map()->orientation() == Map::Staggered)
-                    && ((mapDocument()->map()->staggerAxis() == Map::StaggerY)?
-                         stamp->height() > 1 : stamp->width() > 1)) {
-                stamp = variation.tileLayer()->copy(variation.tileLayer()->bounds());
-                copyFlag = true;
+            if (mapDocument()->map()->isStaggered()
+                    && ((mapStaggerAxis == Map::StaggerY)? stamp->height() > 1 : stamp->width() > 1)) {
 
-                if(mapDocument()->map()->staggerAxis() == Map::StaggerY) {
-                    int rowOfTop = p.y() - stamp->height() / 2;
+                if (mapStaggerAxis == Map::StaggerY) {
+                    bool topIsOdd = (p.y() - stamp->height() / 2) & 1;
 
-                    if ((variation.map->staggerIndex() == mapDocument()->map()->staggerIndex()
-                            && (rowOfTop&1) == 1)
-                               ||
-                           (variation.map->staggerIndex() != mapDocument()->map()->staggerIndex()
-                            && (rowOfTop&1) == 0)) {
-                        stamp->resize(QSize(stamp->width() + 1,stamp->height()),QPoint());
+                    if ((stampStaggerIndex == mapStaggerIndex) == topIsOdd) {
+                        stamp = variation.tileLayer()->copy(variation.tileLayer()->bounds());
+                        shiftedCopies.insert(stamp, stamp);
 
-                        for (int y = (variation.map->staggerIndex()+1)&1; y < stamp->height(); y += 2) {
-                            for(int x = stamp->width() - 2; x >= 0; --x) {
-                                stamp->setCell(x+1,y,stamp->cellAt(x,y));
-                                stamp->setCell(x,y,Cell());
-                            }
+                        stamp->resize(QSize(stamp->width() + 1, stamp->height()), QPoint());
+
+                        for (int y = (stampStaggerIndex + 1) & 1; y < stamp->height(); y += 2) {
+                            for (int x = stamp->width() - 2; x >= 0; --x)
+                                stamp->setCell(x + 1, y, stamp->cellAt(x, y));
+                            stamp->setCell(0, y, Cell());
                         }
                     }
                 } else {
-                    int colOfLeft = p.x() - stamp->width() / 2;
+                    bool leftIsOdd = (p.x() - stamp->width() / 2) & 1;
 
-                    if ((variation.map->staggerIndex() == mapDocument()->map()->staggerIndex()
-                            && (colOfLeft&1) == 1)
-                               ||
-                           (variation.map->staggerIndex() != mapDocument()->map()->staggerIndex()
-                            && (colOfLeft&1) == 0)) {
-                        stamp->resize(QSize(stamp->width(),stamp->height()+1),QPoint());
+                    if ((stampStaggerIndex == mapStaggerIndex) == leftIsOdd) {
+                        stamp = variation.tileLayer()->copy(variation.tileLayer()->bounds());
+                        shiftedCopies.insert(stamp, stamp);
 
-                        for (int x = (variation.map->staggerIndex() + 1)&1; x < stamp->width(); x += 2) {
-                            for(int y = stamp->height() - 2; y >= 0; --y) {
-                                stamp->setCell(x,y+1,stamp->cellAt(x,y));
-                                stamp->setCell(x,y,Cell());
-                            }
+                        stamp->resize(QSize(stamp->width(), stamp->height() + 1), QPoint());
+
+                        for (int x = (stampStaggerIndex + 1) & 1; x < stamp->width(); x += 2) {
+                            for (int y = stamp->height() - 2; y >= 0; --y)
+                                stamp->setCell(x, y + 1, stamp->cellAt(x, y));
+                            stamp->setCell(x, 0, Cell());
                         }
                     }
                 }
@@ -464,7 +457,7 @@ void StampBrush::drawPreviewLayer(const QVector<QPoint> &list)
             if (!paintedRegion.intersects(region)) {
                 paintedRegion += region;
 
-                PaintOperation op = { centered, stamp, copyFlag };
+                PaintOperation op = { centered, stamp };
                 operations.append(op);
             }
         }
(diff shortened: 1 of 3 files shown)


src/tiled/stampbrush.cpp
```diff
@@ -385,7 +388,53 @@ void StampBrush::drawPreviewLayer(const QVector<QPoint> &list)
             const TileStampVariation variation = mStamp.randomVariation();
             mapDocument()->unifyTilesets(variation.map, mMissingTilesets);
 
-            TileLayer *stamp = variation.tileLayer();
+            TileLayer *stamp = variation.tileLayer()->copy(variation.tileLayer()->bounds());
+
+            //If using a hexagonal map, this takes into acount staggering
+            if(mapDocument()->map()->orientation() == 4 && stamp->height()>1)
```
ketanhwr: Code style: Add a space between `if` and `(`

src/tiled/stampbrush.cpp
```diff
@@ -385,7 +388,53 @@ void StampBrush::drawPreviewLayer(const QVector<QPoint> &list)
...
+            {
+                int shiftMode;//-1 do not shift, 0 shift even rows (relatived to the stamp), 1 shift odd rows
+                int rowOfTop = p.y()-stamp->height()/2;
+                if(rowOfTop<0)rowOfTop*=-1; //Only used to check if even or odd, and being negative can mess this up.
```
ketanhwr: Always add spaces to make the code more readable. Also, add the statement `rowOfTop*=-1` in the next line :)
```
if (rowOfTop < 0)
    rowOfTop *= -1;

File: src/tiled/stampbrush.cpp [hunk 3]
```diff
@@ -385,7 +388,53 @@ void StampBrush::drawPreviewLayer(const QVector<QPoint> &list)
...
+
+                if(mCaptureTopCorner == -1)
```
ketanhwr: Space between `if` and `(`

src/tiled/stampbrush.cpp
```diff
@@ -385,7 +388,53 @@ void StampBrush::drawPreviewLayer(const QVector<QPoint> &list)
...
+                    shiftMode = 1;
+                else
```
ketanhwr: Why have you added an `else` construct when you don't have to execute any statements under it?

src/tiled/stampbrush.cpp
```diff
@@ -385,7 +388,53 @@ void StampBrush::drawPreviewLayer(const QVector<QPoint> &list)
...
+                //if what the stamp captured starts by going left
+                if((mCaptureTopCorner+1)%2 == mapDocument()->map()->staggerIndex())
```
ketanhwr: Space between `if` and `(`

src/tiled/stampbrush.cpp
```diff
@@ -385,7 +388,53 @@ void StampBrush::drawPreviewLayer(const QVector<QPoint> &list)
...
+                {
+                    //if where the top is now will go down the correct way
+                    if((rowOfTop+1)%2 != mapDocument()->map()->staggerIndex())
```
ketanhwr: Here too.

src/tiled/stampbrush.cpp
```diff
@@ -385,7 +388,53 @@ void StampBrush::drawPreviewLayer(const QVector<QPoint> &list)
...
+                        shiftMode = 0;
+                    else
+                        shiftMode = -1;
+                }
+                else//if going right
```
ketanhwr: Code style: Shift the `else` to previous line like this:
```
} else {
```

src/tiled/stampbrush.cpp
```diff
@@ -385,7 +388,53 @@ void StampBrush::drawPreviewLayer(const QVector<QPoint> &list)
...
+                {
+                    //if where the top is now will go down the correct way
+                    if((rowOfTop+1)%2 == mapDocument()->map()->staggerIndex())
```
ketanhwr: Here.

src/tiled/stampbrush.cpp
```diff
@@ -385,7 +388,53 @@ void StampBrush::drawPreviewLayer(const QVector<QPoint> &list)
...
+                        shiftMode = 1;
+                    else
+                        shiftMode = -1;
+                }
+
+                if(shiftMode != -1)
```
ketanhwr: Here too.

src/tiled/stampbrush.cpp
```diff
@@ -385,7 +388,53 @@ void StampBrush::drawPreviewLayer(const QVector<QPoint> &list)
...
+                {
+                    stamp->resize(QSize(stamp->width()+1,stamp->height()),QPoint());
+
+                    QRect boundBox = stamp->bounds();
+                    //Actually does the shift of the tiles. Maybe this could be done as
+                    //a method of TileLayer?
+                    for(int i = shiftMode; i <= boundBox.bottom(); i+=2)
```
ketanhwr: Add a space between `for` and `(` too :)

src/tiled/stampbrush.cpp
```diff
@@ -385,7 +388,53 @@ void StampBrush::drawPreviewLayer(const QVector<QPoint> &list)
...
+                    {
+                        for(int j = boundBox.right()-1; j >= 0; --j)
```
ketanhwr: Here as well.

src/tiled/stampbrush.cpp
Bdtrotte: that else should pair up with the if under it (separated by a comment). Though i will clean it up for readability!

src/tiled/stampbrush.cpp
Bdtrotte: haha,DoyouthinkIshouldaddsomespaces? Ill get on cleaning up the style! Thanks for pointing it all out.
Ben

src/tiled/stampbrush.cpp
ketanhwr: Oh, I got confused by that commend :/

src/tiled/stampbrush.cpp
ketanhwr: Also I forgot to mention, but move the curly bracket to this line itself.
```
if (mapDocument()->map()->orientation() == 4 && stamp->height() > 1) {
```

src/tiled/stampbrush.cpp
ketanhwr: Move the curly bracket here too.
```
if ((mCaptureTopCorner+1)%2 == mapDocument()->map()->staggerIndex()) {
```

src/tiled/stampbrush.cpp
ketanhwr: Here as well.
```
for () {
```

src/tiled/stampbrush.cpp
ketanhwr: It's just a style everyone should follow in order to maintain readability and consistency :)

ketanhwr: Btw, move the curly brackets up here as well.

bjorn:
@ketanhwr When you do a review, please use the "Start a review" button instead of "Add a single comment". It's something I had to get used to after they added the feature and I still sometimes forget, but it makes the whole review process a lot less spammy in terms of e-mail notifications. As a bonus, it means you can tweak all your comments until the time when you feel ready to submit your review. But thanks for stepping in to comment!

src/tiled/stampbrush.cpp
```diff
@@ -385,7 +388,47 @@ void StampBrush::drawPreviewLayer(const QVector<QPoint> &list)
...
+                        shiftMode = -1;
+                }
+
+                if (shiftMode != -1) {
+                    stamp->resize(QSize(stamp->width()+1,stamp->height()),QPoint());
+
+                    QRect boundBox = stamp->bounds();
+                    //Actually does the shift of the tiles. Maybe this could be done as
+                    //a method of TileLayer?
+                    for (int i = shiftMode; i <= boundBox.bottom(); i+=2)
```
bjorn: I see no advantage to using `<= boundBox.bottom()` here instead of just `< stamp->height()`.

src/tiled/stampbrush.cpp
```diff
@@ -385,7 +388,47 @@ void StampBrush::drawPreviewLayer(const QVector<QPoint> &list)
...
+            TileLayer *stamp = variation.tileLayer()->copy(variation.tileLayer()->bounds());
+
+            //If using a hexagonal map, this takes into acount staggering
+            if (mapDocument()->map()->orientation() == 4 && stamp->height()>1) {
+                int shiftMode;//-1 do not shift, 0 shift even rows (relatived to the stamp), 1 shift odd rows
+                int rowOfTop = p.y()-stamp->height()/2;
+                if (rowOfTop<0)
+                    rowOfTop*=-1; //Only used to check if even or odd, and being negative can mess this up.
+
+                if (mCaptureTopCorner == -1)
```
bjorn: It's interesting how this breaks the shift adjustment for captures made starting from -1, just outside the map.

In any case, you will need to associate this data with the tile layer instance rather than storing it as a member of the stamp brush. In fact, the tile layer already has a member for exactly that, which is the `staggerIndex`. All you should need to do, is to make sure this property gets set correctly when the stamp is captured, and to account for it when the stamp is used at odd indexes (or at even indexes, when the stamp stagger index != the map's stagger index).

bjorn:
> In any case, you will need to associate this data with the tile layer instance rather than storing it as a member of the stamp brush. In fact, the tile layer already has a member for exactly that, which is the staggerIndex. All you should need to do, is to make sure this property gets set correctly when the stamp is captured, and to account for it when the stamp is used at odd indexes (or at even indexes, when the stamp stagger index != the map's stagger index).

Sorry, I completely forgot that the staggerIndex is stored on the map rather than on the tile layer. Thankfully, at some point I changed the stamp to always store a map with a single tile layer, in order to keep track of exactly this kind of information. So you should be able to use that.

Bdtrotte:
Made all the changes, now works with isometric and x-axis stagger nicely! Feel much better about this version. Hopefully there are not too many style issues this time...

src/tiled/stampbrush.cpp
```diff
@@ -251,7 +250,16 @@ void StampBrush::endCapture()
...
+
+    //gets if the relative stagger should be the same as the base layer
+    int stagIndexOffSet;
+    if (tileLayer->map()->staggerAxis() == Map::StaggerX)
+        stagIndexOffSet = captured.x() % 2;
+    else
+        stagIndexOffSet = captured.y() % 2;
 
     captured &= QRect(tileLayer->x(), tileLayer->y(),
                       tileLayer->width(), tileLayer->height());
```
bjorn: Don't forget to remove these lines, since you moved the code up.

src/tiled/stampbrush.cpp
```diff
@@ -456,8 +474,11 @@ void StampBrush::drawPreviewLayer(const QVector<QPoint> &list)
                                               bounds.x(), bounds.y(),
                                               bounds.width(), bounds.height()));
 
-        for (const PaintOperation &op : operations)
+        for (const PaintOperation &op : operations) {
             preview->merge(op.pos - bounds.topLeft(), op.stamp);
+            if(op.isCopy)
```
bjorn: Coding style: please put a space after `if`

src/tiled/stampbrush.cpp
```diff
@@ -386,6 +401,52 @@ void StampBrush::drawPreviewLayer(const QVector<QPoint> &list)
...
+
+                    if ((variation.map->staggerIndex() == mapDocument()->map()->staggerIndex()
+                            && (rowOfTop&1) == 1)
+                               ||
+                           (variation.map->staggerIndex() != mapDocument()->map()->staggerIndex()
+                            && (rowOfTop&1) == 0)) {
+                        stamp->resize(QSize(stamp->width() + 1,stamp->height()),QPoint());
+
+                        for (int y = (variation.map->staggerIndex()+1)&1; y < stamp->height(); y += 2) {
+                            for(int x = stamp->width() - 2; x >= 0; --x) {
```
bjorn: Coding style: space after `for`

src/tiled/stampbrush.cpp
```diff
@@ -386,6 +401,52 @@ void StampBrush::drawPreviewLayer(const QVector<QPoint> &list)
...
+                                stamp->setCell(x+1,y,stamp->cellAt(x,y));
+                                stamp->setCell(x,y,Cell());
```
bjorn: This line can still be moved out of the inner loop, since it only needs to be done for the left-most cell.

src/tiled/stampbrush.cpp
```diff
@@ -386,6 +401,52 @@ void StampBrush::drawPreviewLayer(const QVector<QPoint> &list)
...
             TileLayer *stamp = variation.tileLayer();
+            bool copyFlag = false;
+
+            //if staggered map, makes sure stamp stays the same
+            if ((mapDocument()->map()->orientation() == Map::Hexagonal
+                    || mapDocument()->map()->orientation() == Map::Staggered)
+                    && ((mapDocument()->map()->staggerAxis() == Map::StaggerY)?
+                         stamp->height() > 1 : stamp->width() > 1)) {
+                stamp = variation.tileLayer()->copy(variation.tileLayer()->bounds());
+                copyFlag = true;
```
bjorn: At this point you still don't know whether it is actually necessary to create a copy. Also, you only ever need a single adjusted copy, since there is only a single possible adjustment depending on the context. So I think you can really just store it in a `QScopedPointer<TileLayer*> copy` (or maybe call it `shiftedCopy`) and create it on-demand.

src/tiled/stampbrush.cpp
```diff
@@ -386,6 +401,52 @@ void StampBrush::drawPreviewLayer(const QVector<QPoint> &list)
...
+
+                if(mapDocument()->map()->staggerAxis() == Map::StaggerY) {
+                    int rowOfTop = p.y() - stamp->height() / 2;
+
+                    if ((variation.map->staggerIndex() == mapDocument()->map()->staggerIndex()
+                            && (rowOfTop&1) == 1)
+                               ||
+                           (variation.map->staggerIndex() != mapDocument()->map()->staggerIndex()
```
bjorn: You're doing `mapDocument()->map()->staggerIndex()` so often that it really makes sense to introduce a local `mapStaggerIndex` (and a `stampStaggerIndex`) to make the code shorter and more readable.

src/tiled/stampbrush.cpp
```diff
@@ -386,6 +401,52 @@ void StampBrush::drawPreviewLayer(const QVector<QPoint> &list)
...
+                    && ((mapDocument()->map()->staggerAxis() == Map::StaggerY)?
+                         stamp->height() > 1 : stamp->width() > 1)) {
+                stamp = variation.tileLayer()->copy(variation.tileLayer()->bounds());
+                copyFlag = true;
+
+                if(mapDocument()->map()->staggerAxis() == Map::StaggerY) {
+                    int rowOfTop = p.y() - stamp->height() / 2;
+
+                    if ((variation.map->staggerIndex() == mapDocument()->map()->staggerIndex()
+                            && (rowOfTop&1) == 1)
```
bjorn: Coding style: please put spaces around `&`.

Also, since you're not using `rowOfTop` for any other computation, you can change that local variable to `bool oddOffset` for example.

src/tiled/stampbrush.cpp
```diff
@@ -386,6 +401,52 @@ void StampBrush::drawPreviewLayer(const QVector<QPoint> &list)
             mapDocument()->unifyTilesets(variation.map, mMissingTilesets);
 
             TileLayer *stamp = variation.tileLayer();
+            bool copyFlag = false;
+
+            //if staggered map, makes sure stamp stays the same
+            if ((mapDocument()->map()->orientation() == Map::Hexagonal
+                    || mapDocument()->map()->orientation() == Map::Staggered)
```
bjorn: Feel free to introduce a `bool Map::isStaggered() const` method to make this code more readable.

src/tiled/stampbrush.cpp
```diff
@@ -386,6 +401,52 @@ void StampBrush::drawPreviewLayer(const QVector<QPoint> &list)
...
+                        }
+                    }
+                } else {
+                    int colOfLeft = p.x() - stamp->width() / 2;
+
+                    if ((variation.map->staggerIndex() == mapDocument()->map()->staggerIndex()
+                            && (colOfLeft&1) == 1)
+                               ||
+                           (variation.map->staggerIndex() != mapDocument()->map()->staggerIndex()
+                            && (colOfLeft&1) == 0)) {
```
bjorn: I believe this whole condition could be written as:
```c++
if ((stampStaggerIndex == mapStaggerIndex) == oddOffset) {
```

src/tiled/stampbrush.cpp
```diff
@@ -386,6 +401,52 @@ void StampBrush::drawPreviewLayer(const QVector<QPoint> &list)
...
+                        stamp->resize(QSize(stamp->width(),stamp->height()+1),QPoint());
+
+                        for (int x = (variation.map->staggerIndex() + 1)&1; x < stamp->width(); x += 2) {
+                            for(int y = stamp->height() - 2; y >= 0; --y) {
+                                stamp->setCell(x,y+1,stamp->cellAt(x,y));
```
bjorn: Coding style: please put spaces after commas.

src/tiled/stampbrush.cpp
```diff
@@ -250,6 +250,17 @@ void StampBrush::endCapture()
 
     // Intersect with the layer and translate to layer coordinates
     QRect captured = capturedArea();
+
+    captured &= QRect(tileLayer->x(), tileLayer->y(),
+                      tileLayer->width(), tileLayer->height());
+
+    //gets if the relative stagger should be the same as the base layer
+    int stagIndexOffSet;
```
bjorn: One more comment: there is no point in shortening "stagger" here, so just call it `staggerIndexOffset`.

src/tiled/stampbrush.cpp
bjorn: Oops, I totally forgot about the randomVariation part. You could still do a cache like:
```c++
QHash<TileLayer*, TileLayer*> shiftedCopies;

...

qDeleteAll(shiftedCopies);
```

Bdtrotte: Is there a disadvantage to how I'm doing it now with the copy flag? As we already have to loop through all the `operations` so I thought deleting the copies there with `delete` would be fastest. Ill go ahead and implement with the QHash and push a new commit, but I'm curious of the difference.

src/tiled/stampbrush.cpp
Bdtrotte: I have done this. Good idea.

Bdtrotte:
Made the latest batch of requested improvements and style updates!

src/tiled/stampbrush.cpp
bjorn: It is only a performance concern. You're making a copy for each point where the stamp is to be drawn. With a somewhat large stamp in combination with a large amount of points (so either drawing a long line or large circle), this consumes unnecessary memory and CPU cycles.

Even though this could count as "premature optimization", I think the more performant code where only a single copy is made per variation (and there is usually only one variation) is trivial to write and then I think why not just do it.

src/tiled/stampbrush.cpp
```diff
@@ -380,13 +391,56 @@ void StampBrush::drawPreviewLayer(const QVector<QPoint> &list)
...
+            //if staggered map, makes sure stamp stays the same
+            if (mapDocument()->map()->isStaggered()
+                    && ((mapStaggerAxis == Map::StaggerY)? stamp->height() > 1 : stamp->width() > 1)) {
+
+                if (mapStaggerAxis == Map::StaggerY) {
+                    bool topIsOdd = (p.y() - stamp->height() / 2) & 1;
+
+                    if ((stampStaggerIndex == mapStaggerIndex) == topIsOdd) {
+                        stamp = variation.tileLayer()->copy(variation.tileLayer()->bounds());
+                        shiftedCopies.insert(stamp, stamp);
```
bjorn: Before you go and make the copy and insert it, you should of course check whether it isn't already there (meaning as well, that you should use the original stamp as the key, and not the copy). If a shifted copy was already made for this stamp, just assign that to `stamp` and don't do the shifting again.

src/tiled/stampbrush.cpp
```diff
@@ -413,8 +467,11 @@ void StampBrush::drawPreviewLayer(const QVector<QPoint> &list)
                                               bounds.x(), bounds.y(),
                                               bounds.width(), bounds.height()));
 
-        for (const PaintOperation &op : operations)
+        for (const PaintOperation &op : operations) {
             preview->merge(op.pos - bounds.topLeft(), op.stamp);
+        }
```
bjorn: Coding style: please remove these brackets again now.

src/tiled/stampbrush.cpp
```diff
@@ -380,13 +391,56 @@ void StampBrush::drawPreviewLayer(const QVector<QPoint> &list)
...
+
+            //if staggered map, makes sure stamp stays the same
+            if (mapDocument()->map()->isStaggered()
+                    && ((mapStaggerAxis == Map::StaggerY)? stamp->height() > 1 : stamp->width() > 1)) {
+
+                if (mapStaggerAxis == Map::StaggerY) {
+                    bool topIsOdd = (p.y() - stamp->height() / 2) & 1;
+
+                    if ((stampStaggerIndex == mapStaggerIndex) == topIsOdd) {
+                        stamp = variation.tileLayer()->copy(variation.tileLayer()->bounds());
```
bjorn: Just use `stamp` instead of `variation.tileLayer()`. It's the same thing after all.

src/tiled/stampbrush.cpp
Bdtrotte: Yep, you're totally right. I'll work on fixing that. Do you think there is a clean way of not having to make to copy but once when the stamp is captured (or a map setting is changed)? I just couldn't think where to store this.

src/tiled/stampbrush.cpp
Bdtrotte: Could it be easier to just use a bool flag to see if we've made the copy? As we are only storing one copy. Or is there an advantage to the QHash?

src/tiled/stampbrush.cpp
bjorn: I don't think that is possible in an easy way. So at that point, the optimization because too much of an engineering effort and it's not worth doing it unless we notice an actual performance issue with the easy approach.

src/tiled/stampbrush.cpp
Bdtrotte: Or for that matter just checking is a `TileLayer * == nullptr` and if so make the copy, and if not set `stamp = copy`

bjorn: No, it's not easier with a bool flag, because you need a bool flag _for each variation_. This also means we're storing potentially multiple copies (one for each variation that ends up being used). Hence we need the QHash. You see if you've made a copy by doing:
```c++
TileLayer *shiftedStamp = shiftedCopies.value(stamp);
if (!shiftedStamp) {
    shiftedStamp = static_cast<TileLayer*>(stamp->clone());
    shiftedCopies.insert(stamp, shiftedStamp);
    // do the shifting
}
stamp = shiftedStamp;
```

Bdtrotte: Forgot to take into account the variations! Sorry, still learning tiled and qt as I go!

src/tiled/stampbrush.cpp
```diff
@@ -250,9 +250,17 @@ void StampBrush::endCapture()
 
     // Intersect with the layer and translate to layer coordinates
     QRect captured = capturedArea();
+
```
bjorn: No need to add this line.

src/tiled/stampbrush.cpp
```diff
@@ -250,9 +250,17 @@ void StampBrush::endCapture()
...
     captured &= QRect(tileLayer->x(), tileLayer->y(),
                       tileLayer->width(), tileLayer->height());
 
+    //gets if the relative stagger should be the same as the base layer
+    int staggerIndexOffSet;
+    if (tileLayer->map()->staggerAxis() == Map::StaggerX)
+        staggerIndexOffSet = captured.x() % 2;
+    else
+        staggerIndexOffSet = captured.y() % 2;
```
bjorn: Please move this bit of code down to after `if (captured.isValid()) {`, since you don't need the variable yet here.

Bdtrotte:
Thank you for being so patient as I get used to the style guidelines, qt, and tiled in general!

bjorn:
> Thank you for being so patient as I get used to the style guidelines, qt, and tiled in general!

No problem! You also need to get used to my rigorous review process, and to look carefully at your code to see if it could be written better, shorter, faster, etc. :-)

Now that we got this in, it occurred to me that of course it could probably have been done without copying and modifying the stamp at all. Instead, there could be some kind of specialized version of `TileLayer::merge` that can apply the shifting as it merges. I'll leave that as a possible exercise to explore later...

Bdtrotte:
Rigorous review is great! Lots of good feed back that made it a lot better.

>Instead, there could be some kind of specialized version of TileLayer::merge that can apply the shifting as it merges. I'll leave that as a possible exercise to explore later...

Haha, probably right! Ill take a breather to focus on rotation for a bit. Which speaking of, passing the stagger index to the stamp's map solved a lot of the problems I was having with rotation, and it now works perfectly! There was, though, some oddities with flipping (which I think is just due to that process not accounting for the stagger) which I need to work out. As well go over the code with a very fine comb to get through review a little easier :).

PR #1499, opened 2017-03-19
Updated French translation.

Bertram25:
Sorry for being late. But here is the update. Cheers,


bjorn:
Better late than never! Pushed to `0.18` as d34d9484a5eceb664e9674f923b990f0cca1956a, thanks!

PR #1500, opened 2017-03-19
Zoom default without Ctrl

cristyhj:
Here are the changes I have made to Tiled! I already wrote on the forum!


src/tiled/mainwindow.cpp
```diff
@@ -316,6 +316,7 @@ MainWindow::MainWindow(QWidget *parent, Qt::WindowFlags flags)
     connect(mUi->actionZoomOut, SIGNAL(triggered()), SLOT(zoomOut()));
     connect(mUi->actionZoomNormal, SIGNAL(triggered()), SLOT(zoomNormal()));
     connect(mUi->actionFullScreen, &QAction::toggled, this, &MainWindow::setFullScreen);
+    connect(mUi->actionZoom_With_Ctrl, SIGNAL(toggled(bool)),preferences, SLOT(setZoomPref(bool)));
```
bjorn: Coding style:
* Please remove the underscores, as done for the other actions (you can edit this name in the Design mode).
* Please prefer the new syntax for signal/slot connections (see line above).

src/tiled/preferences.cpp
```diff
@@ -566,6 +567,14 @@ void Preferences::setPluginEnabled(const QString &fileName, bool enabled)
     mSettings->setValue(QLatin1String("Plugins/Enabled"), enabledPlugins);
 }
 
+void Preferences::setZoomPref(bool mode){
+    mZoomPref = mode;
+}
+
+bool Preferences::getZoomPref(){
```
bjorn: Coding style:
* Getters do not have a "get" prefix, and should be marked `const`.
* The `{` for function bodies goes onto its own line.

src/tiled/preferences.cpp
```diff
@@ -566,6 +567,14 @@ void Preferences::setPluginEnabled(const QString &fileName, bool enabled)
     mSettings->setValue(QLatin1String("Plugins/Enabled"), enabledPlugins);
 }
 
+void Preferences::setZoomPref(bool mode){
+    mZoomPref = mode;
```
bjorn: `mZoomPref`? Please use a more descriptive name like `mWheelZoomsByDefault`.

Please make sure the setting is persistent (see other settings for how to interact with the `mSettings` member).

src/tiled/mainwindow.ui
```diff
@@ -591,6 +592,14 @@
     <string>Edit Commands...</string>
    </property>
   </action>
+  <action name="actionZoom_With_Ctrl">
+   <property name="checkable">
+    <bool>true</bool>
+   </property>
+   <property name="text">
+    <string>Zoom With Ctrl</string>
```
bjorn: I think "Mouse wheel zooms by default" is more descriptive (I cheated by looking up how they call this in Inkscape).

src/tiled/mainwindow.ui
```diff
@@ -156,6 +156,7 @@
     <addaction name="actionZoomIn"/>
     <addaction name="actionZoomOut"/>
     <addaction name="actionZoomNormal"/>
+    <addaction name="actionZoom_With_Ctrl"/>
```
bjorn: It's rare that a user will change this setting, so please don't add it to the menu. Add the checkbox to the Preferences dialog instead.

src/tiled/mapview.cpp
```diff
@@ -214,8 +214,8 @@ void MapView::wheelEvent(QWheelEvent *event)
     auto *hBar = static_cast<FlexibleScrollBar*>(horizontalScrollBar());
     auto *vBar = static_cast<FlexibleScrollBar*>(verticalScrollBar());
 
-    if (event->modifiers() & Qt::ControlModifier
-        && event->orientation() == Qt::Vertical)
+    if ((event->modifiers() & Qt::ControlModifier && event->orientation() == Qt::Vertical)
+        ^ !Preferences::instance()->getZoomPref())
```
bjorn: This check can't be right because the `event->orientation() == Qt::Vertical` check is necessary regardless of whether zooming is the default.

src/tiled/mapview.cpp
```diff
@@ -233,34 +233,35 @@ void MapView::wheelEvent(QWheelEvent *event)
         setTransformationAnchor(QGraphicsView::AnchorViewCenter);
         return;
     }
-
```
bjorn: Please don't remove the empty line and don't turn the rest of the function into an `else` block. It's not necessary because of the `return` above.

PR #1501, opened 2017-03-19
small changes to layer

ApertureScienceInnovators:


ApertureScienceInnovators:
Disregard; intended to merge this into my own master.

bjorn:
@ApertureScienceInnovators I am curious though, why would you add with/height back to Layer? What are your plans?

ApertureScienceInnovators:
@bjorn I am working with Ken VanDine to replicate the functionality [here](https://github.com/Bacon2D/Bacon2D/pull/125/) while still being able to use a nearly verbatim version of tiled.  We have tiled set up as a submodule within Bacon2D and have renamed the classes in Bacon2D to avoid the namespace conflict (instead of the more error prone renaming the tiled classes) but this was a change that (at first glance) seemed unavoidable.

ApertureScienceInnovators:
(for reference, the modified version of Bacon2D I'm working on is [here](https://github.com/arcrowel/Bacon2D))

bjorn:
@ApertureScienceInnovators I do not understand the namespace conflict. Are you saying that qmake can't handle two files with the same name in the same project, even if they are in different directories? Though even if so, that should be easy to fix by building libtiled as a separate project (a static library) and linking to it. Please don't resolve to hacks to get around a hacky build process.

ApertureScienceInnovators:
@bjorn That is what I was led to believe by Ken and by the conversation on the pull request.  I've never used Qt before so I've followed Ken's suggested solution (rename the files/classes to avoid namespace conflicts).
I will look into the static building; thanks!

bjorn:
@ApertureScienceInnovators Since in qmake most things are complicated, as is the including of internal static libraries in a project and linking correctly to them on various platforms, I would suggest to use the Qt Creator wizard to set up all the magic: http://doc.qt.io/qtcreator/creator-project-qmake-libraries.html

I didn't immediately realize this before since I'm mostly using Qbs these days, which makes this stuff so much easier.

Issue #1489, opened 2017-03-15
Make the object selection rectangle more visible


bjorn:
The selected objects are currently highlighted with a thin dashed black line. This doesn't work on dark backgrounds, and it also works very bad in combination with the "highlight current layer" feature.

![selection_153](https://cloud.githubusercontent.com/assets/531764/23952885/3015c3e4-0993-11e7-89f1-442b2315aaf9.png)

Quiz: which objects are selected?

Based on [feedback](http://discourse.mapeditor.org/t/feedback-after-a-couple-months-with-tiled/2090) from @eevee.

ketanhwr:
Answer to quiz: 
![Answer](https://image.ibb.co/ncjRqa/quizlol.png "Answer")

Do I get a cookie? :joy:

PR #1503, opened 2017-03-21
Implemented marching ants effect on selected objects(#1489)

thabetx:
Used a QTimer to repeatedly redraw the marquee, upon every redraw, A solid white line is drawn then black dashes are drawn above it shifted by the offset to simulate the marching ants effect.

Currently when selecting 100 objects simultaneously there was a slight increase in CPU usage to about 7-8%.

Redraws happen every 100ms, and offset increases by one, this numbers were chosen arbitrarily to look ok, but needs feedback.

Here is how it looks:
marchingants

### DIFF
--- a/src/tiled/objectselectionitem.cpp
+++ b/src/tiled/objectselectionitem.cpp
@@ -29,9 +29,9 @@
 #include "objectgroup.h"
 #include "preferences.h"
 #include "tile.h"
-#include <QTimer>
 
 #include <QGuiApplication>
+#include <QTimerEvent>
 
 namespace Tiled {
 namespace Internal {
@@ -128,17 +128,14 @@ static Preferences::ObjectLabelVisiblity objectLabelVisibility()
 }
 
 
-// MapObjectOuline inherits from QObject to use timer signals
-class MapObjectOutline : public QObject , public QGraphicsItem
+class MapObjectOutline : public QGraphicsObject
 {
 public:
     MapObjectOutline(MapObject *object, QGraphicsItem *parent = nullptr)
-        : QGraphicsItem(parent)
+        : QGraphicsObject(parent)
         , mObject(object)
     {
         setZValue(1); // makes sure outlines are above labels
-        connect(updateTimer, &QTimer::timeout, this, [this] {this->update();});
-        updateTimer->start(100);
     }
 
     void syncWithMapObject(MapRenderer *renderer);
@@ -148,14 +145,16 @@ public:
                const QStyleOptionGraphicsItem *,
                QWidget *) override;
 
+protected:
+    void timerEvent(QTimerEvent *event) override;
+
 private:
     QRectF mBoundingRect;
     MapObject *mObject;
 
-    // Timer is used to call update function repeatedly to animate selection
-    QTimer *updateTimer = new QTimer();
-    // Offset simulates the marching ants effect
-    int offset=0;
+    // Marching ants effect
+    int mUpdateTimer = startTimer(250);
+    int offset = 0;
 };
 
 void MapObjectOutline::syncWithMapObject(MapRenderer *renderer)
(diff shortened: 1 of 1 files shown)


src/tiled/objectselectionitem.cpp
```diff
@@ -127,14 +128,17 @@ static Preferences::ObjectLabelVisiblity objectLabelVisibility()
 }
 
 
-class MapObjectOutline : public QGraphicsItem
+// MapObjectOuline inherits from QObject to use timer signals
+class MapObjectOutline : public QObject , public QGraphicsItem
```
bjorn: You could just drive from `QGraphicsObject`.

src/tiled/objectselectionitem.cpp
```diff
@@ -147,6 +151,11 @@ class MapObjectOutline : public QGraphicsItem
 private:
     QRectF mBoundingRect;
     MapObject *mObject;
+
+    // Timer is used to call update function repeatedly to animate selection
+    QTimer *updateTimer = new QTimer();
```
bjorn: Since you're a `QObject`, please instead use `startTimer` and override `timerEvent`, so this can become `int mUpdateTimer;`.

src/tiled/objectselectionitem.cpp
```diff
@@ -127,14 +128,17 @@ static Preferences::ObjectLabelVisiblity objectLabelVisibility()
...
+class MapObjectOutline : public QObject , public QGraphicsItem
 {
 public:
     MapObjectOutline(MapObject *object, QGraphicsItem *parent = nullptr)
         : QGraphicsItem(parent)
         , mObject(object)
     {
         setZValue(1); // makes sure outlines are above labels
+        connect(updateTimer, &QTimer::timeout, this, [this] {this->update();});
+        updateTimer->start(100);
```
bjorn: I think 250ms works better.

Actually using marching ants is more common in graphics applications like GIMP and Photoshop, where the selection area is important to be aware of. In Inkscape there is no animation, for example. However, I think as long as it is at a slow pace, it's a nice thing to add.

src/tiled/objectselectionitem.cpp
```diff
@@ -147,6 +151,11 @@ class MapObjectOutline : public QGraphicsItem
 private:
     QRectF mBoundingRect;
     MapObject *mObject;
+
+    // Timer is used to call update function repeatedly to animate selection
+    QTimer *updateTimer = new QTimer();
+    // Offset simulates the marching ants effect
+    int offset=0;
```
bjorn: Coding style: please put spaces around the `=`.

Also, this would actually be the first time initialization is done at declaration time in Tiled's source code. But that is no problem since the Windows, macOS and Linux builds have all gone through fine.

src/tiled/objectselectionitem.cpp
```diff
@@ -184,17 +193,35 @@ void MapObjectOutline::paint(QPainter *painter,
...
+    dashPen.setStyle(Qt::DashLine);
+    dashPen.setDashOffset(qMax(qreal(0), x())+offset);
     painter->setPen(dashPen);
     painter->drawLines(horizontal, 2);
 
-    dashPen.setDashOffset(qMax(qreal(0), y()));
+    dashPen.setColor(Qt::white);
+    dashPen.setStyle(Qt::SolidLine);
     painter->setPen(dashPen);
     painter->drawLines(vertical, 2);
```
bjorn: Drawing the white vertical lines together with the white horizontal lines saves you from having to do two pen adjustments.

bjorn:
First fan of your feature on Patreon: https://www.patreon.com/posts/pull-requests-8584365

> **Mark Lindhout**
I _love_ the marching ants effect. Finally!

:-)

thabetx:
That's amazing, glad to help :smile: .

PR #1504, opened 2017-03-22
Test branch

jnermano:
Here is one of the ways I have found to approach this issue.

### DIFF
--- a/src/tiled/mapview.cpp
+++ b/src/tiled/mapview.cpp
@@ -20,8 +20,6 @@
 
 #include "mapview.h"
 
-#include <iostream>
-
 #include "flexiblescrollbar.h"
 #include "mapscene.h"
 #include "preferences.h"
--- a/src/tiled/mapview.cpp
+++ b/src/tiled/mapview.cpp
@@ -20,6 +20,8 @@
 
 #include "mapview.h"
 
+#include <iostream>
+
 #include "flexiblescrollbar.h"
 #include "mapscene.h"
 #include "preferences.h"
@@ -287,6 +289,7 @@ void MapView::mouseReleaseEvent(QMouseEvent *event)
         return;
     }
 
+
     QGraphicsView::mouseReleaseEvent(event);
 }
 
@@ -310,6 +313,42 @@ void MapView::mouseMoveEvent(QMouseEvent *event)
         mLastMousePos = event->globalPos();
         return;
     }
+
+    if (event->buttons() & Qt::LeftButton){
+        auto *hBar = static_cast<FlexibleScrollBar*>(horizontalScrollBar());
+        auto *vBar = static_cast<FlexibleScrollBar*>(verticalScrollBar());
+
+        QPoint pos = event->pos();
+        QRectF mRect = viewport()->rect();
+
+        int horizontalValue = 0,  verticalValue = 0;
+
+        if ( (pos.x() + 20) > mRect.width() ) {
+            horizontalValue = hBar->value() + 20;
+        }
+
+        if ( (pos.x() - 20) < 0 ) {
+            horizontalValue = hBar->value() + (hBar->value() <= 0? 0 : -20);
+        }
+
+        if ( (pos.y() + 20) > mRect.height() ) {
+            verticalValue = vBar->value() + 20;
+        }
+
+        if ( (pos.y() - 20) < 0 ) {
+            verticalValue = vBar->value() + (vBar->value() <= 0 ? 0 : -20);
+        }
+
+        if(horizontalValue){
+            hBar->forceSetValue(horizontalValue);
+        }
+
+        if(verticalValue){
+            vBar->forceSetValue(verticalValue);
+        }
+
+    }
+
 
     QGraphicsView::mouseMoveEvent(event);
     mLastMousePos = event->globalPos();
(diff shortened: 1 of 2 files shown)


src/tiled/mapview.cpp
```diff
@@ -287,6 +287,7 @@ void MapView::mouseReleaseEvent(QMouseEvent *event)
         return;
     }
 
+
```
ketanhwr: Why have you added this new line?

src/tiled/mapview.cpp
```diff
@@ -311,6 +312,42 @@ void MapView::mouseMoveEvent(QMouseEvent *event)
         return;
     }
 
+    if (event->buttons() & Qt::LeftButton){
```
ketanhwr: Coding style:
```
if (event->buttons() & Qt::LeftButton) {
```

src/tiled/mapview.cpp
```diff
@@ -311,6 +312,42 @@ void MapView::mouseMoveEvent(QMouseEvent *event)
...
+        auto *hBar = static_cast<FlexibleScrollBar*>(horizontalScrollBar());
+        auto *vBar = static_cast<FlexibleScrollBar*>(verticalScrollBar());
+
+        QPoint pos = event->pos();
+        QRectF mRect = viewport()->rect();
+
+        int horizontalValue = 0,  verticalValue = 0;
+
+        if ( (pos.x() + 20) > mRect.width() ) {
```
ketanhwr: Why add unncecessary spaces here?

src/tiled/mapview.cpp
```diff
@@ -311,6 +312,42 @@ void MapView::mouseMoveEvent(QMouseEvent *event)
...
+
+        if ( (pos.y() + 20) > mRect.height() ) {
+            verticalValue = vBar->value() + 20;
+        }
+
+        if ( (pos.y() - 20) < 0 ) {
+            verticalValue = vBar->value() + (vBar->value() <= 0 ? 0 : -20);
+        }
+
+        if(horizontalValue){
```
ketanhwr: Coding style:
```
if (horizontalValue) {
```

src/tiled/mapview.cpp
```diff
@@ -311,6 +312,42 @@ void MapView::mouseMoveEvent(QMouseEvent *event)
...
+            hBar->forceSetValue(horizontalValue);
+        }
+
+        if(verticalValue){
```
ketanhwr: Here as well.

src/tiled/newtilesetdialog.h
```diff
@@ -68,6 +68,8 @@ private slots:
     void pickColorFromImage();
     void colorSelected(QColor);
 
+    void on_buttonBox_accepted();
```
ketanhwr: Why have you added this line?

ketanhwr: The build is failing probably due to this line. Remove it.

jnermano:
@ketanhwr thank you for your feedback!

src/tiled/mapview.cpp
```diff
@@ -311,6 +311,42 @@ void MapView::mouseMoveEvent(QMouseEvent *event)
...
+
+        if ((pos.y() + 20) > mRect.height()) {
+            verticalValue = vBar->value() + 20;
+        }
+
+        if ((pos.y() - 20) < 0) {
+            verticalValue = vBar->value() + (vBar->value() <= 0 ? 0 : -20);
+        }
+
+        if(horizontalValue) {
```
ketanhwr: You missed this one: `if (`

src/tiled/mapview.cpp
```diff
@@ -311,6 +311,42 @@ void MapView::mouseMoveEvent(QMouseEvent *event)
...
+            hBar->forceSetValue(horizontalValue);
+        }
+
+        if(verticalValue) {
```
ketanhwr: And here too.

jnermano:
> I would prefer it, if you drag the cursor off the edge and keep the mouse still and then the view should pan automatically, which, in this case does not.

I agree with you. I will continue to look at it.

src/tiled/mapview.cpp
```diff
@@ -311,6 +311,46 @@ void MapView::mouseMoveEvent(QMouseEvent *event)
...
+        auto *vBar = static_cast<FlexibleScrollBar*>(verticalScrollBar());
+
+        QPoint pos = event->pos();
+        QRectF mRect = viewport()->rect();
+
+        int horizontalValue = 0,  verticalValue = 0;
+
+        if ((pos.x() + 5) > mRect.width()) {
+            horizontalValue = hBar->value() + 5;
+            cursor().setPos(mLastMousePos - *(new QPoint(5, 0)));
```
ketanhwr: I don't think it's a good idea to change the cursor position. I'm sure there's a better solution.

bjorn:
I agree with @ketanhwr that it's not a good idea to change the cursor position. Please check out how this works for example by dragging off the view in Firefox or Chromium. They both use a timer and adjust the scroll speed by the distance that the mouse moved out of the view.

@jnermano Do you plan to continue work on this patch?

jnermano:
I've tested some other things, like change values of the cursor position (1 instead of 5), but i'm gonna take a look to your suggestions (both of you).

Later I will let you know of my finds.

Yes, I'm still interested @bjorn .

PR #1505, opened 2017-03-22
Add libtiled/libtiled-static.pri

arcrowel:
This pull request seeks to add libtiled-static.pri to the src/libtiled directory.  While this compilation file is being added so a particular project (Bacon2D), can add tiled as a submodule, it should also allow for other projects to do the same due to the presence of this file and thus having compilation rules for the files in the libtiled directory.


src/libtiled/libtiled-static.pri
```diff
@@ -0,0 +1,64 @@
...
+    $$PWD/tilesetformat.cpp \
+    $$PWD/varianttomapconverter.cpp \
+    $$PWD/layer.cpp \
+    $$PWD/imagelayer.cpp \
+    $$PWD/imagereference.cpp
+
+DEFINES += TILED_LIBRARY
+
+#OBJECTS_DIR = tmp/tiled
+#MOC_DIR = tmp/tiled
```
bjorn: What's the point of these lines?

src/libtiled/libtiled-static.pri
```diff
@@ -0,0 +1,64 @@
...
+    $$PWD/properties.cpp \
+    $$PWD/staggeredrenderer.cpp \
+    $$PWD/tile.cpp \
+    $$PWD/tilelayer.cpp \
+    $$PWD/tileset.cpp \
+    $$PWD/tilesetformat.cpp \
+    $$PWD/varianttomapconverter.cpp \
+    $$PWD/layer.cpp \
+    $$PWD/imagelayer.cpp \
+    $$PWD/imagereference.cpp
```
bjorn: Please move the `HEADERS` and `SOURCES` into their own `libtiled-src.pri` file, so that it can be shared between the `libtiled-static.pri` and `libtiled.pro`.

src/libtiled/libtiled-static.pri
arcrowel: Do you mean only the commented out lines (which can be safely removed, and will be when we update our code), or also the "DEFINES += TILED_LIBRARY" line?

bjorn: I meant only the commented out lines. I guess the `DEFINES += TILED_LIBRARY` is necessary because you're not actually building libtiled as a static library (which I think would require a `pro` file as well), but rather including its sources directly. Then you indeed need to tell the library to export its symbols rather than try to import them (as is the case if you don't define TILED_LIBRARY, like when compiling a project that will be linked to the library).

PR #1509, opened 2017-03-22
Added layer properties: horizontal/vertical repeat, move speed, width, height

Zax37:
New layer properties handled(loading, saving, undo/redo...).
Changing layer dimensions directly plugged into changing width/height attributes.
Partially implemented rendering of repeated layers.

Known bugs/issues to solve:
- moving "camera" makes parts of repeated fragments of plane to disappear
- tile preview(eg. Stamp Brush) causes same thing to happen
- width/height attribute field loses focus upon clicking (+/-) buttons
- rescaling map changes all layers' dimensions(which should probably be optional)

TODO:
- use move speed as multiplier in rendering
- option to show layer margins?
- using tools on repeated layers, outside their bounds? (operating on repeated tiles affects te original ones)

### DIFF
--- a/src/libtiled/orthogonalrenderer.cpp
+++ b/src/libtiled/orthogonalrenderer.cpp
@@ -229,6 +229,9 @@ void OrthogonalRenderer::drawTileLayer(QPainter *painter,
     int endX = layer->width() - 1;
     int endY = layer->height() - 1;
 
+    if(layer->repeatedX()) endX = map()->width();
+    if(layer->repeatedY()) endY = map()->height();
+
     if (!exposed.isNull()) {
         QMargins drawMargins = layer->drawMargins();
         drawMargins.setTop(drawMargins.top() - tileHeight);
@@ -241,10 +244,14 @@ void OrthogonalRenderer::drawTileLayer(QPainter *painter,
 
         rect.translate(-layerPos);
 
-        startX = qMax(qFloor(rect.x() / tileWidth), 0);
-        startY = qMax(qFloor(rect.y() / tileHeight), 0);
-        endX = qMin(qCeil(rect.right()) / tileWidth, endX);
-        endY = qMin(qCeil(rect.bottom()) / tileHeight, endY);
+        if(!layer->repeatedX()){
+            startX = qMax(qFloor(rect.x() / tileWidth), 0);
+            endX = qMin(qCeil(rect.right()) / tileWidth, endX);
+        }
+        if(!layer->repeatedY()){
+            startY = qMax(qFloor(rect.y() / tileHeight), 0);
+            endY = qMin(qCeil(rect.bottom()) / tileHeight, endY);
+        }
     }
 
     // Return immediately when there is nothing to draw
--- a/src/libtiled/tilelayer.cpp
+++ b/src/libtiled/tilelayer.cpp
@@ -112,6 +112,8 @@ QRegion TileLayer::region(std::function<bool (const Cell &)> condition) const
  */
 void Tiled::TileLayer::setCell(int x, int y, const Cell &cell)
 {
+    if(repeatedX()){ x %= width(); if(x<0) x+= width(); }
+    if(repeatedY()){ y %= height(); if(y<0) y+= height(); }
     Q_ASSERT(contains(x, y));
 
     Cell &existingCell = mGrid[x + y * mWidth];
--- a/src/libtiled/tilelayer.h
+++ b/src/libtiled/tilelayer.h
@@ -380,6 +380,8 @@ inline QRegion TileLayer::region() const
  */
 inline const Cell &TileLayer::cellAt(int x, int y) const
 {
+    if(repeatedX()){ x %= width(); if(x<0) x+= width(); }
+    if(repeatedY()){ y %= height(); if(x<0) y+= height(); }
     Q_ASSERT(contains(x, y));
     return mGrid.at(x + y * mWidth);
 }
--- a/src/libtiled/layer.cpp
+++ b/src/libtiled/layer.cpp
@@ -151,6 +151,9 @@ QPointF Layer::totalOffset() const
  */
 Layer *Layer::initializeClone(Layer *clone) const
 {
+    clone->mRepeatedY = mRepeatedY;
+    clone->mRepeatedX = mRepeatedX;
+    clone->mMoveSpeed = mMoveSpeed;
     clone->mOffset = mOffset;
     clone->mOpacity = mOpacity;
     clone->mVisible = mVisible;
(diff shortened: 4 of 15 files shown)


src/libtiled/orthogonalrenderer.cpp
```diff
@@ -241,10 +244,14 @@ void OrthogonalRenderer::drawTileLayer(QPainter *painter,
 
         rect.translate(-layerPos);
 
-        startX = qMax(qFloor(rect.x() / tileWidth), 0);
-        startY = qMax(qFloor(rect.y() / tileHeight), 0);
-        endX = qMin(qCeil(rect.right()) / tileWidth, endX);
-        endY = qMin(qCeil(rect.bottom()) / tileHeight, endY);
+        if(!layer->repeatedX()){
```
ketanhwr: Add spaces: `if (!layer->repeatedX()) {`
Do the same for the next construct.

src/libtiled/tilelayer.cpp
```diff
@@ -112,6 +112,8 @@ QRegion TileLayer::region(std::function<bool (const Cell &)> condition) const
  */
 void Tiled::TileLayer::setCell(int x, int y, const Cell &cell)
 {
+    if(repeatedX()){ x %= width(); if(x<0) x+= width(); }
```
ketanhwr: You can format it like this, makes code more readable :-)
```
if (repeatedX()) {
   x %= width();
   if (x < 0)
      x += width();
}
```

src/libtiled/tilelayer.cpp
```diff
@@ -112,6 +112,8 @@ QRegion TileLayer::region(std::function<bool (const Cell &)> condition) const
...
+    if(repeatedY()){ y %= height(); if(y<0) y+= height(); }
```
ketanhwr: Here too.

src/libtiled/tilelayer.h
```diff
@@ -380,6 +380,8 @@ inline QRegion TileLayer::region() const
  */
 inline const Cell &TileLayer::cellAt(int x, int y) const
 {
+    if(repeatedX()){ x %= width(); if(x<0) x+= width(); }
```
ketanhwr: Same format here and the next line as well.

Zax37:
Thanks for feedback @ketanhwr, I've changed style according to Your suggestions.

bjorn:
Hey @Zax37, sorry for looking at this a little late but I was really overwhelmed with pull requests lately. I understand the patch is far from done, and I hope you will continue it because it's a great and much requested feature! However, I would focus on one thing at a time. So I would forget about the movement speed at first and leave that for another patch.

Similarly, I see you've added the repeat flags to the `Layer` class. But I think the repeating implementation and editing behavior is going to be vastly different for tile layers, object layers and image layers, and I don't think we should have those properties on group layers. So for now I would focus on adding it to tile layers only, again to not try to solve all problems at once.

Finally, supporting this feature for tile layers necessarily means we need to support layers that have a different size than the map. I noticed you've made the Width and Height properties editable in the properties view, but I think we should really use the `ResizeHelper` class for this, which is the dialog that pops up when you do _Map -> Resize Map_. Simply changing the size may not leave the right part of the layer intact, and doing many incremental changes to the width or height, like the Properties view encourages, would consume a lot of undo memory (though we could of course optimize this quite a bit).

> option to show layer margins?

Since this will be quit essential, I think it doesn't even need to be an option. Any layer not matching the map size should render a margin when selected.

> using tools on repeated layers, outside their bounds? (operating on repeated tiles affects te original ones)

I see you've tried to implement this already in the `setCell` method, but I think that's the wrong level at which to do this. This is because the code outside of `TileLayer` will need to be aware of affected regions anyway. We could try to do this in the `TilePainter`, though for sure the `StampBrush` and other tools will also need to do something to have their previews working as expected. In general, I think drawing outside the bounds should also be its own patch because we'll have quite a few changes already and we can work without this feature too.

> moving "camera" makes parts of repeated fragments of plane to disappear

That is because of culling. The QGraphicsView won't paint your item when its boundaries are entirely outside of the view. To fix it you need to adjust the bounding rect calculations in `TileLayerItem::syncWithTileLayer`.

> width/height attribute field loses focus upon clicking (+/-) buttons

Yeah, strange issue. I can't immediately say what could be causing that.

> rescaling map changes all layers' dimensions(which should probably be optional)

Yep, by default it should probably not resize layers that didn't previously match the map size. We can also draw from the Canvas Size dialog in GIMP, which has a combo box for this with several options. See also the TODO comment in `MapDocument::resizeMap`.

There are still a lot of code style issues I could comment on, but I'll leave that for when you get further with the patch. Please just try to make an effort to use the same style as the existing code.

One problem I noticed is that you forgot to initialize `mRepeatedX` and `mRepeatedX`, resulting in some random flashing (at least in OpenGL mode) when the preview gets repeated.

So again, I'm looking forward to see how you will continue with this patch!

Zax37:
Thanks for helping me out on this @bjorn,
> To fix it you need to adjust the bounding rect calculations in TileLayerItem::syncWithTileLayer.
I was looking for this little bastard for too long ;) Changing boundingRect fixes most of it, but there are still some artifacts on the edges on borders as it tries to render one tile more than it should... I'll play around with this and apply other suggestions as soon as I find some time to work on this.

ottworks:
Please finish this pull!

bjorn:
@SasikaAT I've implemented the "move speed" feature as "parallax scroll factor" in [Tiled 1.5](https://www.mapeditor.org/2021/03/23/tiled-1-5-0-released.html). There is no repetition feature yet.

This pull request was never in a state where it could be merged, and it is unlikely @Zax37 would still find time to complete it. I've kept it open because indeed the repeat feature would still be nice to add.

SasikaAT:
> 
> 
> @SasikaAT I've implemented the "move speed" feature as "parallax scroll factor" in [Tiled 1.5](https://www.mapeditor.org/2021/03/23/tiled-1-5-0-released.html). There is no repetition feature yet.
> 
> This pull request was never in a state where it could be merged, and it is unlikely @Zax37 would still find time to complete it. I've kept it open because indeed the repeat feature would still be nice to add.

I see. I didn't interpret the move speed as the parallax factor but more as a constantly moving background. Like several layers of clouds could be constantly moving at different speeds in the background. But thinking once more about that there is no real necessity to have that as a live preview while building the level. Nice to have for a quick preview to set up the right speed values, but would be distracting while building the level.

The hor/ver repeat of an image-layer is something I'm desperately missing right now for endlessly looping textures. I have the feeling that feature might be drowned out when being combined with other not directly related features inside this issue. 👀

PR #1510, opened 2017-03-24
Shortcut handler for changing toolTip text on Actions on toolbars

laspencer91:
This is the shortcuthandler addition. This is a Singleton class implimentation so only one can exist. 

This class houses a list of QWidgets and a list of QActions. These can be added one by one or an existing list can be appended to them.

UpdateActionShortcut methods is a string modifying method for the actions tool tips. This method is not complete, but I will be working on it.

Currently all actions are added at the end of the mainwindow constructor. These 4 lines, the include "", and the new class are the only changes.


laspencer91:
One thing I noticed in the debug commands when the program is ran. It adds actions labeled. "Add Property", "Remove Property", and "Rename Property" 3 different times. I do not believe they all reference the same actions because I have my code check to make sure a QAction doesn't already exist in the list before adding a new one. I haven't looked deep into this, so I do not know where it stems from yet, but there could be a simple explanation.

bjorn:
Hmm, there is just a merge commit in this pull request and no changes to any files. Are you sure you opened the pull request from the right branch?

laspencer91:
Oh shoot. I will fix it when I get home. I have actually added a basic implementation for a shortcut editor as well.

Issue #1476, opened 2017-03-07
Rotating of hexagonal tile stamps


bjorn:
Thanks to @nicolaichuk we now have 60-degree rotation steps when rotating hexagonal tiles (#1447).

However, when rotating a tile stamp with multiple tiles, their relative positions are still changed by 90 degrees rather than staying together rotated by 60 degrees. In addition, and this was already the case, the staggering is not taken into account.

This document should really help to get this implemented: http://www.redblobgames.com/grids/hexagons/#rotation

nicolaichuk:
For help  http://www.redblobgames.com/grids/hexagons

Bdtrotte:
Im taking a look at this... Does anyone have a good hex tileset so I could get a better picture of how all this works?
Thanks,
Ben

bjorn:
@Bdtrotte This feature could be implemented while testing with the `hexagonal-mini.tmx` example shipping with Tiled. However, you'll need to disable the actual rotation of the tile graphics since that doesn't work for that tileset. In fact, since this will be the case for many hex tilesets, this would be a useful option to have for the user as well. I'm just not entirely sure yet at which level such an option should exist.

Thanks for looking into this!

Bdtrotte:
>However, you'll need to disable the actual rotation of the tile graphics since that doesn't work for that tileset. In fact, since this will be the case for many hex tilesets, this would be a useful option to have for the user as well.

I was putting some thought into this. It seems the only time one would want the individual hex tile graphics to rotate by 60 (where it would still line up with the gird) would be if the tiles are perfect hexagons. Maybe disable graphic rotation at 60 degree intervals automatically if they aren't perfect (within a margin of error). Then at 90 degree marks (180 and 360) it could flip them? But then this could be done by flipping vertically and horizontally anyways... Either way a manual override could be implemented.

Had some problems with qt which I am resolving now, and should at least have proper 60 stamp rotation by the end of today!

Ben

Bdtrotte:
Started approaching this, and It's ended up being a bigger task than expected. This weekend I should have it ready for initial review.

Do you think it would be useful to make a new file which contains useful tools for dealing with hexagonal stuff. Such as converting between different coordinate systems (staggered to cube), perform rotations, maybe a class definition to package these tools a bit nicer?

With this in mind is there anything I should know or be sure to do as I make a new file for this project, noting I'm new to qt, and open source contributing.

Thanks,
Ben

bjorn:
[...]

PR #1515, opened 2017-03-29
Staggered flipping now works as expected, as well, Hexagonal rotation now works in 60 degree intervals.

Bdtrotte:
Addresses #1476. Adds a hex class, which stores info of the cube coordinates of a hex tile, and implements methods for rotation, and converting to staggered coordinates.

### DIFF
--- a/src/libtiled/hex.cpp
+++ b/src/libtiled/hex.cpp
@@ -126,16 +126,6 @@ Hex Hex::operator -(Hex h) const
     return Hex(mX - h.x(), mY - h.y(), mZ - h.z());
 }
 
-Hex Hex::operator *(float f) const
-{
-    return Hex(mX * f, mY * f, mZ * f);
-}
-
-Hex Hex::operator /(float f) const
-{
-    return Hex(mX / f, mY / f, mZ / f);
-}
-
 Hex& Hex::operator +=(Hex h)
 {
     mX += h.x();
--- a/src/libtiled/hex.cpp
+++ b/src/libtiled/hex.cpp
@@ -128,12 +128,12 @@ Hex Hex::operator -(Hex h) const
 
 Hex Hex::operator *(float f) const
 {
-    return Hex(mX*f, mY*f, mZ*f);
+    return Hex(mX * f, mY * f, mZ * f);
 }
 
 Hex Hex::operator /(float f) const
 {
-    return Hex(mX/f, mY/f, mZ/f);
+    return Hex(mX / f, mY / f, mZ / f);
 }
 
 Hex& Hex::operator +=(Hex h)
--- a/src/libtiled/hex.cpp
+++ b/src/libtiled/hex.cpp
@@ -20,8 +20,7 @@
 
 #include "hex.h"
 
-namespace Tiled {
-namespace Internal {
+using namespace Tiled;
 
 Hex::Hex(int x, int y, int z):
     mX(x),
@@ -32,24 +31,31 @@ Hex::Hex(QPoint point,
          Map::StaggerIndex staggerIndex,
          Map::StaggerAxis staggerAxis)
 {
+    Hex(point.x(),point.y(),staggerIndex,staggerAxis);
+}
+
+Hex::Hex(int col, int row,
+    Map::StaggerIndex staggerIndex,
+    Map::StaggerAxis staggerAxis)
+{
     if (staggerAxis == Map::StaggerX) {
         if (staggerIndex == Map::StaggerEven) {
-            mX = point.x;
-            mZ = point.y - (point.x + (point.x & 1)) / 2;
+            mX = col;
+            mZ = row - (col + (col & 1)) / 2;
             mY = -mX - mZ;
         } else {
-            mX = point.x;
-            mZ = point.y - (point.x - (point.x & 1)) / 2;
+            mX = col;
+            mZ = row - (col - (col & 1)) / 2;
             mY = -mX - mZ;
         }
     } else {
         if (staggerIndex == Map::StaggerEven) {
-            mX = point.x - (point.y + (point.y & 1)) / 2;
-            mZ = point.y;
+            mX = col - (row + (row & 1)) / 2;
+            mZ = row;
             mY = -mX - mZ;
         } else {
-            mX = point.x - (point.y - (point.y & 1)) / 2;
-            mZ = point.y;
+            mX = col - (row - (row & 1)) / 2;
+            mZ = row;
             mY = -mX - mZ;
         }
     }
@@ -62,19 +68,19 @@ QPoint Hex::toStagger(Map::StaggerIndex staggerIndex,
 
     if (staggerAxis == Map::StaggerX) {
         if (staggerIndex == Map::StaggerEven) {
-            point.x = mX;
-            point.y = mZ + (mX + (mX & 1)) / 2;
+            point.setX(mX);
+            point.setY(mZ + (mX + (mX & 1)) / 2);
         } else {
-            point.x = mX;
-            point.y = mZ + (mX - (mX & 1)) / 2;
+            point.setX(mX);
+            point.setY(mZ + (mX - (mX & 1)) / 2);
         }
     } else {
         if (staggerIndex == Map::StaggerEven) {
-            point.x = mX + (mZ - (mZ & 1)) / 2;
-            point.y = mZ;
+            point.setX(mX + (mZ + (mZ & 1)) / 2);
+            point.setY(mZ);
         } else {
-            point.x = mX + (mZ + (mZ & 1)) / 2;
-            point.y = mZ;
+            point.setX(mX + (mZ - (mZ & 1)) / 2);
+            point.setY(mZ);
         }
     }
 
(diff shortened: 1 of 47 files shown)


Bdtrotte:
I think I've done something wrong, as it's showing commits far beyond the context of this pr. Ill work to fix this...

Bdtrotte:
Alright, should be up to date now

bjorn:
> Alright, should be up to date now

Well, you merged my branch with yours, but you should rather rebase your work and in the process you can remove all the commits that have already been merged (and squashed as cdc71f4bf06c2ce9fb754887a8756f3686b18ea2).

src/libtiled/tilelayer.h
```diff
@@ -266,7 +268,7 @@ class TILEDSHARED_EXPORT TileLayer : public Layer
      * are rotated within the layer, and the tiles themselves are rotated. The
      * dimensions of the tile layer are swapped.
      */
-    void rotateHexagonal(RotateDirection direction);
+    void rotateHexagonal(RotateDirection direction, Tiled::Map::StaggerIndex staggerIndex, Tiled::Map::StaggerAxis staggerAxis);
```
bjorn: You're in the `Tiled` namespace here, so you don't need the `Tiled::` qualification.

src/tiled/tilestamp.cpp
```diff
@@ -232,6 +232,22 @@ TileStamp TileStamp::flipped(FlipDirection direction) const
...
         TileLayer *layer = variation.tileLayer();
+
+        Map::StaggerIndex staggerIndex = variation.map->staggerIndex();
+        Map::StaggerAxis staggerAxis = variation.map->staggerAxis();
+
+        if (variation.map->orientation() == Map::Staggered || variation.map->orientation() == Map::Hexagonal) {
+            if (staggerAxis == Map::StaggerY)
+            {
+                if ((direction == FlipVertically && !(layer->height() & 1)) || direction == FlipHorizontally)
+                    variation.map->setStaggerIndex(static_cast<Map::StaggerIndex>((staggerIndex + 1) & 1));
```
bjorn: A little shorter would be:
```c++
variation.map->setStaggerIndex(static_cast<Map::StaggerIndex>(1 - staggerIndex));
```
Though for readability I might try to use:
```c++
static Map::StaggerIndex invert(Map::StaggerIndex staggerIndex)
{
    return static_cast<Map::StaggerIndex>(1 - staggerIndex);
}
```

src/tiled/tilestamp.cpp
```diff
@@ -232,6 +232,22 @@ TileStamp TileStamp::flipped(FlipDirection direction) const
 
     for (const TileStampVariation &variation : flipped.variations()) {
         TileLayer *layer = variation.tileLayer();
+
+        Map::StaggerIndex staggerIndex = variation.map->staggerIndex();
+        Map::StaggerAxis staggerAxis = variation.map->staggerAxis();
+
+        if (variation.map->orientation() == Map::Staggered || variation.map->orientation() == Map::Hexagonal) {
```
bjorn: Remember that we added `map->isStaggered()`.

src/libtiled/hex.cpp
```diff
@@ -0,0 +1,136 @@
...
+Hex::Hex(int x, int y, int z):
+    mX(x),
+    mY(y),
+    mZ(z) { }
+
+Hex::Hex(QPoint point,
+         Map::StaggerIndex staggerIndex,
+         Map::StaggerAxis staggerAxis)
+{
+    Hex(point.x(),point.y(),staggerIndex,staggerAxis);
```
Bdtrotte: Thanks for this. I was thinking logically what I was doing wasn't quite right, but thought I would get an error if it was wrong. But I see now exactly what it was doing. 

> instead you could also define a Hex::setStaggered(int col, int row, ...)

Good advice! Implementing this way.

src/tiled/tilestamp.cpp
Bdtrotte: I forgot to use the function I added... Resolved now!

Bdtrotte:
> Well, you merged my branch with yours, but you should rather rebase your work and in the process you can remove all the commits that have already been merged (and squashed as cdc71f4).

Would it be best to close this pr and start a new one with everything cleaned up? Or just continue working in this one?

Bdtrotte:
Updated all items from the review. I absolutely appreciate good style, just have to get used to being consistently good! As well constantly learning a lot about c++, so hopefully less coding errors as we go.

bjorn:
> Would it be best to close this pr and start a new one with everything cleaned up? Or just continue working in this one?

There is no reason to close this PR, please keep updating this one.

Bdtrotte:
I believe I have fixed all the style issues. Is there anything else you think should be changed within the code?

src/libtiled/tilelayer.cpp
```diff
@@ -360,16 +372,34 @@ void TileLayer::rotateHexagonal(RotateDirection direction)
...
 
     mWidth = newWidth;
     mHeight = newHeight;
     mGrid = newGrid;
+
+    QRect filledRect = region().boundingRect();
+
+    if (staggerAxis == Map::StaggerY) {
+        if (filledRect.y() & 1)
+            filledRect.adjust(0, -1, 0, 0);
```
bjorn: Would it be possible to change the stagger index instead of keeping an empty row of data around?

src/libtiled/tilelayer.cpp
```diff
@@ -312,12 +312,24 @@ void TileLayer::rotate(RotateDirection direction)
     mGrid = newGrid;
 }
 
-void TileLayer::rotateHexagonal(RotateDirection direction)
+void TileLayer::rotateHexagonal(RotateDirection direction, Map::StaggerIndex staggerIndex, Map::StaggerAxis staggerAxis)
 {
-    int newWidth = mHeight;
-    int newHeight = mWidth;
+    Hex botRight(mWidth, mHeight, staggerIndex, staggerAxis);
```
bjorn: Naming style: Please just use `bottomRight`.

src/libtiled/tilelayer.cpp
```diff
@@ -312,12 +312,24 @@ void TileLayer::rotate(RotateDirection direction)
...
+    Hex center(mWidth / 2, mHeight / 2, staggerIndex, staggerAxis);
+
+    botRight -= center;
+    topRight -= center;
+
+    botRight.rotate(RotateRight);
+    topRight.rotate(RotateRight);
+
+    int newWidth = topRight.toStaggered(staggerIndex, staggerAxis).x() * 2 + 2;
+    int newHeight = botRight.toStaggered(staggerIndex, staggerAxis).y() * 2 + 2;
```
bjorn: It is not immediately obvious to me why this works (like, why the new area is both minimal but also large enough). Could you share a few words about it?

src/libtiled/hex.h
```diff
@@ -0,0 +1,79 @@
...
+    void setZ(int z) { mZ = z; }
+
+    void setStaggered(int col, int row,
+                      Map::StaggerIndex staggerIndex,
+                      Map::StaggerAxis staggerAxis);
+
+    Hex operator + (Hex h) const;
+    Hex operator -(Hex h) const;
+    Hex operator *(float f) const;
+    Hex operator /(float f) const;
```
bjorn: You're not actually using the `*` and `/` operators and using them would risk violating the rule that hex cube coordinates always adhere to `x + y + z = 0` (due to integer truncation). I would suggest you leave those out.

src/libtiled/hex.h
```diff
@@ -0,0 +1,79 @@
...
+
+    void setX(int x) { mX = x; }
+    void setY(int y) { mY = y; }
+    void setZ(int z) { mZ = z; }
+
+    void setStaggered(int col, int row,
+                      Map::StaggerIndex staggerIndex,
+                      Map::StaggerAxis staggerAxis);
+
+    Hex operator + (Hex h) const;
```
bjorn: Stray space after the `+` (it'd be fine, but you don't have it on any of the other operators).

src/libtiled/hex.h
```diff
@@ -0,0 +1,79 @@
...
+
+#pragma once
+
+#include "tiled.h"
+#include "map.h"
+#include <QPoint>
+
+namespace Tiled {
+
+class Hex
```
bjorn: Please add a comment like:
```c++
/**
 * Represents a hex position in cube coordinates.
 */
```

src/libtiled/tilelayer.cpp
Bdtrotte: This was admittedly a bit of a lazy solution. Though I believe we don't have the inherent ability to change the stagger index from a TileLayer. The solution could be to omit the staggerIndex and staggerAxis parameters, and in place pass the map reference. I was thinking it would be preferable to keep them separate,  but I'll go ahead and implement that way if you think I'd be better.

src/libtiled/tilelayer.cpp
Bdtrotte: If a rectangle is rotated 60 degrees clockwise around the center, then the top right corner will be the furthest right point, the bottom left will be the furthest left by the same amount, and top left will be the highest, bottom right the lowest. Similar logic applies for anti-clockwise rotation, and the sizes will be the same, so one direction can be chosen, regardless of intended rotation direction. So we take the two corners, transform them around the center, rotate, and take the respective largest axis. This value is of the center, so it is doubled to get the maximum new width/height. The two is added because of potential integer rounding in regards to the center. 

Of course a problem is that sometimes the corners will be empty, so the enlargement is unnecessary. I don't believe there is an easy way to account for this, thus I enlarge, then trim it down later.

In short: This process produces the smallest rectangle which contains the original rotated by 60 degrees.

src/libtiled/tilelayer.h
```diff
@@ -266,7 +267,7 @@ class TILEDSHARED_EXPORT TileLayer : public Layer
      * are rotated within the layer, and the tiles themselves are rotated. The
      * dimensions of the tile layer are swapped.
```
bjorn: Please still adjust this comment, and make a note of the `map` parameter and that the stagger axis may be changed.

bjorn:
Many thanks for this great enhancement!

bjorn:
@Bdtrotte Please read up on Git, in particular how to update your fork without deleting and re-creating it. :-)

Bdtrotte:
Of course! Will do. It was just pretty messy, so what quicker way to clean up..!

PR #1516, opened 2017-03-30
Minimap in a resizing dialog

Acuion:
I think it is useful to allow a user to see what exactly is going to be removed while resizing and visually estimate the proportions.
 
- Rendering of a minimap is now located in a singleton, which can be accessed from both an editor and a resizing window. Maybe it is not a perfect architecture solution, but at lest it works. Waiting for a feedback regarding it.
- I think now there is no reason to scale the preview when the old rect goes outside the view, so a user can see a more detailed picture of a map.

### DIFF
--- a/src/tiled/minimaprenderer.h
+++ b/src/tiled/minimaprenderer.h
@@ -41,11 +41,14 @@ public:
 
     Q_DECLARE_FLAGS(MiniMapRenderFlags, MiniMapRenderFlag)
 
-    static MiniMapRenderer* miniMapRenderer();
+    static MiniMapRenderer* instance();
     void renderMinimapToImage(QImage& image, const MiniMapRenderFlags minimapRenderFlags) const;
     void setMapDocument(MapDocument* map);
 private:
     MiniMapRenderer();
+    ~MiniMapRenderer();
+    MiniMapRenderer(MiniMapRenderer const&) = delete;
+    MiniMapRenderer& operator= (MiniMapRenderer const&) = delete;
 
     MapDocument* mMapDocument;
 };
--- a/src/tiled/minimaprenderer.h
+++ b/src/tiled/minimaprenderer.h
@@ -19,7 +19,6 @@
  * this program. If not, see <http://www.gnu.org/licenses/>.
  */
 
-
 #pragma once
 
 #include "mapdocument.h"
--- a/src/tiled/minimaprenderer.h
+++ b/src/tiled/minimaprenderer.h
@@ -42,7 +42,7 @@ public:
 
     Q_DECLARE_FLAGS(RenderFlags, RenderFlag)
 
-    static MiniMapRenderer* instance();
+    static MiniMapRenderer& instance();
     void renderToImage(QImage& image, RenderFlags renderFlags) const;
     void setMapDocument(MapDocument* map);
 private:
--- a/src/tiled/minimaprenderer.h
+++ b/src/tiled/minimaprenderer.h
@@ -1,0 +1,57 @@
+/*
+ * minimaprenderer.h
+ * Copyright 2017, Yuriy Natarov <natarur@gmail.com>
+ * Based on a minimap by Christoph Schnackenberg and Thorbjørn Lindeijer
+ *
+ * This file is part of Tiled.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the Free
+ * Software Foundation; either version 2 of the License, or (at your option)
+ * any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+
+#pragma once
+
+#include "mapdocument.h"
+
+#include <QImage>
+
+namespace Tiled {
+namespace Internal {
+
+class MiniMapRenderer
+{
+public:
+    enum MiniMapRenderFlag {
+        DrawObjects             = 0x0001,
+        DrawTiles               = 0x0002,
+        DrawImages              = 0x0004,
+        IgnoreInvisibleLayer    = 0x0008,
+        DrawGrid                = 0x0010
+    };
+
+    Q_DECLARE_FLAGS(MiniMapRenderFlags, MiniMapRenderFlag)
+
+    static MiniMapRenderer* miniMapRenderer();
+    void renderMinimapToImage(QImage& image, const MiniMapRenderFlags minimapRenderFlags) const;
+    void setMapDocument(MapDocument* map);
+private:
+    MiniMapRenderer();
+
+    MapDocument* mMapDocument;
+};
+
+Q_DECLARE_OPERATORS_FOR_FLAGS(Tiled::Internal::MiniMapRenderer::MiniMapRenderFlags)
+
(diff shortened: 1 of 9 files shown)


bjorn:
This is something I have wanted to do for years, but somehow never prioritized it over other stuff. I shortly tried it out and it's really cool! I'll try to get around to a detailed code review in the evening, or at least in the coming days.

> I think now there is no reason to scale the preview when the old rect goes outside the view, so a user can see a more detailed picture of a map.

I'm not sure about this. It depends on whether it's more important for the user to see the remaining piece of the map, or what part he is cutting away. If the remaining piece is what he wants to focus on, it should be generally easier to select that area and use the "Crop to Selection" action. So in general I'm inclined to think it's better to make sure the whole map is visible, even if it will be smaller as a result.

Acuion:
- A preview behavior was strange if a new size was smaller than an old one. Now in that case it doesn't scale and a user can select a new rect by moving a frame.
- Objects were not offseted if the "Remove objects outside of the map" was not set

bjorn:
> Objects were not offseted if the "Remove objects outside of the map" was not set

Nice catch! Can you open a separate pull request for the this bug, to the `0.18` branch? It's a nice catch and if we do a 0.18.3 release I'd like to include it there.

> A preview behavior was strange if a new size was smaller than an old one. Now in that case it doesn't scale and a user can select a new rect by moving a frame.

I don't notice strange behavior currently in Tiled, so I assume this was about this pull request?

Acuion:
> I don't notice strange behavior currently in Tiled, so I assume this was about this pull request?
Oh, yes, it becomes strange for me when a minimap is inserted.
> Nice catch! Can you open a separate pull request for the this bug, to the 0.18 branch? It's a nice catch and if we do a 0.18.3 release I'd like to include it there.
Ok, is it still needed here?

Acuion:
And sorry for offtopic, but I still really need a feedback to my proposal. I hope mentors will have some time before the deadline.

src/tiled/minimaprenderer.h
```diff
@@ -0,0 +1,59 @@
...
+        DrawObjects             = 0x0001,
+        DrawTiles               = 0x0002,
+        DrawImages              = 0x0004,
+        IgnoreInvisibleLayer    = 0x0008,
+        DrawGrid                = 0x0010
+    };
+
+    Q_DECLARE_FLAGS(MiniMapRenderFlags, MiniMapRenderFlag)
+
+    static MiniMapRenderer* instance();
```
bjorn: I doesn't makes sense to have the `MiniMapRenderer` as a singleton, since it doesn't store any state that would be useful to keep around. Instead, please just allocate it on the stack where you need it, passing the `MapDocument*` to its constructor instead of having a `setMapDocument` method:
```c++
MiniMapRenderer renderer(mMapDocument);
renderer.renderToImage(image, flags);
```

src/tiled/minimaprenderer.h
```diff
@@ -0,0 +1,59 @@
...
+
+#include <QImage>
+
+namespace Tiled {
+namespace Internal {
+
+class MiniMapRenderer
+{
+public:
+    enum MiniMapRenderFlag {
```
bjorn: Please call it just `RenderFlags`.

src/tiled/minimaprenderer.h
```diff
@@ -0,0 +1,59 @@
...
+        DrawTiles               = 0x0002,
+        DrawImages              = 0x0004,
+        IgnoreInvisibleLayer    = 0x0008,
+        DrawGrid                = 0x0010
+    };
+
+    Q_DECLARE_FLAGS(MiniMapRenderFlags, MiniMapRenderFlag)
+
+    static MiniMapRenderer* instance();
+    void renderMinimapToImage(QImage& image, const MiniMapRenderFlags minimapRenderFlags) const;
```
bjorn: You can call this `renderToImage` since it's clear that it renders a mini-map.

You can remove the `const` on the `flags` parameter (please also shorten its name). It has no meaning since the parameter is passed by value.

src/tiled/minimaprenderer.h
```diff
@@ -0,0 +1,59 @@
+/*
+ * minimaprenderer.h
+ * Copyright 2017, Yuriy Natarov <natarur@gmail.com>
+ * Based on a minimap by Christoph Schnackenberg and Thorbjørn Lindeijer
```
bjorn: Please just copy the copyright lines from `minimap.h` instead of writing a custom message.

src/tiled/resizehelper.cpp
```diff
@@ -118,14 +130,17 @@ void ResizeHelper::paintEvent(QPaintEvent *)
...
 
-    painter.setPen(pen);
-    painter.setBrush(Qt::white);
     painter.setOpacity(0.5);
-    painter.drawRect(oldRect);
+    QImage minimap = QImage(oldRect.width() * mScale, oldRect.height() * mScale, QImage::Format_ARGB32_Premultiplied);
+    mMiniMapRenderer->renderMinimapToImage(minimap, MiniMapRenderer::DrawObjects
+                                           | MiniMapRenderer::DrawImages
+                                           | MiniMapRenderer::DrawTiles
+                                           | MiniMapRenderer::IgnoreInvisibleLayer);
```
bjorn: Rendering the mini-map on each paint event is too slow for large maps. Instead, please keep the image around and only repaint it when the scale changes.

Ideally, we'd do the rendering in a thread. I think we can easily do that here, because the resize helper dialog is modal so there can't be any changes to the underlying data in the meantime. Up to you whether you want to look into that or leave it for later.

src/tiled/minimaprenderer.h
Acuion: The problem is that `ResizeHelper` has no `mMapDocument`. That's why I've implemented the renderer as a singleton.
Or is it better to somehow pass it to `ResizeHelper`? I think it is more complicated than the singleton approach.

Acuion:
Done, but I personally think that my approach was better because currently, in case of a map shrinking, a lot of the space is unused and the minimap become smaller, so as a profit from its usage here.

src/tiled/resizehelper.cpp
```diff
@@ -179,7 +177,13 @@ void ResizeHelper::recalculateScale()
     // Pick the smallest scale
     const double scaleW = _size.width() / (double) width;
     const double scaleH = _size.height() / (double) height;
-    mScale = (scaleW < scaleH) ? scaleW : scaleH;
+    mScale = qMin(scaleW, scaleH);
```
bjorn: That the scale is recalculated doesn't mean it'll change, so please still compare with the previous scale and don't recreate the image when it isn't necessary.

Acuion:
> While you're right that the image is smaller that it'd need to be, I think it is confusing when you for example scale a 100x100 map to 50x200 map, that in one direction the mouse drags the existing map and in the other direction it drags the preview of the new area.

How about zooming as a compromise solution? It allows a user to switch between a general and a detailed picture.

Waiting for your ideas regarding the singleton.

bjorn:
I like the zooming! It's rather hidden, but indeed this way it can help people who want to see more detail.

About the singleton, I struggled a bit to resolve it myself, and then rather than explaining how I would like it to be solved, I decided to just push that in a follow-up change myself. See commit 62a89e1c099d95e769bf1859eacaa624b51ba484. It may make the code a little more convoluted, especially since I needed to forward the renderer callback from the ResizeDialog, but it keeps things more cleanly separated.

Thanks for the nice improvement! I do hope that before release either me or somebody else will get around to making it render the image in a thread so that the UI stays more responsive when resizing larger maps.

bjorn:
Not sure why I didn't think if this before, but I noticed today that the implementation as such only works for orthogonal maps using square tiles. I've opened issue #1554 about extending the functionality to other map configurations.

PR #1518, opened 2017-03-30
Fixed objects offsetting while resizing

Acuion:
Objects were not offseted if the "Remove objects outside of the map" was not set


bjorn:
Thanks!

PR #1521, opened 2017-03-31
Adjusted dash length in object selection rectangle

thabetx:
Used a custom dash line and flat cap to make the black and white dashes have the exact same length, also fixed a typo in the comment.

The chosen length is 5 pixels, but other lengths like 4 or 6 pixels also look fine.


bjorn:
Thanks! It's great how flat the cap style also fixes the rendering in OpenGL mode.

Issue #563, opened 2013-11-14
Command to auto-show/hide all palettes


benstahl:
Similar to the Adobe apps (which use the Tab key), press a key and all palettes, tool bar, and status bar disappear, leaving you with a large "clean" view of your tile map grid. Press again and the palettes re-appear. Not sure whether the "clean" map view should re-position itself to center itself up in the new space (but I suggest yes).

erem2k:
Looks quite straightforward, anything else other than is adding shortcut performing this action needs to be done here?

bjorn:
@erem2k I guess the action should be visible in the menu somewhere, it could go together with _View -> Full Screen_ (F11).

The main implementation work is probably to remember which docks were visible and to restore this state when the shortcut is used again. It could be done either manually or using `QMainWindow::restoreState`, not sure what will work better. Also it needs to be considered what should be done when the user has manually restored some docks, or maybe that should be disallowed. Disallowing is probably easiest (though I'm not sure how that would be done), because care should also be taken that when entering the mode and then closing Tiled, that you can still reopen Tiled either in regular mode, or still in "distraction free" mode and able to switch back to regular mode.

When in doubt about the behavior I tend to also look at other applications that already have similar functionality. They have usually already struggled with the same questions.

erem2k:
@bjorn 
> View -> Full Screen (F11)

Wouldn't something like "Clear View" fit better? It's not supposed to expand window into full-screen mode judging from issue description

> Also it needs to be considered what should be done when the user has manually restored some docks, or maybe that should be disallowed

In Adobe apps, restoring docks is possible during "view" mode, and they remain after interface is turned back on again in addition to docks that were originally hidden. If dock had multiple tabs, was active and had tab open during "interface" mode and was restored during "view", the tab active in "view" mode remains, sort of expand and replace behavior. If app is closed while in "view", it opens back in "interface" mode - think I'll go with that idea

bjorn:
> Yup, will stuff it there. What hotkey is it supposed to use, Tab like suggested?

Yes, I think Tab key and in general the behavior in Adobe apps will work fine.

PR #1522, opened 2017-04-01
Added auto-show/hide all palettes (Clear View) option to View menu

erem2k:
Addresses #563. Added private slot to MainWindow which does all the work, appropriate QAction and modified View Menu to accommodate new option.

### DIFF
--- a/src/tiled/mainwindow.cpp
+++ b/src/tiled/mainwindow.cpp
@@ -1150,12 +1150,9 @@ void MainWindow::toggleClearView(bool clearView)
                 mHiddenToolbars.push_back(toolbar);
             toolbar->hide();
         }      
-        //mUi->menuBar->hide();
     } else {
-        //mUi->menuBar->show();
         for (auto dock : mHiddenDocks)
             dock->show();
-
         for (auto toolbar : mHiddenToolbars)
             toolbar->show();
     }
--- a/src/tiled/mainwindow.cpp
+++ b/src/tiled/mainwindow.cpp
@@ -504,8 +504,9 @@ void MainWindow::closeEvent(QCloseEvent *event)
         toggleClearView(false);
         writeSettings();
         event->accept();
-    } else
+    } else {
         event->ignore();
+    }
 }
 
 void MainWindow::changeEvent(QEvent *event)
@@ -1133,8 +1134,8 @@ void MainWindow::setFullScreen(bool fullScreen)
 
 void MainWindow::toggleClearView(bool clearView)
 {
-    QList<QDockWidget*> docks = this->findChildren<QDockWidget*>();
-    QList<QToolBar*> toolbars = this->findChildren<QToolBar*>();
+    QList<QDockWidget*> docks = findChildren<QDockWidget*>();
+    QList<QToolBar*> toolbars = findChildren<QToolBar*>();
 
     if (clearView) {
         mHiddenDocks.clear();
@@ -1142,12 +1143,12 @@ void MainWindow::toggleClearView(bool clearView)
 
         for (auto dock : docks) {
             if (dock->isVisible())
-                mHiddenDocks.push_back(dock);
+                mHiddenDocks.append(dock);
             dock->hide();
         }
         for (auto toolbar : toolbars) {
             if (toolbar->isVisible())
-                mHiddenToolbars.push_back(toolbar);
+                mHiddenToolbars.append(toolbar);
             toolbar->hide();
         }      
     } else {
@@ -1155,6 +1156,9 @@ void MainWindow::toggleClearView(bool clearView)
             dock->show();
         for (auto toolbar : mHiddenToolbars)
             toolbar->show();
+
+        mHiddenDocks.clear();
+        mHiddenToolbars.clear();
     }
 }
 
--- a/src/tiled/mainwindow.cpp
+++ b/src/tiled/mainwindow.cpp
@@ -316,6 +316,8 @@ MainWindow::MainWindow(QWidget *parent, Qt::WindowFlags flags)
     connect(mUi->actionZoomOut, SIGNAL(triggered()), SLOT(zoomOut()));
     connect(mUi->actionZoomNormal, SIGNAL(triggered()), SLOT(zoomNormal()));
     connect(mUi->actionFullScreen, &QAction::toggled, this, &MainWindow::setFullScreen);
+    /*Add appropriate slot*/
+    connect(mUi->actionClearView, &QAction::toggled, this, &MainWindow::toggleClearView);
 
     CommandManager::instance()->registerMenu(mUi->menuCommand);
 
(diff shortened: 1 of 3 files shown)


erem2k:
There *is* a way to do it more elegant, namely by using QMainWindow states, but it'll require a dedicated parser for state's QByteArray to merge states before and after toggle for keeping docks and toolbars, that were enabled in view mode, open

src/tiled/mainwindow.ui
```diff
@@ -591,6 +592,17 @@
...
   </action>
+  <action name="actionClearView">
+   <property name="checkable">
+    <bool>true</bool>
+   </property>
+   <property name="text">
+    <string>Clear View</string>
+   </property>
+   <property name="shortcut">
+    <string>Q</string>
```
erem2k: Yeah, it was the original idea, but with the way widgets are currently positioned, main toolbar and document tabs receive keypress and can be navigated with Tab, unsure if that's intended behavior or not, so I figured it would be safer to bind it to something else rather than breaking navigation.

src/tiled/mainwindow.cpp
```diff
@@ -1129,6 +1135,46 @@ void MainWindow::setFullScreen(bool fullScreen)
         setWindowState(windowState() & ~Qt::WindowFullScreen);
 }
 
+void MainWindow::toggleClearView(bool clearView)
+{
+    //Static vectors keep list of visible docks and toolbars before toggle
+    static std::vector<bool> visibleDocks, visibleToolbars;
```
erem2k: Yes, makes sense and eliminates need for redundant checks everywhere, will do.

src/tiled/mainwindow.cpp
```diff
@@ -1129,6 +1135,46 @@ void MainWindow::setFullScreen(bool fullScreen)
...
+    QList<QDockWidget*> docks = this->findChildren<QDockWidget*>();
+    QList<QToolBar*> toolbars = this->findChildren<QToolBar*>();
+
+    if(clearView)
+    {
+        visibleDocks.clear();
+        visibleToolbars.clear();
+        mDocumentManager->saveState();
```
erem2k: Some debug leftovers, will clean that up.

src/tiled/mainwindow.cpp
```diff
@@ -1129,6 +1135,46 @@ void MainWindow::setFullScreen(bool fullScreen)
...
+        for(auto toolbar : toolbars)
+        {
+            visibleToolbars.push_back(toolbar->isVisible());
+            toolbar->setVisible(false);
+        }
+    }
+    else
+    {
+        if(visibleDocks.empty() || visibleToolbars.empty())
+            return;
```
erem2k: That was originally to prevent from `visibleDocks` and `visibleToolbars` throwing an exception if uninitialized during loop lookup by index, but with appropriate `QList`-s that would be deprecated

erem2k:
I'll try to implement hiding menu and status bar before merger, and give a go for camera position after.

src/tiled/mainwindow.cpp
```diff
@@ -498,10 +499,15 @@ void MainWindow::commitData(QSessionManager &manager)
 
 void MainWindow::closeEvent(QCloseEvent *event)
 {
-    writeSettings();
-
     if (confirmAllSave())
+    {
```
erem2k: Didn't really understand the last part, should it be like this or...
```
if (statement) {
command;
anothercommand;
}
```
...this?
```
if (statement) {
command;
anothercommand; }
```

src/tiled/mainwindow.ui
bjorn: I don't think the Tab for navigation is actually working very well for navigating the global UI. It's hard to see where the focus is and eventually you simply get stuck in the tileset view. I'm not surprised that other applications have chosen to use Tab for something more useful, and I think we can do the same with Tiled.

src/tiled/mainwindow.cpp
bjorn: Ah, right, I forgot that these are not the containers you were iterating.

bjorn:
> On the other hand, doesn't hiding menu defeat a purpose of being able to open up needed docks/toolbars during view as was described in addressed issue?

I'm undecided. I guess it's a rare action actually, and I personally expect it would be more useful to keep the tools toolbar visible than the menu. Maybe we should focus on the docks for now, since I also agree that the status bar may be useful (it could have its own show/hide toggle somewhere for those who want to get rid of it).

erem2k:
Pushed changes requested, kept behavior for menu and status bar for now.

src/tiled/mainwindow.cpp
```diff
@@ -498,11 +499,12 @@ void MainWindow::commitData(QSessionManager &manager)
...
-    writeSettings();
-
-    if (confirmAllSave())
+    if (confirmAllSave()) {
+        //Make sure user won't end up in view mode on next launch
+        toggleClearView(false);
+        writeSettings();
         event->accept();
-    else
+    } else
```
bjorn: Coding style: Please also wrap the else part with brackets. The general rule is that when any part of a condition has curly brackets, all parts should have them.

src/tiled/mainwindow.cpp
```diff
@@ -1129,6 +1131,33 @@ void MainWindow::setFullScreen(bool fullScreen)
         setWindowState(windowState() & ~Qt::WindowFullScreen);
 }
 
+void MainWindow::toggleClearView(bool clearView)
+{
+    QList<QDockWidget*> docks = this->findChildren<QDockWidget*>();
+    QList<QToolBar*> toolbars = this->findChildren<QToolBar*>();
```
bjorn: Please still remove the `this->` as requested earlier, since it's not needed.

src/tiled/mainwindow.cpp
```diff
@@ -1129,6 +1131,33 @@ void MainWindow::setFullScreen(bool fullScreen)
...
+        for (auto toolbar : toolbars) {
+            if (toolbar->isVisible())
+                mHiddenToolbars.push_back(toolbar);
+            toolbar->hide();
+        }      
+    } else {
+        for (auto dock : mHiddenDocks)
+            dock->show();
+        for (auto toolbar : mHiddenToolbars)
+            toolbar->show();
```
bjorn: Please clear the containers after this as well, to make sure that repeated calls of `toggleClearView(false)` (like may happen during shutdown) don't make in-the-meantime-manually-hidden docks visible again.

src/tiled/mainwindow.cpp
```diff
@@ -1129,6 +1131,33 @@ void MainWindow::setFullScreen(bool fullScreen)
...
+        mHiddenToolbars.clear();
+
+        for (auto dock : docks) {
+            if (dock->isVisible())
+                mHiddenDocks.push_back(dock);
+            dock->hide();
+        }
+        for (auto toolbar : toolbars) {
+            if (toolbar->isVisible())
+                mHiddenToolbars.push_back(toolbar);
```
bjorn: Please prefer to use the Qt-style members, `append` in this case, and not `push_back`, which is only there for STL compatibility.

erem2k:
Sure, done.

bjorn:
Thanks a lot for your contribution! I think it's a very nice addition.

erem2k:
You're welcome! Pleasure working with you.

Issue #1520, opened 2017-03-31
automap behaviour is counter intuitive when some tiles are empty on input_layer


axboureau:
I find the rules for counting matches counter intuitive when there are some empty tiles. I think it is related to issue #1276.

Not sure which ones should be considered bugs:

- inputnot_layera  always fails when layera is not present

It is probably not a big deal, as there would be little reason to define an inputnot_layer in this case, and a blank input layer would be enough, but the result is unexpected, so I think inputnot_layer should return a match if the layer is absent.

- The match fails if there is a tile defined in the region, for which there is no matching tiles in any of the
input layers. 
![tiled_automap_issue1](https://cloud.githubusercontent.com/assets/3704031/24555174/76e0dd5a-1630-11e7-8367-5cda7d57032b.png)
Here, the region is 2 tiles. Only one is checked on input_structure, but if there is nothing on the structure layer at the position of the second, it does not count as a match.
Having anything there on this structure layer counts as a match. 

- Sometimes, having an empty tile in the rule file makes the match fails.

![tiled_automap_issue2](https://cloud.githubusercontent.com/assets/3704031/24555184/7e106992-1630-11e7-80b1-e8fa30eeb3dc.png)

This one is puzzling too: 
This time, there are 2 floor tiles on the wallbases layer, but only one is required according to the input_wallbases layer. This result in a missmatch. But if we add another floor tile on the input_structure layer, it becomes a match, even though we added a condition, and didn't change the structure layer itself.

bjorn:
> It is probably not a big deal, as there would be little reason to define an inputnot_layer in this case, and a blank input layer would be enough, but the result is unexpected, so I think inputnot_layer should return a match if the layer is absent.

I agree.

In general, I hope @stefanbeller could weigh in on the other parts, since they are a bit beyond me and I don't know the details about why the system works like it does.

bjorn:
Related forum threads:

* http://discourse.mapeditor.org/t/automap-rules-input-multiple-layers/438/10
* http://discourse.mapeditor.org/t/trying-to-understand-solve-an-issue-with-automapping-hope-someone-can-help/2228

bjorn:
This behavior will stay counter-intuitive for compatibility reasons, but a more intuitive solution has been implemented as part of #1918 and #3100. The relevant improvements are two-fold:

[...]

PR #1524, opened 2017-04-06
updated automapper.cpp

axboureau:
- A match won't fail if an input_layer has no corresponding layer in the workmap as long as all of the cells of the input_layer for the region are empty. I added a dummy empty layer that gets passed to compareLayerTo 
- A match won't fail if one cell of an input_region is empty and the corresponding workmap layer cell is also empty (it will also work if the cell of the workmap layer is not empty). cf change line 752.
It should address the problems mentioned here:
 http://discourse.mapeditor.org/t/trying-to-understand-solve-an-issue-with-automapping-hope-someone-can-help/2228/2
http://discourse.mapeditor.org/t/automap-rules-input-multiple-layers/438/15
and here:
https://github.com/bjorn/tiled/issues/1520

I also added some comments (maybe some of them are too"trivial"?).

### DIFF
--- a/src/tiled/automapper.cpp
+++ b/src/tiled/automapper.cpp
@@ -206,6 +206,13 @@ bool AutoMapper::setupRuleMapTileLayers()
                 continue;
             }
 
+            // for each index, creates an InputIndex, that will
+            // hold all the InputConditions for each layer that exist
+            // with this index, and index them by name.
+            // So an InputConditions holds all the layer that share
+            // the same name and index.
+            // The InputConditions are divided in listNo and listYes.
+
             mInputRules.names.insert(name);
 
             if (!mInputRules.contains(index))
@@ -221,6 +228,7 @@ bool AutoMapper::setupRuleMapTileLayers()
 
             continue;
         }
+        // Update output tile and object layer list.
 
         if (layerName.startsWith(QLatin1String("output"), Qt::CaseInsensitive)) {
             if (layer->isTileLayer())
@@ -279,6 +287,16 @@ static bool compareRuleRegion(const QRegion &r1, const QRegion &r2)
     return p1.y() < p2.y() || (p1.y() == p2.y() && p1.x() < p2.x());
 }
 
+/* Get the list of coherentRegions from the union of mLayerInputRegions and
+ * mLayerOutputRegions, then checks that inputRegions match with
+ * outputRegions
+ * cf geometry.cpp for coherentRegions,
+ * 'coherent' means that either the rectangle is overlapping the region or
+ * the rectangle contains at least one tile, which is a direct neighbour
+ * to a tile, which belongs to the region.
+ * coherentRegions returns an array of coherentRegions from a region.
+ */
+
 bool AutoMapper::setupRuleList()
 {
     Q_ASSERT(mRulesInput.isEmpty());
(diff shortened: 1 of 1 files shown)


src/tiled/automapper.cpp
```diff
@@ -749,7 +797,7 @@ static bool compareLayerTo(const TileLayer *setLayer,
                 if (listNo.isEmpty()) {
                     if (matchListYes)
                         continue;
-                    if (!ruleDefinedListYes && !cells.contains(c1))
```
bjorn: This change would require an update to the comments, in particular this part:
```
 *      Exception, when having only the listYes:
 *      if at the examined position there are no tiles within all Layers
 *      of the listYes, all tiles except all used tiles within
 *      the layers of that list are considered good.
 *
 *      This exception was added to have a better functionality
 *      (need of less layers.)
```
And apparently, this makes some case worse. Essentially, leaving a tile empty in the `listYes` is taken to mean that there should not be any tile used elsewhere in that same rule (when there are no explicit exclusion rules). Without that logic you need to fill the `listNo` with all those tiles to get the same functionality. That seems like a nice shortcut, but it is indeed rather unexpected. Maybe @stefanbeller could weigh in on this?

src/tiled/automapper.cpp
```diff
@@ -519,16 +558,25 @@ QRect AutoMapper::applyRule(const int ruleIndex, const QRect &where)
...
-                    const TileLayer *setLayer = mMapWork->layerAt(i)->asTileLayer();
-                    allLayerNamesMatch &= compareLayerTo(setLayer,
-                                                         conditions.listYes,
-                                                         conditions.listNo,
-                                                         ruleInputRegion,
-                                                         QPoint(x, y));
+                    dummyCreated = true;
+                    setLayer = new TileLayer(name, 0, 0,
+                                     mMapWork->width(),
+                                     mMapWork->height());
```
axboureau: Indeed, that is a good point.

src/tiled/automapper.cpp
axboureau: Maybe we should add some other layer types (on top of regions, inputyes, inputnot) to handle such cases. Leaving all listYes empty is sometimes necessary when we want to create a region rule where not all input tiles are adjacent(like an X shaped one ), so I think it is better if an empty input tile has the expected result, and use some additional explicit layers for other cases.

bjorn: @axboureau I agree the system should work more intuitively, but do you have a specific suggestion of what alternative to offer for people relying on this behavior?

axboureau: Indeed, maybe the best would be to have a property to handle the way it works when there is no inputnot_layer.
Otherwise, we'd have to add an empty inputnot_layer for each input_layer.

bjorn:
>A match won't fail if an input_layer has no corresponding layer in the workmap as long as all of the cells of the input_layer for the region are empty. I added a dummy empty layer that gets passed to compareLayerTo

@axboureau I've addressed this in change a22476fc49a4f4af5bc16779118156a16b42899d, which does essentially the same without allocating and freeing a lot of memory all the time (it certainly helps that tile layers are infinite now and only allocate chunks of memory on-demand).

> A match won't fail if one cell of an input_region is empty and the corresponding workmap layer cell is also empty (it will also work if the cell of the workmap layer is not empty). cf change line 752.

I've tried to make this change, but it changed the behavior in a way that broke the "sewer_automap" example. It might still be that this behavior is desired in more cases, but I didn't want to break compatibility at this point. I'm hopeful that the new "StrictEmpty" mode would also solve your use-case, see the description of change eb1e85044f309d1cd9132a15baf7eaab1be6d4a5 for more information.

If that new option resolves your use-case, then I think we can close this PR and issue #1520. But I would also still be open to introduce another option for this if the option doesn't turn out to be so convenient (which I could definitely imagine).

bjorn:
@axboureau I'll close this PR, but feel free to let me know anytime when there is still something to be improved about the automapping.

Issue #409, opened 2013-03-31
Change tile after inserting a tile object.


mimpedro:
When creating a new object with the "Insert Tile Objects" tool, you can't change its tile anymore. 

In the game I am currently making, there is a huge amount of objects that take appearance of different terrain tiles, but they all have the same properties inside. It would be a big hassle to create each individual object with the Insert Tile Object tool and then write all of its properties by hand, as opposed to simply copying and pasting one of them and changing its corresponding tile.

Anyway, I've already solved this by using a tile layer, but I believe this issue could be easily fixed and would be very useful for other developers.

Sorry if it sounded confusing, English is not my first language.

bjorn:
Also [suggested on the forum](http://discourse.mapeditor.org/t/how-can-i-get-image-names-on-objects/1964/2?u=bjorn), with the menu item in the context menu of the tile:

> 1. selecting the Image Object/s
> 2. selecting the Tile in the tileset window
> 3. RMB > Replace image in selected object/s

thabetx:
I started working on this issue, I'm implementing the workflow in the previous comment.

The basic functionality is done, but it still needs some adjustments when changing multiple objects.

PR #1525, opened 2017-04-09
Replace object tile

thabetx:
Implemented replacing objects tiles function, solves #409 .

Here is a demo of replacing then undoing
replace07

Currently, when no object is selected the menu item is available, choosing it won't do anything, I tired to disable it but couldn't figure out the best way to achieve it as the tileset view doesn't reference the map document.


thabetx:
When an object's tile changes, I changed the size of the object to match the tile's dimensions.

But I updated it, so now every object keeps its size, this feels more natural, but don't know if it's the required behavior.

On a different note, the build currently fails but works locally, this is the error

>C:/projects/tiled/release/tiled.qt.6b4bc486/.obj/3a52ce780950d4d9/tilesetdock.cpp.o:tilesetdock.cpp:(.text+0x1c26): undefined reference to `Tiled::Internal::ReplaceObjectsTile::ReplaceObjectsTile(Tiled::Internal::MapDocument*, QList<Tiled::MapObject*> const&, Tiled::Tile*)'

I'm trying to fix it but couldn't figure it out yet.

thabetx:
I amended the last commit and fixed the build, the files weren't included in tiled.qbs, so I included them and Appveyor's build worked.

But Travis failed while initializing its environment so I closed and reopened the pull request to trigger the build again.

src/tiled/replaceobjectstile.cpp
```diff
@@ -0,0 +1,51 @@
...
+{}
+
+void ReplaceObjectsTile::replace()
+{
+    Tile * const tile = mTile;
+
+    mOriginalTiles.clear();
+
+    for (MapObject *object : mMapObjects) {
+        mOriginalTiles.append(object->cell().tile());
```
bjorn: Please initialize `mOriginalTiles` once in the constructor.

src/tiled/tilesetdock.cpp
```diff
@@ -939,3 +942,17 @@ void TilesetDock::swapTiles(Tile *tileA, Tile *tileB)
...
+    if (!mMapDocument)
+        return;
+
+    auto &selectedObjects = mMapDocument->selectedObjects();
+
+    if (selectedObjects.isEmpty())
+        return;
+
+    QUndoStack *undoStack = mMapDocument->undoStack();
+    undoStack->push(new ReplaceObjectsTile(mMapDocument, selectedObjects, tile));
```
bjorn: Not all selected objects may be tile objects. I would suggest to only support this action on existing tile objects, because the behavior when setting a tile on a rectangle, circle, polygon or text object isn't well defined (or at least, this is currently not a supported scenario).

src/tiled/replaceobjectstile.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+void ReplaceObjectsTile::restore()
+{
+    for (int i = 0; i < mMapObjects.size(); ++i)
+        swapObjectTile(mMapObjects[i], mOriginalTiles[i]);
+
+    emit mMapDocument->mapObjectModel()->objectsChanged(mMapObjects);
+}
+
+void ReplaceObjectsTile::swapObjectTile(MapObject *object, Tile *tile) {
```
bjorn: Coding style: `{` should be on its own line for functions.

Naming: this function should just be called `setObjectTile`.

Also, this function can be a local static because it doesn't use any members of `ReplaceObjectsTile` and it's not used from elsewhere.

src/tiled/replaceobjectstile.h
```diff
@@ -0,0 +1,42 @@
...
+
+namespace Tiled {
+
+class Tile;
+
+namespace Internal {
+
+class MapDocument;
+
+class ReplaceObjectsTile : public QUndoCommand
```
bjorn: I think the class should be called `ChangeMapObjectsTile` and it can be moved into the existing `changemapobject.h/cpp` files.

thabetx:
Thank you for the great feedback, I have addressed the issues.

The tile size changing was counter intuitive because it's affecting the size of objects which preserved there size, but I got the point now, the swap two tiles code was doing something similar but I didn't know why it was doing it that way.

Changing the tile of tile objects only was done by filtering the objects before passing them.

To simplify the code of replacing the tiles, I saved the state in two list(tile, size) and one helper list for specifying whether to change the size or not.

src/tiled/changemapobject.h
```diff
@@ -87,5 +87,27 @@ class ChangeMapObjectCells : public QUndoCommand
...
+
+private:
+    void changeTiles();
+    void restoreTiles();
+
+    MapDocument *mMapDocument;
+    const QList<MapObject *> mMapObjects;
+    Tile * const mTile;
+    QList<Tile *> mOldTiles;
+    QList<QSizeF> mOldSize;
```
bjorn: Please use a `QVector<QSizeF>`,  because a `QSizeF` does not fit in `sizeof(void*)` (making QList very inefficient for this type, since it will allocate each size on the heap and store a pointer to it). Actually you could use a QVector for all containers, except for `mMapObjects` (since it would require changing too much other code to avoid copying from QList, and there is no reason not to use a QList for storing pointers.

src/tiled/tilesetdock.cpp
```diff
@@ -939,3 +942,24 @@ void TilesetDock::swapTiles(Tile *tileA, Tile *tileB)
...
+{
+    if (!mMapDocument)
+        return;
+
+    auto &selectedObjects = mMapDocument->selectedObjects();
+
+    // Only change tiles of tile objects
+    QList<MapObject *> filterdObjects;
+
+    for(auto object : selectedObjects)
```
bjorn: Coding style: Please put a space in `for(`.

src/tiled/tilesetdock.cpp
```diff
@@ -939,3 +942,24 @@ void TilesetDock::swapTiles(Tile *tileA, Tile *tileB)
...
+
+void TilesetDock::changeSelectedMapObjectsTile(Tile *tile)
+{
+    if (!mMapDocument)
+        return;
+
+    auto &selectedObjects = mMapDocument->selectedObjects();
+
+    // Only change tiles of tile objects
+    QList<MapObject *> filterdObjects;
```
bjorn: Typo: `filteredObjects`, but I'd call it `tileObjects` instead.

src/tiled/tilesetview.cpp
```diff
@@ -699,6 +699,16 @@ void TilesetView::contextMenuEvent(QContextMenuEvent *event)
             swapTilesAction->setEnabled(exactlyTwoTilesSelected);
             connect(swapTilesAction, SIGNAL(triggered()),
                     SLOT(swapTiles()));
+
+            bool onlyOneTileSelected =
+                    (selectionModel()->selectedIndexes().size() == 1);
+            // TODO: check that at least one object is selected
+
+            QAction *changeSelectedMapObjectsTileAction =
+                    menu.addAction(tr("&Replace Tile of Selected Object/s"));
```
bjorn: Just write "&Replace Tile of Selected Objects".

src/tiled/tilesetview.cpp
```diff
@@ -699,6 +699,16 @@ void TilesetView::contextMenuEvent(QContextMenuEvent *event)
...
+            changeSelectedMapObjectsTileAction->setEnabled(onlyOneTileSelected);
+            connect(changeSelectedMapObjectsTileAction , SIGNAL(triggered()),
+                    SLOT(changeSelectedMapObjectsTile()));
```
bjorn: No reason not to use a modern style connect here:
```c++
connect(changeSelectedMapObjectsTileAction, &QAction::triggered,
        this, &TilesetView::changeSelectedMapObjectsTile);
```

src/tiled/tilesetview.cpp
```diff
@@ -699,6 +699,16 @@ void TilesetView::contextMenuEvent(QContextMenuEvent *event)
             swapTilesAction->setEnabled(exactlyTwoTilesSelected);
             connect(swapTilesAction, SIGNAL(triggered()),
                     SLOT(swapTiles()));
+
+            bool onlyOneTileSelected =
+                    (selectionModel()->selectedIndexes().size() == 1);
+            // TODO: check that at least one object is selected
```
bjorn: Right. I think it's fine to leave a todo here for now, cause I'm not sure what's the best way. Maybe the context menu should be moved out of the TilesetView so that it can be separately implemented in the TilesetEditor and the TilesetDock.

thabetx:
Thank you for the feedback, I adjusted the style, I'm still getting used to it :sweat_smile: .

I didn't know that if the size of the stored objects was larger than pointer size the elements won't be contiguous, anyway, I found out that the list wasn't used as the size is implicitly stored in the tile, so I removed the list.

Regarding the cell check I extracted a function for choosing only tile objects before calling the change function, so the change function assumes that all the passed objects are valid tile objects, this isn't enforced, so I wonder if it is enough?

src/tiled/tilesetdock.cpp
```diff
@@ -939,3 +942,29 @@ void TilesetDock::swapTiles(Tile *tileA, Tile *tileB)
     QUndoStack *undoStack = mMapDocument->undoStack();
     undoStack->push(new SwapTiles(mMapDocument, tileA, tileB));
 }
+
+static QList<MapObject*> chooseTileObjects(QList<MapObject*> objects)
+{
+    QList<MapObject*> tileObjects;
+
+    for (auto object : objects)
+        if (!object->cell().isEmpty() && object->cell().tile() != nullptr)
```
bjorn: You could simplify this check to just `if (object->cell().tile())`, however I think it's not the behavior we want. I think this code should check only `if (!object->cell().isEmpty()`, since when the tile happens to be set but is invalid, we still want to allow the user to replace it with another tile.

One problem with that though is undo, which should then restore the Tileset* and (invalid) tile ID, rather than setting the tile explicitly to null. So I think in `ChangeMapObjectsTile` you'll need to replace `mOldTiles` with `QVector<Cell> mOldCells`. And of course that code can't use the Tile* unconditionally.

thabetx:
Alright, I didn't think of that. I've updated the code, it now works on tile objects which reference non existing tilesets.

The null-tile objects doesn't have a tile so the object size is overwritten, I had to bring back the mOldSizes vector to be able to restore their sizes.

src/tiled/changemapobject.cpp
```diff
@@ -99,35 +99,39 @@ ChangeMapObjectsTile::ChangeMapObjectsTile(MapDocument *mapDocument,
...
 
     emit mMapDocument->mapObjectModel()->objectsChanged(mMapObjects);
 }
 
 void ChangeMapObjectsTile::changeTiles()
 {
-    for (int i = 0; i < mMapObjects.size(); i++)
-        setObjectTile(mMapObjects[i], mTile, mSizePreserved[i]);
+    for (int i = 0; i < mMapObjects.size(); ++i)
+        setObjectCell(mMapObjects[i], mCell, mUpdateSize[i], mCell.tile()->size());
```
bjorn: I'm afraid you can't use a single cell for all objects, since that way any horizontal or vertical flipping will not be preserved. You need to take the cell from the object and change its tile, like you used to do in `setObjectTile`.

src/tiled/changemapobject.cpp
```diff
@@ -99,35 +99,39 @@ ChangeMapObjectsTile::ChangeMapObjectsTile(MapDocument *mapDocument,
...
+    mCell = Cell(mTile);
     for (MapObject *object : mMapObjects) {
-        Tile *tile = object->cell().tile();
-        mOldTiles.append(tile);
-        mSizePreserved.append(object->size() == tile->size());
+        Cell cell = object->cell();
+        mOldCells.append(cell);
+        mOldSizes.append(object->size());
+        Tile *tile = cell.tile();
+        mUpdateSize.append(tile == nullptr || object->size() == tile->size());
```
bjorn: If the previous tile was no longer valid, you don't know whether the object used to match the size or not. I think the decision about whether to update the size is arbitrary in that case, and I may have chosen to leave the size alone instead. That way you also don't need `mOldSizes`.

Please leave out explicit comparisons to `nullptr`, btw. You can just write `!tile`, or if you would change the condition, `tile && object->size() == tile->size()`.

thabetx: Thank you, my initial implementation left the object size as is, but I wasn't sure so I just updated the size to match the tile, should've discussed that.

src/tiled/changemapobject.cpp
thabetx: Oh, I missed that, will update it.

bjorn:
Thanks!

thabetx:
You're welcome and thank you for your help.

PR #1528, opened 2017-04-10
Add qml.qrc for tiledquick's QML files

mitchcurtis:
This avoids the need to mess around with import paths for the
application's QML files.


mitchcurtis:
This simplifies the process a bit until you get around to working on those issues you mentioned. I didn't add the Android files since I'm not testing there.

bjorn:
That's great, thanks!

PR #1534, opened 2017-04-12
Added Hungarian translation

urbalazs:
Hi @bjorn 
Here is the complete Hungarian translation for Tiled. Is it enough to sent it as pull request, or should I also upload it to Transifex and/or Weblate?

### DIFF
--- a/translations/tiled_hu.ts
+++ b/translations/tiled_hu.ts
@@ -2110,7 +2110,7 @@ Line %1, column %2</source>
         <location line="-13"/>
         <source>Duplicate %n Object(s)</source>
         <translation>
-            <numerusform>%n objektum kettőzése</numerusform>
+            <numerusform>Objektum kettőzése</numerusform>
             <numerusform>%n objektum kettőzése</numerusform>
         </translation>
     </message>
@@ -2118,7 +2118,7 @@ Line %1, column %2</source>
         <location line="+2"/>
         <source>Remove %n Object(s)</source>
         <translation>
-            <numerusform>%n objektum eltávolítása</numerusform>
+            <numerusform>Objektum eltávolítása</numerusform>
             <numerusform>%n objektum eltávolítása</numerusform>
         </translation>
     </message>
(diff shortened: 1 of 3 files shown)


translations/tiled_hu.ts
```diff
@@ -0,0 +1,4831 @@
...
+        <source>%1 x %2</source>
+        <translation>%1 x %2</translation>
+    </message>
+    <message>
+        <location line="+96"/>
+        <source>Width</source>
+        <translation>Szélesség</translation>
+    </message>
+    <message>
+        <location line="+8
```
bjorn: I notice that you basically always used the same translation for both plural forms. Doesn't Hungarian have different plural forms? In any case, the first form is used when there is only 1, which generally means you can leave the `%n` out of the translation.

urbalazs: Hungarian has no different plural forms (e. g. the English "1 tile" and "2 tiles" in Hungarian "1 csempe" and "2 csempe").

bjorn: Ok, makes sense. What about leaving the number out in case it is 1? It helps to keep the texts in the UI simpler since 1 is usually the most common case.

urbalazs: I don't really know the internal structure of .ts file. But I think, we can leave the translation in this form, and executing "lrelease" command will drop these numerusforms when it is unneeded for generating the .qm file. I tried it, and put the tiled_hu.qm into the "translations" folder. It works fine.
There is plural form in Hungarian language when using no number in the sentence. For example the English "Delete tile?" or "Delete tiles?" are also different is Hungarian: "Törli a csempét?" or "Törli a csempé**ket**?". The latter is the plural form.

bjorn: I wasn't talking about the structure of a .ts file, just that it looks nicer in the UI when using Tiled, to read "Duplicate Object" instead of "Duplicate 1 Object". So I was just suggesting to leave the %n out in the singular case.

urbalazs:
I now understand, what are you talking about. I modified my pull request. Thanks for the suggestion!

bjorn:
Thanks! I've pushed the translation to the `0.18` branch as 5343bed3b32d9607fbb54f1f39a1e875e5e4114d. I'll probably make a Tiled 0.18.3 release in the next days and merge this to `master`.

Note that I plan to make a Tiled 1.0 release in some weeks. You'll get an e-mail from me at least 2 weeks in advance, so that there will be time for all the translation updates needed since Tiled 0.18.

PR #1535, opened 2017-04-14
Isometric drawing

mitchcurtis:


mitchcurtis:
Continued on from this pull request: https://github.com/bjorn/tiled/pull/1529

This is still kinda work in progress, and I'm guessing these commits will need to be squashed into one? :p

The map is probably not rendered in the correct place, but I'm sure you'll have other comments too, so I'll wait for you to try it out.

mitchcurtis:
Now it's looking a bit better.

mitchcurtis:
Can this be merged?

mitchcurtis:
With https://github.com/bjorn/tiled/pull/1540 cherry-picked onto this, the tile coordinates don't match what is rendered. I don't know what's wrong.

In IsometricRenderer, there is lots of extra stuff going on. For example, I tried adding [this](https://github.com/bjorn/tiled/blob/wip/qtquick/src/libtiled/isometricrenderer.cpp#L212) to `IsometricRenderHelper::appendTileData()`, but it just made the coordinates even more incorrect. Doing [this](https://github.com/bjorn/tiled/blob/wip/qtquick/src/libtiled/isometricrenderer.cpp#L216) would have no effect, since the layer's x and y are 0. Then there's [this](https://github.com/bjorn/tiled/blob/wip/qtquick/src/libtiled/isometricrenderer.cpp#L218), which seems to address the case where rect.x/rect.y are non-zero, so I assume it's not relevant here, where the code currently *requires* the rect to start at (0, 0).

mitchcurtis:
If I make this change, the tiles are rendered in the "correct" place:

```
diff --git a/src/tiledquickplugin/tilelayeritem.cpp b/src/tiledquickplugin/tilelayeritem.cpp
index 5b62b00..4fce77c 100644
--- a/src/tiledquickplugin/tilelayeritem.cpp
+++ b/src/tiledquickplugin/tilelayeritem.cpp
@@ -249,7 +249,9 @@ void IsometricRenderHelper::appendTileData(int index)
     const QPointF screenPos = mRenderer->tileToScreenCoords(mapPos).toPoint();
     TileData data;
     data.x = screenPos.x();
+    data.x -= mTileWidth / 2;
     data.y = screenPos.y();
+    data.y -= mTileHeight / 2;
     const QSize size = cell.tile()->size();
     data.width = size.width();
     data.height = size.height();
```

Using this tileset:

isometric_grass_and_water

And adding some code to main.qml:

```
diff --git a/examples/isometric_grass_and_water.png b/examples/isometric_grass_and_water.png
index b342a0c..35012ec 100644
Binary files a/examples/isometric_grass_and_water.png and b/examples/isometric_grass_and_water.png differ
diff --git a/src/tiledquick/main.qml b/src/tiledquick/main.qml
index 7037916..9039fda 100644
--- a/src/tiledquick/main.qml
+++ b/src/tiledquick/main.qml
@@ -131,6 +131,32 @@ ApplicationWindow {
                             mapView.width / scale,
                             mapView.height / scale);
                 }
+
+                Rectangle {
+                    anchors.fill: parent
+                    color: "transparent"
+                    border.color: "darkorange"
+                }
+
+                Rectangle {
+                    id: player
+                    z: 10000
+                    width: 32
+                    height: width
+                    radius: width / 2
+                    color: "darkorange"
+
+                    Timer {
+                        running: true
+                        interval: 100
+                        onTriggered: {
+                            var p = mapItem.tileToScreenCoords(0, 0);
+                            print(p)
+                            player.x = p.x - player.width / 2;
+                            player.y = p.y - player.height / 2;
+                        }
+                    }
+                }
             }
         }
     }
```

I get this result:

rendering

So the "fix" might have worked, if it weren't for the fact that any game object is going to be rendered at the incorrect position, because it's using the (apparently incorrect?) coordinate conversion functions in MapItem.

mitchcurtis:
> any game object is going to be rendered at the incorrect position, because it's using the (apparently incorrect?) coordinate conversion functions in MapItem.

Hmmm, hang on.. it looks right. It's centred on 0, 0. If I use this code:


                            var p = mapItem.tileToScreenCoords(0.5, 0.5);
                            print(p)
                            player.x = p.x - player.width / 2;
                            player.y = p.y - player.height / 2;

It looks correct. So maybe the fix is fine?

bjorn:
Note that you just have to iterate y from 0 to "height - 1", starting x at 0 each time while iterating the row, and then iterate x from 1 to "width - 1", starting y at "height - 1" each time. You iterate the row by doing ++x, --y and checking x < width and y >= 0. So there are two basic loops to iterate all the tiles, and then you don't need `indexToMapPos` either. This should simplify `addTilesToNode` massively.

src/tiledquickplugin/tilelayeritem.cpp
```diff
@@ -173,6 +174,206 @@ static void drawTileLayer(QSGNode *parent,
...
+}
+
+void IsometricRenderHelper::appendTileData(int index)
+{
+    const QPoint mapPos = indexToMapPos(index);
+    const Cell &cell = mLayer->cellAt(mapPos.x(), mapPos.y());
+    if (cell.isEmpty())
+        return;
+
+    Tileset *tileset = cell.tile()->tileset();
```
bjorn: You get the tileset by doing `cell.tileset()`, because `cell.tile()` may return null when no tile with the `cell.tileId()` is found in the tileset (maybe I should call it `cell.findTile()`...)

src/tiledquickplugin/tilelayeritem.cpp
```diff
@@ -173,6 +174,206 @@ static void drawTileLayer(QSGNode *parent,
...
+    }
+
+    if (!mTilesetHelper.texture())
+        return;
+
+    const QPointF screenPos = mRenderer->tileToScreenCoords(mapPos).toPoint();
+    TileData data;
+    data.x = screenPos.x() - mTileWidth / 2;
+    data.y = screenPos.y() - mTileHeight / 2;
+    const QSize size = cell.tile()->size();
```
bjorn: Since this may return null you need to check the returned tile pointer, like the orthogonal rendering code is doing (it's doing that since the big merge from master).

mitchcurtis:
> Note that you just have to iterate y from 0 to "height - 1", starting x at 0 each time while iterating the row, and then iterate x from 1 to "width - 1", starting y at "height - 1" each time. You iterate the row by doing ++x, --y and checking x < width and y >= 0. So there are two basic loops to iterate all the tiles, and then you don't need indexToMapPos either. This should simplify addTilesToNode massively.

Sorry, but this made no sense to me. :D

bjorn:
> Could you provide some pseudo code or something?

The problem being that if I wrote the code I might as well do it in C++. See change f8527ed9851f30a9a3509c57ef10eaed7fdf731c. :-)

mitchcurtis:
> The problem being that if I wrote the code I might as well do it in C++. See change f8527ed. :-)

Yeah, sorry. Sometimes it's just easier that way. :x

Thanks! Do you think the tile coordinates are a bit off? If I hover slightly past the "x centre" of a tile, it says 1; it should only say 1 when I've gone onto the next tile.

mitchcurtis:
Ah, never mind. Fixed it: https://github.com/bjorn/tiled/pull/1547

PR #1536, opened 2017-04-14
Fix MSVC warning about struct that is forward declared as class

mitchcurtis:


bjorn:
I hate this warning. Thanks!

PR #1539, opened 2017-04-15
Add invokable coordinate conversion functions to MapItem

mitchcurtis:


bjorn:
Looks good! :-)

bjorn:
Will you add a little hover coordinate display to Tiled Quick now?

mitchcurtis:
> Will you add a little hover coordinate display to Tiled Quick now?

Haha, I'm currently writing some tests to ensure that the coordinate mapping is correct. I could definitely add it after that. :)

mitchcurtis:
By the way, I'm guessing we have to support Qt Quick 2.4 since the minimum Qt version supported is 5.4?

bjorn:
> By the way, I'm guessing we have to support Qt Quick 2.4 since the minimum Qt version supported is 5.4?

Yes, unless there is some reason to depend on a more recent version?

mitchcurtis:
Nope, just wanted to make sure.

PR #1540, opened 2017-04-15
Add tile coords to status bar

mitchcurtis:


mitchcurtis:
I had a lot of trouble with errors while copying the required dependencies into the build directory of the auto test... I was using this code:

```
CONFIG(debug, debug|release) {
    VARIANT = debug
} else {
    VARIANT = release
}

# Adpated from http://stackoverflow.com/a/39234363/904422
win32 {
    copyQmlImport.commands = $(COPY_DIR) \"$$clean_path($$OUT_PWD/../../qml)\" \"$$shell_path($$OUT_PWD)/$$VARIANT\"
    copyTiledLibs.commands = $(COPY_DIR) \"$$clean_path($$OUT_PWD/../../lib)\" \"$$shell_path($$OUT_PWD)/$$VARIANT\"
} else {
    copyQmlImport.commands = $(COPY_DIR) \"$$clean_path($$OUT_PWD/../../qml)\" \"$$shell_path($$OUT_PWD)\"
    copyTiledLibs.commands = $(COPY_DIR) \"$$clean_path($$OUT_PWD/../../lib)\" \"$$shell_path($$OUT_PWD)\"
}
first.depends = $(first) copyQmlImport copyTiledLibs
export(first.depends)
export(copyQmlImport.commands)
export(copyTiledLibs.commands)
QMAKE_EXTRA_TARGETS += first copyQmlImport copyTiledLibs
```

I hope QBS can do this, because it was (as always) super difficult with qmake, and only worked half the time, for some weird reason.

Anyway, I figured your comment about the tile coordinates made sense if I didn't get the whole auto test thing working; it would be an easy way to verify that the isometric rendering is correct.

PR #1541, opened 2017-04-15
Check for null renderer in coordinate conversion functions

mitchcurtis:


bjorn:
Maybe it would make sense to have a dummy renderer, or default to the orthogonal renderer...

mitchcurtis:
I think that once MapLoader's status signal is emitted at the correct time (currently it's emitted before e.g. an onStatusChanged connection is even established, so initial map loads won't produce warnings if you're relying on statusChanged to be emitted), not having a renderer won't be an issue. If game code only adds/positions its entities after the status is "Ready", it should just work... I think. :)

The fix for the signal emission timing is something I would like to have auto tested with the new tests that I'm trying to get working.

bjorn:
Right, looks like `MapLoader` needs to be adjusted to wait with loading until `componentComplete` is called.

PR #1542, opened 2017-04-15
Debian/Ubuntu Build Instructions README

daemonburrito:
Debian has a `dnf builddep` equivalent.

`apt-get build-dep $package` == `dnf builddep $package`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,13 +31,21 @@ Compiling
 
 Make sure the Qt (>= 5.4) development libraries are installed:
 
-* In Ubuntu/Debian: `apt-get install qt5-default qttools5-dev-tools zlib1g-dev`
+* In Ubuntu/Debian: `apt-get build-dep tiled`
 * In Fedora:        `sudo dnf builddep tiled`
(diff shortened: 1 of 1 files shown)


bjorn:
Hmm, whether that works is going to depend on the version of Debian somebody is using. I think the existing command will work on Jessie (stable) while this command will only work on more recent Debian versions, since Jessie still shipped with Tiled 0.10.1 built against Qt 4.

Maybe we should wait at least until it works on Debian stable?

daemonburrito:
Hmm, interesting.

I also added a note about the `qt${version}-default` meta packages before I got your comment.

+1 for waiting, I wasn't aware there was a chance that this wouldn't work on Debian stable.

bjorn:
Closing this in preference of listing the actual dependencies, since it's more useful for documentation purposes and some changes might happen in the dependencies of the current `master` compared to the Debian Tiled package. It might be better to transition the Fedora dependencies also to an explicit list instead.

PR #1543, opened 2017-04-16
Copy qmldir for in-source builds as well

mitchcurtis:
Applications that want to use quicktiledplugin will be expecting
the qmldir to be with all of the other plugin-related files.


bjorn:
Looks fine and I guess you tested it. :-)

I just hope we can stop using qmake soon, but that's what issue #1484 is about.

PR #1544, opened 2017-04-16
Update .gitignore

mitchcurtis:


.gitignore
```diff
@@ -24,6 +24,13 @@ automappingconverter.exe
 *.lib
 *.exp
 *.intermediate.manifest
+src/libtiled/debug/
+src/tiledquick/debug/
+src/tiledquick/uic_wrapper.bat
+src/tiledquickplugin/debug/
+tiledquick.exe
+tiledquick.ilk
+tiledquick.pdb
```
bjorn: I guess we can just have *.exe, *.ilk and *.pdb.

.gitignore
```diff
@@ -24,6 +24,13 @@ automappingconverter.exe
 *.lib
 *.exp
 *.intermediate.manifest
+src/libtiled/debug/
+src/tiledquick/debug/
+src/tiledquick/uic_wrapper.bat
+src/tiledquickplugin/debug/
```
bjorn: Why the `debug/` directory only for these three projects? And why `uic_wrapper.bat` for `tiledquick`, which doesn't use `uic` at all? (it's actually the first time I see that file)

mitchcurtis: Dunno why uic_wrapper.bat is there; I just added everything that I saw that looked out of place.

mitchcurtis:
There are a bunch of .ui files in tiled, so I think it doesn't hurt having it in wip/qtquick, in case someone forgets to comment out the tiled subdir project in the .pro, for example. :p

bjorn:
Thanks for the updates!

The .gitignore generally goes out of date because I never do an in-source build. Qt Creator performs shadow builds by default for QMake and Qbs always builds in another folder anyway (well, that's why the "default/" entry is there now).

mitchcurtis:
> Thanks for the updates!

No worries!

> The .gitignore generally goes out of date because I never do an in-source build. Qt Creator performs shadow builds by default for QMake and Qbs always builds in another folder anyway (well, that's why the "default/" entry is there now).

Yeah, I generally do shadow builds too, except when I get stuck on build system stuff and need to simplify it. :p

PR #1545, opened 2017-04-16
Add QBS-based quicktiledplugin auto tests

mitchcurtis:


mitchcurtis:
I'm getting an error when building the tests:

```
15:14:33: Running steps for project tests...
compiling tst_quicktiledplugin.cpp
linking tst_quicktiledplugin.exe
C:\dev\Microsoft Visual Studio 14.0\VC\BIN\x86_amd64/link.exe /nologo /DEBUG /PDB:tst_quicktiledplugin.pdb /MACHINE:X64 /OSVERSION:5.02 "/SUBSYSTEM:CONSOLE,5.02" /MANIFEST /MANIFESTFILE:C:\dev\tiled-qt5_dev_debug-Debug\qtc_qt5_dev__07a3bce7-debug\tests\qtc_qt5_dev__07a3bce7-debug\tst-quicktiledplugin.qtc-qt5-dev--07a3bce7.d1623c1c\intermediate.tst_quicktiledplugin.exe.manifest C:\dev\tiled-qt5_dev_debug-Debug\qtc_qt5_dev__07a3bce7-debug\tests\qtc_qt5_dev__07a3bce7-debug\tst-quicktiledplugin.qtc-qt5-dev--07a3bce7.d1623c1c\.obj\3a52ce780950d4d9\tst_quicktiledplugin.cpp.obj Shell32.lib C:\dev\qt5-dev-d\qtbase\lib\qtmaind.lib C:\dev\qt5-dev-d\qtbase\lib\Qt5Cored.lib C:\dev\qt5-dev-d\qtbase\lib\Qt5Guid.lib C:\dev\qt5-dev-d\qtbase\lib\Qt5Networkd.lib C:\dev\qt5-dev-d\qtbase\lib\Qt5Qmld.lib C:\dev\qt5-dev-d\qtbase\lib\Qt5Quickd.lib C:\dev\qt5-dev-d\qtbase\lib\Qt5Testd.lib /OUT:C:\dev\tiled-qt5_dev_debug-Debug\qtc_qt5_dev__07a3bce7-debug\tests\qtc_qt5_dev__07a3bce7-debug\tst-quicktiledplugin.qtc-qt5-dev--07a3bce7.d1623c1c\intermediate.tst_quicktiledplugin.exe /LIBPATH:C:\dev\qt5-dev-d\qtbase\lib
tst_quicktiledplugin.cpp.obj : error LNK2019: unresolved external symbol "__declspec(dllimport) int __cdecl quick_test_main(int,char * *,char const *,char const *)" (__imp_?quick_test_main@@YAHHPEAPEADPEBD1@Z) referenced in function main
C:\dev\tiled-qt5_dev_debug-Debug\qtc_qt5_dev__07a3bce7-debug\tests\qtc_qt5_dev__07a3bce7-debug\tst-quicktiledplugin.qtc-qt5-dev--07a3bce7.d1623c1c\intermediate.tst_quicktiledplugin.exe : fatal error LNK1120: 1 unresolved externals
Process failed with exit code 1120.
The following products could not be built for configuration qtc_qt5_dev__07a3bce7-debug:
	tst_quicktiledplugin
Error while building/deploying project tests (kit: qt5-dev-debug)
When executing step "Qbs Build"
15:14:35: Elapsed time: 00:01.
```

Do you know what I'm doing wrong? :)

mitchcurtis:
That is, I can see that it's failing to link to Qt5QuickTestd.lib, but I don't know what the submodule name for that is. I tried "quicktest", but no cigar. http://doc.qt.io/qbs/qt-modules.html#list-of-submodules doesn't mention anything.

mitchcurtis:
I'm going to assume it doesn't exist: https://bugreports.qt.io/browse/QBS-1120

I wonder if I can just link to it manually...

mitchcurtis:
I'm making progress slowly. I'm now finding a variable that I can use to set up the QML2_IMPORT_PATH. For now I've hard-coded it. Same with tiled.dll, which quicktiledplugin requires. Hard-coding a PATH entry doesn't seem to work, so I guess I'll have to find out how to copy the DLL. Any tips welcome. :)

mitchcurtis:
Can you elaborate a bit on why you don't like installing stuff to the Qt's "qml"/"lib" folders? I feel like all of these issues could be solved if that was how it was done. Qt would look in the standard places and sort it all out automatically. Best of all, users wouldn't need to add this file-copying code to their projects that use tiledquickplugin.

tests/quicktiledplugin/quicktiledplugin.qbs
```diff
@@ -0,0 +1,26 @@
...
+    Depends { name: "Qt"; submodules: ["gui", "testlib", "quick"] }
+
+    cpp.dynamicLibraries: {
+        var suffix = qbs.targetOS.contains("windows") && qbs.buildVariant === "debug" ? "d" : "";
+        return [ "Qt5QuickTest" + suffix ];
+    }
+
+    qbs.setupRunEnvironment: {
+        // TODO: don't hard-code qml directory
+        Environment.putEnv("QML2_IMPORT_PATH", "C:/dev/tiled-qt5_dev_debug-Debug/qtc_qt5_dev__07a3bce7-debug/install-root/qml")
```
jakepetroules: You want:

    import qbs.FileInfo
    ...
    Environment.putEnv("QML2_IMPORT_PATH", FileInfo.joinPaths(qbs.installRoot, "qml"))

bjorn:
> Can you elaborate a bit on why you don't like installing stuff to the Qt's "qml"/"lib" folders? I feel like all of these issues could be solved if that was how it was done.

Well, there's two reasons why I think it's not nice to rely on files installed in Qt's folders:

* On Linux, those folders are usually part of the system, so you'd require installing files as root, which is something you'd avoid whenever possible during development. It is also not nice for people who want to try it out and compile it themselves, since it means they need to mess with their system and it makes it hard for them to clean up everything afterwards.

* I often switch between multiple branches checked out in different directories. When the files are installed into the Qt folder, this means the QML plugin would need to be overwritten each time, and I couldn't easily run two versions in parallel.

> Qt would look in the standard places and sort it all out automatically. Best of all, users wouldn't need to add this file-copying code to their projects that use tiledquickplugin.

I think there is actually no need for any file copying code apart from the usual installation rules that are needed anyway. In the end, also Tiled Quick will need to be distributed in various platform-specific ways, though on all supported platforms it means packaging all needed files into some kind of archive. You can never rely on the user installing the QML plugin into Qt's folders, except on Linux when Tiled is packaged by the distribution.

Thanks a lot for working to add autotests for the Tiled Quick plugin, btw!

tests/quicktiledplugin/quicktiledplugin.qbs
mitchcurtis: Ah, thanks Jake!

mitchcurtis:
> Well, there's two reasons why I think it's not nice to rely on files installed in Qt's folders:

OK, those are pretty good reasons. :p

> I think there is actually no need for any file copying code apart from the usual installation rules that are needed anyway

What about tiled.dll? I still have no idea how to copy that sucker somewhere where the auto tests can see it. Or perhaps there's a way to add it to some environment variable so that I don't need to copy it? Either way, I'm stuck at this point.

> Thanks a lot for working to add autotests for the Tiled Quick plugin, btw!

No worries. :)

mitchcurtis:
> What about tiled.dll? I still have no idea how to copy that sucker somewhere where the auto tests can see it. Or perhaps there's a way to add it to some environment variable so that I don't need to copy it? Either way, I'm stuck at this point.

Ahhh, I think I see it now... everything is installed to "install-root". It looks like tiledquick does this with the following QBS:

    Group {
        qbs.install: true
        qbs.installDir: {
            if (qbs.targetOS.contains("windows") || qbs.targetOS.contains("osx"))
                return ""
            else
                return "bin"
        }
        fileTagsFilter: product.type
    }

Is that correct?

Either way, that's probably not appropriate for auto tests, since they have to stay where they're built...

bjorn:
> Or perhaps there's a way to add it to some environment variable so that I don't need to copy it? Either way, I'm stuck at this point.

Windows will find the `tiled.dll` if it is in the `PATH` environment variable, so I guess you could add it there using `qbs.setupRunEnvironment`.

mitchcurtis:
> Windows will find the tiled.dll if it is in the PATH environment variable, so I guess you could add it there using qbs.setupRunEnvironment.

Hmm, [last time I tried that](https://github.com/bjorn/tiled/pull/1545#issuecomment-294356813) it didn't work.

jakepetroules:
> Is that correct?

"osx" is now "macos" in qbs >= 1.5.2. "Correct" is also subjective here.

> Windows will find the tiled.dll if it is in the PATH environment variable, so I guess you could add it there using qbs.setupRunEnvironment.

And you need DYLD_FRAMEWORK_PATH for Darwin (macOS) and LD_LIBRARY_PATH for other Unix-like platforms. Good example being: https://github.com/qt-labs/qbs/blob/master/src/lib/qtprofilesetup/templates/core.qbs#L211

jakepetroules:
...and if setupRunEnvironment isn't working, it could be a bug, please [report it](https://bugreports.qt.io/browse/QBS/).

mitchcurtis:
Here's what I've currently got:

```
import qbs
import qbs.Environment

QtGuiApplication {
    name: "tst_quicktiledplugin"
    type: ["application", "autotest"]
    targetName: name

    Depends { name: "cpp" }
    Depends { name: "Qt"; submodules: ["gui", "testlib", "quick"] }

    cpp.dynamicLibraries: {
        var suffix = qbs.targetOS.contains("windows") && qbs.buildVariant === "debug" ? "d" : "";
        return [ "Qt5QuickTest" + suffix ];
    }

    qbs.setupRunEnvironment: {
        // TODO: don't hard-code qml directory
        Environment.putEnv("QML2_IMPORT_PATH", "C:/dev/tiled-qt5_dev_debug-Debug/qtc_qt5_dev__07a3bce7-debug/install-root/qml")

        var installRootPath = "../../install-root";
        var env;
        if (qbs.targetOS.contains('windows')) {
            env = new ModUtils.EnvironmentVariable("PATH", qbs.pathListSeparator, true);
            env.append(installRootPath);
            env.set();
        }
    }

    files: [
        "tst_quicktiledplugin.cpp"
    ]
}
```

I'm getting a lot of

> Error retrieving run environment: C:\dev\tiled\tests\quicktiledplugin\quicktiledplugin.qbs:17:30 Error while setting up run environment: TypeError: Result of expression 'qbs.targetOS' [undefined] is not an object.

Do you know how I can get an absolute path to the install-root directory of a parent project? For example, my test is in C:\dev\tiled-qt5_dev_debug-Debug\qtc_qt5_dev__07a3bce7-debug\tests\qtc_qt5_dev__07a3bce7-debug\tst-quicktiledplugin.qtc-qt5-dev--07a3bce7.d1623c1c, and it needs to have C:\dev\tiled-qt5_dev_debug-Debug\qtc_qt5_dev__07a3bce7-debug\install-root in its PATH before running.

mitchcurtis:
Getting closer:

    qbs.setupRunEnvironment: {
        // TODO: don't hard-code qml directory
        Environment.putEnv("QML2_IMPORT_PATH", "C:/dev/tiled-qt5_dev_debug-Debug/qtc_qt5_dev__07a3bce7-debug/install-root/qml")

        if (targetOS.contains('windows')) {
            var PATH = Environment.getEnv("PATH");
            Environment.putEnv("PATH", PATH + ";" + project.buildDirectory + "/install-root");
            console.info(Environment.getEnv("PATH"))
        }
    }

> Error retrieving run environment: C:\dev\tiled\tests\quicktiledplugin\quicktiledplugin.qbs:17:30 Error while setting up run environment: ReferenceError: Can't find variable: project

I see `project.buildDirectory` and `project.destinationDirectory` mentioned [here](http://doc.qt.io/qbs/product-item.html), but I can't use them apparently.

mitchcurtis:
If I do something like this:

    property string buildDir: project.buildDirectory

    qbs.setupRunEnvironment: {
        // TODO: don't hard-code qml directory
        Environment.putEnv("QML2_IMPORT_PATH", "C:/dev/tiled-qt5_dev_debug-Debug/qtc_qt5_dev__07a3bce7-debug/install-root/qml")

        if (targetOS.contains('windows')) {
            var PATH = Environment.getEnv("PATH");
            Environment.putEnv("PATH", PATH + ";" + buildDir + "/install-root");
            console.info(Environment.getEnv("PATH"))
        }
    }

> Error retrieving run environment: C:\dev\tiled\tests\quicktiledplugin\quicktiledplugin.qbs:19:30 Error while setting up run environment: ReferenceError: Can't find variable: buildDir

Why does it not work like standard QML? That property should definitely be visible.

mitchcurtis:
As mentioned before, even if I hard-code the PATH entry

    qbs.setupRunEnvironment: {
        // TODO: don't hard-code qml directory
        Environment.putEnv("QML2_IMPORT_PATH", "C:/dev/tiled-qt5_dev_debug-Debug/qtc_qt5_dev__07a3bce7-debug/install-root/qml")

        if (targetOS.contains('windows')) {
            var PATH = Environment.getEnv("PATH");
//            Environment.putEnv("PATH", PATH + ";" + buildDir + "/install-root");
            Environment.putEnv("PATH", PATH + ";" + "C:/dev/tiled-qt5_dev_debug-Debug/qtc_qt5_dev__07a3bce7-debug/install-root");
            console.info(Environment.getEnv("PATH"))
        }
    }

It still fails to find tiled.dll (though the path is definitely correct: `C:/dev/tiled-qt5_dev_debug-Debug/qtc_qt5_dev__07a3bce7-debug/install-root`).

bjorn:
> And you need DYLD_FRAMEWORK_PATH for Darwin (macOS) and LD_LIBRARY_PATH for other Unix-like platforms. Good example being: https://github.com/qt-labs/qbs/blob/master/src/lib/qtprofilesetup/templates/core.qbs#L211

At least on Linux this problem is solved by RPATH. I think that's what makes the existing autotests run fine.

mitchcurtis:
Can re-open if anyone ever finds out how to get it to find the .dll.

PR #1546, opened 2017-04-16
Make MapLoader support qrc sources

mitchcurtis:


mitchcurtis:
The docs for [`isLocalFile`](http://doc.qt.io/qt-5/qurl.html#isLocalFile) say:

> A URL is a local file path if the scheme is "file"

src/tiledquickplugin/maploader.cpp
```diff
@@ -46,7 +46,9 @@ void MapLoader::setSource(const QUrl &source)
 
     Tiled::MapReader mapReader;
 
-    Tiled::Map *map = mapReader.readMap(source.toLocalFile());
+    const QString qrcScheme = QStringLiteral("qrc");
+    const bool isQrcScheme = source.scheme() == qrcScheme;
+    Tiled::Map *map = mapReader.readMap(isQrcScheme ? source.toString().mid(qrcScheme.length()) : source.toLocalFile());
```
bjorn: I don't understand. If the file to be read is a `qrc` file, you're taking the `qrc` prefix out? Why not just keep it? Or maybe you could try `source.path()` instead?

mitchcurtis: My experience has been that you generally only use the qrc scheme in QML - when it comes to C++, it's always ":/my/path". That's probably a simplified understanding of the problem, but it is supported by e.g. [this code](url). If you think the way I've done it is weird, I could copy that code.

bjorn: Right, I'd include a copy of `urlToLocalFileOrQrc` then and use that. It's unfortunate the QQmlFile class is `\internal`, so we shouldn't use it directly even though it's in public headers.

PR #1561, opened 2017-05-06
Expose map and tile size as properties

mitchcurtis:


bjorn:
That's cool, but please keep all the simple getters inline.

bjorn:
Thanks! :-)

PR #1562, opened 2017-05-07
MapItem: emit mapChanged() after calling refresh()

mitchcurtis:
This allows QML code to be guaranteed that MapItem has a renderer by
the time onMapChanged is called, which is necessary to be able to
e.g. set a transform using e.g. pixelToScreenCoords().


bjorn:
Good idea!

PR #1566, opened 2017-05-10
Add objects filter to the objects dock

### LINKED ISSUE #1467: Being able to search objects

thabetx:
Fixes #1467 .

filter06

The filter text box supports two shortcuts

- Escape clears the filter.
- Return focuses the first filtered item, pressing return while an object is focused still changes the focus to the properties dock.

Also each map has its own search string so switching between maps will change the filter.

All columns are matched (name, type, id, position) even if they are hidden. Group names are not matched.

Not entirely sure about the above decisions so would like some feedback regarding them.

Some stuff that are still missing
- Expanded groups that didn't match will collapse after clearing the search, I'm still looking into it.
- Matching custom properties.


thabetx:
Now, any group having a matching object will be expanded while filtering, so the above the GIF is not very accurate.

I also updated the behavior of the return key to just focus the tree instead of focusing the first item.

I will fix expanding pre-filter expanded groups that didn't match.

src/tiled/objectsdock.cpp
```diff
@@ -454,3 +509,44 @@ void ObjectsView::synchronizeSelectedItems()
...
+bool ObjectsFilterModel::filterAcceptsRow(int sourceRow, const QModelIndex &sourceParent) const
+{
+    // Show all if filter string is empty. Prevents hiding empty groups
+    if(filterRegExp().isEmpty())
+        return true;
+
+    QModelIndex index = sourceModel()->index(sourceRow, 0, sourceParent);
+
+    // sourceParent of a group has no model
+    if(!sourceParent.model())
```
bjorn: This is a somewhat interesting and not entirely correct check. There is no model, because the parent index is invalid, so a more sensible check may be `if (!sourceParent.isValid())`. However, either check won't work since layers now form a hierarchy and an object layer within a group layer _will_ have a valid parent, which would lead those object layers to be filtered as if they were an object.

Of course, group layers are a problem of their own as well, since whether they should be filtered or not would depend on whether any of their child layers are object groups that contain any non-filtered objects.

Fortunately, we know we're dealing with a `MapObjectModel` and I would suggest working through its interface directly instead of requesting data through `Qt::DisplayRole`. That way, you can use `toGroupLayer`, `toObjectGroup` and `toMapObject`. And to avoid having to convert between `ReversingProxyModel` indexes here, I would change the order in which the two proxies are applied so that the filtering happens before the reversing.

src/tiled/objectsdock.cpp
```diff
@@ -454,3 +509,44 @@ void ObjectsView::synchronizeSelectedItems()
...
+        return groupHasAnyMatchingObjects(index);
+    else
+        return objectContainsFilterString(index);
+}
+
+bool ObjectsFilterModel::groupHasAnyMatchingObjects(const QModelIndex index) const
+{
+    for (int i = 0; i < sourceModel()->rowCount(index); ++i) {
```
bjorn: In my previous comment I suggested to replace this code, but just a note here: since `sourceModel()->rowCount(index)` is relatively slow the result should really be saved in a local variable for use in the loop.

src/tiled/objectsdock.cpp
```diff
@@ -454,3 +509,44 @@ void ObjectsView::synchronizeSelectedItems()
...
+        QModelIndex childIndex = sourceModel()->index(i, 0, index);
+        if (childIndex.isValid() && objectContainsFilterString(childIndex))
```
bjorn: The `childIndex.isValid()` should not be necessary since you're only requesting data from valid indexes.

src/tiled/objectsdock.cpp
```diff
@@ -454,3 +509,44 @@ void ObjectsView::synchronizeSelectedItems()
...
+    }
+    return false;
+}
+
+bool ObjectsFilterModel::objectContainsFilterString(const QModelIndex index) const
+{
+    for (int i = 0; i < sourceModel()->columnCount(index); ++i) {
+        QModelIndex objectIndex = sourceModel()->index(index.row(), i, index.parent());
+        QString type = sourceModel()->data(objectIndex, Qt::DisplayRole).toString();
+        if (type.contains(filterRegExp()))
```
bjorn: "type" is a confusing variable name here.

src/tiled/objectsdock.h
```diff
@@ -57,23 +58,32 @@ private slots:
...
     MapDocument *mMapDocument;
+
+    // These maps store the state of the objects dock for each opened map document
     QMap<MapDocument*, QList<ObjectGroup*> > mExpandedGroups;
+    QMap<MapDocument*, QList<ObjectGroup*> > mPrefilterExpandedGroups;
+    QMap<MapDocument*, QString> mFilterStrings;
+    QMap<MapDocument*, bool> mFilterWasEmpty;
+
     QMenu *mMoveToMenu;
+    QModelIndex getGroupIndex(ObjectGroup *og);
```
bjorn: Would be called `groupIndex` according to naming style, but I think it should instead be a convenience override to the previously suggested `mapToViewModel` helper function. Should be marked `const`.

src/tiled/objectsdock.h
```diff
@@ -107,9 +119,35 @@ private slots:
...
+{
+    return mProxyModel;
+}
+
+class ObjectsFilterModel : public QSortFilterProxyModel
+{
+    Q_OBJECT
+
+public:
+    ObjectsFilterModel(QObject *parent = 0);
```
bjorn: Please use `nullptr` instead of `0`.

src/tiled/objectsdock.h
```diff
@@ -107,9 +119,35 @@ private slots:
...
+
+protected:
+    bool filterAcceptsRow(int sourceRow, const QModelIndex &sourceParent) const override;
+
+private :
+    bool groupHasAnyMatchingObjects(const QModelIndex index) const;
+    bool objectContainsFilterString(const QModelIndex index) const;
```
bjorn: Arguments are missing a `&`.

thabetx:
Alright, thank you for the review, I will be updating the pull request.

I've a question though, suppose we have an expanded group, then we applied a filter and one of the objects of this group matched the search, then we collapsed this group, when cancelling the search it is restored to the initial state(expanded) although we just collapsed it, is this behavior expected?

I saw one application (gitkraken) that prevented changing the expansion state of any group while a filter is applied, this will solve this case but is kind of restrictive, what do you think?

bjorn:
> I've a question though, suppose we have an expanded group, then we applied a filter and one of the objects of this group matched the search, then we collapsed this group, when cancelling the search it is restored to the initial state(expanded) although we just collapsed it, is this behavior expected?

I guess it could be helpful to keep it collapsed given the user's disinterest in this group. So instead of collapsing everything and restoring from `mPrefilterExpandedGroups`, you could instead just collapse any group that isn't in it that list.

> I saw one application (gitkraken) that prevented changing the expansion state of any group while a filter is applied, this will solve this case but is kind of restrictive, what do you think?

I agree that this is too restrictive.

Btw, it occurred to me that for expanding the tree when the search filter is changed you should just call `mObjectsView->expandAll()`. The current method of expanding all object groups won't work in case of group layers.

bjorn:
Actually on second thought, it occurred to me that allowing the user to collapse the tree is not very useful when on any typed character we will automatically expand the whole thing again. So maybe disabling `itemsExpandable` on the view while a filter is active isn't a bad idea after all...

thabetx:
Ok I will test disabling collapsing. This might also simplify the code as mExpandedGroups and mPrefilterExpandedGroups can be merged into one in that case again.

src/tiled/objectsdock.cpp
```diff
@@ -221,47 +233,90 @@ void ObjectsDock::objectProperties()
...
+        mFilterStrings.remove(mapDocument);
         mExpandedGroups.remove(mapDocument);
+    }
+}
+
+void ObjectsDock::filterObjects()
+{
+    // Save the prefilter expansion state if this is the first applied filter
+    if(mFilterWasEmpty[mMapDocument]) {
+        mFilterWasEmpty[mMapDocument] = false;
```
thabetx: I tried doing that but couldn't get it to work properly.

The filter is shared between documents, so suppose we switched from a non-filtered map to a filtered map, the above condition will be satisfied and the state of the new map will be saved overwriting the original saved state, I couldn't figure a clean way to prevent this from happening.

I changed the mFilterWasEmpty map to a single bool variable and got it work properly but had to introduce a small hack for this particular case.

bjorn: Ah, makes sense, thanks for the explanation!

src/tiled/objectsdock.cpp
thabetx: I agree on working directly with the objects, I think it  will be better(using toMapObject), I will try doing that and will refactor the proxies code.

src/tiled/objectsdock.h
thabetx: I'm not sure what you mean by convenience override.

But here is what I did so far, I moved the getGroupIndex function from the objects dock to the objects view (it uses mapToViewModel for conversion), then made it private and made public helper functions in the objects view(isGroupExpanded, setGroupExpanded), so that the objectsDock can use them without worrying about the conversion.

thabetx:
Hey @bjorn, thank you for the great review, I've addressed the requests, the small notes were also helpful.

Filtering now works directly on the objects, it made the filtering code much simpler. I think this will make it possible to filter by custom properties.

I would like to receive some usability reviews, I think the behavior of the filtering can be improved.

thabetx:
Things don't work properly when moving objects up and down, will investigate it.

thabetx:
For some reason reversing the proxy models stopped deleting the objects or moving them up an down from working, I tried restoring it to the previous order and the functions worked fine. This will require adding the missing translation step when filtering, although it works without it.

I thought of introducing two extra QAbsractProxyModel pointers that point to the two proxies, called mFrontEndProxyModel and mBackEndProxyModel, I think this makes the code more readable and changing the order of the proxies is just a matter of changing what those pointers point to, I'm not sure if this is a good practice though.

Another thing to notice when moving objects up and down is that the order of the objects is the same as when the objects are not filtered, so moving-up an object while filtering might not visually move it up in the view because there might be objects above that are hidden, I think this is expected from the code, but don't know what to do about it.

bjorn:
> I would like to receive some usability reviews, I think the behavior of the filtering can be improved.

Can you elaborate a bit about this?

> Things don't work properly when moving objects up and down, will investigate it.

I'll have a look at this as well. Sounds like an annoying thing to debug... Of course the order of the proxy models shouldn't matter at all. :-/

> I thought of introducing two extra QAbsractProxyModel pointers that point to the two proxies, called mFrontEndProxyModel and mBackEndProxyModel, I think this makes the code more readable and changing the order of the proxies is just a matter of changing what those pointers point to, I'm not sure if this is a good practice though.

I prefer the more explicit names.

> Another thing to notice when moving objects up and down is that the order of the objects is the same as when the objects are not filtered, so moving-up an object while filtering might not visually move it up in the view because there might be objects above that are hidden, I think this is expected from the code, but don't know what to do about it.

I think it is reasonable to not expect those actions to adjust to any active search filter.

thabetx:
About the improvements, I thought it'd be useful to receive feedback from the community, I don't have much in mind other than keeping groups with selected objects expanded when cancelling the search, adding a shortcut to focus the search bar and currently when filtering, collapsing is disabled, but there is no indication for that, so needed some opinions about it.

bjorn:
Ohoh, those conflicts in `objectsdock.cpp` are looking quite severe...

bjorn:
After fixing the issue with the `ReversingProxyModel` by basing it on the `QSortFilterProxyModel` in change e24860400597c4e2d4309faabcf1fd06275a82ee, I decided to reimplement this change instead of trying to resolve the conflicts.

My patch does not remember a search filter for each document, but I'm anyway not sure what is the better behavior and it was easier to leave that functionality out for now.

Issue #642, opened 2014-04-16
Add autocrop


alphaonex86:
Hello,
As gimp, can you add auto crop to crop the empty (no object, no tile, no image, ...) and useless border.
It's more fast to map than select only the util zone and crop the selected zone.
Cheers,

ketanhwr:
The simple solution would be to resize the map based on (minX,minY) to (maxX,maxY), right? (Where min and max would be based on extreme coordinates of any placed tile.)

Altough I'm a bit confused as in how this would work when we're using multiple layers.

bjorn:
> Although I'm a bit confused as in how this would work when we're using multiple layers.

You could just go through all layers and compute min/max based on that, but that won't work very nicely when in the future we decouple layer size and tile size from the map.

But, I just checked what GIMP does and actually it crops the image only based on the currently selected layer. So, surely that is acceptable behavior as well.

Then, the only remaining question is how to treat object layers. We could either just not support this option for now, or determine the boundaries of the objects on the current object layer and then somehow mapping that also to any potential tile layers. I think at first we could just disable this action when an object layer is selected.

ketanhwr:
Yes, alright. So I'll add an `Autocrop` option in the same menu where the resize menu is placed. What should be the suitable shortcut key for this, if needed?

bjorn:
> What should be the suitable shortcut key for this, if needed?

I think it is a quite specific and rare action, so no shortcut is needed. GIMP also doesn't have a shortcut for it.

PR #1574, opened 2017-05-21
Adds Autocrop feature

ketanhwr:
This addresses #642. The option is available in the 'Map' menu and is available only when a Tile Layer is selected. The implementation is pretty straightforward:

- First compute the `(minX, minY)` and `(maxX, maxY)` by checking all the cells of the current layer.
- Resize the map using this region.
- Works similar to `Crop to Selection`, except, the region is calculated using `TileLayer::computeAutocropRegion()`

### DIFF
--- a/src/libtiled/tilelayer.cpp
+++ b/src/libtiled/tilelayer.cpp
@@ -572,6 +572,30 @@ QRegion TileLayer::computeDiffRegion(const TileLayer *other) const
     return ret;
 }
 
+QRegion TileLayer::computeAutocropRegion() const
+{
+    int minX = mWidth, minY = mHeight;
+    int maxX = 0, maxY = 0;
+
+    if (isEmpty()) {
+        return QRegion();
+    }
+
+    for (int y = 0; y < mHeight; ++y) {
+        for (int x = 0; x < mWidth; ++x) {
+
+            if (!cellAt(x, y).isEmpty()) {
+                minX = qMin(x, minX);
+                maxX = qMax(x, maxX);
+                minY = qMin(y, minY);
+                maxY = qMax(y, maxY);
+            }
+        }
+    }
+
+    return QRegion(minX, minY, maxX-minX+1, maxY-minY+1);
+}
+
 bool TileLayer::isEmpty() const
 {
     for (const Cell &cell : mGrid)
--- a/src/libtiled/tilelayer.cpp
+++ b/src/libtiled/tilelayer.cpp
@@ -572,29 +572,6 @@ QRegion TileLayer::computeDiffRegion(const TileLayer *other) const
     return ret;
 }
 
-QRegion TileLayer::computeAutocropRegion() const
-{
-    int minX = mWidth, minY = mHeight;
-    int maxX = 0, maxY = 0;
-
-    if (isEmpty()) {
-        return QRegion();
-    }
-
-    for (int y = 0; y < mHeight; ++y) {
-        for (int x = 0; x < mWidth; ++x) {
-            if (!cellAt(x, y).isEmpty()) {
-                minX = qMin(x, minX);
-                maxX = qMax(x, maxX);
-                minY = qMin(y, minY);
-                maxY = qMax(y, maxY);
-            }
-        }
-    }
-
-    return QRegion(minX, minY, maxX-minX+1, maxY-minY+1);
-}
-
 bool TileLayer::isEmpty() const
 {
     for (const Cell &cell : mGrid)
--- a/src/libtiled/tilelayer.cpp
+++ b/src/libtiled/tilelayer.cpp
@@ -583,7 +583,6 @@ QRegion TileLayer::computeAutocropRegion() const
 
     for (int y = 0; y < mHeight; ++y) {
         for (int x = 0; x < mWidth; ++x) {
-
             if (!cellAt(x, y).isEmpty()) {
                 minX = qMin(x, minX);
                 maxX = qMax(x, maxX);
--- a/src/tiled/mainwindow.cpp
+++ b/src/tiled/mainwindow.cpp
@@ -246,7 +246,7 @@ MainWindow::MainWindow(QWidget *parent, Qt::WindowFlags flags)
     mUi->menuEdit->insertSeparator(mUi->actionPreferences);
 
     mUi->menuMap->insertAction(mUi->actionOffsetMap,
-                                mActionHandler->actionCropToSelection());
+                               mActionHandler->actionCropToSelection());
 
     mUi->menuMap->insertAction(mUi->actionOffsetMap,
                                 mActionHandler->actionAutocrop());
--- a/src/tiled/mainwindow.cpp
+++ b/src/tiled/mainwindow.cpp
@@ -249,7 +249,7 @@ MainWindow::MainWindow(QWidget *parent, Qt::WindowFlags flags)
                                mActionHandler->actionCropToSelection());
 
     mUi->menuMap->insertAction(mUi->actionOffsetMap,
-                                mActionHandler->actionAutocrop());
+                               mActionHandler->actionAutocrop());
 
     mLayerMenu = new QMenu(tr("&Layer"), this);
     mNewLayerMenu = mActionHandler->createNewLayerMenu(mLayerMenu);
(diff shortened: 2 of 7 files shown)


src/libtiled/tilelayer.cpp
```diff
@@ -572,6 +572,30 @@ QRegion TileLayer::computeDiffRegion(const TileLayer *other) const
...
+    int minX = mWidth, minY = mHeight;
+    int maxX = 0, maxY = 0;
+
+    if (isEmpty()) {
+        return QRegion();
+    }
+
+    for (int y = 0; y < mHeight; ++y) {
+        for (int x = 0; x < mWidth; ++x) {
+
```
Ablu: There is an unneeded empty line here

src/tiled/mainwindow.cpp
```diff
@@ -246,7 +246,10 @@ MainWindow::MainWindow(QWidget *parent, Qt::WindowFlags flags)
     mUi->menuEdit->insertSeparator(mUi->actionPreferences);
 
     mUi->menuMap->insertAction(mUi->actionOffsetMap,
-                               mActionHandler->actionCropToSelection());
+                                mActionHandler->actionCropToSelection());
```
Ablu: you changed the indentation here

src/tiled/mapdocument.cpp
```diff
@@ -335,6 +335,19 @@ void MapDocument::resizeMap(const QSize &size, const QPoint &offset, bool remove
...
+void MapDocument::autocropMap()
+{
+    // Since autocrop option is allowed only for Tile Layer
+    TileLayer* tileLayer = static_cast<TileLayer*>(mCurrentLayer);
+
+    const QRect bounds = tileLayer->computeAutocropRegion().boundingRect();
+    if (bounds.isNull())
+        return;
+
+    resizeMap(bounds.size(), -bounds.topLeft(), true);
```
Ablu: Hm... The boundary check only considers tiles... @bjorn Is it fine to silently delete objects (the third parameter) which are not in the area here?

src/tiled/mapdocument.cpp
```diff
@@ -335,6 +335,19 @@ void MapDocument::resizeMap(const QSize &size, const QPoint &offset, bool remove
     // TODO: Handle layers that don't match the map size correctly
 }
 
+
+void MapDocument::autocropMap()
+{
+    // Since autocrop option is allowed only for Tile Layer
+    TileLayer* tileLayer = static_cast<TileLayer*>(mCurrentLayer);
```
Ablu: It should be `TileLayer *tileLayer`. We try to stick the pointer to the variablename.

src/tiled/mainwindow.cpp
ketanhwr: Yes it wasn't matching up with the others..so that's why.

ketanhwr: I'll revert this one.

src/tiled/mapdocument.cpp
ketanhwr: The option of `Crop to Selection` behaves in a similar way. It also deletes the objects outside the selected region. I took that under consideration.

bjorn: Yes, I think it's reasonable for this to be consistent with "Crop to Selection". There isn't really an opportunity to ask the user here, unless we'd introduce a popup, which I don't really want to.

src/tiled/mapdocument.cpp
```diff
@@ -335,6 +335,19 @@ void MapDocument::resizeMap(const QSize &size, const QPoint &offset, bool remove
     // TODO: Handle layers that don't match the map size correctly
 }
 
+
+void MapDocument::autocropMap()
+{
+    // Since autocrop option is allowed only for Tile Layer
+    TileLayer *tileLayer = static_cast<TileLayer*>(mCurrentLayer);
+
+    const QRect bounds = tileLayer->computeAutocropRegion().boundingRect();
```
bjorn: I don't see the need for a special `computeAutocropRegion` function. Just use `tileLayer->region().boundingRect()`.

src/tiled/mapdocument.cpp
```diff
@@ -335,6 +335,19 @@ void MapDocument::resizeMap(const QSize &size, const QPoint &offset, bool remove
     // TODO: Handle layers that don't match the map size correctly
 }
 
+
+void MapDocument::autocropMap()
+{
+    // Since autocrop option is allowed only for Tile Layer
```
bjorn: Since you never know when this function may get called in the future, at this level I'm usually inclined to insert an explicit check here, like:
```c++
if (!mCurrentLayer || !mCurrentLayer->isTileLayer())
    return;
```

src/tiled/mapdocumentactionhandler.cpp
```diff
@@ -425,6 +428,14 @@ void MapDocumentActionHandler::cropToSelection()
     mMapDocument->resizeMap(bounds.size(), -bounds.topLeft(), true);
 }
 
+void MapDocumentActionHandler::autocrop()
+{
+    if (!mMapDocument)
+        return;
+
+    mMapDocument->autocropMap();
```
bjorn: Please write it like this, mostly because it's shorter and consistent with other functions in this file:
```c++
if (mMapDocument)
    mMapDocument->autocropMap();
```
The reason why I often will do an "early exit" like you did here is to avoid indenting a lot of code or to make the function more readable, but that does not apply here.

src/tiled/mapdocumentactionhandler.cpp
```diff
@@ -655,13 +666,17 @@ void MapDocumentActionHandler::updateActions()
...
         } else if (currentLayer->asObjectGroup()) {
             mActionSelectNone->setEnabled(selectedObjectsCount  > 0);
+            mActionAutocrop->setEnabled(false);
         } else {
             mActionSelectNone->setEnabled(false);
+            mActionAutocrop->setEnabled(false);
         }
     } else {
         mActionSelectNone->setEnabled(false);
+        mActionAutocrop->setEnabled(false);
```
bjorn: It will be more readable to put one line below together with `mActionCropToSelection` like this:
```c++
mActionAutocrop->setEnabled(currentLayer && currentLayer->isTileLayer());
```
The `mActionSelectNone` needed all this special code only because the condition on whether it should be enabled or not varies depending on the type of the selected layer, but let's not "abuse" all those cases for setting the state of `mActionAutocrop`.

src/tiled/mapdocumentactionhandler.h
```diff
@@ -104,6 +105,8 @@ public slots:
 
     void cropToSelection();
```
bjorn: I group similar functions together here, and I think there shouldn't be an empty line between `cropToSelection` and `autocrop` given their similar function.

src/tiled/mapdocumentactionhandler.cpp
ketanhwr: Nice idea :slightly_smiling_face:

src/tiled/mapdocument.cpp
ketanhwr: Oh, I totally missed this function!

src/tiled/mainwindow.cpp
```diff
@@ -248,6 +248,9 @@ MainWindow::MainWindow(QWidget *parent, Qt::WindowFlags flags)
     mUi->menuMap->insertAction(mUi->actionOffsetMap,
                                mActionHandler->actionCropToSelection());
 
+    mUi->menuMap->insertAction(mUi->actionOffsetMap,
+                                mActionHandler->actionAutocrop());
```
bjorn: The indentation rule for wrapping function parameters is to align them to the opening `(`, so there is one space too many here.

src/tiled/mapdocument.cpp
```diff
@@ -335,6 +335,21 @@ void MapDocument::resizeMap(const QSize &size, const QPoint &offset, bool remove
     // TODO: Handle layers that don't match the map size correctly
 }
 
+
```
bjorn: I generally only separate methods by one empty line, so there is essentially one empty line too much here.

ketanhwr:
@bjorn I think you forgot this one :slightly_smiling_face:

bjorn:
@ketanhwr Indeed, I haven't created an `1.0` branch yet, but I implicitly ended the string freeze by merging #1580, so this one can go in now as well. Thanks for the reminder! :-)

PR #1580, opened 2017-05-29
Allow setting working directory

ketanhwr:
This addresses #941. I've also split 'Command' to 'Executable' and 'Arguments'. I've shown the design in the issue and one of the reasons for this was `..the "Browse" button to choose the executable currently resets the command line arguments.`

A screenshot of what I've done:
tiled-wd

### DIFF
--- a/src/tiled/command.cpp
+++ b/src/tiled/command.cpp
@@ -34,56 +34,58 @@ QString Command::finalWorkingDirectory() const
 {
     QString finalWorkingDirectory = workingDirectory;
 
-    // Perform variable replacement
-    if (Document *document = DocumentManager::instance()->currentDocument()) {
-        const QString fileName = document->fileName();
-
-        QFileInfo fileInfo(fileName);
-        QString mapPath = fileInfo.absolutePath();
-        finalWorkingDirectory.replace(
-            QLatin1String("%mappath"),
-            QString(QLatin1String("%1")).arg(mapPath));
-        // QProcess::setWorkingDirectory doesn't work with quoted paths
-
-        // TODO: %executablepath
-    }
-
-    return finalWorkingDirectory;
+    return replaceVariables(finalWorkingDirectory, true);
 }
 
 QString Command::finalCommand() const
 {
     QString finalCommand = QString(QLatin1String("%1 %2")).arg(executable).arg(arguments);
 
+    return replaceVariables(finalCommand);
+}
+
+QString Command::replaceVariables(const QString &string, bool flag) const
+{
+    QString finalString = string;
+
     // Perform variable replacement
     if (Document *document = DocumentManager::instance()->currentDocument()) {
         const QString fileName = document->fileName();
 
-        finalCommand.replace(QLatin1String("%mapfile"),
+        finalString.replace(QLatin1String("%mapfile"),
                              QString(QLatin1String("\"%1\"")).arg(fileName));
 
         QFileInfo fileInfo(fileName);
         QString mapPath = fileInfo.absolutePath();
-        finalCommand.replace(
-            QLatin1String("%mappath"),
-            QString(QLatin1String("\"%1\"")).arg(mapPath));
+
+        if (flag) {
+            finalString.replace(
+                QLatin1String("%mappath"),
+                QString(QLatin1String("%1")).arg(mapPath));
+        } else {
+            finalString.replace(
+                QLatin1String("%mappath"),
+                QString(QLatin1String("\"%1\"")).arg(mapPath));
+        }
 
         if (MapDocument *mapDocument = qobject_cast<MapDocument*>(document)) {
             if (const Layer *layer = mapDocument->currentLayer()) {
-                finalCommand.replace(QLatin1String("%layername"),
+                finalString.replace(QLatin1String("%layername"),
                                      QString(QLatin1String("\"%1\"")).arg(layer->name()));
             }
         }
 
         if (MapObject *currentObject = dynamic_cast<MapObject *>(document->currentObject())) {
-            finalCommand.replace(QLatin1String("%objecttype"),
+            finalString.replace(QLatin1String("%objecttype"),
                                  QString(QLatin1String("\"%1\"")).arg(currentObject->type()));
-            finalCommand.replace(QLatin1String("%objectid"),
+            finalString.replace(QLatin1String("%objectid"),
                                  QString(QLatin1String("\"%1\"")).arg(currentObject->id()));
         }
+
+        // TODO: %executablename
     }
 
-    return finalCommand;
+    return finalString;
 }
 
 void Command::execute(bool inTerminal) const
--- a/src/tiled/command.cpp
+++ b/src/tiled/command.cpp
@@ -37,7 +37,7 @@ QString Command::finalWorkingDirectory() const
     finalWorkingDirectory = replaceVariables(finalWorkingDirectory, false);
 
     QString finalExecutable = replaceVariables(executable);
-    QFileInfo mFile(finalExecutable);   //Check if executable is a path or not
+    QFileInfo mFile(finalExecutable);
 
     finalWorkingDirectory.replace(QLatin1String("%executablepath"), mFile.absolutePath());
 
--- a/src/tiled/command.cpp
+++ b/src/tiled/command.cpp
@@ -32,7 +32,7 @@ using namespace Tiled::Internal;
 
 QString Command::finalCommand() const
 {
-    QString finalCommand = command;
+    QString finalCommand = QString(QLatin1String("%1 %2")).arg(executable).arg(arguments);
 
     // Perform variable replacement
     if (Document *document = DocumentManager::instance()->currentDocument()) {
@@ -82,7 +82,9 @@ QVariant Command::toQVariant() const
     QHash<QString, QVariant> hash;
     hash[QLatin1String("Enabled")] = isEnabled;
     hash[QLatin1String("Name")] = name;
-    hash[QLatin1String("Command")] = command;
+    hash[QLatin1String("Executable")] = executable;
+    hash[QLatin1String("Arguments")] = arguments;
+    hash[QLatin1String("WorkingDirectory")] = workingDirectory;
     hash[QLatin1String("Shortcut")] = shortcut;
     hash[QLatin1String("SaveBeforeExecute")] = saveBeforeExecute;
     return hash;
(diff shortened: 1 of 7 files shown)


src/tiled/command.cpp
```diff
@@ -30,9 +30,30 @@
...
+        const QString fileName = document->fileName();
+
+        QFileInfo fileInfo(fileName);
+        QString mapPath = fileInfo.absolutePath();
+        finalWorkingDirectory.replace(
+            QLatin1String("%mappath"),
+            QString(QLatin1String("%1")).arg(mapPath));
+        // QProcess::setWorkingDirectory doesn't work with quoted paths
+
+        // TODO: %executablepath
```
Ablu: hm. If there is an issue to track it we do not need the todo here I think...

ketanhwr:
I was thinking the same, but to extend this would easy when we add new variables. New variables would normally be unique to Executable or Working Directory. So, unifying them won't be helpful, imo.

Ablu:
Well... I think if we have one central place to add new variables that should be just as easy. (if the quote handling is also done automatically). So I would think of one function which does all the replacement is then invoked for both, the command and the working directory.

ketanhwr:
Alright, I'll think up of some solution for this. I actually made a function `substituteVariables()` but discarded it. Nvm, I'll add it again

ketanhwr:
@Ablu I've done both the things which were left, you can review now. The build is failing due to libtiled-java :confused:

src/tiled/command.cpp
```diff
@@ -34,56 +34,58 @@ QString Command::finalWorkingDirectory() const
...
+{
+    QString finalString = string;
+
     // Perform variable replacement
     if (Document *document = DocumentManager::instance()->currentDocument()) {
         const QString fileName = document->fileName();
 
-        finalCommand.replace(QLatin1String("%mapfile"),
+        finalString.replace(QLatin1String("%mapfile"),
                              QString(QLatin1String("\"%1\"")).arg(fileName));
```
Ablu: The identation is broken here by one space now.

src/tiled/command.cpp
```diff
@@ -34,56 +34,58 @@ QString Command::finalWorkingDirectory() const
...
 }
 
 QString Command::finalCommand() const
 {
     QString finalCommand = QString(QLatin1String("%1 %2")).arg(executable).arg(arguments);
 
+    return replaceVariables(finalCommand);
+}
+
+QString Command::replaceVariables(const QString &string, bool flag) const
```
Ablu: I think the name `flag` is not really revealing what the variable is for... Maybe something like `quotePaths` would be more intuitive?

src/tiled/command.cpp
```diff
@@ -34,56 +34,58 @@ QString Command::finalWorkingDirectory() const
...
+
+        if (flag) {
+            finalString.replace(
+                QLatin1String("%mappath"),
+                QString(QLatin1String("%1")).arg(mapPath));
+        } else {
+            finalString.replace(
+                QLatin1String("%mappath"),
+                QString(QLatin1String("\"%1\"")).arg(mapPath));
+        }
```
Ablu: I think this does not really solve the problem. My idea with having a single function for replacement of the variables was that each variable can be used in both (the command and the working dir) settings. Here you only do special handling for this specific variable. But extending this approach to the other variables does not really fly either i think... If we wrap each replacement with an `if` we could simply stick to two functions as before. So I think this is better solved by doing a final replacement of `"` characters with an empty string (after all the previous variables are replaced) maybe?

@bjorn: any opinions here?

ketanhwr: Yes, while merging into two functions I had the same doubt. What you suggest might work, I'm not sure though.

src/tiled/command.cpp
ketanhwr: `quotePaths` doesn't make sense when using `%executablepath`. Something like `workingDirectoryVariables` was making sense but it was too long a literal name.

Ablu: Hm... I am talking about the boolean variable here... I do not understand why `workingDirectoryVariables` would make sens here 🤔. The variable controls whether the variable values are quoted, right? So maybe `quoteVariableValues`?

ketanhwr: It also controls `%executablepath` actually.

Ablu: well... Let's see what the result of discussion about the quoting-handling is first...

ketanhwr:
@Ablu what you suggested works nicely! Have a look now :slightly_smiling_face:

src/tiled/command.cpp
```diff
@@ -30,39 +30,69 @@
...
-            finalCommand.replace(QLatin1String("%objectid"),
-                                 QString(QLatin1String("\"%1\"")).arg(currentObject->id()));
+            finalString.replace(QLatin1String("%objecttype"),
+                                QString(QLatin1String("\"%1\"")).arg(currentObject->type()));
+            finalString.replace(QLatin1String("%objectid"),
+                                QString(QLatin1String("\"%1\"")).arg(currentObject->id()));
+        }
+    }
+
+    if (flag) {
```
ketanhwr: Also, I placed this inside `if (Document *document = DocumentManager::instance()->currentDocument())` previously. `%executablepath` is independent of this therefore I moved it ouside that control.

Ablu: Oh... Now I get what you intended to do with the flag variable... I did not realize that it makes no sense to replace the executablepath variable in the executable path setting... 😁. However, I think that this flag is a bad idea. The replacement of the `%executablepath` variable can be done in the `finalWorkingCommand` function. The replacement of `"` can be then made conditional using a parameter.

ketanhwr: You mean `finalWorkingDirectory`, right?

Ablu: yep. Sorry.

ketanhwr: >  The replacement of " can be then made conditional using a parameter.

Didn't understand this part.

Ablu: Ok. Let me try to explain what I was thinking about: I would move the logic for `executablepath` into `finalWorkingDirectory` and then rename `flag` to something like `quoteVariableValues`, which is then false for the working directory case --> quotes are not replaced. Is that more clear?

ketanhwr: Yes okay.

src/tiled/command.h
```diff
@@ -56,6 +62,10 @@ struct Command
      */
     QString finalCommand() const;
 
+    QString finalWorkingDirectory() const;
+
+    QString replaceVariables(const QString &string, bool quotePaths = true) const;
```
bjorn: Judging by the implementation, it seems `quotePaths` should be called `quoteValues`.

src/tiled/command.cpp
```diff
@@ -30,39 +30,70 @@
...
 using namespace Tiled::Internal;
 
+QString Command::finalWorkingDirectory() const
+{
+    QString finalWorkingDirectory = workingDirectory;
+
+    finalWorkingDirectory = replaceVariables(finalWorkingDirectory, false);
+
+    QString finalExecutable = replaceVariables(executable);
+    QFileInfo mFile(finalExecutable);   //Check if executable is a path or not
```
ketanhwr: Yes exactly. That's why I left the `PATH` case out. This was also mentioned in the issue ticket.

src/tiled/command.cpp
```diff
@@ -30,39 +30,70 @@
...
+
+    if (mFile.exists() && mFile.isFile()) {
+        finalWorkingDirectory.replace(QLatin1String("%executablepath"),
+                            QString(QLatin1String("%1")).arg(mFile.absolutePath()));
```
ketanhwr: :man_facepalming:

src/tiled/command.cpp
```diff
@@ -30,39 +30,70 @@
...
+            finalString.replace(QLatin1String("%objecttype"),
+                                QString(QLatin1String("\"%1\"")).arg(currentObject->type()));
+            finalString.replace(QLatin1String("%objectid"),
+                                QString(QLatin1String("\"%1\"")).arg(currentObject->id()));
         }
     }
 
-    return finalCommand;
+    if (!quotePaths)
+        finalString.replace(QLatin1String("\""), QString(QLatin1String("")));
```
ketanhwr: I thought the same but couldn't figure out a case where a variable would contain quote. I'll do this just to be safe, though.

src/tiled/command.cpp
bjorn: I still don't really understand why you need this check. What would go wrong, if you left it out and just always replaced `%executablepath` with `mFile.absolutePath()`? And in what way is it more useful for the user to replace it with an empty string instead?

src/tiled/command.cpp
```diff
@@ -30,39 +30,70 @@
...
+
+    QString finalExecutable = replaceVariables(executable);
+    QFileInfo mFile(finalExecutable);   //Check if executable is a path or not
+
+    if (mFile.exists() && mFile.isFile()) {
+        finalWorkingDirectory.replace(QLatin1String("%executablepath"),
+                            mFile.absolutePath());
+    } else {
+        finalWorkingDirectory.replace(QLatin1String("%executablepath"),
+                            QString(QLatin1String("")));
```
bjorn: In case you leave this here, note that to create an empty string you should just write `QString()`, and to remove something from a string, you would use `finalWorkingDirectory.remove(...)`.

src/tiled/command.cpp
```diff
@@ -30,39 +30,70 @@
...
+    }
+
+    return finalWorkingDirectory;
+}
+
 QString Command::finalCommand() const
 {
-    QString finalCommand = command;
+    QString finalCommand = QString(QLatin1String("%1 %2")).arg(executable).arg(arguments);
```
bjorn: Please use one call to `arg`, like `.arg(executable, arguments)`.

src/tiled/command.cpp
```diff
@@ -30,39 +30,70 @@
...
+    QString finalWorkingDirectory = workingDirectory;
+
+    finalWorkingDirectory = replaceVariables(finalWorkingDirectory, false);
+
+    QString finalExecutable = replaceVariables(executable);
+    QFileInfo mFile(finalExecutable);   //Check if executable is a path or not
+
+    if (mFile.exists() && mFile.isFile()) {
+        finalWorkingDirectory.replace(QLatin1String("%executablepath"),
+                            mFile.absolutePath());
```
bjorn: Coding style: please align wrapped function arguments (so `mFile.absolutePath());` should start at the same position as `QLatin1String` in the line above).

src/tiled/command.cpp
ketanhwr: Aah, didn't think that through.

bjorn:
Looking good, thanks!

PR #1582, opened 2017-05-29
Wang Tiles

Bdtrotte:
An iterative pr for code review on the GSOC wang tiles project.


Bdtrotte:
In the most recent push, I've fixed the changes requested in your first review, as well as added wangsets to Tilesets in the way you suggested (as a list). For the most part I modeled after the Terrains list, with the major exception of not tracking the id within the wangsets themselves. I don't think it will be necessary, though I've yet to look to far into how the interface will be built (which may need more certain ids?). 

For now I've kept my method of finding wang tile variations as it was, as I still think despite the overhead of setting it up,  in most cases it will run faster than searching through all the tiles (as most sets should be full), though this could require further investigation by me.

On that note further testing is still required, though Ill get file IO, and wang assignment working before I test too rigorously. (I did some small tests and all worked as expected).

Sorry for the delay in making progress, but it should be more steady from here on out!

src/libtiled/wangset.cpp
```diff
@@ -0,0 +1,166 @@
...
+    if (potentials.length() > 0)
+        return potentials[qrand() % potentials.length()];
+    else
+        return NULL;
+}
+
+typedef struct WangWildCard
+{
+    int index, colorCount;
+} WangWildCard;
```
bjorn: This way of defining a struct was necessary in C times, to avoid needing to repeat the `struct` keyword whenever you would use `WangWildCard`. In C++, we can just write:
```c++
struct WangWildCard
{
    int index;
    int colorCount;
};
```

src/libtiled/wangset.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+        int max = wildCards.size();
+
+        while (!stack.isEmpty()) {
+            if (stack.size() == max) {
+                int idVariation = 0;
+
+                for (int i = 0; i < max; ++i) {
+                    idVariation |= stack[i].colorCount << stack[i].index;
+                }
```
bjorn: Coding style: please leave out the braces for one-line bodies (also below).

src/libtiled/wangset.h
```diff
@@ -0,0 +1,170 @@
...
+{
+public:
+    WangId(unsigned id) : mId(id) {}
+    WangId() : mId(0) {}
+
+    operator unsigned() const { return mId; }
+    inline void setId(unsigned id) { mId = id; }
+
+    /**
+     * @brief getColor returns the color of a desired edge/corner of a given wang id
```
bjorn: Docs still need updating.

src/libtiled/wangset.h
```diff
@@ -0,0 +1,170 @@
...
+     * @param index index 0-3 with zero being the top right, and 3 left top
+     * @return
+     */
+    int edgeColor(int index) const;
+    int cornerColor(int index) const;
+
+    /**
+     * @brief rotateWangId rotates the wang id 90 * rotations degrees ccw
```
bjorn: Docs seem to mention old function name? I'd suggest to leave out the `@brief` and the function name.

src/libtiled/wangset.h
```diff
@@ -0,0 +1,170 @@
...
+     * @param rotations 1-3
+     * @return
```
bjorn: Returns nothing, so just leave out the `@return`.

src/libtiled/wangset.h
```diff
@@ -0,0 +1,170 @@
...
+     */
+    void rotate(int rotations);\
```
bjorn: Stray `\` at end of line.

src/libtiled/wangset.h
```diff
@@ -0,0 +1,170 @@
...
+    /**
+     * @brief wangIdOfTile returns the wangId of a given tileId
+     * @param tileId
+     */
+    WangId wangIdOfTile(Tile *tile) const;
+
+    WangSet *clone(Tileset *tileset) const;
+
+private:
+    Tileset *mTileSet;
```
bjorn: Naming: "tileset" is always a single word in Tiled source code, so this should should be `mTileset`.

src/libtiled/wangset.h
```diff
@@ -0,0 +1,170 @@
...
+     * 0s in the id are wild cards, and can be filled with any color.
+     * @param wangId
+     * @return
+     */
+    QList<Tile*> findMatchingTiles(WangId wangId) const;
+
+
+    /**
+     * @brief wangIdOfTile returns the wangId of a given tileId
+     * @param tileId
```
bjorn: All docs seem outdated. Also, if you don't provide any comment on a parameter, there is no need to mention it with `@param`.

src/libtiled/wangset.h
```diff
@@ -0,0 +1,170 @@
...
+    WangSet *clone(Tileset *tileset) const;
+
+private:
+    Tileset *mTileSet;
+    QString mName;
+    int mImageTileId;
+    int mEdgeColors;
+    int mCornerColors;
+    QMultiHash<WangId, Tile*> mWangIdToTile;
+    QHash<int, WangId> mTileIdToWangId; //This could be stored in the tile object.
```
bjorn: Whether it could be stored in the tile object depends on whether we allow a tile to be part of multiple wang sets. Would limiting a tile to a single wang set be useful in some way, beyond the small possible performance improvement?

src/libtiled/wangset.cpp
```diff
@@ -0,0 +1,166 @@
...
+                 QString name,
+                 int imageTileId):
+    Object(Object::TerrainType), //for now, will add unique type soon
+    mTileSet(tileset),
+    mName(std::move(name)),
+    mImageTileId(imageTileId),
+    mEdgeColors(edgeColors),
+    mCornerColors(cornerColors)/*,
+    mWangIdToTile(QMultiHash<WangId, Tile*>()),
+    mTileIdToWangId(QHash<int, WangId>())*/
```
bjorn: Just remove these lines.

Bdtrotte: Knew I forgot something...

src/libtiled/wangset.cpp
Bdtrotte: What a wonderful time we live in! Thanks for informing me

src/libtiled/wangset.h
Bdtrotte: Yeah, for all the docs I just use qt's auto formatting. Though its a bit burdensome. Ill go through and update all those.

src/libtiled/wangset.cpp
```diff
@@ -53,58 +53,69 @@ void WangSet::addTile(Tile *tile, WangId wangId)
...
     if (wildCards.isEmpty()) {
         list.append(mWangIdToTile.values(wangId));
     } else {
-        QStack<QPoint> stack;
+        QStack<WangWildCard> stack;
 
-        for (int i = 0; i < wildCards.size(); ++i) {
-            stack.push(wildCards[i]);
-        }
+        stack.append(wildCards);
```
leonidax: I checked the documentation, check the following link.

[http://doc.qt.io/qt-5/qvector.html#append-2](http://doc.qt.io/qt-5/qvector.html#append-2 )

The method to append another qvector was introduced in Qt 5.5 so there is a difference in the Qt version that you are using and the one that the Travis CI uses, I would suggest you to get Qt 5.4.2 which is the officially minimal supported to build Tiled.

leonidax: You might be able to use += or << operators instead.

bjorn: I guess we could also just raise the minimum Qt version to 5.5. I wanted to quickly look up if those operators would work as an alternative, but http://doc.qt.io/archives/index.html does not even host Qt 5.4 documentation anymore.

In general, it does not matter which version is used for development but I'd recommend simply using the latest version, since that is likely the version the next Tiled release is going to be made against. We can check the minimum supported version using Travis CI.

Bdtrotte: So for the time being, should I change that part of the code to comply with the older version? Or just leave it as is?

leonidax: Bjorn has more a word on this than I, he is the product owner, but I would say that if += work in older versions and in newers aswell, then just use the operator.

bjorn: Right, looking at the Qt 5.4 docs in Qt Creator, the operator+= does work with another QVector, so please just use that instead of append, so we can have it in a single line without raising the Qt dependency.

Bdtrotte:
I've implemented file saving and loading, though I've not yet tested rigorously... I wanted to first make sure I'm doing things correctly, and that the format I've used is good, and is in line with current formatting.

bjorn:
> I've implemented file saving and loading, though I've not yet tested rigorously... I wanted to first make sure I'm doing things correctly, and that the format I've used is good, and is in line with current formatting.

Please change the naming style to keep the tags all lowercase. About the format in general, I found it somewhat unexpected that there is a `wanginfo` tag nested in the `tile` tags. I had rather expected all the information to be stored along with the Wang set. For making the reading easier, this would mean storing the Wang sets after the tiles.

Bdtrotte:
Yes, that was my first thought for how to do it, but figured it would be easier for it to be read right after the tile is created. Also I wanted to avoid reprinting the tile Id and such for each wangset, though it probably ends up being about the same number of lines. I mostly liked that it maintains all the info for a given tile in the single block.

Did you think one way was better than the other?

bjorn:
> Did you think one way was better than the other?

I think keeping all the information together in the `wangset` is better, because it keeps the feature in general cleanly separated from the rest, both in the format and in the code.

Bdtrotte:
Not yet in a "working" state, but the foundation for displaying, creating, and editing wangSets.

Bdtrotte:
The latest commit is ready for review! It completes the support for creating and editing the basic properties of a wangset from the UI.

Features still needed for this are assigning the image for a wang set.

Next to come is of course actually assigning the wangIds to the tiles. I see this as an iterative thing; Ill get a very basic implementation working first, then build it up into a more elegant UI. This will be done over the next week or so.

Before I do that Ill get a basic implementation of the wang bucket fill working to test out, and show off those methods.

Now that this is done, I feel progress will continue more quickly!

Zireael07:
"addremocewangset"? Did you mean "remove"?

Bdtrotte:
A working test of the wang fill bucket is in this pr! https://github.com/Bdtrotte/tiled/pull/1

src/libtiled/mapreader.cpp
```diff
@@ -564,11 +568,48 @@ void MapReaderPrivate::readTilesetTerrainTypes(Tileset &tileset)
...
+
+            while (xml.readNextStartElement()) {
+                if (xml.name() == QLatin1String("properties"))
+                    wangSet->mergeProperties(readProperties());
+                else if (xml.name() == QLatin1String("wangsettile")) {
+                    const QXmlStreamAttributes tileAtts = xml.attributes();
+                    int tileId = tileAtts.value(QLatin1String("tileid")).toInt();
+                    int wangId = tileAtts.value(QLatin1String("wangid")).toInt();
+
+                    wangSet->addTile(tileset.tileAt(tileId), wangId);
```
bjorn: This should be `findOrCreateTile`, since you can't assume the tile is already allocated. In my case this was crashing with your test tileset, since the tiles are usually allocated later, when the image is loaded.

Bdtrotte:
Pending further review of everything, I believe the work so far is mergable into the work in progress branch.  Currently all the base data, and core UI is set up and should be ready to be used in further expansion of the feature. My next step is going to be to focus on making the wangsets available in the map editor, and a tool bar button for activating wangfilling methods. After which I will focus back on the assignment of tiles to a wangId.

src/tiled/tileseteditor.cpp
```diff
@@ -826,6 +861,40 @@ void TilesetEditor::removeTerrainType()
...
+        return;
+
+    if (wangSet)
+        view->setWangSet(wangSet);
+}
+
+void TilesetEditor::addWangSet()
+{
+    Tileset *tileset = currentTileset();
+    if(!tileset)
```
bjorn: Coding style: space after `if`.

src/tiled/tileseteditor.cpp
```diff
@@ -826,6 +861,40 @@ void TilesetEditor::removeTerrainType()
...
+        return;
+
+    //2 and 0 are default values for number of edges and corners TODO define this some where better?
```
bjorn: This is fine for now, though we should probably add a "NewWangSetDialog" class so that the user can set up his wang set upon creation (where there would be space to explain the corner/edges count a bit and show the presets).

src/tiled/tileseteditor.cpp
```diff
@@ -826,6 +861,40 @@ void TilesetEditor::removeTerrainType()
...
+
+    mCurrentTilesetDocument->undoStack()->push(new AddWangSet(mCurrentTilesetDocument,
+                                                              wangSet));
+
+    mWangDock->editWangSetName(wangSet);
+}
+
+void TilesetEditor::removeWangSet()
+{
+    WangSet *wangSet = mWangDock->currentWangSet();
```
bjorn: Please do check that there is a current wang set, even if the button to trigger this action should be disabled at this point.

src/tiled/tilesetwangsetmodel.h
```diff
@@ -0,0 +1,99 @@
...
+ * THIS SOFTWARE IS PROVIDED BY THE CONTRIBUTORS ``AS IS'' AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO
+ * EVENT SHALL THE CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+ * PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+ * WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR
+ * OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF
+ * ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```
bjorn: Please make sure your files use the right copyright header. This is the header used for libtiled, but this file is part of Tiled itself and should have the GPL header.

src/tiled/tilesetwangsetmodel.h
```diff
@@ -0,0 +1,99 @@
...
+class TilesetWangSetModel : public QAbstractListModel
+{
+    Q_OBJECT
+
+public:
+    enum UserRoles {
+        WangSetRole = Qt::UserRole
+    };
+
+    TilesetWangSetModel(TilesetDocument *mapDocument,
```
bjorn: This constructor should be marked `explicit`, to avoid potential implicit conversion from `TilesetDocument *` to `TilesetWangSetModel` (however unlikely to run into).

src/libtiled/wangset.cpp
```diff
@@ -0,0 +1,304 @@
...
+void WangTile::flipVertically()
+{
+    int map[] = {2, 5, 0, 7, 6, 1, 4, 3};
+
+    translate(map);
+}
+
+Cell WangTile::makeCell() const
+{
+    if(!mTile)
```
bjorn: Please make coding style adjustments in all needed places, not just where I point them out. This is still one of 10 places where there is no space after the `if`.

src/tiled/tilesetwangsetmodel.h
```diff
@@ -0,0 +1,92 @@
...
+    QVariant data(const QModelIndex &index,
+                  int role = Qt::DisplayRole) const override;
+
+    bool setData(const QModelIndex &index, const QVariant &value, int role) override;
+
+    Qt::ItemFlags flags(const QModelIndex &index) const override;
+
+    WangSet *wangSetAt(const QModelIndex &index) const;
+
+    void insertWangSet(WangSet *wangSet, int index);
```
bjorn: All the other functions take `index` as first parameter, please be consistent and do the same here.

src/tiled/mapdocument.h
```diff
@@ -315,6 +328,8 @@ private slots:
     MapRenderer *mRenderer;
     Layer* mCurrentLayer;
     MapObjectModel *mMapObjectModel;
+    TerrainModel *mTerrainModel;
```
bjorn: You've removed the getter, but this member should not be added either, nor should the forward declaration and the `tilesetTerrain...` set of signals.

Since your `WangSetModel` now follows the same logic as the `TerrainModel`, it should also not be part of the `MapDocument`, nor should the `tilesetWangSet...` set of signals.

src/tiled/changewangsetdata.cpp
```diff
@@ -0,0 +1,102 @@
...
+}
+
+void ChangeWangSetEdges::redo()
+{
+    mWangSetModel->setWangSetEdges(mIndex, mNewValue);
+}
+
+ChangeWangSetCorners::ChangeWangSetCorners(TilesetDocument *tilesetDocument,
+                                            int index,
+                                            int newValue)
```
bjorn: Indentation off by one.

bjorn:
Thanks for your work so far!

Please open a new pull request today or tomorrow for the next steps. You should decide on a work package that you estimate will take about one week, so that we can aim for another merge on Monday or Tuesday next week.

Please don't add merge commits to the pull request. It is confusing on GitHub and makes the history less clean. I had to forward the `wip/wangtiles` branch before doing the squashed merge to make sure the history still made sense. I don't think we will need further changes that will be made on `master` in the meantime, but if we do then I will do a merge in between pull requests. You should base further changes on the `wip/wangtiles` branch in my repository, and rebase any commits you may have already done onto it.

I think the general set up for the wang tiles is really promising, so I'm really looking forward to see the rest of the UI and the tools materialize!

Issue #1084, opened 2015-09-15
Rotating buttons at toolbar


lab313ru:
It would be great to add these buttons at toolbar. Because rotating only implemented with shortcuts, and there is no visual buttons for that.

bjorn:
Ah yes, same for flipping horizontally / vertically.

bjorn:
I think before implementing this, we need to add the possibility of a tool-specific tool bar, otherwise the UI will get too cluttered. There's other tools that could benefit from such a tool bar as well.

ketanhwr:
@bjorn, what all other tools should be added to this tool bar as well, according to you?

bjorn:
@ketanhwr Well, there's a lot of tools with options that could use visualization in the UI. However, before this can be implemented a tool-specific tool bar needs to be introduced that switches content based on the currently selected tool.

**Stamp Tool** (and probably Bucket Fill tool as well)
* Flip horizontally / vertically
* Rotate left / right
* Random Mode
 
**Terrain Brush**
* "Single corner edit mode" (currently Ctrl modifier)
* "Mirror mode" (current Alt modifier)

**Rectangular Select** (and Magic Wand, Select Same Tile)
* Subtract / Add / Exclusive-Or (currently Ctrl, Shift and Ctrl+Shift modifiers)

And there's some new features that could be a good fit for this as well, like a paint tool where you can control the size of the brush.

ketanhwr:
Oh, this one would kind-of be like a mini-project then.

bjorn:
@ketanhwr Yes, the complex bit is not actually part of this issue, but this issue depends on it. It's the part where each tool can somehow set up tool bar widgets on a context-sensitive tool bar (or a context-sensitive dock widget, which is how GIMP solved it).

lab313ru:
Wow, thank you very much!)

PR #1586, opened 2017-06-01
Adds Tool Specific Toolbar

ketanhwr:
#1084

@bjorn, @Ablu : I've added a tool specific toolbar. It's **not complete** right now.

I've added the tools specific to stamp brush and bucket fill tool only, i.e. Flipping and Rotating, and the Random tool as well. I've attached a screenshot as well. The toolbar is visible after the Offset Layer tool. If the selected tool is changed to something else, the tools are not visible anymore.

toolspecific

I will have to extend this only to add specific tools for other tools, so I thought of making this pull request.


src/tiled/mapeditor.cpp
```diff
@@ -508,6 +492,67 @@ void MapEditor::showMessage(const QString &text, int timeout)
...
+        mToolSpecificToolBar->addWidget(mFlipVerticalButton);
+        mToolSpecificToolBar->addWidget(mRotateLeft);
+        mToolSpecificToolBar->addWidget(mRotateRight);
+
+        connect(mRandomButton, &QToolButton::toggled, mStampBrush, &StampBrush::setRandom);
+        connect(mRandomButton, &QToolButton::toggled, mBucketFillTool, &BucketFillTool::setRandom);
+        connect(mFlipHorizontalButton, &QToolButton::clicked, this, &MapEditor::flipHorizontally);
+        connect(mFlipVerticalButton, &QToolButton::clicked, this, &MapEditor::flipVertically);
+        connect(mRotateLeft, &QToolButton::clicked, this, &MapEditor::rotateRight);
+        connect(mRotateRight, &QToolButton::clicked, this, &MapEditor::rotateLeft);
```
Ablu: hm. you create a lot of objects in the code before... Are those ever deleted? Does `mToolSpecificToolBar->clear();` do this?

ketanhwr: Hmm..looks like I will need to store all the objects too in a list probably and use `qDeleteAll`

ketanhwr:
I'm thinking it would be a better option to make a new class `ToolSpecificToolBar`.

Ablu:
I think one option could be to simply create one toolbar for each editing context and then simply toggle the visibility according to the current context. That would be easier to handle regarding deletion and recreation of buttons. Also if later other editing contexts might require something different than a toolbar, or require special toolbars that should be easier since there is no dependency on a toolbar then. But maybe @bjorn should comment on this...

ketanhwr:
@Ablu you can have a look. I also tried the other way round. Basically, I instantiated a `QToolButton` for each of the tool (Random, Flipping, Rotating) in the constructor of the `ToolSpecificToolBar` and I added or removed the same instances when the selected tool was changed. But it didn't seem to work. The `QToolButton` would be visible only for the first time. I guess `QToolbar::clear()` actually clears the memory too.

Ablu:
I guess @bjorn has to do this review... I do not have too much experience with Qt... I still think that having multiple toolbars and only toggling the visibility might be another clean approach... But I do not want to rate them against each other :)

src/tiled/mapeditor.cpp
```diff
@@ -669,6 +609,12 @@ void MapEditor::rotate(RotateDirection direction)
     }
 }
 
+void MapEditor::random(bool value)
```
bjorn: Please call this `setRandom`.

src/tiled/stampbrush.h
```diff
@@ -64,6 +64,8 @@ class StampBrush : public AbstractTileTool
      */
     const TileStamp &stamp() const { return mStamp; }
 
+    bool random() { return mIsRandom; }
```
bjorn: I would call the getter `isRandom` as well. Also it should be marked `const`.

src/tiled/toolspecifictoolbar.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+    addWidget(mRandomButton);
+    mButtons.append(mRandomButton);
+
+    connect(mRandomButton, &QToolButton::toggled, mMapEditor, &MapEditor::random);
+}
+
+void ToolSpecificToolBar::addFlippingTools()
+{
+    QToolButton *mFlipHorizontalButton = new QToolButton(this);
```
bjorn: Try to rely on `QAction` when possible (and I think it's possible for all actions you're currently adding). That way, you can instantiate the action once and just add it to the tool bar when applicable. It won't get deleted upon calling `clear`, which is good.

src/tiled/toolspecifictoolbar.cpp
```diff
@@ -0,0 +1,162 @@
...
+    mButtons.append(mRotateLeft);
+    mButtons.append(mRotateRight);
+
+    connect(mRotateLeft, &QToolButton::clicked, mMapEditor, &MapEditor::rotateRight);
+    connect(mRotateRight, &QToolButton::clicked, mMapEditor, &MapEditor::rotateLeft);
+}
+
+void ToolSpecificToolBar::retranslateUi()
+{
+    // TODO
```
bjorn: If you rely on `QAction`, you can set their tool tips here.

src/tiled/toolspecifictoolbar.h
```diff
@@ -0,0 +1,69 @@
...
+namespace Tiled {
+namespace Internal {
+
+class AbstractTool;
+class MapEditor;
+
+class ToolSpecificToolBar : public QToolBar
+{
+public:
+    ToolSpecificToolBar(QWidget *parent = nullptr, MapEditor *mapEditor = nullptr);
```
bjorn: Please make the parent the last parameter, which is a general convention followed in Qt.

Also, probably the MapEditor parameter should not default to `nullptr`?

src/tiled/toolspecifictoolbar.cpp
ketanhwr: Oh that's exactly what I was looking for!

src/tiled/toolspecifictoolbar.h
ketanhwr: Then what should be the default parameter for it?

bjorn: It would simply have no default.

ketanhwr: It gave off an error when I did that :thinking:

bjorn: It would do that if you forget the first part of my initial comment, which was to make the parent parameter the last one. In C++, once some parameter is given a default value, all parameters after that also need a default value.

ketanhwr: Oh yes, shit :man_facepalming:

ketanhwr:
Alright so how about this:

- Add a function `virtual void populateToolBar(ToolSpecificToolBar* toolBar)` in `AbstractTool`
- The `ToolSpecificToolBar` would hold the `QAction` instances of all the tools like Flipping/Rotating/etc.
- Whenever a tool is activated, the `populateToolBar` would call the functions like `addFlippingTools`, etc to add the `QAction` instances to the toolbar.

bjorn:
That would be a good intermediate step, though eventually I think the parameter should be simply a `QToolBar` pointer and the `QAction` instances should be owned by the tools.

ketanhwr:
So there's no need for an intermediate class `ToolSpecificToolBar` then, I guess. It can be created in `MapEditor` only like the other toolbars.

bjorn:
Right, with the important difference, that the content of the tool bar is determined by the currently selected tool.

ketanhwr:
Currently, all the functions handling rotation and flipping are in `MapEditor` class. If these are shifted to `StampBrush` and `BucketFillTool` then we won't be able to share common tile stamps. How do I handle this? Should I pass a reference of `MapEditor` instance to `populateToolBar` too? In this way, we can connect the `QAction` to the functions in `populateToolBar` which solves the problem

bjorn:
> Currently, all the functions handling rotation and flipping are in MapEditor class. If these are shifted to StampBrush and BucketFillTool then we won't be able to share common tile stamps. How do I handle this? Should I pass a reference of MapEditor instance to populateToolBar too? In this way, we can connect the QAction to the functions in populateToolBar which solves the problem

You do not not need to use `MapEditor` directly to share the same stamp between these tools. Instead, you could rely on a connection like this one:
```c++
connect(mStampBrush, &StampBrush::stampCaptured, this, &MapEditor::setStamp);
```
So when the `StampBrush` flips the stamp, it could also emit `stampCaptured` (which could probably be renamed to `stampChanged`), and a similar signal can be added to `BucketFillTool`.

ketanhwr:
@bjorn, I've added a new class `StampActions`. I first started with `AbstractStampTool` class but there was nothing common except the newly added `QActions` so I reverted and made the class `StampActions`.

src/tiled/stampactions.cpp
```diff
@@ -0,0 +1,97 @@
...
+using namespace Tiled;
+using namespace Tiled::Internal;
+
+StampActions::StampActions(QObject *parent) : QObject(parent)
+{
+    QIcon mDiceIcon(QLatin1String(":images/24x24/dice.png"));
+    QIcon mFlipHorizontalIcon(QLatin1String(":images/24x24/flip-horizontal.png"));
+    QIcon mFlipVerticalIcon(QLatin1String(":images/24x24/flip-vertical.png"));
+    QIcon mRotateLeftIcon(QLatin1String(":images/24x24/rotate-left.png"));
+    QIcon mRotateRightIcon(QLatin1String(":images/24x24/rotate-right.png"));
```
bjorn: Naming style: don't name local variables with the member prefix.

src/tiled/stampactions.h
```diff
@@ -0,0 +1,58 @@
...
+
+    void languageChanged();
+
+    void populateToolBar(QToolBar *toolBar, bool isRandom);
+
+    QAction *random() { return mRandom; }
+    QAction *flipHorizontal() { return mFlipHorizontal; }
+    QAction *flipVertical() { return mFlipVertical; }
+    QAction *rotateLeft() { return mRotateLeft; }
+    QAction *rotateRight() { return mRotateRight; }
```
bjorn: All these getters should be `const`. However, when `mStamp` and `mRandom` are moved into this class, you only need getters (and signals) for those and not for all the actions.

src/tiled/bucketfilltool.h
```diff
@@ -64,9 +65,18 @@ class BucketFillTool : public AbstractTileTool
      */
     const TileStamp &stamp() const { return mStamp; }
 
+    bool isRandom() { return mIsRandom; }
```
bjorn: Please remove (since it's not used) or mark `const` if kept.

src/tiled/stampbrush.h
```diff
@@ -64,6 +65,10 @@ class StampBrush : public AbstractTileTool
      */
     const TileStamp &stamp() const { return mStamp; }
 
+    bool isRandom() { return mIsRandom; }
```
bjorn: Please remove (since it's not used) or mark `const` if kept.

bjorn:
Thanks! I'm looking forward to see the use of this toolbar extended to more tools. :-)

Issue #70, opened 2011-09-06
Object templates


zamp:
Would be nice if you could set up object templates so when you create new objects you could just pick one from a list and drag n drop it on the map.
Maybe double click on list to edit a template.
The templates should have the object type and properties with default values.
http://juvo.se/tbb3.jpg here's a mashup of what it could look like.

downstroy:
Damn I would love to see this feature implemented!

Imagine you define an object (eg: a platform), you give it some basic properties (like minX, minY, maxX, maxY) to what you want it to do. 

Now imagine you are doing a platformer, you will probably use dozen if not hundred of those objects.

And after using lots of them, you have the idea of adding custom collision to your platform (like it should collide from above but not from under. You will have to cycle through all your platform object and edit them all to add the collision : top / collision : bottom / collision : both custom property. What a hassle to maintain!

Of course you could just do it in your code, checking the id of the object and adding the behaviour, but it would really be a mess and not faster.

I think having template of object (aka prefab (cf unity3D)) would really be a neat feature for any "medium to big" sized project.

bjorn:
@zamp Is there any chance you would be able to restore the mashup you linked to in your original request? I think it could be very helpful for anybody who could be interested to work on this feature (which could be this summer, as a Google Summer of Code project).

i-ka:
In Unity, unchanged instance properties synchronizes with prefab properties. I think tracking prefab property changes will good feature. In addition it can reduce file size a little bit, by saving only changed properties.

zamp:
@bjorn sorry. I don't have the file anymore nor can I recall what it looked like.

bjorn:
@zamp Alright, no problem. Thanks for letting us know!

bjorn:
I'm closing this issue since the feature is now implemented, thanks to @thabetx's Google Summer of Code project! Check out [the documentation](http://doc.mapeditor.org/en/latest/manual/using-templates/) and be sure to try it out by using a development snapshot (available [at itch.io](https://thorbjorn.itch.io/tiled)).

Any feedback would be very welcome as we will now be working towards the Tiled 1.1 release.

bjorn:
For reference, here are the issues I've already opened covering remaining tasks:

* Allow replacing existing objects with template instances #1722
* Allow deleting object templates from a template group #1723
* Allow unloading a template group #1724
* Allow resetting overridden properties individually for template instances #1725
* Allow locking template properties #1726

PR #1587, opened 2017-06-01
Reusable object templates

thabetx:
This pull request is about the reusable object templates project in GSoC 2017, addresses #70 .

The entry point for testing is selecting some objects and choosing "Save as a template group", the selected objects will be saved to a file, then the file will be immediately read and the sizes of the saved templates will be printed, this is just for testing purposes.

When saving tile objects, the tileset info will be saved as well, this encourages having standalone template groups that can be used across multiple objects, although for now, when reading the file, the tilesets are ignored.

The next step will be creating a template group document that will allow loading the template group into tiled.

Code wise, I've followed tileset format and map format for creating the new classes, the template group is till now similar to object group, maybe I will need to create a new class for a template instead of using the object class.

Reading and writing reuses some of the available code, I had to extract some of the object reading and writing code (the part that handles different types of objects), otherwise things are pretty similar to writing object groups albeit with some changes like not writing the coordinates of the object or its ID as they won't be relevant when creating new copies of the template.

So currently this hypothetical use case is simple and some parts of the code are not complete, I think having the template group ready and loading it into Tiled will require doing more checking and will reveal more cases that require better handling.


src/libtiled/mapreader.cpp
```diff
@@ -834,6 +873,48 @@ void MapReaderPrivate::readImageLayerImage(ImageLayer &imageLayer)
...
+//    const unsigned gid = atts.value(QLatin1String("gid")).toUInt();
+    const qreal width = atts.value(QLatin1String("width")).toDouble();
+    const qreal height = atts.value(QLatin1String("height")).toDouble();
+    const QString type = atts.value(QLatin1String("type")).toString();
+
+    const QSizeF size(width, height);
+
+    MapObject *object = new MapObject(name, type, {0, 0}, size);
+
+    readObjectData(object);
```
thabetx: Alright, I didn't understand what you meant, I thought you referred to the template group as template.

So just to be clear the format will look like this
```
<templategroup ...>
 <template ...>
  <object ...>
    ...
  </object ...>
 </template ...>

 <template ...>
  <object ...>
    ...
  </object ...>
 </template>
</templategroup>
```
This looks more effective, I will update it and the template class makes more sense now.

And for writing, is there a better way other than creating a new function just to prevent writing the id and the coordinates? Does sending a parameter to tell it whether to write them or not be appropriate?

bjorn: I think it would be fine to prevent the writing of the ID by just making sure it is 0 and adding a check for this in `MapWriterPrivate::writeObject`. For now, I would not bother with preventing the writing of the position.

thabetx: Okay that would allow more code reuse, but what do you mean by setting id to 0, wouldn't that change the original object?

bjorn: Well, only the object owned by the template, which should be a copy, would have its ID set to 0.

thabetx: Okay right, thanks!

thabetx: `template` is a keyword in C++, would `objectTemplate` as the name of the class be fine?

thabetx:
I've update the format, added the tileset references to the templategroup and fixed some review points.

I'm working on changing the writing function to work directly on a template group instead of an object list, and will add more information to the template group and the templates.

thabetx:
I updated the saving function and will start working on the `TemplateGroupDocument`.

I added a name to `templateGroup` and a name and ID to `objectTemplate`, not sure yet how the ID will be used.

To prevent writing the ID and the coordinates of the `mapObject` in the `objectTemplate`, I used the suggestion of setting the ID of the cloned `mapObject` to 0 and used it as a check.

thabetx:
I implemented the `TemplateGroupDocument` functions and now it's used for testing. I've also squashed some of the commits, testing is still done by selecting some objects then choosing "save as template group".

src/tiled/templategroupdocument.cpp
```diff
@@ -0,0 +1,96 @@
...
+FileFormat *TemplateGroupDocument::writerFormat() const
+{
+    return mTemplateGroup->format();
+}
+
+QString TemplateGroupDocument::displayName() const
+{
+    QString displayName = mTemplateGroup->name();
+
+    if(displayName.isEmpty())
```
bjorn: Coding style: please put a space after `if`.

src/tiled/tmxmapformat.cpp
```diff
@@ -166,3 +166,56 @@ bool TsxTilesetFormat::supportsFile(const QString &fileName) const
...
+    : TemplateGroupFormat(parent)
+{
+}
+
+TemplateGroup *TtxTemplateGroupFormat::read(const QString &fileName)
+{
+    mError.clear();
+
+    MapReader reader;
+    TemplateGroup *templateGroup =  reader.readTemplateGroup(fileName);
```
bjorn: Coding style: one space too many after `=`.

src/tiled/mapdocument.cpp
```diff
@@ -125,6 +127,58 @@ MapDocument::~MapDocument()
     delete mMap;
 }
 
+/**
+ * Currently, this is the entry point for testing templates,
+ * it creates a templateGroup from the selected objects except the last object,
+ * then creates a templatGroupDocument from the templateGroup,
```
bjorn: Typo: templatGroupDocument (also, I think it would be better to match the casing with the class name).

src/tiled/mapdocument.cpp
```diff
@@ -125,6 +127,58 @@ MapDocument::~MapDocument()
...
+
+    for (int i = 0; i < mSelectedObjects.size() - 1; ++i) {
+        ObjectTemplate *objectTemplate = new ObjectTemplate(i+1, QLatin1String("template ") + QString::number(i+1));
+        objectTemplate->setObject(mSelectedObjects[i]);
+        templateGroup->addTemplate(objectTemplate);
+    }
+
+    QString fileName = QLatin1String("templateGroup.ttx");
+
+    TemplateGroupDocument *templateGroupDocument = new TemplateGroupDocument(templateGroup, fileName);
```
bjorn: No change required, but just to note, that to have less repetition I am often using `auto` for such lines these days, since the type is completely obvious:
```c++
auto templateGroupDocument = new TemplateGroupDocument(templateGroup, fileName);
```

PR #1588, opened 2017-06-01
Rename Polygon tool icon names

ketanhwr:
This fixes the issue mentioned here: http://discourse.mapeditor.org/t/tiled-1-0-released/2402/5


bjorn:
Thanks!

PR #1591, opened 2017-06-02
Fixed relative paths in variants when symlinks are involved (fixed #1589)

### LINKED ISSUE #1589: Exported relative paths don't respect symlinks
On my machine, there is a directory named `srv` in my home directory which is a symlink to `/srv/http`.

Right now, this is part of the directory tree of `~/srv/map` (realpath: `/srv/http/map`):

```
.
└── assets
|      └── tilesets
|             └── tileset.png
└── intro.tmx
```

The file `~/srv/map/intro.tmx` uses `~/srv/map/assets/tilesets/tileset.png`. Therefore, when I export `~/srv/map/intro.tmx` to `~/srv/map/assets/maps/intro.json`, I'd expect the relative path in the JSON file to become `../tilesets/tileset.png`.

The current version, however, creates `../../../../../../srv/http/map/assets/tilesets/tileset.png`, which essentially makes the file kind of unusable on my target platform without messing with the values.

Could you please respect symlinks and not force `realpath`?

kdex:
This PR addresses and fixes #1589.

Before this commit, the `source` variant could be messed up easily by having the tmx project as well as a tileset and its corresponding image in a directory that is really just a symlink.

In the process of exporting, `MapToVariantConverter` will at some point try to make a relative path between the output file and the tileset image.

Creating portable relative paths, however, only makes sense if the input paths are canonicalized, otherwise you may end up with garbage (refer to the example shown in the issue for details).

### DIFF
--- a/src/libtiled/maptovariantconverter.cpp
+++ b/src/libtiled/maptovariantconverter.cpp
@@ -45,7 +45,7 @@ static QString colorToString(const QColor &color)
 
 QVariant MapToVariantConverter::toVariant(const Map &map, const QDir &mapDir)
 {
-    mMapDir = mapDir;
+    mMapDir = mapDir.canonicalPath();
     mGidMapper.clear();
 
     QVariantMap mapVariant;
@@ -94,7 +94,7 @@ QVariant MapToVariantConverter::toVariant(const Map &map, const QDir &mapDir)
 QVariant MapToVariantConverter::toVariant(const Tileset &tileset,
                                           const QDir &directory)
 {
-    mMapDir = directory;
+    mMapDir = directory.canonicalPath();
     return toVariant(tileset, 0);
 }
 


bjorn:
Are you sure this fix is only needed in `MapToVariantConverter`? I wonder about other export formats, which all refer to image files using relative paths, and well as about plain saving of the map or tileset files in TMX format.

Also, could the same problem happen the other way around? So, suppose you have the following structure:
```
.
└── assets -> symlink to /usr/share/tilesets
└── intro.tmx
```
Then, when `/home/bjorn/projects/intro.tmx` referred to `./assets/tileset.png`, it should not upon saving make a reverence like `../../../usr/share/tilesets/tileset.png`, which would be what happens if we're going to use canonical paths, right?

kdex:
Closing due to [this](https://github.com/bjorn/tiled/issues/1589#issuecomment-306225819).

PR #1593, opened 2017-06-04
Show output of commands in Debug Console

### LINKED ISSUE #1552: Write output from custom commands to Debug Console window
Tiled has a Debug Console window, but it isn't really used. The only thing it is used for at the moment is to allow writing some debug lines from Python plugins.

It would be useful to write the output from custom commands also to the Debug Console.

This would avoid the need for workarounds like [described in this forum topic](http://discourse.mapeditor.org/t/running-python-from-the-command-tool-behaves-wierdly/2277?u=bjorn).

ketanhwr:
This closes #1552. I've attached a screenshot of the same.

outputconsole

### DIFF
--- a/src/tiled/command.cpp
+++ b/src/tiled/command.cpp
@@ -217,8 +217,7 @@ CommandProcess::CommandProcess(const Command &command, bool inTerminal, bool sho
 
     if (showOutput) {
         CommandManager::instance()->logger()->log(LoggingInterface::INFO,
-                                                  QString(QLatin1String("Executing: "))
-                                                  + mFinalCommand);
+                                                  tr("Executing: %1").arg(mFinalCommand));
 
         connect(this, &QProcess::readyReadStandardError, this, &CommandProcess::consoleError);
         connect(this, &QProcess::readyReadStandardOutput, this, &CommandProcess::consoleOutput);
--- a/src/tiled/command.cpp
+++ b/src/tiled/command.cpp
@@ -157,7 +157,6 @@ CommandProcess::CommandProcess(const Command &command, bool inTerminal, bool sho
     , mName(command.name)
     , mFinalCommand(command.finalCommand())
     , mFinalWorkingDirectory(command.finalWorkingDirectory())
-    , mLogger(CommandManager::instance()->logger())
 #ifdef Q_OS_MAC
     , mFile(QDir::tempPath() + QLatin1String("/tiledXXXXXX.command"))
 #endif
@@ -216,8 +215,14 @@ CommandProcess::CommandProcess(const Command &command, bool inTerminal, bool sho
 
     connect(this, SIGNAL(finished(int)), SLOT(deleteLater()));
 
-    if (showOutput)
+    if (showOutput) {
+        CommandManager::instance()->logger()->log(LoggingInterface::INFO,
+                                                  QString(QLatin1String("Executing: "))
+                                                  + mFinalCommand);
+
+        connect(this, &QProcess::readyReadStandardError, this, &CommandProcess::consoleError);
         connect(this, &QProcess::readyReadStandardOutput, this, &CommandProcess::consoleOutput);
+    }
 
     if (!mFinalWorkingDirectory.trimmed().isEmpty())
         setWorkingDirectory(mFinalWorkingDirectory);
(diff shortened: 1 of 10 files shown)


src/tiled/command.cpp
```diff
@@ -140,17 +144,20 @@ Command Command::fromQVariant(const QVariant &variant)
...
     return command;
 }
 
-CommandProcess::CommandProcess(const Command &command, bool inTerminal)
+CommandProcess::CommandProcess(const Command &command, bool inTerminal, bool showOutput)
     : QProcess(DocumentManager::instance())
     , mName(command.name)
     , mFinalCommand(command.finalCommand())
     , mFinalWorkingDirectory(command.finalWorkingDirectory())
+    , mLogger(CommandManager::instance()->logger())
```
bjorn: If the logger is available from the global `CommandManager` instance, I don't see a reason to also have it as member of `CommandProcess`.

src/tiled/command.cpp
```diff
@@ -209,12 +216,20 @@ CommandProcess::CommandProcess(const Command &command, bool inTerminal)
 
     connect(this, SIGNAL(finished(int)), SLOT(deleteLater()));
 
+    if (showOutput)
+        connect(this, &QProcess::readyReadStandardOutput, this, &CommandProcess::consoleOutput);
```
bjorn: You'll want to also connect to `readyReadStandardError` or use [`setProcessChannelMode`](http://doc.qt.io/qt-5/qprocess.html#setProcessChannelMode) to merge the standard error channel into the standard out channel.

Of course, it would be nice to report the error channel as appropriate to the `LoggingInterface`, so that the Debug Console will show errors in red.

src/tiled/command.cpp
```diff
@@ -209,12 +216,20 @@ CommandProcess::CommandProcess(const Command &command, bool inTerminal)
...
+
     if (!mFinalWorkingDirectory.trimmed().isEmpty())
         setWorkingDirectory(mFinalWorkingDirectory);
 
     start(mFinalCommand);
 }
 
+void CommandProcess::consoleOutput()
+{
+    mLogger->log(LoggingInterface::INFO, QLatin1String(readAllStandardOutput()));
```
bjorn: You shouldn't wrap the returned `QByteArray` with a `QLatin1String`, cause that will break whenever the encoding of the output is not actually latin1. Since we can't really know the output encoding, we should probably assume local encoding, which means using [`QString::fromLocal8Bit`](https://doc.qt.io/qt-5/qstring.html#fromLocal8Bit-1).

src/tiled/commanddialog.cpp
```diff
@@ -132,7 +143,7 @@ void CommandDialog::updateWidgets(const QModelIndex &current, const QModelIndex
     mUi->exBrowseButton->setEnabled(enable);
     mUi->keySequenceEdit->setEnabled(enable);
     mUi->clearButton->setEnabled(enable);
-    mUi->saveBox->setEnabled(enable);
```
bjorn: This line shouldn't be deleted, right?

src/tiled/commanddialog.ui
```diff
@@ -151,6 +151,13 @@
        </property>
       </widget>
      </item>
+     <item row="4" column="0" colspan="3">
+      <widget class="QCheckBox" name="outputBox">
+       <property name="text">
+        <string>Show Output on Console</string>
```
bjorn: Please use "Show output in Debug Console".

src/tiled/commanddialog.cpp
ketanhwr: Actually I noticed that this line is already there in top :joy:

ketanhwr: So I deleted this one.

src/tiled/commanddialog.ui
```diff
@@ -154,7 +154,7 @@
      <item row="4" column="0" colspan="3">
       <widget class="QCheckBox" name="outputBox">
        <property name="text">
-        <string>Show Output on Console</string>
+        <string>Show Output in Debug Console</string>
```
bjorn: Small nit: please lowercase "output"

src/tiled/command.cpp
```diff
@@ -216,8 +215,14 @@ CommandProcess::CommandProcess(const Command &command, bool inTerminal, bool sho
 
     connect(this, SIGNAL(finished(int)), SLOT(deleteLater()));
 
-    if (showOutput)
+    if (showOutput) {
+        CommandManager::instance()->logger()->log(LoggingInterface::INFO,
+                                                  QString(QLatin1String("Executing: "))
```
bjorn: This should really be a translatable string, like `tr("Executing: %1").arg(...)`

bjorn:
It's working well, nicely done!

Issue #650, opened 2014-04-26
Add Move Selection feature.


bryanedds:
This is really fundamental to building stages in side-scrolling games. In that context, you often need to move slices of the stages back and forth. The current tools are insufficient to do this smoothly.

Simple usage of Move Selection feature would be -
- Select desired tiles with existing Selection tool.
- Drag selection with right-click.

Usage in a side-scrolling level editing scenario would be -
- Select tiles you wish to relocate with existing Selection tool.
- Cut.
- Select tiles you need to scoot out of the way.
- Drag selection with right-click.
- Paste, then place relocated tiles appropriately.

Additional polish feature would be -
- Press escape to go from Stamp / Paste Mode back to previous mode (here would be Select Mode).

Implementation of main feature and polish should be straight-forward enough.

joshmmo:
I need this

bjorn:
Another +1 from @eevee in [his forum post](http://discourse.mapeditor.org/t/feedback-after-a-couple-months-with-tiled/2090). About time to raise the priority.

sanchaez:
Ok I'm on it :)

Bobjt:
My crap-a-rific map editor implemented multi-layer movement with a global check box titled "current layer only" - if that was off, every operation applied to every layer. I had a proper move tool (and of course, selection), and so could move portions of my multi layered map around, Yayyyyy.  Or just the current selected layer. Rudimentary, but it worked like a charm. I like Tiled so much, can't do both make a map editor AND game (i'll die before then), but man, what a big wall this is not having a move tool... holding out for this one, Thorbjørn, my maps are gonna suffer until then. Might even be willing to up my patronage for this one. If I do though, I'm gonna pm you directly and let you know what it's for with wide eyes and a ridiculously huge wink ;-)

sanchaez:
@Bobjt I've been working on this, but it's current state is not really good and I don't have much free time :(
I will push the changes to my fork today.

ketanhwr:
Should this be specific to a particular tile layer, because the right now, the selection tool selects only from the currently active layer?

Another doubt: Should this be a new tool? Or should be added to the current selection tool itself (will work only when a particular key/mouse button is pressed)?

bjorn:
@ketanhwr Multi-layer move would be awesome, but I suggest you implement single-layer move first, which would already be a nice addition.

[...]

PR #1607, opened 2017-06-10
Added Move selection Feature

ketanhwr:
This addresses #650.

It works fine as of now, but I need feedback regarding the flow of this tool. You need to select an area with the tile selection tool, and then select this tool and then move the selection to someplace else. I have a few doubts in mind regarding the flow but I think a feedback would be great!


Ablu:
A few points I realized while testing:
- When starting to drag, but still staying on the same tile (so only drag like a pixel to the left or right), the whole map is overlayed with the light blue/gray
- Sometimes I get a situation like this:
screenshot from 2017-06-10 17-22-20
The lower row is nothing what I selected... seems to be a small graphical glitch... It seems to preview one tile to much in width and height. Also the display of those lines seems to lag behind while dragging...

Generally I think the flow of the tool is fine... The only thing which could be annoying is that there is no thing like a floating layer like in Gimp... So if you move tiles on crowded layers you might accidently delete / override tiles...

src/tiled/moveselectiontool.cpp
```diff
@@ -0,0 +1,197 @@
...
+}
+
+MoveSelectionTool::~MoveSelectionTool()
+{
+}
+
+void MoveSelectionTool::tilePositionChanged(const QPoint &pos)
+{
+    if (mDragging) {
+
```
Ablu: unneeded empty line

src/tiled/moveselectiontool.cpp
```diff
@@ -0,0 +1,197 @@
...
+}
+
+void MoveSelectionTool::cut()
+{
+    Layer *currentLayer = mapDocument()->currentLayer();
+
+    TileLayer *tileLayer = dynamic_cast<TileLayer*>(currentLayer);
+    const QRegion &selectedArea = mapDocument()->selectedArea();
+
+    ClipboardManager::instance()->copySelection(mapDocument());
```
Ablu: I still think that there should be no `ClipboardManager` involved here... One would not expect to loose the current clipboard content when using the tool. It should be possible to store the moved selection in the tool.

src/tiled/moveselectiontool.cpp
```diff
@@ -0,0 +1,197 @@
...
+                                       source->x()+offset.x(),
+                                       source->y()+offset.y(),
+                                       source));
+
+    undoStack->push(new ChangeSelectedArea(mapDocument(), selectedArea));
+
+    if (map)
+        tilesetManager->removeReferences(map->tilesets());
+
+    brushItem()->clear();
```
Ablu: I think the preview layer should be cleared here too.

src/tiled/moveselectiontool.cpp
Ablu: Can't you simply use the `mPreviewLayer ` variable for this?

ketanhwr: Oh yeah. I didn't realise that, I'll look into this!

ketanhwr:
@Ablu: The clipboard is no longer needed now. I've used the preview layer to paint only. There are still a few bugs. 

> So if you move tiles on crowded layers you might accidently delete / override tiles...

This can be solved by: Whenever the tool is selected, the selected area remains same, as in, even if it overrides the tiles when you move the selection, it won't actually delete them _unless_ you choose some other tool. I will have to save the preview layer when the tool is first activated and paint when the tool is deactivated.

ketanhwr:
@Ablu, I've probably fixed all the issues that you've mentioned.

> Sometimes I get a situation like this:
I'm not able to reproduce it, it might as well be a graphical error only.

src/tiled/moveselectiontool.h
```diff
@@ -0,0 +1,65 @@
...
+{
+    Q_OBJECT
+
+public:
+    MoveSelectionTool(QObject *parent = nullptr);
+    ~MoveSelectionTool();
+
+    void activate(MapScene *scene) override;
+    void deactivate(MapScene *scene) override;
+
```
bjorn: You'll want to override the default implementations of `mouseEntered` and `mouseLeft`, since they're showing/hiding the preview item, which is not what you want for this tool.

src/tiled/moveselectiontool.cpp
```diff
@@ -0,0 +1,184 @@
...
+}
+
+void MoveSelectionTool::cut()
+{
+    Layer *currentLayer = mapDocument()->currentLayer();
+
+    TileLayer *tileLayer = dynamic_cast<TileLayer*>(currentLayer);
+    const QRegion &selectedArea = mapDocument()->selectedArea();
+
+    TileLayer *brushLayer = tileLayer->copy(tileLayer->bounds());
```
bjorn: You should not copy the entirely layer (and `Layer::bounds` is technically not in the right coordinate space to use as a parameter to `Layer::copy`, though it won't run into issues with layers at 0,0 which is all layers at the moment). Essentially, I think you want:
```c++
mPreviewLayer = SharedTileLayer(tileLayer->copy(selectedArea.translated(-tileLayer->position())));
```
And since this only copied the relevant region, you'll need to set the initial position:
```c++
mPreviewLayer->setPosition(selectedArea.boundingRect().topLeft());
```

src/tiled/moveselectiontool.cpp
```diff
@@ -0,0 +1,184 @@
...
+    mPreviewLayer = SharedTileLayer(brushLayer);
+
+    brushItem()->setTileLayer(mPreviewLayer);
+    brushItem()->setTileRegion(mapDocument()->selectedArea());
```
bjorn: Please use the version of `setTileLayer` that also takes a region, otherwise `setTileLayer` will first compute the region from the tile layer, which is a waste of time:
```c++
brushItem()->setTileLayer(mPreviewLayer, selectedArea);
```

src/tiled/moveselectiontool.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+    QUndoStack *stack = mapDocument()->undoStack();
+    stack->beginMacro(tr("Move Selection"));
```
bjorn: I think it's a bad idea to leave this macro open, since it will be hard to ensure that it is eventually ended, and even harder to make sure that no other changes are pushed to the stack in between. Maybe it's better just to end this macro here and deal with the fact that the user may undo after starting a move operation.

src/tiled/moveselectiontool.cpp
```diff
@@ -0,0 +1,184 @@
...
+        QRegion selectedArea = mapDocument()->selectedArea();
+        selectedArea.translate(offset);
+
+        mapDocument()->undoStack()->push(new ChangeSelectedArea(mapDocument(), selectedArea));
+
+        if (mPreviewLayer) {
+            mPreviewLayer->setX(mPreviewLayer->x()+offset.x());
+            mPreviewLayer->setY(mPreviewLayer->y()+offset.y());
+            brushItem()->setTileLayer(mPreviewLayer);
+            brushItem()->setTileRegion(mapDocument()->selectedArea());
```
bjorn: Please use the version of `setTileLayer` that takes the region as second parameter. Also, since we shouldn't rely on the selected area, you could instead translate the current tile region by the offset as well: `brushItem()->tileRegion().translated(offset)`.

ketanhwr:
@bjorn I've made the changes, have a look now :slightly_smiling_face:

bjorn:
Just to add an explanatory note: this PR was closed because @ketanhwr introduced an `AbstractTileSelectionTool` in #1620 and will now merge the functionality of the Move Selection tool added here with that class.

PR #1612, opened 2017-06-12
Templates dock

thabetx:
This pull request represents the second checkpoint of the Reusable Object Templates project, it's mainly concerned with creating the dock and managing the created templates.

These are the main items I will focus on doing:
- [x]  Create the templates dock.
- [x]  Display the templates in a tree structure.
- [x]  Load the templates on startup from a predefined path.
- [x]  Create a new template group via a button in the templates dock.
- [x]  Save an object to a specific template group.
- [ ]  Load an existing template group.
- [ ]  Add a button to the dock to unload a template group


src/tiled/objecttemplatemodel.cpp
```diff
@@ -0,0 +1,109 @@
...
+ObjectTemplateModel::ObjectTemplateModel(const QList<TemplateGroup*> &templateGroups, QObject *parent):
+    QAbstractItemModel(parent),
+    mTemplateGroups(templateGroups)
+{
+}
+
+QModelIndex ObjectTemplateModel::index(int row, int column,
+                                       const QModelIndex &parent) const
+{
+    if (parent == QModelIndex())
```
bjorn: Please use `if (!parent.isValid())`

src/tiled/objecttemplatemodel.cpp
```diff
@@ -0,0 +1,109 @@
...
+            return createIndex(row, column, templateGroup->templateAt(row));
+    }
+
+    return QModelIndex();
+}
+
+QModelIndex ObjectTemplateModel::parent(const QModelIndex &index) const
+{
+    if (ObjectTemplate *objectTemplate = toObjectTemplate(index))
+        return createIndex(0, 0, objectTemplate->templateGroup());
```
bjorn: Row 0 can't be always right, you need to determine the index of the template group in `mTemplateGroups`.

src/tiled/objecttemplatemodel.h
```diff
@@ -0,0 +1,52 @@
...
+#include <templategroup.h>
+
+namespace Tiled {
+
+class ObjectTemplateModel : public QAbstractItemModel
+{
+    Q_OBJECT
+
+public:
+    ObjectTemplateModel(const QList<TemplateGroup *> &templateGroups, QObject *parent = 0);
```
bjorn: Please use `nullptr`.

bjorn:
> I think renaming it to templategroups.xml and have internal elements as templategroups might be a good idea but not sure.

Yes, I think these would be better names as well. The "Document" is only an internal thing referring to the fact that it is a file of which the changes and format are tracked. I'm considering to rename this to "Asset", which is a common name for this as well.

> One particular part that I don't like is getting a parent of a template group in the model, I loop through all the documents to find the correct parent.

That's fine.

src/tiled/templategroupdocument.cpp
```diff
@@ -94,3 +102,125 @@ void TemplateGroupDocument::addTemplate(ObjectTemplate *objectTemplate)
...
+                    "TemplateDocuments", "File doesn't contain template documents.");
+        return;
+    }
+
+    while (reader.readNextStartElement()) {
+        if (reader.name() == QLatin1String("templatedocument")) {
+            const QXmlStreamAttributes atts = reader.attributes();
+            const QString path(atts.value(QLatin1String("path")).toString());
+
+            auto templateGroupFormat = new TtxTemplateGroupFormat();
```
bjorn: Hmm, there should be only one instance of `TtxTemplateGroupFormat`. Like this it is a memory leak. I would suggest to just make a singleton for now, by adding a static method:
```c++
TtxTemplateGroupFormat *TtxTemplateGroupFormat::instance()
{
    static TtxTemplateGroupFormat format;
    return &format;
}
```

src/tiled/templategroupdocument.cpp
```diff
@@ -94,3 +102,125 @@ void TemplateGroupDocument::addTemplate(ObjectTemplate *objectTemplate)
...
+
+            // TODO: handle errors that might happen while loading
```
bjorn: Right, at least don't put `nullptr` into `templateDocuments`. :-)

Also, don't forget to initialize `templateGroup` in `MapReaderPrivate::readTemplateGroup` with `nullptr` or you're going to have trouble detecting errors.

src/tiled/objecttemplatemodel.cpp
```diff
@@ -0,0 +1,128 @@
...
+}
+
+QModelIndex ObjectTemplateModel::parent(const QModelIndex &index) const
+{
+    if (!index.isValid())
+        return QModelIndex();
+
+    if (ObjectTemplate *objectTemplate = toObjectTemplate(index)) {
+        auto templateGroup = objectTemplate->templateGroup();
+        // TODO: this doesn't work properly when multiple documents have the same group
```
bjorn: This can't happen. You're comparing pointers here and at no point in the code would it insert the same pointer twice.

src/tiled/preferences.cpp
```diff
@@ -436,6 +443,13 @@ void Preferences::setObjectTypes(const ObjectTypes &objectTypes)
     emit objectTypesChanged();
 }
 
+void Preferences::setTemplateDocuments(const TemplateDocuments &templateDocuments)
+{
+    mTemplateDocuments = templateDocuments;
+    // This signal isn't used yet
```
bjorn: Connect to it in `TemplatesDock`. :-)

src/tiled/templategroupdocument.cpp
thabetx: This is a new way for implementing a singleton :), I thought of creating the static object as a member and manually checking every time when the instance function is called.

bjorn: A member is only needed when the object needs to be manually deleted. This static variable is automatically destroyed at some point when the program exists, which I think will not cause any problems.

thabetx:
Currently groups can be created and templates can be saved and viewed in the dock, the workflow still need some refining I will work on it in the upcoming days. 

There parts that I'm not sure how to do yet so I left TODOs there.

src/tiled/templategroupdocument.cpp
```diff
@@ -94,3 +101,124 @@ void TemplateGroupDocument::addTemplate(ObjectTemplate *objectTemplate)
...
+    writer.setAutoFormatting(true);
+    writer.setAutoFormattingIndent(1);
+
+    writer.writeStartDocument();
+    writer.writeStartElement(QLatin1String("templatedocuments"));
+
+    for (const TemplateGroupDocument *templateDocument : templateDocuments) {
+        writer.writeStartElement(QLatin1String("templatedocument"));
+
+        writer.writeAttribute(QLatin1String("path"), templateDocument->fileName());
```
bjorn: Please use a relative path here (can check the `MapWriter` for examples regarding external tilesets and tileset images).

src/tiled/templatesdock.h
```diff
@@ -0,0 +1,70 @@
...
+class TemplatesView : public QTreeView
+{
+    Q_OBJECT
+
+public:
+    QSize sizeHint() const override;
+    TemplatesView(QWidget *parent = nullptr);
+
+    void applyTemplateGroups();
+    ObjectTemplateModel *objectTemplateModel();
```
bjorn: Should be `const`.

thabetx:
I've done a punch of improvements

- The new template dialog behavior is now similar to the new tileset dialog.
- Insertion now works correctly and the expansion state of the tree will not change.
- Overwriting a template group will also delete the old file from the tree.
- Relative paths are used for the stored document.

I've also handled the tilesets so a template group will have no duplicates, this required storing a reference upon saving any template that had a new tileset.

The references then were required to be deleted upon exit which initially violated some assertions. I had to change the destruction order in the main window as the preferences object which stored the documents was deleted after the tileset manager.

src/tiled/objecttemplatemodel.cpp
```diff
@@ -96,17 +103,53 @@ QVariant ObjectTemplateModel::data(const QModelIndex &index, int role) const
 
 bool ObjectTemplateModel::addNewDocument(QString fileName, QString *error)
 {
-    // TODO: remove the old file if it was already loaded
+    // Remove old document if the new document overwrites it
+    QMutableListIterator<TemplateGroupDocument*> it(mTemplateDocuments);
+    while (it.hasNext()) {
+        if (it.next()->fileName() == fileName) {
+            int pos = mTemplateDocuments.indexOf(it.value());
+            delete it.value();
```
bjorn: To avoid potential invalid memory access, you should move this delete to after the `beginRemoveRows`.

src/tiled/objecttemplatemodel.cpp
```diff
@@ -96,17 +103,53 @@ QVariant ObjectTemplateModel::data(const QModelIndex &index, int role) const
 
 bool ObjectTemplateModel::addNewDocument(QString fileName, QString *error)
 {
-    // TODO: remove the old file if it was already loaded
+    // Remove old document if the new document overwrites it
+    QMutableListIterator<TemplateGroupDocument*> it(mTemplateDocuments);
+    while (it.hasNext()) {
+        if (it.next()->fileName() == fileName) {
+            int pos = mTemplateDocuments.indexOf(it.value());
```
bjorn: Instead of iterating the list forwards and for each match, finding the index, you could instead iterate the list backwards like `for (int i = mTemplateDocuments.size() - 1; i >= 0; --i) {`.

src/tiled/templategroupdocument.cpp
```diff
@@ -46,10 +47,16 @@ TemplateGroupDocument::TemplateGroupDocument(TemplateGroup *templateGroup, const
...
     mFileName = fileName;
+    // TODO: Tileset references are added with each template, if a tileset is not used it will be added again
+    // TilesetManager *tilesetManager = TilesetManager::instance();
+    // tilesetManager->addReferences(mTemplateGroup->tilesets());
 }
 
 TemplateGroupDocument::~TemplateGroupDocument()
 {
+    TilesetManager *tilesetManager = TilesetManager::instance();
+    tilesetManager->removeReferences(mTemplateGroup->tilesets());
```
bjorn: If you add references in the `TemplateGroup`, then please remove them there as well.

thabetx:
The last two items in the list will be moved to other PRs.

PR #1618, opened 2017-06-19
enable setting views through tiled

willt:
This should allow views in gm rooms to be configured through tiled. They syntax is a little long but there isn't really a better way to do it that I can think of short of having a custom drop down.

### DIFF
--- a/src/plugins/gmx/gmxplugin.cpp
+++ b/src/plugins/gmx/gmxplugin.cpp
@@ -111,7 +111,7 @@ bool GmxPlugin::write(const Map *map, const QString &fileName)
     stream.writeStartElement("views");
     for (int c=0; c < 8; c++) {
         QString visible = QString::number(optionalProperty(map, QString("view_%1_visible").arg(c), 0));
-        QString objName = QString::number(optionalProperty(map, QString("view_%1_objName").arg(c), ""));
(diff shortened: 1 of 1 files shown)


bjorn:
> They syntax is a little long but there isn't really a better way to do it that I can think of short of having a custom drop down.

Actually, wouldn't it be nicer to support rectangular objects of with type "View"? That way, at least the x, y, width and height can be visually edited, and only all the other properties would need to be done as custom properties. And they would no longer need to be numbered.

Also, would it be possible to set `enableViews` based on whether any views have been configured? Or would there be a reason to enable this without views, or to disable it even though you defined views?

willt:
Yes rectangular objects to represent views would be nice. It would make
defining the extra properties much cleaner too. I was trying to work with
whats there. I know once I setup a view I really never change it.

Yes you could set enableViews based on if any views were configured. I
don't think there is ever a reason to disable it if you are using views. If
you don't want it enabled then don't define views.

bjorn:
> Yes rectangular objects to represent views would be nice. It would make
> defining the extra properties much cleaner too. I was trying to work with
> whats there. I know once I setup a view I really never change it.

Are you interested in adjusting your patch to work with rectangle objects?

bjorn:
Superseded by #1621.

PR #1620, opened 2017-06-21
Adds specific tools for selection tool

ketanhwr:
Should the tool buttons be just fine? Because using along with modifiers feels a bit weird.


bjorn:
> Should the tool buttons be just fine? Because using along with modifiers feels a bit weird.

The modifiers should definitely still work. What feels weird about it?

The only difference I noticed with GIMP is that GIMP will lock into a certain mode when a button is clicked, whereas when a modifier is held it will only temporarily switched to that mode. In contract, your patch currently always changes back to the "replace" mode when all modifiers are released. Maybe the locking would make it feel less weird?

ketanhwr:
Oh I previously misunderstood what you were trying to say. Then I tried out GIMP and got what you meant. The behaviour is similar now, have a look.

src/tiled/tileselectiontool.cpp
```diff
@@ -190,8 +194,13 @@ void TileSelectionTool::modifiersChanged(Qt::KeyboardModifiers modifiers)
         mSelectionMode = Intersect;
         mIntersect->setChecked(true);
     } else {
-        mSelectionMode = Replace;
-        mReplace->setChecked(true);
+        mSelectionMode = mDefaultMode;
+        switch (mDefaultMode) {
```
bjorn: If you make this switch based on `mSelectionMode`, you can move it out of the condition and remove all the `setChecked(true)` lines above. Then, all the bodies are just one line so you can even remove all the brackets. :-)

src/tiled/tileselectiontool.cpp
```diff
@@ -43,6 +46,47 @@ TileSelectionTool::TileSelectionTool(QObject *parent)
...
+    mIntersect->setToolTip(tr("Intersect Selection"));
+
+    mActionGroup = new QActionGroup(this);
+    mActionGroup->addAction(mReplace);
+    mActionGroup->addAction(mAdd);
+    mActionGroup->addAction(mSubtract);
+    mActionGroup->addAction(mIntersect);
+
+    connect(mReplace, &QAction::triggered,
+        [this]() { mSelectionMode = mDefaultMode = Replace; });
```
bjorn: Needs some more indentation.

src/tiled/tileselectiontool.cpp
```diff
@@ -43,6 +46,47 @@ TileSelectionTool::TileSelectionTool(QObject *parent)
...
+    mReplace->setIcon(replaceIcon);
+    mReplace->setCheckable(true);
+    mReplace->setChecked(true);
+    mReplace->setToolTip(tr("Replace Selection"));
+
+    mAdd = new QAction(this);
+    mAdd->setIcon(addIcon);
+    mAdd->setCheckable(true);
+    mAdd->setToolTip(tr("Add Selection"));
+    mReplace->setShortcut(QKeySequence(tr("Ctrl")));
```
bjorn: Ah, missed this earlier. That should be removed.

src/tiled/tileselectiontool.cpp
```diff
@@ -43,6 +46,47 @@ TileSelectionTool::TileSelectionTool(QObject *parent)
...
+    QIcon replaceIcon(QLatin1String(":images/16x16/selection-replace.png"));
+    QIcon addIcon(QLatin1String(":images/16x16/selection-add.png"));
+    QIcon subtractIcon(QLatin1String(":images/16x16/selection-subtract.png"));
+    QIcon intersectIcon(QLatin1String(":images/16x16/selection-intersect.png"));
+
+    mReplace = new QAction(this);
+    mReplace->setIcon(replaceIcon);
+    mReplace->setCheckable(true);
+    mReplace->setChecked(true);
+    mReplace->setToolTip(tr("Replace Selection"));
```
bjorn: Please just call `languageChanged` at the end of the constructor, so you can remove all these lines here.

bjorn:
Found some nice alternative icons:

https://github.com/shimmerproject/elementary-xfce/blob/master/elementary-xfce/actions/22/selection-break.svg
https://github.com/shimmerproject/elementary-xfce/blob/master/elementary-xfce/actions/22/selection-union.svg
https://github.com/shimmerproject/elementary-xfce/blob/master/elementary-xfce/actions/22/selection-exclude.svg

Only problem is, there is no icon for "intersect", but maybe it won't be too difficult to derive it from these.

Also, as I said on IRC, these buttons should really also be added to the Magic Wand and the Select Same Tile tools. I think a similar approach as done for the Stamp Brush and Fill Tools would work, though actually I had plans to try writing an AbstractSelectionTool anyway...

ketanhwr:
> Also, as I said on IRC, these buttons should really also be added to the Magic Wand and the Select Same Tile tools. I think a similar approach as done for the Stamp Brush and Fill Tools would work, though actually I had plans to try writing an AbstractSelectionTool anyway...

So should I do as of now then?

bjorn:
> So what should I do as of now then?

Well, you could introduce the `AbstractSelectionTool` now, just as way to share the implementation of these tool bar actions. I would later try to merge the move selection tool into it.

bjorn:
Seems like a really nice cleanup of duplicate functionality. Even better now that all selection tools have their selection modes visible!

PR #1621, opened 2017-06-22
Added support for defining gm views with objects

willt:
Okay this adds the ability to define views in game maker by creating an object in tiled.
Set it to the size you want your view. Make sure x,y,width,height, etc are all Ints because thats what GM uses. Numbers will be rounded but that could lead to unexpected results.

If this looks good i can add some better docs and maybe some screen shots
Basically create an object set type to "view"
properties can be defined on the object the match gms properties like so:

igzo9iw

Views must be enabled in the map properties since this config comes before views are defined unless you can think of a better way of auto detecting them. I didn't want to iterate through the layers again just to find out if the views were defined. But we could?
Map Properties:
fq9h23d

### DIFF
--- a/src/plugins/gmx/gmxplugin.cpp
+++ b/src/plugins/gmx/gmxplugin.cpp
@@ -111,15 +111,9 @@ bool GmxPlugin::write(const Map *map, const QString &fileName)
     writeProperty(stream, map, "persistent", false);
     writeProperty(stream, map, "clearDisplayBuffer", true);
     writeProperty(stream, map, "clearViewBackground", false);
-    writeProperty(stream, map, "enableViews", false);
-
-    stream.writeTextElement("isometric", toString(map->orientation() == Map::Orientation::Isometric));
-
-    // Write out views
-    // Last view in Object layer is the first view in the room
-    stream.writeStartElement("views");
+    // Check if views are defined
+    bool enableViews = false;
     LayerIterator iterator(map);
-    int viewCount = 0;
     while (const Layer *layer = iterator.next()) {
 
         if (layer->layerType() != Layer::ObjectGroupType)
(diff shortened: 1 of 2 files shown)


src/plugins/gmx/gmxplugin.cpp
```diff
@@ -110,16 +110,69 @@ bool GmxPlugin::write(const Map *map, const QString &fileName)
...
+        if (layer->layerType() != Layer::ObjectGroupType)
+            continue;
+
+        const ObjectGroup *objectLayer = static_cast<const ObjectGroup*>(layer);
+
+          for (const MapObject *object : objectLayer->objects()) {
+            const QString type = effectiveObjectType(object);
+            if (type.isEmpty())
+                continue;
+            if (type.operator !=("view"))
```
bjorn: Please just write this instead:
```c++
if (type != "view")
```

src/plugins/gmx/gmxplugin.cpp
```diff
@@ -110,16 +110,69 @@ bool GmxPlugin::write(const Map *map, const QString &fileName)
...
+
+            // GM only has 8 views so drop anything more than that
+            if (viewCount > 7)
+                break;
+
+            viewCount++;
+            stream.writeStartElement("view");
+
+            stream.writeAttribute("visible", toString(object->isVisible()));
+            stream.writeAttribute("objName", QString(optionalProperty(object, "objName", QString(""))));
```
bjorn: Please construct an empty string with `QString()` instead of `QString("")`. It is less verbose and more efficient. :-)

src/plugins/gmx/gmxplugin.cpp
```diff
@@ -132,6 +185,8 @@ bool GmxPlugin::write(const Map *map, const QString &fileName)
             const QString type = effectiveObjectType(object);
             if (type.isEmpty())
                 continue;
+            if (type.operator ==("view"))
```
bjorn: Just write:
```c++
if (type == "view")
```

src/plugins/gmx/gmxplugin.cpp
```diff
@@ -110,16 +110,69 @@ bool GmxPlugin::write(const Map *map, const QString &fileName)
...
+    while (const Layer *layer = iterator.next()) {
+
+        if (layer->layerType() != Layer::ObjectGroupType)
+            continue;
+
+        const ObjectGroup *objectLayer = static_cast<const ObjectGroup*>(layer);
+
+          for (const MapObject *object : objectLayer->objects()) {
+            const QString type = effectiveObjectType(object);
+            if (type.isEmpty())
```
bjorn: An additional empty check isn't needed, cause it is already covered by the next check.

willt:
Okay things are refactored. Much happier with enabling the views by default now. Just was worried about iterating too many times.
Some example images:
viewproperties
mapproperties

tiledviews

roomwithviewingm

runningingm

src/plugins/gmx/gmxplugin.cpp
```diff
@@ -110,16 +110,87 @@ bool GmxPlugin::write(const Map *map, const QString &fileName)
...
+    bool enableViews = false;
+    LayerIterator iterator(map);
+    while (const Layer *layer = iterator.next()) {
+
+        if (layer->layerType() != Layer::ObjectGroupType)
+            continue;
+
+        const ObjectGroup *objectLayer = static_cast<const ObjectGroup*>(layer);
+
+          for (const MapObject *object : objectLayer->objects()) {
```
bjorn: Coding style: please fix up the indentation (this line is indented too much, and lines below don't use 4 spaces indent)

src/plugins/gmx/gmxplugin.cpp
```diff
@@ -110,16 +110,87 @@ bool GmxPlugin::write(const Map *map, const QString &fileName)
...
+        iterator.toFront();
+        int viewCount = 0;
+        while (const Layer *layer = iterator.next()) {
+
+            if (layer->layerType() != Layer::ObjectGroupType)
+                continue;
+
+            const ObjectGroup *objectLayer = static_cast<const ObjectGroup*>(layer);
+
+              for (const MapObject *object : objectLayer->objects()) {
```
bjorn: Coding style: indentation is off here.

src/plugins/gmx/gmxplugin.cpp
```diff
@@ -110,16 +110,87 @@ bool GmxPlugin::write(const Map *map, const QString &fileName)
...
+        if (layer->layerType() != Layer::ObjectGroupType)
+            continue;
+
+        const ObjectGroup *objectLayer = static_cast<const ObjectGroup*>(layer);
+
+          for (const MapObject *object : objectLayer->objects()) {
+            const QString type = effectiveObjectType(object);
+            if (type == "view") {
+                enableViews = true;
+                break;
```
bjorn: This break only aborts the inner loop, but it will continue iterating the other layers. Can be solved by either another check whether `enableViews` is already true and then a `break`, or by extracting this "check if there are any views" out into a function (my preference).

willt:
Let me know if this works. Thanks for the feedback. strange i missed the indentation.

src/plugins/gmx/gmxplugin.cpp
```diff
@@ -83,6 +83,26 @@ GmxPlugin::GmxPlugin()
 {
 }
 
+bool GmxPlugin::checkIfViewsDefined(LayerIterator iterator)
```
bjorn: Since this function makes no use of `GmxPlugin` state, it can be a local static function:
```c++
static bool checkIfViewsDefined(const Map *map)
```
Also since there is no performance gain in re-using iterator, I would suggest just passing in the map and having a local iterator in the function.

src/plugins/gmx/gmxplugin.cpp
```diff
@@ -83,6 +83,26 @@ GmxPlugin::GmxPlugin()
...
+        if (layer->layerType() != Layer::ObjectGroupType)
+            continue;
+
+        const ObjectGroup *objectLayer = static_cast<const ObjectGroup*>(layer);
+
+        for (const MapObject *object : objectLayer->objects()) {
+            const QString type = effectiveObjectType(object);
+            if (type == "view") {
+                return true;
+            }
```
bjorn: Coding style: if there is only a single line body, please leave out the braces.

willt:
No worries your feedback is most appreciated.

bjorn:
Looking good, thanks a lot for this improvement!

Issue #1560, opened 2017-05-02
Add support for editing Stardew Valley maps


bjorn:
Stardew Valley maps are stored inside `.xnb` files (produced by Microsoft XNA Game Studio) in `.tbin` format (a [binary tile map format](https://tide.codeplex.com/SourceControl/latest#TileMapEditor/xTile/Format/TbinFormat.cs) used by [tIDE](https://tide.codeplex.com)). It should be possible to write a Tiled plugin that can directly load and save such maps.

* [SimpleXNBDemapper](https://github.com/Inari-Whitebear/SimpleXNBDemapper) shows how `.tbin` and `.png` files can be unpacked from (and packed into) the `.xnb` files (see [Program.cs](https://github.com/Inari-Whitebear/SimpleXNBDemapper/blob/master/SimpleXNBDemapper/Program.cs))

* [TbinFormat.cs](https://tide.codeplex.com/SourceControl/latest#TileMapEditor/xTile/Format/TbinFormat.cs) shows how a `.tbin` file can be parsed

On the Tiled forum [somebody reported](http://discourse.mapeditor.org/t/copying-and-pasting-between-maps-broken/2318) to have successfully extracted the necessary `.tbin` and `.png` files and then used tIDE to open the map and save it as `.tmx` so that it could be opened in Tiled. It would be cool to simplify this process by supporting `.tbin` and maybe even `.xnb` files containing `.tbin` files directly.

spacechase0:
So I got my XNB loader working for most of the Stardew files (and certainly the ones Tiled would need). It's [here](https://github.com/spacechase0/XNB). I re-ported the library I mentioned earlier [here](https://github.com/spacechase0/libmspack-lzx).

If you aren't happy with that license for that compressor, someone apparently wrote their own [here](https://github.com/LeonBlade/xnbcli/blob/master/app/Presser/Lzx.js). (It doesn't have a license right this moment, but the author tells me he'll put one up in the next day or two. He mentioned "MIT or whatever" earlier today.)

I'm probably going to be writing a tbin loader soon to feature-creep my own project. I'll throw that under MIT as well.

bjorn:
@spacechase0 That's so awesome! Let me know if I can help in any way. Right now, I've got a lot of other things with higher priority, like fixing things for Tiled 1.0.1, mentoring the GSoC students and updating the manual, but I'd really like to work on this as well.

spacechase0:
[...]

PR #1622, opened 2017-06-22
Implement .tbin file format

spacechase0:
(#1560)

At the moment, loading is fully implemented (except tilesheet tile properties - they're stored as properties on the tilesheet with tbins, I forgot about that (EDIT: And it looks like individual tile properties are currently stored in the tilesheet?)) besides needing to manually fix tilesheet image sources. (It says it can't find it in '.', do I need to put the full path?)

Next I'm going to do saving, and then take out `FakeSfml.hpp` and fix things to use the primitive types directly, as well as `QPointI` (?) instead of `sf::Vector2i` (plus any other cleanup wanted).

### DIFF
--- a/src/plugins/tbin/tbin.pro
+++ b/src/plugins/tbin/tbin.pro
@@ -1,6 +1,6 @@
 include(../plugin.pri)
 
-DEFINES += JSON_LIBRARY
+DEFINES += TBIN_LIBRARY
 
 SOURCES += tbinplugin.cpp \
     tbin/Map.cpp
@@ -11,4 +11,5 @@ HEADERS += tbin_global.h \
     tbin/Map.hpp \
     tbin/PropertyValue.hpp \
     tbin/Tile.hpp \
-    tbin/TileSheet.hpp
+    tbin/TileSheet.hpp \
+    tbin/FakeSfml.hpp
--- a/src/plugins/tbin/tbin.pro
+++ b/src/plugins/tbin/tbin.pro
@@ -1,0 +1,14 @@
+include(../plugin.pri)
+
+DEFINES += JSON_LIBRARY
+
+SOURCES += tbinplugin.cpp \
+    tbin/Map.cpp
+
+HEADERS += tbin_global.h \
+    tbinplugin.h \
+    tbin/Layer.hpp \
+    tbin/Map.hpp \
+    tbin/PropertyValue.hpp \
+    tbin/Tile.hpp \
+    tbin/TileSheet.hpp
--- a/src/plugins/tbin/tbin/Map.cpp
+++ b/src/plugins/tbin/tbin/Map.cpp
@@ -1,0 +1,381 @@
+#include "tbin/Map.hpp"
+
+#include <fstream>
+#include <stdexcept>
+
+#include "FakeSfml.hpp"
+
+namespace tbin
+{
+    template< typename T >
+    T read( std::istream& in )
+    {
+        T t;
+        in.read( reinterpret_cast< char* >( &t ), sizeof( T ) );
+        return t;
+    }
+
+    template<>
+    sf::Vector2i read< sf::Vector2i >( std::istream& in )
+    {
+        sf::Int32 x = read< sf::Int32 >( in );
+        sf::Int32 y = read< sf::Int32 >( in );
+        return sf::Vector2i( x, y );
+    }
+
+    template<>
+    std::string read< std::string >( std::istream& in )
+    {
+        auto len = read< sf::Int32 >( in );
+        std::string str( len, 0 );
+        in.read( &str[ 0 ], len );
+        return str;
+    }
+
+    template< typename T >
+    void write( std::ostream& out, const T& t )
+    {
+        out.write( reinterpret_cast< const char* >( &t ), sizeof( T ) );
+    }
+
+    template<>
+    void write< sf::Vector2i >( std::ostream& out, const sf::Vector2i& vec )
+    {
+        write< sf::Int32 >( out, vec.x );
+        write< sf::Int32 >( out, vec.y );
+    }
+
+    template<>
+    void write< std::string >( std::ostream& out, const std::string& str )
+    {
+        write< sf::Int32 >( out, str.length() );
+        out.write( &str[ 0 ], str.length() );
+    }
+
+    Properties readProperties( std::istream& in )
+    {
+        Properties ret;
+
+        int count = read< sf::Int32 >( in );
+        for ( int i = 0; i < count; ++i )
+        {
+            std::string key;
+            PropertyValue value;
+
+            key = read< std::string >( in );
+            value.type = static_cast< PropertyValue::Type >( read< sf::Uint8 >( in ) );
+            switch ( value.type )
+            {
+                case PropertyValue::Bool:    value.data.b  = read< sf::Uint8   >( in ) > 0; break;
+                case PropertyValue::Integer: value.data.i  = read< sf::Int32   >( in );     break;
+                case PropertyValue::Float:   value.data.f  = read< float       >( in );     break;
+                case PropertyValue::String:  value.dataStr = read< std::string >( in );     break;
+                default: throw std::invalid_argument( util::format( "Bad property type: $", static_cast< int >( value.type ) ) );
+            }
+
+            ret[ key ] = value;
+        }
+
+
+        return std::move( ret );
+    }
+
+    void writeProperties( std::ostream& out, const Properties& props )
+    {
+        write< sf::Int32 >( out, props.size() );
+        for ( const auto& prop : props )
+        {
+            write( out, prop.first );
+            write< sf::Uint8 >( out, prop.second.type );
+            switch ( prop.second.type )
+            {
+                case PropertyValue::Bool: write< sf::Uint8 >( out, prop.second.data.b ? 1 : 0 ); break;
+                case PropertyValue::Integer: write( out, prop.second.data.i ); break;
+                case PropertyValue::Float: write( out, prop.second.data.f ); break;
+                case PropertyValue::String: write( out, prop.second.dataStr ); break;
+                default: throw std::invalid_argument( util::format( "Bad property type: $", prop.second.type ) );
+            }
+        }
+    }
+
+    TileSheet readTilesheet( std::istream& in )
+    {
+        TileSheet ret;
+        ret.id = read< std::string >( in );
+        ret.desc = read< std::string >( in );
+        ret.image = read< std::string >( in );
+        ret.sheetSize = read< sf::Vector2i >( in );
+        ret.tileSize = read< sf::Vector2i >( in );
+        ret.margin = read< sf::Vector2i >( in );
+        ret.spacing = read< sf::Vector2i >( in );
+        ret.props = readProperties( in );
+        return std::move( ret );
+    }
+
+    void writeTilesheet( std::ostream& out, const TileSheet& ts )
+    {
+        write( out, ts.id );
+        write( out, ts.desc );
+        write( out, ts.image );
+        write( out, ts.sheetSize );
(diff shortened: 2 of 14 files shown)


bjorn:
> besides needing to manually fix tilesheet image sources. (It says it can't find it in '.', do I need to put the full path?)

Normally the paths in a Tiled map file are to be interpreted as relative from the location of the map file. Isn't that the case for .tbin as well? In any case you'll have to somehow derive an absolute path from which to load the tileset images, yes.

> Next I'm going to do saving, and then take out FakeSfml.hpp and fix things to use the primitive types directly, as well as QPointI (?) instead of sf::Vector2i (plus any other cleanup wanted).

The equivalent of `sf:Vector2i` would be `QPoint`. However, if you'd like to keep that part of your code more generic and not depending on Qt that'd be fine and you can just define your own point class. It could even be set up as a small static library that doesn't include Qt at all (I could help you set up the Qbs project that way if you like).

spacechase0:
Got the writing done. I tested it with tIDE, and it can read them as well. I'll do more thorough testing once I get the tilesheet properties fixed and image sources fixed.

> Normally the paths in a Tiled map file are to be interpreted as relative from the location of the map file. Isn't that the case for .tbin as well?
Yes. (At least for Stardew Valley, and the modloader we use that can load custom maps.)

> It could even be set up as a small static library that doesn't include Qt at all (I could help you set up the Qbs project that way if you like).
I already have it as a library in the main repo. Thanks for the offer though. :)

src/plugins/tbin/tbin.pro
```diff
@@ -0,0 +1,14 @@
+include(../plugin.pri)
+
+DEFINES += JSON_LIBRARY
```
bjorn: Should say `TBIN_LIBRARY`.

src/plugins/tbin/tbinplugin.h
```diff
@@ -0,0 +1,68 @@
+/*
+ * JSON Tiled Plugin
```
bjorn: It's the TBIN plugin. :-)

src/plugins/tbin/tbinplugin.cpp
```diff
@@ -0,0 +1,358 @@
+/*
+ * JSON Tiled Plugin
```
bjorn: TBIN Tiled Plugin.

src/plugins/tbin/tbinplugin.cpp
```diff
@@ -0,0 +1,358 @@
...
+#include "tile.h"
+#include "tilelayer.h"
+
+#include <cmath>
+#include <fstream>
+#include <map>
+#include <QFile>
+#include <QFileInfo>
+#include <QJsonDocument>
+#include <QJsonObject>
```
bjorn: These includes can be removed.

src/plugins/tbin/tbinplugin.cpp
```diff
@@ -0,0 +1,358 @@
...
+QString TbinMapFormat::shortName() const
+{
+    return QLatin1String("tbin");
+}
+
+bool TbinMapFormat::supportsFile(const QString &fileName) const
+{
+    // ...
+
+    return false;
```
bjorn: This should probably return whether the extension of the file is `tbin`, and/or it could open the file and check for a certain expected first few bytes. That should allow Tiled to open .tbin files without explicitly selecting that file format in the open file dialog.

src/plugins/tbin/tbinplugin.cpp
```diff
@@ -0,0 +1,358 @@
...
+    }
+
+    QByteArray contents = file.readAll();
+    std::istringstream ss( std::string( contents.constData(), contents.length() ) );
+
+    tbin::Map tmap;
+    Tiled::Map* map;
+    try
+    {
+        tmap.loadFromStream( ss );
```
bjorn: Coding style: please don't put spaces on the inner sides of parentheses (you're fine to keep them in the tbin library part, of course).

src/plugins/tbin/tbinplugin.cpp
```diff
@@ -0,0 +1,358 @@
...
+        {
+            mError = tr( "Could not open file for writing" );
+            return false;
+        }
+        tmap.saveToStream( file );
+        file.close();
+    }
+    catch ( std::exception& e )
+    {
+        mError = tr((std::string("Exception: ") + e.what()).c_str());
```
bjorn: Passing a dynamic string into `tr` is not going to work, since the `lupdate` tool will not be able to extract the source string this way. It will work however, if the possible strings are wrapped elsewhere with [`QT_TR_NOOP`](http://doc.qt.io/qt-5/qtglobal.html#QT_TR_NOOP). Something like the following may work:
```c++
throw std::invalid_argument(QT_TR_NOOP("Only object and tile layers supported."));
```
and then
```c++
mError = tr("Exception: %1").arg(tr(e.what()));
```

spacechase0:
> Hmm, I don't know exactly what ImageSource contains, but I would expect it to be a full path?
>
> Hmm, I see one file has proper tile sizes and neither of them seem to have a valid source? But what are those screenshots from?

The screenshots are from the tileset page inside Tiled. The second one (the one exported by tIDE as aTMX) works fine.

I ended up with: `tilesheet->setImageSource((dir.absolutePath().toStdString() + QDir::separator().toLatin1() + ttilesheet.image).c_str());`. This appears to set the correct *path* on the area where those screenshots were taken (for the top part, where I press 'Edit...'), but "Source" is still blank, and the tile width/height is still 0. [Image](http://i.imgur.com/iwGC7PV.png)

I noticed `MapReaderPrivate` uses `MapReader::resolveReference` for this. Would I need to use this, and if so how would I access this?

> Right, but I meant setting it up to build as a static library as part of the Tiled plugin. It would keep things more cleanly separated.
Ah. Sure, I'd love help with that.

src/plugins/tbin/tbinplugin.cpp
spacechase0: Would you prefer both or just one of the two?

bjorn: I did not check whether the `tbin` format has a recognizable and unique first few bytes. If it has, then you can leave out a check on the file extension.

bjorn:
> The screenshots are from the tileset page inside Tiled. The second one (the one exported by tIDE as aTMX) works fine.

Ok I knew what I was looking at as such, but I didn't know what files were being shown. I'll try to have a look at the tile size issue, cause if it's showing 0 even if things are working then there is probably an existing bug in Tiled.

> I ended up with: tilesheet->setImageSource((dir.absolutePath().toStdString() + QDir::separator().toLatin1() + ttilesheet.image).c_str());. This appears to set the correct path on the area where those screenshots were taken (for the top part, where I press 'Edit...'), but "Source" is still blank,

I think you will need to call `tileset->loadImage()` at some point.

Btw, you should use this instead to get a clean absolute path to the tileset image:
```c++
QDir::cleanPath(dir.filePath(QString::fromStdString(ttilesheet.image)))
```

src/plugins/tbin/tbinplugin.cpp
```diff
@@ -0,0 +1,353 @@
...
+                //tlayer.visible = ???;
+                for (int iy = 0; iy < tlayer.layerSize.y; ++iy)
+                {
+                    for (int ix = 0; ix < tlayer.layerSize.x; ++ix)
+                    {
+                        Tiled::Cell cell = layer->cellAt(ix, iy);
+                        tbin::Tile ttile;
+                        ttile.staticData.tileIndex = -1;
+
+                        if (cell.tile() != NULL)
```
bjorn: Please do this to avoid looking up the tile many times:
```c++
if (Tile *tile = cell.tile()) {
```

Coding style: please put the `{` after `if`, `for`, `catch`, etc. on the same line.

src/plugins/tbin/tbin/Map.cpp
```diff
@@ -0,0 +1,384 @@
...
+
+            if ( nulls > 0 )
+            {
+                write< sf::Uint8 >( out, 'N' );
+                write( out, nulls );
+            }
+        }
+    }
+
+    constexpr const char* MAGIC_1_0 = "tBIN10";
```
bjorn: Alright, so there is a proper magic string. :-)

Unfortunately, I'm building the Windows version of Tiled with VS 2013, which does not appear to support `constexpr`:
```
C:\projects\tiled\src\plugins\tbin\tbin\Map.cpp(299) : error C2143: syntax error : missing ';' before 'const'
C:\projects\tiled\src\plugins\tbin\tbin\Map.cpp(299) : error C4430: missing type specifier - int assumed. Note: C++ does not support default-int
```
Since you already started to use Qt stuff here, you could replace it with the [`Q_DECL_CONSTEXPR`](http://doc.qt.io/qt-5/qtglobal.html#Q_DECL_CONSTEXPR) macro.

spacechase0: Are there any other issues relating to C++11, or is `constexpr` the only thing (so far)?

bjorn: I think VS 2013 is also not supporting member initialization at declaration, but in this case, we can just fix up this line and see if the build on AppVeyor finds anything else to complain about.

spacechase0:
I noticed that tIDE can't load the tbin files exported. My original version can load it, and if it loads the original tbin and resaves it tIDE can read that. Reading and saving the tiled one doesn't work though.

The [differences ](https://www.diffchecker.com/19LxDwUv) between the versions are rather minimal though. I'll need to do more digging.

bjorn:
@spacechase0 Did you get stuck on this problem? Anything I can help with?

spacechase0:
I was, but looking at it again this morning I figured it out. All that's left is per-tile properties.

bjorn:
Are you aware that clang finds your usage of `std::move` non-optimal? It says it's preventing it from doing more efficient handling of return values:
```
src/plugins/tbin/tbin/Map.cpp:81:16: warning: moving a local object in a return statement prevents copy elision [-Wpessimizing-move]
        return std::move( ret );
               ^
src/plugins/tbin/tbin/Map.cpp:81:16: note: remove std::move call here
        return std::move( ret );
               ^~~~~~~~~~~    ~
```

spacechase0:
I had no clue. I knew compilers could optimize that case, but I didn't realize an explicit move would prevent that.

bjorn:
Note that I had added more warnings about `std::move` in other places in an edit. :-)

bjorn:
Btw, it would be good to add copyright headers to the files from your library so that the license of that code is clear.

spacechase0:
You know, I suddenly don't know when I'm ever supposed to use move. :P Guess I'll need to read up on that stuff again.

bjorn:
> You know, I suddenly don't know when I'm ever supposed to use move. :P Guess I'll need to read up on that stuff again.

Well, the most common scenario is when you have some kind of large object that you construct locally, and then want to pass it as a parameter, but you no longer care for that object locally after that. Since normally the local would stay around, the compiler generally would have to make a copy (unless the parameter is `const &`, but in that case a copy may need to be made inside of the function you're calling). You can avoid any copying in this case by taking the parameter "by value" and passing it with `std::move`.

Just note that the above case has nothing to do with return values or passing of temporary values, which are apparently both cases that the compiler will already optimize for you in a more optimal way than move constructor / assignment could do.

spacechase0:
Oh, and I still see a 'changes requested', but when I go there it only shows outdated diffs. Am I missing one?

bjorn:
@spacechase0 I think the "changes requested" stays until I did a "approved" review. But since I'm the only one who needs to approve, I usually don't bother because by that point I will anyway merge it.

Unfortunately I am not able to test your plugin right now, but it looks good from a code point of view and I will try to get around to trying it out soon!

PR #1623, opened 2017-06-23
Template instances

thabetx:
This is the third checkpoint in the reusable object templates project, it will be mainly concerned with reading template instance in a map.

The format will look like this

```
 <templategroup firsttid="1" source="group1.ttx"/>
 <templategroup firsttid="11" source="group2.ttx"/>

 <object id="66" tid="2" x="191" y="342"/>
```
The instance in the map file would be the same as an object except that it would have an extra attribute similar to the gid called the tid standing for template id. Template ids would work like tileset ids, the map assigns the firsttid to each template group and the template manager would resolve those ids.

If a template object has properties they will override the template.

- [x] Create TemplatesManager to aid in mapping.
- [x] Create the tid mapping functionality.
- [x] Handle reading errors.
- [x] Write template group references inside the map.
- [x] Write template instances inside the map.
- [x] Reduce redundant data when writing.

Currently the pull request is focused on reading an instance but may be expanded to also cover creation.


thabetx:
The code can now read template instances and load them into the map. I will do more testing and handle possible errors.

src/libtiled/mapwriter.cpp
```diff
@@ -32,12 +32,14 @@
 
 #include "compression.h"
 #include "gidmapper.h"
+#include "tidmapper.h"
```
bjorn: Duplicate include.

src/libtiled/mapwriter.cpp
```diff
@@ -618,74 +652,100 @@ void MapWriterPrivate::writeObjectGroup(QXmlStreamWriter &w,
     w.writeEndElement();
 }
 
+static bool shouldWrite(bool holdsInfo, bool isTemplateInstance, bool changed) {
+    return (!isTemplateInstance && holdsInfo) ||
+           (isTemplateInstance && changed);
```
bjorn: I think it is easier to understand as:
```c++
return isTemplateInstance ? changed : holdsInfo;
```

src/libtiled/mapwriter.cpp
```diff
@@ -618,74 +652,100 @@ void MapWriterPrivate::writeObjectGroup(QXmlStreamWriter &w,
     w.writeEndElement();
 }
 
+static bool shouldWrite(bool holdsInfo, bool isTemplateInstance, bool changed) {
```
bjorn: Coding style: `{` should be on the next line for function definitions.

src/libtiled/mapwriter.cpp
```diff
@@ -618,74 +652,100 @@ void MapWriterPrivate::writeObjectGroup(QXmlStreamWriter &w,
...
     const int id = mapObject.id();
     const QString &name = mapObject.name();
     const QString &type = mapObject.type();
-    if (id != 0)
+
+    const QPointF pos = mapObject.position();
+
+    TemplateRef templateRef = mapObject.templateRef();
+    const int templateId = templateRef.templateId;
+    const auto group = templateRef.templateGroup;
```
bjorn: You're not frequently referring to these variables, so please just refer to them using `templateRef.templateId` and `templateRef.templateGroup` instead of having these locals.

src/libtiled/mapwriter.cpp
```diff
@@ -618,74 +652,100 @@ void MapWriterPrivate::writeObjectGroup(QXmlStreamWriter &w,
...
+    bool isTemplateInstance = group;
+
+    if (isTemplateInstance) {
+        int groupFirstTid = mTidMapper.templateGroupToFirstTid(group);
```
bjorn: I would suggest to have a `mTidMapper.templateRefToTid(templateRef)` instead, which is more convenient.

src/libtiled/mapwriter.cpp
```diff
@@ -618,74 +652,100 @@ void MapWriterPrivate::writeObjectGroup(QXmlStreamWriter &w,
...
+        int tid = templateId + groupFirstTid;
+        w.writeAttribute(QLatin1String("tid"), QString::number(tid));
+    }
+
+    if (id != 0 && !isTemplateInstance)
```
bjorn: Template instances need to write their ID as well.

src/libtiled/mapwriter.cpp
```diff
@@ -618,74 +652,100 @@ void MapWriterPrivate::writeObjectGroup(QXmlStreamWriter &w,
...
-    if (rotation != 0.0)
+    if (shouldWrite(rotation != 0.0, isTemplateInstance, mapObject.propertyChanged(MapObject::RotationProperty)))
         w.writeAttribute(QLatin1String("rotation"), QString::number(rotation));
 
-    if (!mapObject.isVisible())
+    if (shouldWrite(!mapObject.isVisible(), isTemplateInstance, mapObject.propertyChanged(MapObject::VisibleProperty)))
         w.writeAttribute(QLatin1String("visible"), QLatin1String("0"));
 
-    writeProperties(w, mapObject.properties());
+    if (!isTemplateInstance)
```
bjorn: Properties should always be written, since this will only include overridden values and any properties only added to the instance.

src/tiled/objecttemplatemodel.h
```diff
@@ -49,21 +49,21 @@ class ObjectTemplateModel : public QAbstractItemModel
...
-    TemplateDocuments mTemplateDocuments;
+    TemplateDocuments *mTemplateDocuments;
     ObjectTemplate *toObjectTemplate(const QModelIndex &index) const;
     TemplateGroup *toTemplateGroup(const QModelIndex &index) const;
 };
 
-inline void ObjectTemplateModel::setTemplateDocuments(const TemplateDocuments &templateDocuments)
-{ mTemplateDocuments = templateDocuments; }
+inline void ObjectTemplateModel::setTemplateDocuments(TemplateDocuments &templateDocuments)
+{ mTemplateDocuments = &templateDocuments; }
```
bjorn: `mTemplateDocuments` should not be a pointer. Rather, just store the list here by value and when you need it in the `TemplatesDock`, have it call `ObjectTemplateModel::templateDocuments` and remove `TemplatesDock::mTemplateDocuments`. The function should take `const TemplateDocuments &templateDocuments`.

Also, since this affects the model, it should call `beginResetModel` and `endResetModel`, even though currently the function may only be called while the model is not yet used by a view, this could change later on.

It should probably also own the documents, so it should call `qDeleteAll(mTemplateDocuments)` before the assignment, and this should be added to the destructor (instead of doing it in `TemplatesDock`).

src/tiled/objecttemplatemodel.h
```diff
@@ -49,21 +49,21 @@ class ObjectTemplateModel : public QAbstractItemModel
...
 
 inline const TemplateDocuments ObjectTemplateModel::templateDocuments() const
-{ return mTemplateDocuments; }
+{ return *mTemplateDocuments; }
```
bjorn: This function should return `const TemplateDocuments &`.

src/tiled/templatesdock.cpp
```diff
@@ -62,6 +63,36 @@ TemplatesDock::TemplatesDock(QWidget *parent):
...
+
+    TemplateGroups templateGroups;
+
+    for (auto document : mTemplateDocuments)
+        templateGroups.append(document->templateGroup());
+
+    manager->setTemplateGroups(templateGroups);
+}
+
+TemplatesDock::~TemplatesDock() {
```
bjorn: Coding style: `{` should be on the next line.

src/libtiled/mapreader.cpp
```diff
@@ -921,36 +964,72 @@ MapObject *MapReaderPrivate::readObject()
     const QPointF pos(x, y);
     const QSizeF size(width, height);
 
-    MapObject *object = new MapObject(name, type, pos, size);
+    MapObject *object = new MapObject;
+
+    if (tid) { // This object is a template instance
+        ObjectTemplate *objectTemplate = templateForTid(tid);
+        if (objectTemplate) {
+            object->setTemplateRef({objectTemplate->templateGroup(), objectTemplate->id()});
```
thabetx: I think with how I implemented the tids, keeping them valid will be more complicated because If a tid wasn't mapped to any object, the templateRef won't be valid, and the tid will be lost forever.

I think the cause of the problem is, unlike tiles in a tileset, every template in the group has it's own id, which is part of the templateRef, which is used when the tid is created, so maybe we need a new way of how the templateRef works.

bjorn: This was explained on IRC, right?

thabetx: Yes.

src/tiled/templatesdock.h
```diff
@@ -49,6 +52,7 @@ private slots:
     void retranslateUi();
 
     TemplatesView *mTemplatesView;
+    TemplateDocuments mTemplateDocuments;
```
bjorn: It seems you forgot to remove this  member, and the above include?

src/tiled/flipmapobjects.cpp
```diff
@@ -64,15 +64,27 @@ FlipMapObjects::FlipMapObjects(MapDocument *mapDocument,
...
 void FlipMapObjects::flip()
 {
     //flip objects
-    for (MapObject *object : mMapObjects)
-        object->flip(mFlipDirection, mObjectsCenter);
+    for (int i = 0; i < mMapObjects.size(); ++i) {
+        mMapObjects[i]->flip(mFlipDirection, mObjectsCenter);
+
+        mMapObjects[i]->setPropertyChanged(MapObject::CellProperty, mNewCellStates[i]);
+        mMapObjects[i]->setPropertyChanged(MapObject::RotationProperty, mNewCellStates[i]);
```
bjorn: `mNewCellStates` should be `mNewRotationStates`

src/libtiled/map.cpp
```diff
@@ -288,6 +289,33 @@ bool Map::isTilesetUsed(const Tileset *tileset) const
...
+    for (auto group : objectGroups()) {
+        for (auto o : group->objects()){
+            if (o->templateRef().templateGroup == oldTemplateGroup) {
+                o->setTemplateRef({newTemplateGroup, o->templateRef().templateId});
+                o->syncWithTemplate();
+            }
+        }
+    }
+
+    delete oldTemplateGroup;
```
bjorn: Maybe somebody should delete the old template group, but it's not the map.

src/tiled/brokenlinks.cpp
```diff
@@ -73,6 +78,22 @@ Tileset *BrokenLink::tileset() const
...
+
+TemplateGroup *BrokenLink::templateGroup() const
+{
+    switch (type) {
+    case TilesetImageSource:
+    case MapTilesetReference:
+    case TilesetTileImageSource:
+        return nullptr;
+    case TemplateGroupReference:
+        return _templateGroup;
```
bjorn: Well, that is a lot of lines for writing:
```c++
return type == TemplateGroupReference ? _templateGroup : nullptr;
```
:-)

src/tiled/brokenlinks.cpp
```diff
@@ -161,8 +194,21 @@ void BrokenLinksModel::refresh()
...
             }
+
+            for (TemplateGroup *templateGroup : mapDocument->map()->templateGroups()) {
+                if (!templateGroup->fileName().isEmpty() && !templateGroup->loaded()) {
+                    BrokenLink link;
+                    link.type = TemplateGroupReference;
+                    link._templateGroup = templateGroup;
+                    mBrokenLinks.append(link);
+                } else {
+                    processTemplateGroup(templateGroup);
```
bjorn: This else makes no sense, because it does essentially the same as the code above, just without the check on the file name being empty (which should be there, I guess).

src/tiled/brokenlinks.cpp
```diff
@@ -404,6 +460,7 @@ void BrokenLinksWidget::selectionChanged()
     mLocateButton->setEnabled(!selection.isEmpty());
 
     bool isTileset = qobject_cast<TilesetDocument*>(mBrokenLinksModel->document()) != nullptr;
+    bool isTemplateGroup = qobject_cast<TemplateGroupDocument*>(mBrokenLinksModel->document()) != nullptr;
```
bjorn: The broken links model document can never be a template group document, since we don't currently have a template group editor.

src/tiled/brokenlinks.cpp
```diff
@@ -420,6 +477,11 @@ void BrokenLinksWidget::selectionChanged()
             else
                 mLocateButton->setText(tr("Open Tileset..."));
             break;
+        case Tiled::Internal::TemplateGroupReference:
+            if (isTemplateGroup)
+                mLocateButton->setText(tr("Locate File..."));
+            else
+                mLocateButton->setText(tr("Open Template Group..."));
```
bjorn: The button should always read "Locate File..." at this point, since we can't open template groups directly.

bjorn:
@thabetx Looks like your last commit should go to the `1.0` branch. :-)

thabetx:
Okay, I will test it on master.

src/libtiled/templategroup.cpp
```diff
@@ -66,9 +74,18 @@ void TemplateGroup::addTileset(const SharedTileset &tileset)
     if (mTilesets.contains(tileset))
         return;
 
-    // TODO: If a tileset was  not used by any  template it will be added again
+    // TODO: If a tileset was not used by any template it will be added again
```
bjorn: What is the thing that needs to be done here?

src/libtiled/templatemanager.h
```diff
@@ -0,0 +1,60 @@
...
+#include "templategroup.h"
+
+namespace Tiled {
+
+class TILEDSHARED_EXPORT TemplateManager : public QObject
+{
+    Q_OBJECT
+
+public:
+    static TemplateManager *instance();
```
bjorn: It seems like nobody deletes the template manager. Not that it matters much, but it would be nicer to delete it on program exit.

src/tiled/brokenlinks.cpp
```diff
@@ -161,8 +184,19 @@ void BrokenLinksModel::refresh()
...
+                    BrokenLink link;
+                    link.type = TemplateGroupReference;
+                    link._templateGroup = templateGroup;
+                    mBrokenLinks.append(link);
+                }
+            }
         } else if (auto tilesetDocument = qobject_cast<TilesetDocument*>(mDocument)) {
             processTileset(tilesetDocument->tileset());
+        } else if (auto templateGroupDocument = qobject_cast<TemplateGroupDocument*>(mDocument)) {
+            processTemplateGroup(templateGroupDocument->templateGroup());
```
bjorn: I think you can just leave out this part as well as `processTemplateGroup`. The `mDocument` will never be a `TemplateGroupDocument`, since currently template group documents can't be opened as a stand-alone file. And even when it could, checking the file itself for being loaded makes no sense. The only thing that would make sense then, is to check whether its tileset references and any images referenced by those tilesets have been loaded.

It would be fine to make some preparations, but I think it's a little early. We will get to that point when we actually implement a TemplateGroupEditor.

src/tiled/brokenlinks.cpp
```diff
@@ -563,6 +612,38 @@ void BrokenLinksWidget::tryFixLink(const BrokenLink &link)
...
+                return;
+            }
+
+            ObjectTemplateModel::instance()->addTemplateGroup(templateGroup);
+        }
+
+        MapDocument *mapDocument = static_cast<MapDocument*>(document);
+        int index = mapDocument->map()->templateGroups().indexOf(link._templateGroup);
+        if (index != -1)
+            mapDocument->replaceTemplateGroup(index, templateGroup);
```
bjorn: This should be done with an undo command, like `ReplaceTileset`.

src/tiled/flipmapobjects.cpp
```diff
@@ -64,15 +64,27 @@ FlipMapObjects::FlipMapObjects(MapDocument *mapDocument,
...
 
 void FlipMapObjects::flip()
 {
     //flip objects
-    for (MapObject *object : mMapObjects)
-        object->flip(mFlipDirection, mObjectsCenter);
+    for (int i = 0; i < mMapObjects.size(); ++i) {
+        mMapObjects[i]->flip(mFlipDirection, mObjectsCenter);
+
+        mMapObjects[i]->setPropertyChanged(MapObject::CellProperty, mNewCellStates[i]);
```
bjorn: I'm a little concerned about this. I think flipping some tile object is not actually expected to override the referenced tile, but just its flipped state. I think it's not very important to deal with this now, but it's something to keep in mind.

src/tiled/mainwindow.cpp
```diff
@@ -608,7 +609,37 @@ bool MainWindow::openFile(const QString &fileName, FileFormat *fileFormat)
     Document *document = nullptr;
 
     if (MapFormat *mapFormat = qobject_cast<MapFormat*>(fileFormat)) {
-        document = MapDocument::load(fileName, mapFormat, &error);
+        MapDocument *mapDocument =  MapDocument::load(fileName, mapFormat, &error);
+        document = mapDocument;
+
+        bool hasNonEmbeddedTemplateGroups;
```
bjorn: Variable is missing an initialization. Not that it matters, because the variable isn't really used. Remove?

src/tiled/mapdocument.cpp
```diff
@@ -670,6 +672,17 @@ SharedTileset MapDocument::replaceTileset(int index, const SharedTileset &tilese
     return oldTileset;
 }
 
+void MapDocument::replaceTemplateGroup(int index, TemplateGroup *templateGroup)
+{
+    TemplateGroup *oldTemplateGroup = mMap->templateGroups().at(index);
+    auto changedObjects = mMap->replaceTemplateGroup(oldTemplateGroup, templateGroup);
+
+    // Update the objects in the map scene
+    emit(objectsChanged(changedObjects));
```
bjorn: Coding style: please use the empty `emit` macro as follows:
```c++
emit objectsChanged(changedObjects);
```

src/tiled/templategroupdocument.cpp
```diff
@@ -45,7 +45,7 @@ TemplateGroupDocument::TemplateGroupDocument(TemplateGroup *templateGroup, const
     : Document(TemplateGroupDocumentType, fileName)
```
bjorn: Should this `fileName` also be changed to `mTemplateGroup->fileName()`?

src/libtiled/mapobject.h
```diff
@@ -99,16 +104,22 @@ class TILEDSHARED_EXPORT MapObject : public Object
...
+        VisibleProperty          = 1 << 2,
+        TextProperty             = 1 << 3,
+        TextFontProperty         = 1 << 4,
+        TextAlignmentProperty    = 1 << 5,
+        TextWordWrapProperty     = 1 << 6,
+        TextColorProperty        = 1 << 7,
+        CellProperty             = 1 << 8,
+        SizeProperty             = 1 << 9,
+        ShapeProperty            = 1 << 10,
+        RotationProperty         = 1 << 11
```
bjorn: Since you've added several entries here, I'm getting warnings in `MapObject::setMapObjectProperty` and `MapObject::mapObjectProperty`. I think size and rotation could simply be handled, but "shape" and "cell" are too special and it should probably catch those with a `Q_ASSERT(false);`.

src/libtiled/templategroup.cpp
thabetx: Probably, at some point, unused tilesets in a template group were not registered, so saving a template that uses an unused tileset would think that the tileset was new, so the tileset was added again, but currently this case causes no problems, so the todo should be removed.

src/tiled/templategroupdocument.cpp
thabetx: I think so. The template group should always contain the file name, so the constructor arguments can be the template group only.

Issue #734, opened 2014-08-29
Ability to lock layers


xerosugar:
I think this should be added to prevent the user from working on the wrong layer.

ncareau:
+1, currently our game has multiple layer and i throw an error when wrong tile are inserted into the special layer. Being able to lock the layer would be nice to prevent any unwanted tile inserted in that layer.

superqix:
I don't usually +1 something, but now that we have scaleable objects, this feature is getting more important. We use lots of gradients on the topmost and bottommost layers that make it hard to nudge other objects in a map.

bjorn:
Alright, I've added it to the [roadmap](https://trello.com/b/yl3PAtN0/tiled-roadmap). You can vote on it there. I'm still trying to find out how to prioritize features against each other. I definitely agree this would be a good to have feature, no question about it.

Sometimes I tell people they can support me on [Patreon](http://www.patreon.com/bjorn) to help me get around to more features sooner, but I see you already are @superqix, so thanks a lot for your support! And to @kheftel as well of course!

superqix:
Thanks for adding to the roadmap. I've been super impressed with updates since you went with Patreon funding, so keep up the good work.

I'm working toward releasing a new Lua loader for CoronaSDK/Love2D so hopefully that will open you up to a new group of Patreons.

bjorn:
That's great to hear, thanks!

As for growing the funding, this seems really hard from the current level. The pace rather stopped once the first goal was reached, but I'd still like to reach the second goal as well. I've opened up issue #1064 for this. :-)

EdenIndustries:
I've actually already implemented layer locking in our local version of Tiled. Do you want me to try a pull request for it? Or did you have a plan in mind for it already?

bjorn:
@EdenIndustries This is one of those features I'd like to get around to sooner rather than later, but I did not do any work on it yet since there are many other things I'd like to get done as well. So feel free to open a pull request for it, I'm looking forward to see your changes!

EdenIndustries:
@bjorn Ok I'll see what I can do! To my embarassment, I admit Git's notion of branches and pulling and forking and such is not something I'm all that familiar with yet (I've only really used SVN and Perforce), but I'll try to figure it out and get the pull request sorted out :)

bjorn:
@EdenIndustries Good luck and feel free to ask questions! No need to be embarrassed!

JesseTG:
[...]

PR #1627, opened 2017-06-26
Adds option to lock/unlock layer

ketanhwr:
Iterative PR for #734.


src/tiled/createobjecttool.cpp
```diff
@@ -195,7 +195,7 @@ void CreateObjectTool::finishNewMapObject()
     Q_ASSERT(mNewMapObjectItem);
 
     ObjectGroup *objectGroup = currentObjectGroup();
-    if (!objectGroup) {
+    if (!objectGroup || !objectGroup->isUnlocked()) {
```
bjorn: The check is too late at this point. It should not be possible to start the creation of a new object.

src/tiled/layeroffsettool.cpp
```diff
@@ -153,7 +153,8 @@ void LayerOffsetTool::finishDrag()
     if (!mapDocument())
         return;
 
-    if (Layer *layer = mapDocument()->currentLayer()) {
+    Layer *layer = mapDocument()->currentLayer();
+    if (layer && layer->isUnlocked()) {
```
bjorn: This check is too late (as is the check in `mouseMoved`). Rather, `startDrag` should refuse to start dragging when the layer is locked.

src/tiled/mapdocument.cpp
```diff
@@ -586,6 +586,8 @@ void MapDocument::moveLayerDown(Layer *layer)
 void MapDocument::removeLayer(Layer *layer)
 {
     Q_ASSERT(layer->map() == mMap);
+    if (!layer->isUnlocked())
+        return;
```
bjorn: You should be able to remove locked layers. Also this check would be in the wrong place since this function is not expected to fail.

ketanhwr: Locked layer can be removed? I doubt this..

src/tiled/layeroffsettool.cpp
```diff
@@ -136,7 +136,8 @@ void LayerOffsetTool::startDrag(const QPointF &pos)
...
```
bjorn: I noticed that this change is not enough, since after calling `startDrag`, the `mouseMoved` function continues to change the layer position even if `mDragging` is not set to `true`. So I would suggest to remove the `return` in the `else` in that function and instead check `mDragging` in addition to `currentLayer` before moving the layer.

PR #1628, opened 2017-06-26
Introduced TilesetDocumentsModel and its sort-filter model companion

bjorn:
This model lists the tileset documents that are currently open, and the sort-filter version sorts them by name and filters out tilesets that are embedded in other maps than the current one.

This model extracts part of the logic from TilesetDock, so that it could be reused by an updated TerrainModel. The TerrainModel currently only lists terrains from tilesets that are already part of the map, but it should display all loaded external tilesets.

This is for addressing the following issue raised on the forum:
http://discourse.mapeditor.org/t/created-a-terrain-but-its-not-showing-up-in-the-terrain-tab/2478

### DIFF
--- a/src/libtiled/tileset.h
+++ b/src/libtiled/tileset.h
@@ -620,3 +620,5 @@ inline bool Tileset::imageLoaded() const
 }
 
 } // namespace Tiled
+
+Q_DECLARE_METATYPE(Tiled::SharedTileset)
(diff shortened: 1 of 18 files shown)


bjorn:
@ketanhwr, @thabetx: I'm doing this rather big change on the 1.0 branch to fix up something I had forgotten, namely that the Terrains view should display terrains from all loaded tilesets and not just those that are already part of the map, in similar fashion to the Tilesets view. This commit is the first step in fixing that, and a change to the `TerrainModel` will follow.

If you find some time, it would be great if you could help with reviewing and/or testing.

Issue #1626, opened 2017-06-26
IndexOutOfBoundsException when loading tmx maps


Shorttail:
Inside the test resources for libtiled-java, create a new folder named "other" and copy desert.tmx and desert.tsx into it. Add the following test case:

```java
@Test
public void testErrorReadingMap() throws Exception {
    new TMXMapReader().readMap("other/desert.tmx");
}
```

The test should fail with an IndexOutOfBoundsException similar to this:

```
java.lang.IndexOutOfBoundsException: Index: 16, Size: 1
	at java.util.ArrayList.rangeCheck(ArrayList.java:653)
	at java.util.ArrayList.get(ArrayList.java:429)
	at tiled.core.TileSet.getTile(TileSet.java:419)
	at tiled.io.TMXMapReader.getTileForTileGID(TMXMapReader.java:727)
	at tiled.io.TMXMapReader.setTileAtFromTileId(TMXMapReader.java:713)
	at tiled.io.TMXMapReader.readLayer(TMXMapReader.java:626)
	at tiled.io.TMXMapReader.buildMap(TMXMapReader.java:812)
	at tiled.io.TMXMapReader.unmarshal(TMXMapReader.java:846)
	at tiled.io.TMXMapReader.readMap(TMXMapReader.java:873)
```

When the tmx and tsx files are in their original position, which is the same as the package for TMXMapReader, this exception is now thrown.

bjorn:
Thanks @Shorttail!

Actually after looking into the issue for a while, it occurred to me that you mention only copying the `desert.tmx` and `desert.tsx` files. However, the `desert.tsx` file refers to `tmw_desert_spacing.png`. If that image file isn't present the map won't load. However, it isn't expected to throw the above exception in that case, it should exit as follows:
```
Error while reading the map:
Can't read input file!
```
At least, `TMXViewer` does. When I do also copy the image file, things seem to work fine.

That `getTileForTileGID` is catching the wrong exception though. It's trying to catch `ArrayIndexOutOfBoundsException` but the exception raised is `IndexOutOfBoundsException`, since `tilesets` was changed from a `Vector` to an `ArrayList`. However, it shouldn't run into either exception in this case.

Shorttail:
Good catch. I copied tmx_desert_spacing.png as well, and now it fails with this error instead:

```
java.io.FileNotFoundException: some\path\tiled\util\java\libtiled-java\other\desert.tmx (The system cannot find the path specified)
[...]

PR #1629, opened 2017-06-26
Fixed wrong exception being caught in TileSet.

Shorttail:
Issue #1626 happens because an incorrect exception is caught in TileSet#getTile(). This makes it catch the correct exception.


bjorn:
Thanks. :-)

PR #1630, opened 2017-06-26
Preparations for loading files from remote locations

bjorn:
Initially, the idea is just to support loading of images via http/https. Later it could be extended to support also loading tilesets from remote locations and finally maybe even the maps.

These changes introduce a new `LoadingStatus` enum to replace some `isLoaded` booleans already used, and change most image references from `QString` to `QUrl` (not including `ImageLayer` yet). It's a work-in-progress, especially since I think it breaks something about these image references in the Properties view (at the very least the file type filter, maybe more).


bjorn:
@thabetx, @Bdtrotte: Just putting this here so that you are aware of this change, which may soon be merged into `master`, possibly before the actual loading of remote images is finished to avoid it leading to too many conflicts later.

src/plugins/python/python.qbs
```diff
@@ -51,5 +51,7 @@ TiledPlugin {
         "pythonplugin.cpp",
         "pythonplugin.h",
         "pythonbind.cpp",
+        "tiledbinding.py",
+        "qtbinding.py",
```
Ablu: I think the order should be switched here, right?

bjorn: Hmm, I guess it should. Probably added these manually and forgot about that...

PR #1633, opened 2017-06-28
Update tiled_pt.ts

olueiro:
Lua is not an acronym. Fix shortcut for 'Copy current position of mouse cursor to clipboard' (Alt+C)


bjorn:
Thanks! I've cherry-picked it to the `1.0` branch as change f88c9f5b960a41b3e56846c78365d4d14c94a258 so that it will be included in Tiled 1.0.3 eventually.

Issue #260, opened 2012-08-12
Automatically resized maps


Asmageddon:
In other words, maps that are automatically resized when you paint outside of map boundaries.

ketanhwr:
What exactly would be the point of this? If by mistake someone paints a tile outside the boundary, then the map would automatically be resized even if the user does not want so. And while drawing on the boundary, this is something that would frequently happen.

bjorn:
> What exactly would be the point of this? If by mistake someone paints a tile outside the boundary, then the map would automatically be resized even if the user does not want so. And while drawing on the boundary, this is something that would frequently happen.

When somebody would enable automatic resizing of maps, then he would probably not care at all about "painting outside the boundary". The whole point after all is that you don't want to think about such a boundary while mapping, and just have the idea of an infinite canvas. Check out [MyPaint](http://mypaint.org/about/) for an example of a painting application that has no boundaries on its canvas.

ketanhwr:
Oh, I'll try to work on this one now. Should I add an option to enable/disable it in preferences?

bjorn:
@ketanhwr Hmm, I think you're vastly underestimating the complexity of this. It will need efficient data structures (probably allocating tile layer memory in blocks), and the automatic expansion should work per tile layer and not on the map. It needs to be thought out well before you start. I think it's a project of a month at least.

Just saying this because simply an automatic invocation of the "resize map" action is unlikely to cut it. It would quickly consume a huge amount of memory and probably there would be noticeable performance issues when you go over the "invisible edge" while painting.

ketanhwr:
Oh. I didn't really think of it that way, thanks for your feedback! :-)

bjorn:
> I do chunks and infinite maps in all tile-based code I do with almost no exceptions, you're gonna be fine with just allocating 16x16 chunks and storing/indexing them from a hashmap.

Maybe it's something @ketanhwr could consider adding to his Google Summer of Code proposal. While the data changes may not be so complicated, it may affect large amounts of existing code to actually fully support an infinite map.

ketanhwr:
@bjorn, yes, thanks!

ketanhwr:
> .. blocks), and the automatic expansion should work per tile layer and not on the map.

@bjor, what exactly do you mean by this? If we were to resize the layer, we will have to resize the map too nevertheless.

ketanhwr:
Have a look at this: https://github.com/Squarific/TiledCanvas

Is this what we're looking for?

[...]

PR #1635, opened 2017-06-30
Supporting Infinite Maps

ketanhwr:
In progress PR for #260.

Currently I've added a class `Block` which is a 16x16 grid of cells. This is used in `TileLayer` and now, TileLayer no longes needs `mGrid`, and only stores instances of `Block`. The blocks are stored in a `QMap` and these blocks get deleted when they're empty.

### DIFF
--- a/src/libtiled/tilelayer.cpp
+++ b/src/libtiled/tilelayer.cpp
@@ -148,11 +148,11 @@ QRegion TileLayer::region(std::function<bool (const Cell &)> condition) const
 {
     QRegion region;
 
-    QHashIterator<QPoint, Chunk* > it(mChunks);
+    QHashIterator<QPoint, Chunk> it(mChunks);
     while (it.hasNext()) {
         it.next();
-        region += it.value()->region(condition).translated(it.key().x() * CHUNK_SIZE + mX,
-                                                           it.key().y() * CHUNK_SIZE + mY);
+        region += it.value().region(condition).translated(it.key().x() * CHUNK_SIZE + mX,
+                                                          it.key().y() * CHUNK_SIZE + mY);
     }
 
     return region;
@@ -165,13 +165,10 @@ void Tiled::TileLayer::setCell(int x, int y, const Cell &cell)
 {
     Q_ASSERT(contains(x, y));
 
-    Chunk *_chunk = chunk(x, y);
-
-    Cell existingCell;
-    existingCell = _chunk->cellAt(x & CHUNK_MASK, y & CHUNK_MASK);
+    Chunk &_chunk = chunk(x, y);
 
     if (!mUsedTilesetsDirty) {
-        Tileset *oldTileset = existingCell.isEmpty() ? nullptr : existingCell.tileset();
+        Tileset *oldTileset = _chunk.cellAt(x & CHUNK_MASK, y & CHUNK_MASK).tileset();
         Tileset *newTileset = cell.tileset();
         if (oldTileset != newTileset) {
             if (oldTileset)
@@ -181,10 +178,10 @@ void Tiled::TileLayer::setCell(int x, int y, const Cell &cell)
         }
     }
 
-    _chunk->setCell(x & CHUNK_MASK, y & CHUNK_MASK, cell);
+    _chunk.setCell(x & CHUNK_MASK, y & CHUNK_MASK, cell);
 
-    if (_chunk->isEmpty())
-        delete mChunks.take(chunkCoordinates(x, y));
+    if (_chunk.isEmpty())
+        mChunks.remove(chunkCoordinates(x, y));
 }
 
 TileLayer *TileLayer::copy(const QRegion &region) const
@@ -275,7 +272,7 @@ void TileLayer::flip(FlipDirection direction)
 
     Q_ASSERT(direction == FlipHorizontally || direction == FlipVertically);
 
-    QHashIterator<QPoint, Chunk* > it(mChunks);
+    QHashIterator<QPoint, Chunk> it(mChunks);
     while (it.hasNext()) {
         it.next();
         for (int y = 0; y < CHUNK_SIZE; ++y) {
@@ -286,7 +283,7 @@ void TileLayer::flip(FlipDirection direction)
                 if (!contains(_x, _y))
                     continue;
 
-                const Cell &source = it.value()->cellAt(x, y);
+                const Cell &source = it.value().cellAt(x, y);
                 Cell dest(source);
 
                 if (direction == FlipHorizontally) {
@@ -316,7 +313,7 @@ void TileLayer::flipHexagonal(FlipDirection direction)
 
     const char (&flipMask)[16] = (direction == FlipHorizontally ? flipMaskH : flipMaskV);
 
-    QHashIterator<QPoint, Chunk* > it(mChunks);
+    QHashIterator<QPoint, Chunk> it(mChunks);
     while (it.hasNext()) {
         it.next();
         for (int y = 0; y < CHUNK_SIZE; ++y) {
@@ -327,7 +324,7 @@ void TileLayer::flipHexagonal(FlipDirection direction)
                 if (!contains(_x, _y))
                     continue;
 
-                Cell dest(it.value()->cellAt(x, y));
+                Cell dest(it.value().cellAt(x, y));
 
                 unsigned char mask =
                         (static_cast<unsigned char>(dest.flippedHorizontally()) << 3) |
@@ -364,7 +361,7 @@ void TileLayer::rotate(RotateDirection direction)
     int newHeight = mWidth;
     TileLayer *newLayer = new TileLayer(QLatin1String(""), 0, 0, newWidth, newHeight);
 
-    QHashIterator<QPoint, Chunk* > it(mChunks);
+    QHashIterator<QPoint, Chunk> it(mChunks);
     while (it.hasNext()) {
         it.next();
         for (int y = 0; y < CHUNK_SIZE; ++y) {
@@ -375,7 +372,7 @@ void TileLayer::rotate(RotateDirection direction)
                 if (!contains(_x, _y))
                     continue;
 
-                Cell dest(it.value()->cellAt(x, y));
+                Cell dest(it.value().cellAt(x, y));
 
                 unsigned char mask =
                         (dest.flippedHorizontally() << 2) |
@@ -447,7 +444,7 @@ void TileLayer::rotateHexagonal(RotateDirection direction, Map *map)
     const char (&rotateMask)[16] =
             (direction == RotateRight) ? rotateRightMask : rotateLeftMask;
 
-    QHashIterator<QPoint, Chunk* > it(mChunks);
+    QHashIterator<QPoint, Chunk> it(mChunks);
     while (it.hasNext()) {
         it.next();
         for (int y = 0; y < CHUNK_SIZE; ++y) {
@@ -458,7 +455,7 @@ void TileLayer::rotateHexagonal(RotateDirection direction, Map *map)
                 if (!contains(_x, _y))
                     continue;
 
-                Cell dest(it.value()->cellAt(x, y));
+                Cell dest(it.value().cellAt(x, y));
 
                 unsigned char mask =
                         (static_cast<unsigned char>(dest.flippedHorizontally()) << 3) |
@@ -509,8 +506,8 @@ QSet<SharedTileset> TileLayer::usedTilesets() const
     if (mUsedTilesetsDirty) {
         QSet<SharedTileset> tilesets;
 
-        for (Chunk *chunk : mChunks) {
-            for (const Cell &cell : *chunk)
+        for (const Chunk &chunk : mChunks) {
+            for (const Cell &cell : chunk)
                 if (const Tile *tile = cell.tile())
                     tilesets.insert(tile->sharedTileset());
 
@@ -524,8 +521,8 @@ QSet<SharedTileset> TileLayer::usedTilesets() const
 
 bool TileLayer::hasCell(std::function<bool (const Cell &)> condition) const
 {
-    for (Chunk *chunk : mChunks) {
-        if (chunk->hasCell(condition))
+    for (const Chunk &chunk : mChunks) {
+        if (chunk.hasCell(condition))
             return true;
     }
 
@@ -539,9 +536,8 @@ bool TileLayer::referencesTileset(const Tileset *tileset) const
 
 void TileLayer::removeReferencesToTileset(Tileset *tileset)
 {
-    for (Chunk *chunk : mChunks) {
-        chunk->removeReferencesToTileset(tileset);
-    }
+    for (Chunk &chunk : mChunks)
+        chunk.removeReferencesToTileset(tileset);
 
     mUsedTilesets.remove(tileset->sharedPointer());
 }
@@ -549,9 +545,8 @@ void TileLayer::removeReferencesToTileset(Tileset *tileset)
 void TileLayer::replaceReferencesToTileset(Tileset *oldTileset,
                                            Tileset *newTileset)
 {
-    for (Chunk *chunk : mChunks) {
-        chunk->replaceReferencesToTileset(oldTileset, newTileset);
-    }
+    for (Chunk &chunk : mChunks)
+        chunk.replaceReferencesToTileset(oldTileset, newTileset);
 
     if (mUsedTilesets.remove(oldTileset->sharedPointer()))
         mUsedTilesets.insert(newTileset->sharedPointer());
@@ -594,7 +589,7 @@ void TileLayer::offsetTiles(const QPoint &offset,
 {
     TileLayer *newLayer = clone();
 
-    QHashIterator<QPoint, Chunk* > it(newLayer->mChunks);
+    QHashIterator<QPoint, Chunk> it(newLayer->mChunks);
     while (it.hasNext()) {
         it.next();
         for (int y = 0; y < CHUNK_SIZE; ++y) {
@@ -691,10 +686,9 @@ QRegion TileLayer::computeDiffRegion(const TileLayer *other) const
 
 bool TileLayer::isEmpty() const
 {
-    for (Chunk *chunk : mChunks) {
-        if (!chunk->isEmpty())
+    for (const Chunk &chunk : mChunks)
+        if (!chunk.isEmpty())
             return false;
-    }
 
     return true;
 }
@@ -712,11 +706,7 @@ TileLayer *TileLayer::clone() const
 TileLayer *TileLayer::initializeClone(TileLayer *clone) const
 {
     Layer::initializeClone(clone);
-    QHashIterator<QPoint, Chunk* > it(mChunks);
-    while (it.hasNext()) {
-        it.next();
-        clone->mChunks.insert(it.key(), new Chunk(*it.value()));
-    }
+    clone->mChunks = mChunks;
     clone->mUsedTilesets = mUsedTilesets;
     clone->mUsedTilesetsDirty = mUsedTilesetsDirty;
     return clone;
--- a/src/libtiled/tilelayer.cpp
+++ b/src/libtiled/tilelayer.cpp
@@ -34,11 +34,71 @@
 
 using namespace Tiled;
 
+QRegion Block::region(std::function<bool (const Cell &)> condition) const
+{
+    QRegion region;
+
+    for (int y = 0; y < 16; ++y) {
+        for (int x = 0; x < 16; ++x) {
+            if (condition(cellAt(x, y))) {
+                const int rangeStart = x;
+                for (++x; x <= 16; ++x) {
+                    if (x == 16 || !condition(cellAt(x, y))) {
+                        const int rangeEnd = x;
+                        region += QRect(rangeStart, y, rangeEnd - rangeStart, 1);
+                        break;
+                    }
+                }
+            }
+        }
+    }
+
+    return region;
+}
+
+void Block::setCell(int x, int y, const Cell &cell)
+{
+    int index = x + y * 16;
+
+    if (mGrid[index].isEmpty() && !cell.isEmpty())
+        mCells++;
+    else if (!mGrid[index].isEmpty() && cell.isEmpty())
+        mCells--;
+
+    mGrid[index] = cell;
+}
+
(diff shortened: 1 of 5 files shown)


src/libtiled/tilelayer.cpp
```diff
@@ -114,7 +165,12 @@ void Tiled::TileLayer::setCell(int x, int y, const Cell &cell)
 {
     Q_ASSERT(contains(x, y));
 
-    Cell &existingCell = mGrid[x + y * mWidth];
+    Cell existingCell = Cell();
+
+    if (mMap.contains(block(x, y)))
+        existingCell = mMap[block(x, y)]->cellAt(x%16, y%16);
```
Ablu: `x % 16` <<--- there should be spaces between the operators

src/libtiled/tilelayer.cpp
```diff
@@ -390,7 +463,11 @@ void TileLayer::rotateHexagonal(RotateDirection direction, Map *map)
 
     mWidth = newWidth;
     mHeight = newHeight;
-    mGrid = newGrid;
+    for (int y = 0; y < mHeight; ++y) {
+        for (int x = 0; x < mWidth; ++x) {
+            setCell(x, y, newGrid[x + y * mWidth]);
+        }
+    }
```
Ablu: 4th copy of this code snipped now... Maybe time for a helper function?

src/libtiled/tilelayer.cpp
```diff
@@ -127,7 +183,12 @@ void Tiled::TileLayer::setCell(int x, int y, const Cell &cell)
         }
     }
 
-    existingCell = cell;
+    mMap[block(x, y)]->setCell(x%16, y%16, cell);
```
Ablu: spaces around operators again

src/libtiled/tilelayer.cpp
```diff
@@ -574,9 +672,11 @@ QRegion TileLayer::computeDiffRegion(const TileLayer *other) const
 
 bool TileLayer::isEmpty() const
 {
-    for (const Cell &cell : mGrid)
-        if (!cell.isEmpty())
+    QMapIterator< QPair<int, int>, Block* > it(mMap);
+    while (it.hasNext()) {
+        if (!it.value()->isEmpty())
             return false;
+    }
```
Ablu: there seems to be missing a call to `it.next()` here

src/libtiled/tilelayer.cpp
```diff
@@ -594,7 +694,9 @@ TileLayer *TileLayer::clone() const
 TileLayer *TileLayer::initializeClone(TileLayer *clone) const
 {
     Layer::initializeClone(clone);
-    clone->mGrid = mGrid;
+    for (int x = 0; x < mWidth; ++x)
+        for (int y = 0; y < mHeight; ++y)
```
Ablu: In other cases you first iterate over the `y`, here you first iterate over `x`. Generally first iterating over `y` should be faster due to the access being `x + y * 16` regarding cache hits.

src/libtiled/tilelayer.h
```diff
@@ -371,6 +421,11 @@ inline bool TileLayer::contains(const QPoint &point) const
     return contains(point.x(), point.y());
 }
 
+inline QPair<int, int> TileLayer::block(int x, int y) const
```
Ablu: I find the name of this function not really describing... it sounds like it `blocks` something... But it actually calculates the block coordinates...

src/libtiled/tilelayer.h
```diff
@@ -331,19 +386,14 @@ class TILEDSHARED_EXPORT TileLayer : public Layer
...
-
 protected:
     TileLayer *initializeClone(TileLayer *clone) const;
 
 private:
     int mWidth;
     int mHeight;
-    QVector<Cell> mGrid;
+    Cell mEmptyCell;
+    QMap< QPair<int, int>, Block* > mMap;
```
Ablu: I do not really like this `QPair` here... Why not QPoint? Or introduce an own struct like `BlockCoordinate`?

src/tiled/bucketfilltool.cpp
```diff
@@ -369,9 +369,13 @@ void BucketFillTool::updateRandomListAndMissingTilesets()
         mapDocument()->unifyTilesets(variation.map, mMissingTilesets);
 
         if (mIsRandom) {
-            for (const Cell &cell : *variation.tileLayer()) {
-                if (const Tile *tile = cell.tile())
-                    mRandomCellPicker.add(cell, tile->probability());
+            const TileLayer &tileLayer = *variation.tileLayer();
+            for (int x = 0; x < tileLayer.width(); ++x) {
```
Ablu: iteration order is wrong here again.

src/tiled/bucketfilltool.cpp
```diff
@@ -369,9 +369,13 @@ void BucketFillTool::updateRandomListAndMissingTilesets()
...
+                for (int y = 0; y < tileLayer.height(); ++y) {
+                    const Cell cell = tileLayer.cellAt(x, y);
```
Ablu: `&cell =`, even if it probably does not matter

src/tiled/tilesetdock.cpp
```diff
@@ -315,24 +315,27 @@ void TilesetDock::selectTilesInStamp(const TileStamp &stamp)
...
-                if (tilesetIndex != -1) {
-                    TilesetView *view = tilesetViewAt(tilesetIndex);
-                    if (!view->model()) // Lazily set up the model
-                        setupTilesetModel(view, tileset);
-
-                    const TilesetModel *model = view->tilesetModel();
-                    const QModelIndex modelIndex = model->tileIndex(tile);
-                    QItemSelectionModel *selectionModel = view->selectionModel();
-                    selections[selectionModel].select(modelIndex, modelIndex);
+        for (int x = 0; x < tileLayer.width(); ++x) {
```
Ablu: iteration order again.

ketanhwr:
Can you suggest some other name other than `block(x, y)`? I couldn't think of one. And regarding `QPair`, I personally think that there won't be any point of adding a different data type there since it already completes all the task easily.

src/libtiled/tilelayer.cpp
```diff
@@ -34,11 +34,71 @@
...
+}
+
+void Block::setCell(int x, int y, const Cell &cell)
+{
+    int index = x + y * CHUNK_SIZE;
+
+    if (mGrid[index].isEmpty() && !cell.isEmpty())
+        mCells++;
+    else if (!mGrid[index].isEmpty() && cell.isEmpty())
+        mCells--;
```
ketanhwr: Awesome!

src/libtiled/tilelayer.cpp
```diff
@@ -88,20 +148,11 @@ QRegion TileLayer::region(std::function<bool (const Cell &)> condition) const
...
-                    if (x == mWidth || !condition(cellAt(x, y))) {
-                        const int rangeEnd = x;
-                        region += QRect(rangeStart + mX, y + mY,
-                                        rangeEnd - rangeStart, 1);
-                        break;
-                    }
-                }
-            }
-        }
+    QMapIterator< QPair<int, int>, Block* > it(mMap);
```
ketanhwr: In a map, the key should of such a data type such that it can be easily compared. QPair has operators for `<` and `>` while QPoint has not overloaded these operators. If no comparison between two values can be made, then what's the point of inserting a value in a red black tree? It gives compilation error anyway.

src/libtiled/tilelayer.h
```diff
@@ -331,19 +386,16 @@ class TILEDSHARED_EXPORT TileLayer : public Layer
...
 
 protected:
     TileLayer *initializeClone(TileLayer *clone) const;
 
 private:
     int mWidth;
     int mHeight;
-    QVector<Cell> mGrid;
+    Cell mEmptyCell;
+    QMap< QPair<int, int>, Block* > mMap;
```
ketanhwr: >  We should really still provide a way to iterate over the cells of a tile layer, especially now that it is allocated in chunks. Iterating only the chunks can mean a significant performance boost when iterating layers with large empty areas.

A `QMap` would help in this case I think.

src/libtiled/tilelayer.h
```diff
@@ -191,6 +244,8 @@ class TILEDSHARED_EXPORT TileLayer : public Layer
     bool contains(int x, int y) const;
     bool contains(const QPoint &point) const;
 
+    QPair<int, int> block(int x, int y) const;
```
ketanhwr: I guess, floor() will solve the problem for negative values?

ketanhwr:
> At the moment this patch leaves many operations that iterate the entire area and that will even allocate the entire layer as huge array again, like the resize and offset operations. In each case we should consider ways to perform those operations more efficiently.

The simplest solution I can think of is just like the previous one. Creating a `newChunks` instead of `newGrid` and then iterating over the chunks and performing the operations, and in the end copy and `newChunks` to `mChunks`.

src/libtiled/tilelayer.cpp
bjorn: That's a good point, however we don't need this container to be sorted and if you use a `QHash` then `QPoint` will work fine.

src/libtiled/tilelayer.h
bjorn: I don't see how it matters whether `QMap` or `QHash` is used. Both can be iterated just fine.

ketanhwr: No I mean, when we'll iterate over the chunks, I think we'll prefer to iterate over the chunks in sorted order. Using `QHash`, we cannot iterate in a sorted order.

src/libtiled/tilelayer.h
bjorn: Yes, though that would require doing the division with doubles and converting back to integer. Alternatively you could special-case negative values:
```c++
x < 0 ? (x + 1) / CHUNK_SIZE - 1 : x / CHUNK_SIZE
```
I'm not sure what's best performance-wise, but I would guess this one to be faster.

src/libtiled/tilelayer.h
bjorn: I can't think of any use-case where we would care about the iteration order over the chunks.

ketanhwr: Well, `QPoint` doesn't have `qHash` defined. So should I go with `QHash< QPair<int, int>, Chunk* >` only?

bjorn: Based on the `qHash` implementation for `QPair`, we can derive the following implementation for `QPoint`:
```c++
inline uint qHash(const QPoint &key, uint seed = 0) Q_DECL_NOTHROW
{
    uint h1 = qHash(key.x(), seed);
    uint h2 = qHash(key.y(), seed);
    return ((h1 << 16) | (h1 >> 16)) ^ h2 ^ seed;
}
```

src/libtiled/tilelayer.h
```diff
@@ -371,6 +425,24 @@ inline bool TileLayer::contains(const QPoint &point) const
...
+        return mChunks[chunkCoordinates] = new Chunk();
+}
+
+inline Chunk* TileLayer::findChunk(int x, int y) const
+{
+    QPair<int, int> chunkCoordinates(x / CHUNK_SIZE, y / CHUNK_SIZE);
+    if (mChunks.contains(chunkCoordinates))
+        return mChunks[chunkCoordinates];
+    else
+        return nullptr;
```
bjorn: The above four lines can be written as:
```c++
return mChunks.value(chunkCoordinates);
```

src/libtiled/tilelayer.h
```diff
@@ -371,6 +425,24 @@ inline bool TileLayer::contains(const QPoint &point) const
     return contains(point.x(), point.y());
 }
 
+inline Chunk* TileLayer::chunk(int x, int y)
+{
+    QPair<int, int> chunkCoordinates(x / CHUNK_SIZE, y / CHUNK_SIZE);
+    if (mChunks.contains(chunkCoordinates))
+        return mChunks[chunkCoordinates];
```
bjorn: This does a look-up twice, which we can avoid:
```c++
if (Chunk *chunk = mChunks.value(chunkCoordinates))
    return chunk;
return mChunks[chunkCoordinates] = new Chunk;
```

src/libtiled/tilelayer.cpp
```diff
@@ -114,7 +164,9 @@ void Tiled::TileLayer::setCell(int x, int y, const Cell &cell)
 {
     Q_ASSERT(contains(x, y));
 
-    Cell &existingCell = mGrid[x + y * mWidth];
+    Cell existingCell;
+
+    existingCell = chunk(x, y)->cellAt(x % CHUNK_SIZE, y % CHUNK_SIZE);
```
bjorn: This `x % CHUNK_SIZE` won't work for negative coordinates. Instead, with a chunk size of 16, we could do `x & 0xF`. We could call this `x & CHUNK_MASK` and define:
```c++
const int CHUNK_MASK = CHUNK_SIZE - 1;
```
Of course, this relies on the chunk size always being a power of 2.

src/libtiled/tilelayer.cpp
```diff
@@ -594,8 +659,27 @@ TileLayer *TileLayer::clone() const
 TileLayer *TileLayer::initializeClone(TileLayer *clone) const
 {
     Layer::initializeClone(clone);
-    clone->mGrid = mGrid;
+    QHashIterator<QPoint, Chunk* > it(mChunks);
+    while (it.hasNext()) {
+        it.next();
+        for (int x = 0; x < CHUNK_SIZE; ++x) {
+            for (int y = 0;y < CHUNK_SIZE; ++y) {
```
Ablu: here a space is missing in front of `y`

src/libtiled/tilelayer.cpp
```diff
@@ -594,8 +659,31 @@ TileLayer *TileLayer::clone() const
...
+                if (!contains(it.key().x() * CHUNK_SIZE + x,
+                              it.key().y() * CHUNK_SIZE + y))
+                    continue;
+
+                clone->setCell(it.key().x() * CHUNK_SIZE + x,
+                               it.key().y() * CHUNK_SIZE + y,
+                               it.value()->cellAt(x, y));
```
bjorn: Actually this is a needlessly slow way of cloning. You should just copy over each chunk:
```c++
    QHashIterator<QPoint, Chunk* > it(mChunks);
    while (it.hasNext()) {
        it.next();
        clone->mChunks.insert(it.key(), new Chunk(*it.value()));
    }
```

src/libtiled/tilelayer.cpp
```diff
@@ -127,7 +180,10 @@ void Tiled::TileLayer::setCell(int x, int y, const Cell &cell)
         }
     }
 
-    existingCell = cell;
+    chunk(x, y)->setCell(x & CHUNK_MASK, y & CHUNK_MASK, cell);
+
+    if (chunk(x, y)->isEmpty())
```
bjorn: This is the third call to `chunk(x, y)` in this function. You should really call it just once and remember the return value.

src/libtiled/tilelayer.cpp
ketanhwr: Okay, I totally forgot that we can access the private members of another object in the same class. Also, forgot about the implicit copy constructor. Nice..!

src/libtiled/tilelayer.cpp
```diff
@@ -305,28 +316,40 @@ void TileLayer::flipHexagonal(FlipDirection direction)
...
+
+                mask = flipMask[mask];
+
+                dest.setFlippedHorizontally((mask & 8) != 0);
+                dest.setFlippedVertically((mask & 4) != 0);
+                dest.setFlippedAntiDiagonally((mask & 2) != 0);
+                dest.setRotatedHexagonal120((mask & 1) != 0);
+
+                (direction == FlipHorizontally) ? newLayer->setCell(mWidth - _x - 1, _y, dest)
+                                                : newLayer->setCell(_x, mHeight - _y - 1, dest);
```
bjorn: Please write this as an `if`, `else`

src/libtiled/tilelayer.cpp
```diff
@@ -305,28 +316,40 @@ void TileLayer::flipHexagonal(FlipDirection direction)
...
+    QHashIterator<QPoint, Chunk* > it(mChunks);
+    while (it.hasNext()) {
+        it.next();
+        for (int y = 0; y < CHUNK_SIZE; ++y) {
+            for (int x = 0; x < CHUNK_SIZE; ++x) {
+                int _x = it.key().x() * CHUNK_SIZE + x;
+                int _y = it.key().y() * CHUNK_SIZE + y;
+
+                if (!contains(_x, _y))
+                    continue;
```
bjorn: I think instead of doing this contains check each time, it would be better to check `dest.isEmpty()` below and continue then. This applies to the other functions as well.

src/libtiled/tilelayer.cpp
```diff
@@ -305,28 +316,40 @@ void TileLayer::flipHexagonal(FlipDirection direction)
...
+
+                (direction == FlipHorizontally) ? newLayer->setCell(mWidth - _x - 1, _y, dest)
+                                                : newLayer->setCell(_x, mHeight - _y - 1, dest);
+            }
         }
     }
 
-    copyGrid(newGrid);
+    mChunks = newLayer->mChunks;
+    delete newLayer;
```
bjorn: I realized this only worked because we were never deleting the chunks. I've fixed that leak by pushing a change where the chunk instances are managed by the QHash.

Btw, instead of manual deletion, please declare the `newLayer` as:
```c++
QScopedPointer<TileLayer> newLayer
```

src/libtiled/tilelayer.cpp
```diff
@@ -551,45 +592,56 @@ void TileLayer::offsetTiles(const QPoint &offset,
...
+            for (int x = 0; x < CHUNK_SIZE; ++x) {
+                int _x = it.key().x() * CHUNK_SIZE + x;
+                int _y = it.key().y() * CHUNK_SIZE + y;
+
+                if (!contains(_x, _y))
+                    continue;
+
+                // Skip out of bounds tiles
+                if (!bounds.contains(_x, _y)) {
+                    newLayer->setCell(x, y, cellAt(_x, _y));
```
bjorn: You don't need to do this, since the `newLayer` is already a clone of the current layer. Actually, this means that you should only iterate the area indicated by `bounds`. Iterating just the chunks is not enough, since it could fail to include the needed area.

src/libtiled/tilelayer.cpp
```diff
@@ -271,31 +271,42 @@ void TileLayer::erase(const QRegion &area)
...
+
+                const Cell &source = it.value()->cellAt(x, y);
+                Cell dest(source);
+
+                if (direction == FlipHorizontally) {
+                    dest.setFlippedHorizontally(!source.flippedHorizontally());
+                    newLayer->setCell(mWidth - _x - 1, _y, dest);
+                } else if (direction == FlipVertically) {
+                    dest.setFlippedVertically(!source.flippedVertically());
+                    newLayer->setCell(_x, mHeight -y - 1, dest);
```
ketanhwr: And it should be `_y` too.

src/libtiled/tilelayer.cpp
ketanhwr: That would mean iterating like this, right?
```
    for (int x = bounds.x(); x < bounds.x() + bounds.width(); ++x) {
        for (int y = bound.y(); y < bounds.y() + bounds.height; ++y) {
            ...
        }
    }
```

bjorn:
Btw, I noticed the main reason for the slow load times is due to the following lines in `setCell`:

```c++
    if (_chunk.isEmpty())
        mChunks.remove(chunkCoordinates(x, y));
```

In particular, it is slow because setting an empty cell at a location that is already entirely empty will cause the chunk to get allocated and then immediately de-allocated again, and this will happen for each cell of an empty tile layer.

While this confirms that your idea to avoid setting empty cells in the map reader would resolve that particular slowness, I think instead it may be good to avoid the allocation and de-allocation in general by inserting the following check at the top:

```c++
    if (cell.isEmpty() && !findChunk(x, y))
        return;
```

Additionally, we could not bother with de-allocating chunks at all. I think it's a bit dubious whether we really need to free up memory there, because the main point of the chunks is to avoid massive memory allocation. In contrast, the user is probably not going to erase large areas and then care about whether that frees up some memory. If we don't bother with de-allocation, then there is also no need for keeping track of `mUsedCells` in the chunks.

src/libtiled/tilelayer.cpp
bjorn: Yes, though you can use `top`, `left`, `right` and `bottom` of `QRect` instead (noting the `<=`):
```c++
    for (int x = bounds.left(); x <= bounds.right(); ++x) {
        for (int y = bounds.top(); y <= bounds.bottom(); ++y) {
            ...
        }
    }
```

src/libtiled/tilelayer.cpp
```diff
@@ -165,6 +170,9 @@ void Tiled::TileLayer::setCell(int x, int y, const Cell &cell)
 {
     Q_ASSERT(contains(x, y));
 
+    if (cell.isEmpty() && !findChunk(x, y))
+    return;
```
bjorn: Indentation is off.

src/libtiled/tilelayer.cpp
```diff
@@ -280,30 +285,29 @@ void TileLayer::flip(FlipDirection direction)
                 int _x = it.key().x() * CHUNK_SIZE + x;
                 int _y = it.key().y() * CHUNK_SIZE + y;
 
-                if (!contains(_x, _y))
-                    continue;
-
                 const Cell &source = it.value().cellAt(x, y);
                 Cell dest(source);
 
+                if (dest.isEmpty())
```
bjorn: You can do this check a little earlier based on `source`, but in fact I wonder why you have two variables at all. I'd just do:
```c++
Cell cell = it.value().cellAt(x, y);
if (cell.isEmpty())
    continue;
```

HeadClot:
Happy that this got merged in.

ketanhwr:
@HeadClot, the feature is not complete yet. Currently the only thing completed is storing the Tile Layer in 16*16 blocks. I'll now proceed to work on automatically resizing maps.

HeadClot:
@ketanhwr Thanks for the info. I am still excited to see this feature be worked on :)

PR #1637, opened 2017-07-03
Automatically generate Java classes from XSD

mikepthomas:
Created XML Schema file using the [tiled map format documentation](http://doc.mapeditor.org/reference/tmx-map-format/) to automatically generate Java classes and javadoc documentation and bring in attributes that have been introduced into newer versions of Tiled


util/java/tmxviewer-java/src/main/java/TMXViewer.java
```diff
@@ -119,10 +132,12 @@ public void paintComponent(Graphics g) {
...
         // Draw each tile map layer
-        for (MapLayer layer : map) {
-            if (layer instanceof TileLayer) {
-                renderer.paintTileLayer(g2d, (TileLayer) layer);
-            }
+        for (TileLayer layer : map.getLayers()) {
+            renderer.paintTileLayer(g2d, layer);
+        }
+        // Draw each object group
+        for (ObjectGroup group : map.getObjectGroups()) {
```
bjorn: Hm, this is not how layers work, since layers of different types are supposed to get interleaved. Would it still be possible to set it up that way?

bjorn:
Ah, maybe we should not commit `util/java/libtiled-java/src/main/resources/map.htm`? It would be nice to host this as part of the documentation website, though.

mikepthomas:
> As far as I understand, much of the code is actually no longer there, because it will get automatically generated as part of the build?

Yes that is correct, I have updated the appveyor build script to build the java projects using maven so that the code can be generated as the qbs task cannot do this, however the build is still failing even though the maven task is saying it is successful... not sure why at the moment

> I just noticed one problem with the layers. I hope it is still possible to have all layers in a single array, since the order is important.

This makes perfect sense now you mention it, I have now fixed it so that TileLayer, ObjectGroup and ImageLayer all live in the same list as requested

> Ah, maybe we should not commit util/java/libtiled-java/src/main/resources/map.htm? It would be nice to host this as part of the documentation website, though.

I have now removed the generated documentation, im not quite sure where you would want it put to be part of the documentation website though... let me know if you need anything more

bjorn:
Well, that fixes the AppVeyor build. The fact that now the Travis CI build is broken is not relevant to this PR. Do you think this change should be ready for merging now, @mikepthomas?

mikepthomas:
I believe so, If you are happy with the content it can be merged. Once merged, I can start the process to deploy to maven central. Let me know if you need anything more :)

bjorn:
Alright, I've just tried to get this to work locally. While compiling `libtiled-java`, I noticed many warnings about deprecated `x`, `y`, `width` and `height` in `LayerData`. Is that a currently expected state or did you forget something?

When I tried to get `tmxviewer-java` to work, `mvn clean install` gave me:
```
[ERROR] Failed to execute goal on project tmxviewer: Could not resolve dependencies for project org.mapeditor:tmxviewer:jar:1.0-SNAPSHOT: Failed to collect dependencies at org.mapeditor:libtiled:jar:1.0-SNAPSHOT: Failed to read artifact descriptor for org.mapeditor:libtiled:jar:1.0-SNAPSHOT: Could not find artifact org.mapeditor:tiled-parent:pom:1.0-SNAPSHOT -> [Help 1]
```
But I did do the `mvn clean install` in the `libtiled-java` directory before as mentioned in the `README`. What can I do?

mikepthomas:
> Alright, I've just tried to get this to work locally. While compiling libtiled-java, I noticed many warnings about deprecated x, y, width and height in LayerData. Is that a currently expected state or did you forget something?

This is expected because these values have been marked as deprecated in the [tmx map format docs](http://doc.mapeditor.org/reference/tmx-map-format/):
"Attributes or elements that are deprecated or unsupported by the current version of Tiled are formatted in italics."

The warnings are being displayed as [the showDeprecation flag](https://github.com/mikepthomas/tiled/blob/master/util/java/pom.xml#L105) is set in the maven compiller plugin, this can be turned off if you want to surpress the warnings during compile time.

> But I did do the mvn clean install in the libtiled-java directory before as mentioned in the README. What can I do?

The command "mvn clean install" should be done in the util/java directory so that it builds both projects at the same time. The error you are receiving is due to the parent pom not being in your local maven repository and I have yet to deploy a 1.0-SNAPSHOT build to the [oss snapshots repository](https://oss.sonatype.org/content/repositories/snapshots/org/mapeditor/tiled-parent/). When a snapshot build has been deployed, then building from each project directory should work fine.

bjorn:
> This is expected because these values have been marked as deprecated in the tmx map format docs:
> "Attributes or elements that are deprecated or unsupported by the current version of Tiled are formatted in italics."

Hmm, well technically I wrote "deprecated or unsupported", and for these attributes it actually means more like "unsupported". Tiled can read and write maps using those attributes and will generally work fine, but it can't move or resize tile layers individually. I realized now that `width` and `height` are even required, since leaving them out would mean an empty (0x0) tile layer (the properties don't default to the map size, like I may have planned at some point). So I need to fix up those docs...

> The command "mvn clean install" should be done in the util/java directory so that it builds both projects at the same time. The error you are receiving is due to the parent pom not being in your local maven repository and I have yet to deploy a 1.0-SNAPSHOT build to the oss snapshots repository. When a snapshot build has been deployed, then building from each project directory should work fine.

Ok, that makes sense. The `README` file in the `tmxviewer-java` should probably mention the part about running the command in the parent directory.

mikepthomas:
> Hmm, well technically I wrote "deprecated or unsupported", and for these attributes it actually means more like "unsupported". Tiled can read and write maps using those attributes and will generally work fine, but it can't move or resize tile layers individually. I realized now that width and height are even required, since leaving them out would mean an empty (0x0) tile layer (the properties don't default to the map size, like I may have planned at some point). So I need to fix up those docs...

Even though they are deprecated in libtiled-java they should be able to be read and written to without issues too, and the warnings are just being generated due to the property being marked with the \@deprecated annotation, I can either remove the annotation or ignore the warning, whichever you feel best.

> Ok, that makes sense. The README file in the tmxviewer-java should probably mention the part about running the command in the parent directory.

I can add this to the documentation, however it will only be required when the pom does not exist in the snapshots repository. I will need to add it as part to the deployment to maven central and once it is there, building each project individually should work just fine. So I am not sure the documentation will be relevant after deployment. I can add a snapshot build to oss tonight from my forked repository if you like however the last one was created with a clean checkout from your master branch.

bjorn:
Sorry for the delay. It looks like a great improvement and I've just given it another test run and it was working fine, so it's merged now!

PR #1638, opened 2017-07-04
Enabling wangfilling for BucketFillTool and StampBrush

Bdtrotte:
and hopefully implementing a new dock/view/model for assigning tiles to a wang set


Bdtrotte:
The latest changes assert a new important rule for wangIds. If the edge count of a wangSet == 1, then for a wangId to be valid, all the edges must == 0. The same applies for corners. This is because for the purposes of finding tiles with wangIds, it is much faster if half the Id is already known; without the standardization all possibilities must be checked. Furthermore, when moving from a set with no edges, to one with them, it makes sense that the previously non-existent edges be thought of as wildcards.

Bdtrotte:
With painting working, and a whole lot more refined, I feel this PR is one step away from being merged (not including reviews). All thats left is to alter things to support undo commands, and cleaning up some quarks with how the UI behaves.

Bdtrotte:
Undo now works when painting wangIds onto tiles. There is still the case where one removes wangIds from tiles by changing the edge/corner counts of the wangSet. Currently there is no way to retrieve this once done, in the next push this will be fixed.

Also, some more test files have been added with a real tileset to show off some of the fun!

src/tiled/wangtemplatemodel.cpp
```diff
@@ -0,0 +1,136 @@
...
+            return QModelIndex();
+
+        wangId = wangId - 0x10101010;
+    }
+
+    int row = 0;
+    int cornerEdgePermutations = edges * corners;
+
+    for (int i = 0; i < 8; ++i) {
+        int bellowPermutations = qPow(cornerEdgePermutations, i/2) * ((i&1)? edges : 1);
```
bjorn: Typo: "below"

src/tiled/wangtemplatemodel.cpp
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    return index(row, 0);
+}
+
+void WangTemplateModel::setWangSet(WangSet *wangSet)
+{
+    mWangSet = wangSet;
+
+    resetModel();
```
bjorn: While sometimes it may be unavoidable, please implement this properly when possible:
```c++
beginResetModel();
mWangSet = wangSet;
endResetModel();
```

src/tiled/wangtemplatemodel.h
```diff
@@ -0,0 +1,68 @@
...
+public:
+    enum UserRoles {
+        WangIdRole = Qt::UserRole
+    };
+
+    WangTemplateModel(WangSet *wangSet, QObject *parent = nullptr);
+
+    int rowCount(const QModelIndex &parent = QModelIndex()) const override;
+
+    int columnCount(const QModelIndex &parent = QModelIndex()) const override;
```
bjorn: Since you derive from `QAbstractListModel`, you can remove this override because the list model already returns a column count of 1 by default.

src/tiled/wangtemplateview.cpp
```diff
@@ -0,0 +1,403 @@
...
+    : QListView(parent)
+    , mZoomable(new Zoomable(this))
+{
+    setWrapping(true);
+    setFlow(QListView::LeftToRight);
+    setResizeMode(QListView::Adjust);
+    setSelectionMode(QAbstractItemView::SingleSelection);
+    setVerticalScrollMode(QAbstractItemView::ScrollPerPixel);
+    setItemDelegate(new WangTemplateDelegate(this, this));
+    setTabKeyNavigation(true);
```
bjorn: Please add:
```c++
setUniformItemSizes(true);
```
I didn't notice an actual performance difference, but the view may be able to avoid getting the size hint very often.

You can remove the `setTabKeyNavigation(true);`. That is the default, but actually it won't work anyway because Tab is used to show/hide the all menu bars and dock widgets. Alternatively it could be set to `false` explicitly.

src/tiled/wangtemplateview.cpp
```diff
@@ -0,0 +1,403 @@
...
+
+qreal WangTemplateView::scale() const
+{
+    return mZoomable->scale();
+}
+
+void WangTemplateView::setModel(QAbstractItemModel *model)
+{
+    QListView::setModel(model);
+    updateBackgroundColor();
```
bjorn: Since `updateBackgroundColor` does not change color based on the model, this override can be removed. It could be called from the constructor and you should connect it to `StyleHelper::styleApplied`, as done in `TilesetView`.

src/libtiled/wangset.cpp
```diff
@@ -288,11 +473,62 @@ bool WangSet::wangIdIsValid(WangId wangId) const
...
+bool WangSet::wangIdIsUsed(WangId wangId) const
+{
+    unsigned mask = 0;
+
+    if (mEdgeColors > 1)
+        mask |= 0x0f0f0f0f;
+    if (mCornerColors > 1)
+        mask |= 0xf0f0f0f0;
+
+    for (WangId usedId : mWangIdToWangTile.keys()) {
```
Bdtrotte: That is originally what I was doing, but ran into some problems when there was less certainty about the wangId (I was searching for something with at least all ones in a corner only set, but the wangSet had the corners as wildcards (0)). The mask just allows that, though it should now never happen, the spaces in the wangId which don't matter can be different. Though now that I have made more clear rules about wangIds, this could be changed.

src/libtiled/wangset.cpp
```diff
@@ -288,11 +473,62 @@ bool WangSet::wangIdIsValid(WangId wangId) const
...
+        int bellowPermutations = qPow(cornerEdgePermutations, i/2) * ((i&1)? mEdgeColors : 1);
+        int value = n / bellowPermutations;
+        n -= value * bellowPermutations;
+
+        wangId |= value << i * 4;
+    }
+
+    //before this is like a base 10 range (0 - 9) where we want (1 - 10) for each digit
+    wangId += 0x11111111;
+    //If edges/corners don't have variations then those spots should be wild.
```
Bdtrotte: The math above treats each digit like a traditional base n number, having range [0 - (n-1)]. Though wangId digits range from [1 - n]. So adding 1 to each digit converts between the systems. Though I suppose this is an area where having 0xf be the wild card may look a bit nicer.

src/tiled/bucketfilltool.cpp
```diff
@@ -338,13 +346,36 @@ void BucketFillTool::setRandom(bool value)
...
     // Don't need to recalculate fill region if there was no fill region
     if (!mFillOverlay)
         return;
 
     tilePositionChanged(tilePosition());
 }
 
+void BucketFillTool::setWangFill(bool value)
+{
+    if(mIsWangFill == value)
```
Bdtrotte: And I thought I was getting better about that!!!

src/tiled/bucketfilltool.cpp
```diff
@@ -360,6 +391,59 @@ void BucketFillTool::randomFill(TileLayer &tileLayer, const QRegion &region) con
...
+                              const TileLayer &backgroundtileLayer,
+                              const QRegion &region) const
+{
+    if (region.isEmpty() || !mWangSet)
+        return;
+
+    for (const QRect &rect : region.translated(-tileLayerToFill.position()).rects()) {
+        for (int _x = rect.left(); _x <= rect.right(); ++_x) {
+            for (int _y = rect.top(); _y <= rect.bottom(); ++_y) {
+                tileLayerToFill.setCell(_x, _y, Cell());
```
Bdtrotte: I did find otherwise, without this, the fill won't change after the first generation. I was wondering if there was an easier way to empty a tileLayer?

bjorn: We should probably add a `TileLayer::clear` method, but let's not do that now because its implementation will need to be different on `master`. So we should wait with that until after the merge.

src/tiled/changetilewangid.cpp
```diff
@@ -0,0 +1,91 @@
...
+    setText(QCoreApplication::translate("Undo Commands", "Change Tile WangId"));
+}
+
+ChangeTileWangId::ChangeTileWangId(TilesetDocument *tilesetDocument, WangSet *wangSet, Tile *tile, WangId wangId)
+    : mTilesetDocument(tilesetDocument)
+    , mWangSet(wangSet)
+    , mMergeable(true)
+{
+    setText(QCoreApplication::translate("Undo Commands", "Change Tile WangId"));
+    if (mWangSet)
```
Bdtrotte: However unlikely, could a nullptr not be passed for the wangSet?

src/tiled/changewangsetdata.cpp
```diff
@@ -42,43 +43,107 @@ ChangeWangSetEdges::ChangeWangSetEdges(TilesetDocument *tilesetDocument,
...
+                //when the size is reduced, some wang assignments can be lost.
+                const QList<Tile *> &changedTiles = wangSet->tilesChangedOnSetEdgeColors(mNewValue);
+
+                if (!changedTiles.isEmpty()) {
+                    QList <ChangeTileWangId::WangIdChange> changes;
+
+                    for (Tile *tile : changedTiles)
+                        changes.append(ChangeTileWangId::WangIdChange(wangSet->wangIdOfTile(tile), 0, tile));
+
+                    mClearTiles = new ChangeTileWangId(mTilesetDocument, wangSet, changes, this);
```
bjorn: I understand that part. I meant that it doesn't need to be referenced by a member variable.

src/tiled/changetilewangid.cpp
bjorn: When this happens, I think it would indicate an error in the code calling the constructor. You can consider writing `Q_ASSERT(wangSet);` at the start of the constructor instead so we can catch that.

src/tiled/changewangsetdata.cpp
Bdtrotte: ohhh very true, Ill remove that

src/tiled/tilesetview.cpp
```diff
@@ -241,6 +241,279 @@ static QTransform tilesetGridTransform(const Tileset &tileset, QPoint tileCenter
...
 }
 
+static void setWangStyle(QPainter *painter,int index, int max)
+{
+    QColor c;
+
+    if (index == 0)
+        c = QColor(0,0,0,0);
+    float hue = ((float)index-1)/(float)max;
+    c = QColor::fromHsvF(hue,1,1);
```
Bdtrotte: Need the `if (index == 0)` part to make alpha zero. But otherwise I cleaned up the casts.

src/tiled/tilesetview.cpp
```diff
@@ -609,40 +968,52 @@ void TilesetView::mouseMoveEvent(QMouseEvent *event)
...
+
+                mHoveredCorner = hoveredCorner;
+            }
+        }
 
         if (previousHoveredIndex.isValid())
             update(previousHoveredIndex);
-        if (previousHoveredIndex != mHoveredIndex && mHoveredIndex.isValid())
+        if (mHoveredIndex.isValid() &&
+                (previousHoverCorner != mHoveredCorner || mHoveredIndex != previousHoveredIndex))
```
Bdtrotte: In fact it updates the `previousHoverIndex` every time the mouse moves (and it's valid). I've changed it to the intended behavior.

Bdtrotte:
There is one more change I would like to put on this pr, which is to send a signal from the TilesetView to the wangDock when it changes the current WangId used, and when it changes the used status of a certain wangId, so that the template view updates accordingly. After that it will be complete, and ready for merge.

PR #1640, opened 2017-07-06
Adds iterator and const_iterator for TileLayer

ketanhwr:
Follow up for #1635.

### DIFF
--- a/src/libtiled/tilelayer.h
+++ b/src/libtiled/tilelayer.h
@@ -212,6 +212,168 @@ inline const Cell &Chunk::cellAt(const QPoint &point) const
 class TILEDSHARED_EXPORT TileLayer : public Layer
 {
 public:
+    class iterator
+    {
+    public:
+        iterator(QHash<QPoint, Chunk>::iterator it, QHash<QPoint, Chunk>::iterator end)
+            : mChunkPointer(it)
+            , mChunkEndPointer(end)
+        {
+            if (it != end)
+                mCellPointer = it.value().begin();
+        }
+
+        iterator operator++()
+        {
+            iterator it = *this;
+
+            if (mChunkPointer != mChunkEndPointer) {
+                if (mCellPointer == mChunkPointer.value().end()) {
+                    mChunkPointer++;
+                    if (mChunkPointer != mChunkEndPointer)
+                        mCellPointer = mChunkPointer.value().begin();
+                } else {
+                    mCellPointer++;
+                }
+            }
+
+            return it;
+        }
+
+        iterator operator++(int)
+        {
+            if (mChunkPointer != mChunkEndPointer) {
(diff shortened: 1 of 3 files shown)


src/libtiled/tilelayer.h
```diff
@@ -212,6 +212,160 @@ inline const Cell &Chunk::cellAt(const QPoint &point) const
...
+
+        iterator &operator++()
+        {
+            if (mChunkPointer != mChunkEndPointer) {
+                if (++mCellPointer == mChunkPointer.value().end()) {
+                    mChunkPointer++;
+                    if (mChunkPointer != mChunkEndPointer)
+                        mCellPointer = mChunkPointer.value().begin();
+                }
+            }
```
bjorn: This code is duplicated in the postfix version and I think would be good to separate out in a private `advance` function.

src/libtiled/tilelayer.h
```diff
@@ -212,6 +212,160 @@ inline const Cell &Chunk::cellAt(const QPoint &point) const
...
+        {
+            if (lhs.mChunkPointer == lhs.mChunkEndPointer || rhs.mChunkPointer == rhs.mChunkEndPointer)
+                return lhs.mChunkPointer != rhs.mChunkPointer;
+            else
+                return lhs.mCellPointer != rhs.mCellPointer;
+        }
+
+        Cell &value() { return *mCellPointer; }
+
+        const QPoint key()
```
bjorn: The function itself should be `const`, not the return value.

src/libtiled/tilelayer.h
```diff
@@ -212,6 +212,160 @@ inline const Cell &Chunk::cellAt(const QPoint &point) const
...
+
+        friend bool operator!=(const iterator& lhs, const iterator& rhs)
+        {
+            if (lhs.mChunkPointer == lhs.mChunkEndPointer || rhs.mChunkPointer == rhs.mChunkEndPointer)
+                return lhs.mChunkPointer != rhs.mChunkPointer;
+            else
+                return lhs.mCellPointer != rhs.mCellPointer;
+        }
+
+        Cell &value() { return *mCellPointer; }
```
bjorn: This this operator doesn't change the iterator, it should still be marked `const`.

src/libtiled/tilelayer.h
```diff
@@ -212,6 +212,160 @@ inline const Cell &Chunk::cellAt(const QPoint &point) const
...
+                        mCellPointer = mChunkPointer.value().begin();
+                }
+            }
+
+            return *this;
+        }
+
+        Cell &operator*() { return *mCellPointer; }
+
+        QVector<Cell>::iterator operator->() { return mCellPointer; }
```
bjorn: operator should be `const` since it doesn't modify the iterator.

bjorn:
Thanks!

PR #1641, opened 2017-07-06
Fix tile replacement

thabetx:
When the tile of an object is replaced with a tile from an unused tileset, the new tileset wasn't added to the map. This PR fixes that.


bjorn:
Thanks! Please change the target branch to `1.0` since it is a bugfix.

PR #1646, opened 2017-07-08
Fix .tbin exporting tile properties.

spacechase0:
I swear I tested this stuff.


bjorn:
No problem, seems like a bug that krept in by code modifications done after testing. :-)

bjorn:
I made a small fix for a crash that could potentially happen (c164e24a28d5e94af57ff1c7c7a9f2118822c3e8).

PR #1647, opened 2017-07-10
Adds move selection feature

ketanhwr:
#650 

The tool works when an area is selected and then Ctrl+Alt is used to move the selection. Ctrl+Alt+Shift is used to duplicate the selection but this modifer works weirdly (You need to hold Ctrl+Alt and after that the shift).

### DIFF
--- a/src/tiled/abstracttileselectiontool.h
+++ b/src/tiled/abstracttileselectiontool.h
@@ -23,6 +23,8 @@
 
 #include "abstracttiletool.h"
 
+#include "tilelayer.h"
+
 class QAction;
 class QActionGroup;
 
@@ -41,6 +43,7 @@ public:
                               const QKeySequence &shortcut,
                               QObject *parent = nullptr);
 
+    void mouseMoved(const QPointF &pos,Qt::KeyboardModifiers modifiers) override;
     void mousePressed(QGraphicsSceneMouseEvent *event) override;
     void mouseReleased(QGraphicsSceneMouseEvent *event) override;
 
@@ -51,6 +54,8 @@ public:
     void populateToolBar(QToolBar *toolBar) override;
 
 protected:
+    void tilePositionChanged(const QPoint &pos) override;
+
     enum SelectionMode {
         Replace,
         Add,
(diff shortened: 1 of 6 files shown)


bjorn:
I'm a little bit surprised by that it automatically applies the move when I release the mouse button. Why doesn't it stay in a floating state, like the movement tool used to do?

Also I wonder if it's really needed to have Ctrl+Alt as modifier for this. I realize GIMP uses the same modifier, but they also handle just Alt for moving the selection without affecting the image. Either we could eventually also support Alt like that in Tiled, or we could try using just Alt for doing the drag. Or, we could try whether it would work without modifier, just going by "if you start dragging on the selection, without modifier". What do you think?

ketanhwr:
I like the option of just using Alt. Without using the modifer, might get a bit weird and a lot of tricky situations might come I think.

src/tiled/abstracttileselectiontool.h
```diff
@@ -41,6 +43,7 @@ class AbstractTileSelectionTool : public AbstractTileTool
                               const QKeySequence &shortcut,
                               QObject *parent = nullptr);
 
+    void mouseMoved(const QPointF &pos,Qt::KeyboardModifiers modifiers) override;
```
Ablu: space missing after `,` here

Ablu:
What is the status on this by the way?

bjorn:
I've pushed a commit with several changes. Unfortunately, I still don't think we can merge this to `master` because of two remaining issues:

* It is hardly clear to the user that his selection is floating. Probably the BrushItem should be enhanced with an option to render an animated border around its selected area.

* Like this, it does not work nicely with the undo system. When you start a drag and then undo, the expected behavior is that the drag is aborted with the tile layer restored. Currently, the layer is restored but the floating selection remains.

* When you finish a move by anchoring the selection, then undo will only undo the final Paint operation, and doesn't restore the floating selection.

Maybe I'm making it too complicated by allowing repeated drags, but I think it needs to be supported because users will expect it to work and they will need it to work when they need to take the selection somewhere outside of their screen (or revert back to using copy/paste of course).

I think we should look again at other editors to see how they behave regarding such a feature.

ketanhwr:
Maybe the move selection can have a logic similar to object selection tool. It can have the rotating ants animation as well, which would make the feature quite easy to use and understand.

PR #1648, opened 2017-07-10
Use QStandardPaths::findExecutable when executable is not a file

ketanhwr:


bjorn:
I shortly tested this, but it didn't work. It is because `QStandardPaths::findExecutable` returns the full path of the located executable (including the file name), and this is not a valid working directory.

bjorn:
Well, you should always test things of course. :-)

PR #1649, opened 2017-07-13
WangTiles: Better filling methods, wang brush

Bdtrotte:
A work in progress ppull request for the GSOC Wang Tiles project:

In this PR I will commit, first improvements to how wang filling and brushing is done. Next will be a major project to add in a new tool, the wang brush. This tool will function very closely to how the current terrain brush works.

Along with this there are sure to be a host of improvements to the general architecture of wangTiles.

### DIFF
--- a/src/libtiled/wangset.cpp
+++ b/src/libtiled/wangset.cpp
@@ -101,6 +101,76 @@ bool WangId::hasCornerWildCards() const
     return false;
 }
 
+struct WangWildCard
+{
+    int index, colorCount;
+};
+
+QVector<WangId> WangId::variations(int edgeColors, int cornerColors) const
+{
+    QVector<WangId> wangIds;
+    QVector<WangWildCard> wildCards;
+
+    if (edgeColors > 1) {
+        for (int i = 0; i < 4; ++i) {
+            if (!edgeColor(i)) {
+                WangWildCard w;
+                w.colorCount = edgeColors;
+                w.index = i * 8;
+                wildCards.append(w);
+            }
+        }
+    }
+
+    if (cornerColors > 1) {
+        for (int i = 0; i < 4; ++i) {
+            if (!cornerColor(i)) {
+                WangWildCard w;
+                w.colorCount = cornerColors;
+                w.index = (i * 8) + 4;
+                wildCards.append(w);
+            }
+        }
+    }
+
+    if (wildCards.isEmpty()) {
+        wangIds.append(mId);
+    } else {
+        QStack<WangWildCard> stack;
+
+        stack += wildCards;
+
+        int max = wildCards.size();
+
+        while (!stack.isEmpty()) {
+            if (stack.size() == max) {
+                unsigned idVariation = 0;
+
+                for (int i = 0; i < max; ++i)
+                    idVariation |= stack[i].colorCount << stack[i].index;
+
+                wangIds.append(idVariation | mId);
+
+                WangWildCard top = stack.pop();
+                top.colorCount -= 1;
+                if (top.colorCount >= 0)
+                    stack.push(top);
+            } else {
+                WangWildCard top = stack.pop();
+                top.colorCount -= 1;
+                if (top.colorCount >= 0) {
+                    stack.push(top);
+
+                    for (int i = stack.size(); i < max; ++i)
+                        stack.push(wildCards[i]);
+                }
+            }
+        }
+    }
+
+    return wangIds;
+}
+
 void WangId::rotate(int rotations)
 {
     if (rotations < 0)
@@ -306,85 +376,18 @@ WangTile WangSet::findMatchingWangTile(WangId wangId) const
         return WangTile();
 }
 
-struct WangWildCard
-{
-    int index, colorCount;
-};
-
 QList<WangTile> WangSet::findMatchingWangTiles(WangId wangId) const
 {
+    if (wangId == 0)
+        return mWangIdToWangTile.values();
+
     QList<WangTile> list;
+    QVector<WangId> wangIds = wangId.variations(mEdgeColors, mCornerColors);
 
-    //Stores the space of a wild card, followed by how many colors that space can have.
-    QVector<WangWildCard> wildCards;
-
-    if (mEdgeColors > 1) {
-        for (int i = 0; i < 4; ++i) {
-            if (!wangId.edgeColor(i)) {
-                WangWildCard w;
-                w.index = i * 8;
-                w.colorCount = mEdgeColors;
-
-                wildCards.append(w);
-            }
-        }
-    }
-
-    if (mCornerColors > 1) {
-        for (int i = 0; i < 4; ++i) {
-            if (!wangId.cornerColor(i)) {
-                WangWildCard w;
-                w.index = i * 8 + 4;
-                w.colorCount = mCornerColors;
-
-                wildCards.append(w);
-            }
-        }
-    }
-
-    if (wildCards.isEmpty()) {
-        list.append(mWangIdToWangTile.values(wangId));
-    } else {
-        QStack<WangWildCard> stack;
-
-        stack += wildCards;
-
-        int max = wildCards.size();
-
-        while (!stack.isEmpty()) {
-            if (stack.size() == max) {
-                int idVariation = 0;
-
-                for (int i = 0; i < max; ++i)
-                    idVariation |= stack[i].colorCount << stack[i].index;
-
-                list.append(mWangIdToWangTile.values(idVariation | wangId));
-
-                WangWildCard top = stack.pop();
-                top.colorCount -= 1;
-                if (top.colorCount >= 0)
-                    stack.push(top);
-            } else {
-                WangWildCard top = stack.pop();
-                top.colorCount -= 1;
-                if (top.colorCount >= 0) {
-                    stack.push(top);
-
-                    for (int i = stack.size(); i < max; ++i)
-                        stack.push(wildCards[i]);
-                }
-            }
-        }
-    }
+    for (WangId id : wangIds)
+        list.append(mWangIdToWangTile.values(id));
 
     return list;
-}
-
-Cell WangSet::findMatchingCell(WangId wangId) const
-{
-    WangTile wangTile = findMatchingWangTile(wangId);
-
-    return wangTile.makeCell();
 }
 
 WangId WangSet::wangIdFromSurrounding(WangId surroundingWangIds[]) const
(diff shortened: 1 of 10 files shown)


src/libtiled/wangset.cpp
```diff
@@ -134,6 +187,71 @@ void WangId::flipVertically()
     rotate(2);
 }
 
+WangIdVariationIterator::WangIdVariationIterator(int edgeColors, int cornerColors, WangId wangId)
```
Bdtrotte: The point of it is to make "WildWangIdUsed" a lot faster. That function is pretty important, so Im still working on speeding it up more.

In regards to the iterator itself, I know it isn't proper with end being valid, but the problem is in the 15 x 15 case, all unsigned ints are valid, so I'm not sure there is a way to make it invalid to dereference. So I'm not sure I can make it proper? Though I see your point it would be nice to avoid the allocation... Though currently, the variations function is unused (it is in the findMatchingWangTiles function, but that can be changed to use the iterator as is), so It could be removed all together. Ill work on a more elegant solution though,

src/libtiled/wangset.cpp
```diff
@@ -470,6 +521,42 @@ bool WangSet::wangIdIsUsed(WangId wangId) const
...
+    }
+
+    return false;
+}
+
+bool WangSet::isComplete() const
+{
+    if (isEmpty())
+        return false;
+    return (unsigned) mWangIdToWangTile.uniqueKeys().size() == completeSetSize();
```
Bdtrotte: I also realized since it should be possible to assign a wangId with 0s (wildcards) this has to check for the count of unique keys with no wild cards. So I'll write a specific function for this in WangSet based on your suggestion and that.

src/libtiled/wangset.cpp
```diff
@@ -470,6 +521,42 @@ bool WangSet::wangIdIsUsed(WangId wangId) const
...
+}
+
+unsigned WangSet::completeSetSize() const
+{
+    return qPow(mEdgeColors, 4) * qPow(mCornerColors, 4);
```
Bdtrotte: > We can't go out of int range here anyway due to limitations on the color counts, right?

I think we can. Max color is 15, so max return is 15^8 = 2562890625 > 0x7fffffff

To avoid casting, I the new function for unique keys will return unsigned as well.

src/libtiled/wangset.cpp
Bdtrotte: So the solution I've put together implements a proper iterator to replace to old messy one. It works for the end case by using the biggest wangId for given input + 1 to the underlying id. This works great for all sets, other than those 15x15, where this value will be zero, and immediately fail. As a work around, everywhere this iterator is used must be wrapped with an if testing if it is the 15x15 case, and if so takes on slightly different behavior.

bjorn: In that case, it will be better for it to fail immediately, so maybe you don't need that if...

src/tiled/wangdock.cpp
```diff
@@ -284,6 +294,14 @@ void WangDock::eraseWangIdsButtonClicked()
     emit currentWangIdChanged(mCurrentWangId);
 }
 
+void WangDock::switchTemplateViewButtonClicked()
+{
+    if (mWangTemplateView->isVisible())
+        setColorView(true);
+    else
+        setTemplateView(true);
```
bjorn: This way of switching appears to behave a little strange (with the height changing upon switching). I would suggest to use a QStackedWidget (or QStackedLayout if possible) to switch between the two views.

bjorn:
I think we'll need to sort by tile ID when saving the wang tiles, because their order is essentially randomized every time the file is saved currently due to using a QHash. I think using a QHash is still better for performance than a QMap, but upon saving we'll need to get the keys, sort them and then use that to write out the file.

Bdtrotte:
At this point this PR should be ready for final review and merge! Also completes the major mile stones of Wang Tiles in whole. Next to come will be a host of features and improvements for wang tiles to polish off the feature and get it totally ready for merge into master.

src/tiled/wangbrush.cpp
```diff
@@ -0,0 +1,447 @@
...
+        if (tilePos != mPaintPoint) {
+            mPaintPoint = tilePos;
+            stateChanged();
+        }
+    } else {
+        tilePos.setX(floor(tilePosF.x()));
+        tilePos.setY(floor(tilePosF.y()));
+
+        QPointF tileLocalPoint(modf(tilePosF.x(), nullptr),
+                              modf(tilePosF.y(), nullptr));
```
bjorn: This is crashing for me. Reading the documentation for `modf`, I don't see anything about it being valid to pass in a null pointer.

src/tiled/wangbrush.cpp
```diff
@@ -0,0 +1,447 @@
...
+        //Checks when painting which would avoid change.
+        if (mBrushBehavior == Paint && tilePos == mPaintPoint) {
+            if (abs(tileLocalPoint.x() - 0.5f) < MIDDLE_DEAD_ZONE
+                    && abs(tileLocalPoint.y() - 0.5f) < MIDDLE_DEAD_ZONE)
+                return;
+
+            switch (mEdgeDir) {
+            case 0:
+                if (tileLocalPoint.y() < EDGE_DEAD_ZONE)
+                    return;
```
bjorn: My compiler is warning about the fall-through here. Is it intensional?

src/tiled/wangbrush.cpp
Bdtrotte: Interesting... I've had no problems on mine... I'll pass a local int to be safe.

src/tiled/wangbrush.cpp
Bdtrotte: It was not, forgot my breaks

src/tiled/wangdock.cpp
```diff
@@ -333,13 +431,31 @@ void WangDock::setCurrentWangSet(WangSet *wangSet)
...
 void WangDock::retranslateUi()
 {
     setWindowTitle(tr("Wang Sets"));
 
     mEraseWangIdsButton->setText(tr("Erase WangIds"));
     mAddWangSet->setText(tr("Add Wang Set"));
     mRemoveWangSet->setText(tr("Remove Wang Set"));
+
+    if (mWangColorView->isVisible())
+        mSwitchTemplateViewButton->setText(tr("Switch to Template View"));
```
bjorn: Maybe you could try to use a `QTabWidget` and just have "Templates" and "Colors" tabs? It just feel like this would be a more natural way than having this toggle button. Though if you want to speed up the merging of this patch, feel free to just change the minimum based on my feedback for now and leave this for later.

Also, it occurred to me that maybe having two things called "Templates" would be confusing (Mohamed is also working on adding templates, after all). So maybe we can find a different name for this view...

src/tiled/wangfiller.cpp
```diff
@@ -0,0 +1,227 @@
...
+void WangFiller::setWangSet(WangSet *wangSet)
+{
+    mWangSet = wangSet;
+}
+
+Cell WangFiller::findFittingCell(const TileLayer &back,
+                                 const TileLayer &front,
+                                 const QRegion &fillRegion,
+                                 QPoint point,
+                                 bool lookForward) const
```
bjorn: For both `findFittingCell` and `fillRegion`, the `lookForward` boolean is based on `!mWangFiller->wangSet()->isComplete()`. Maybe just remove the parameter and check completeness locally?

src/tiled/wangfiller.cpp
```diff
@@ -0,0 +1,227 @@
...
+    } else if (!wangTiles.isEmpty()) {
+        wangTile = wangTiles.at(qrand() % wangTiles.size());
+    }
+
+    return wangTile.makeCell();
+}
+
+TileLayer *WangFiller::fillRegion(const TileLayer &back,
+                            const QRegion &fillRegion,
+                            bool lookForward) const
```
bjorn: Indentation is off.

src/tiled/wangfiller.cpp
```diff
@@ -0,0 +1,227 @@
...
+                                fill = wangTiles.isEmpty();
+
+                                break;
+                            }
+                        }
+                    }
+
+                    if (fill) {
+                        tileLayer->setCell(currentPoint.x() - tileLayer->x(),
+                                          currentPoint.y() - tileLayer->y(),
```
bjorn: Indentation off by one.

src/tiled/wangfiller.cpp
```diff
@@ -0,0 +1,227 @@
...
+                                          QPoint point) const
+{
+    Cell surroundingCells[8];
+
+    for (int i = 0; i < 8; ++i) {
+        QPoint adjacentPoint = point + adjacentPoints[i];
+        if (!fillRegion.contains(adjacentPoint) && back.contains(adjacentPoint))
+            surroundingCells[i] = back.cellAt(adjacentPoint);
+        else
+            surroundingCells[i] = Cell();
```
bjorn: This assignment shouldn't be necessary.

tests/wangtiles/grassAndWater.tsx
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<tileset name="GrassAndWater" tilewidth="64" tileheight="64" tilecount="24" columns="4">
+ <image source="C:/Users/topha/Desktop/grass_and_water.png" width="256" height="384"/>
```
bjorn: Please make sure there is a relative reference here.

src/tiled/stampbrush.cpp
```diff
@@ -242,6 +243,18 @@ void StampBrush::populateToolBar(QToolBar *toolBar)
     mStampActions->populateToolBar(toolBar, mIsRandom, mIsWangFill);
 }
 
+void StampBrush::setWangSet(WangSet *wangSet)
+{
+    mWangFiller->setWangSet(wangSet);
+
+    mMissingTilesets.clear();
+
+    const SharedTileset &tileset = wangSet->tileset()->sharedPointer();
```
bjorn: This is crashing when the `wangSet` is null (happens to me when I create a new map, for example).

src/tiled/wangfiller.h
```diff
@@ -0,0 +1,93 @@
...
+                         bool lookForward = true) const;
+
+    /* Returns a tilelayer which has fillRegion filled with wang methods.
+     * If lookForward is true, this will only choose a cell
+     * which allows all empty adjacent cells to also
+     * be filled. If non exist, then no cell will be choosen.
+     * */
+    TileLayer *fillRegion(const TileLayer &back,
+                    const QRegion &fillRegion,
+                    bool lookForward = true) const;
```
bjorn: Indentation is off.

src/tiled/wangbrush.cpp
```diff
@@ -0,0 +1,453 @@
...
+        brushItem()->setLayerOffset(layer->totalOffset());
+    }
+
+    const MapRenderer *renderer = mapDocument()->renderer();
+    const QPointF tilePosF = renderer->screenToTileCoords(offsetPos);
+    QPoint tilePos;
+
+    if (mBrushMode == PaintVertex) {
+        tilePos = tilePosF.toPoint();
+        if (tilePos != mPaintPoint) {
```
Bdtrotte: I was using that at first, but I think the behavior I wanted needed something slightly different from PaintVertex. In any case Ill take another look over it and make sure it couldn't be made better.

src/tiled/wangdock.cpp
Bdtrotte: I'll definitely save it for a later patch. As for the name change thats a good point, I'll think of something different.

src/tiled/wangdock.cpp
```diff
@@ -289,6 +345,38 @@ void WangDock::refreshCurrentWangId()
...
+{
+    QItemSelectionModel *selectionModel = mWangColorView->selectionModel();
+
+    if (!selectionModel->currentIndex().isValid()) {
+        emit wangColorChanged(0, true);
+        return;
+    }
+
+    bool edgeColor;
+    //If not using edges, then corners becomes the first row, and isEdgeColorAt becomes invalid.
```
Bdtrotte: Well this is incredibly useful! Thanks

PR #1651, opened 2017-07-14
Automatically Resize Maps - #260

ketanhwr:
`mRegion` keeps track of the non-empty region of layer. The bounds function is modified so that it calculates the bounds using `mRegion` itself.

### DIFF
--- a/src/libtiled/hexagonalrenderer.cpp
+++ b/src/libtiled/hexagonalrenderer.cpp
@@ -88,6 +88,11 @@ QSize HexagonalRenderer::mapSize() const
     }
 }
 
+QPoint HexagonalRenderer::mapStart() const
+{
+    return QPoint(0, 0);
+}
+
 QRect HexagonalRenderer::boundingRect(const QRect &rect) const
 {
     const RenderParams p(map());
--- a/src/libtiled/hexagonalrenderer.cpp
+++ b/src/libtiled/hexagonalrenderer.cpp
@@ -63,8 +63,7 @@ HexagonalRenderer::RenderParams::RenderParams(const Map *map)
     rowHeight = sideOffsetY + sideLengthY;
 }
 
-
-QSize HexagonalRenderer::mapSize() const
+QRect HexagonalRenderer::mapBoundingRect() const
 {
     const RenderParams p(map());
 
@@ -76,7 +75,7 @@ QSize HexagonalRenderer::mapSize() const
         if (map()->width() > 1)
             size.rheight() += p.rowHeight;
 
-        return size;
+        return QRect(0, 0, size.width(), size.height());
     } else {
         QSize size(map()->width() * (p.tileWidth + p.sideLengthX),
                    map()->height() * p.rowHeight + p.sideOffsetY);
@@ -84,13 +83,8 @@ QSize HexagonalRenderer::mapSize() const
         if (map()->height() > 1)
             size.rwidth() += p.columnWidth;
 
-        return size;
+        return QRect(0, 0, size.width(), size.height());
     }
-}
-
-QPoint HexagonalRenderer::mapStart() const
-{
-    return QPoint(0, 0);
 }
 
 QRect HexagonalRenderer::boundingRect(const QRect &rect) const
--- a/src/libtiled/hexagonalrenderer.cpp
+++ b/src/libtiled/hexagonalrenderer.cpp
@@ -254,7 +254,7 @@ void HexagonalRenderer::drawTileLayer(QPainter *painter,
     QRect rect = exposed.toAlignedRect();
 
     if (rect.isNull())
-        rect = boundingRect(layer->rect());
+        rect = boundingRect(layer->bounds());
 
     QMargins drawMargins = layer->drawMargins();
     drawMargins.setBottom(drawMargins.bottom() + p.tileHeight);
--- a/src/libtiled/hexagonalrenderer.cpp
+++ b/src/libtiled/hexagonalrenderer.cpp
@@ -86,7 +86,7 @@ QRect HexagonalRenderer::mapBoundingRect() const
     if (p.staggerX) {
         QSize size(mapBounds.width() * p.columnWidth + p.sideOffsetX,
                    mapBounds.height() * (p.tileHeight + p.sideLengthY));
-        QPoint origin(mapBounds.x() * p.columnWidth + p.sideOffsetX,
+        QPoint origin(mapBounds.x() * p.columnWidth,
                       mapBounds.y() * (p.tileHeight + p.sideLengthY));
 
 
@@ -98,7 +98,7 @@ QRect HexagonalRenderer::mapBoundingRect() const
         QSize size(mapBounds.width() * (p.tileWidth + p.sideLengthX),
                    mapBounds.height() * p.rowHeight + p.sideOffsetY);
         QPoint origin(mapBounds.x() * (p.tileWidth + p.sideLengthX),
-                      mapBounds.y() * p.rowHeight + p.sideOffsetY);
+                      mapBounds.y() * p.rowHeight);
 
         if (mapBounds.height() > 1)
             size.rwidth() += p.columnWidth;
--- a/src/libtiled/hexagonalrenderer.cpp
+++ b/src/libtiled/hexagonalrenderer.cpp
@@ -254,7 +254,7 @@ void HexagonalRenderer::drawTileLayer(QPainter *painter,
     QRect rect = exposed.toAlignedRect();
 
     if (rect.isNull())
-        rect = boundingRect(layer->bounds());
+        rect = boundingRect(layer->rect());
 
     QMargins drawMargins = layer->drawMargins();
     drawMargins.setBottom(drawMargins.bottom() + p.tileHeight);
--- a/src/libtiled/hexagonalrenderer.cpp
+++ b/src/libtiled/hexagonalrenderer.cpp
@@ -309,8 +309,8 @@ void HexagonalRenderer::drawTileLayer(QPainter *painter,
 
     CellRenderer renderer(painter, CellRenderer::HexagonalCells);
 
-    const int endX = map()->infinite() ? layer->bounds().width() : layer->width();
-    const int endY = map()->infinite() ? layer->bounds().height() : layer->height();
+    const int endX = map()->infinite() ? layer->bounds().right() - layer->x() + 1 : layer->width();
+    const int endY = map()->infinite() ? layer->bounds().bottom() - layer->y() + 1 : layer->height();
 
     if (p.staggerX) {
         if (!map()->infinite()) {
@@ -365,7 +365,7 @@ void HexagonalRenderer::drawTileLayer(QPainter *painter,
         if (p.doStaggerY(startTile.y() + layer->y()))
             startPos.rx() -= p.columnWidth;
 
-        for (; startPos.y() < rect.bottom() && endY; startTile.ry()++) {
+        for (; startPos.y() < rect.bottom() && startTile.y() < endY; startTile.ry()++) {
             QPoint rowTile = startTile;
             QPoint rowPos = startPos;
 
--- a/src/libtiled/hexagonalrenderer.cpp
+++ b/src/libtiled/hexagonalrenderer.cpp
@@ -143,8 +143,10 @@ void HexagonalRenderer::drawGrid(QPainter *painter, const QRectF &exposed,
     if (inLeftHalf)
         startTile.rx()--;
 
-    startTile.setX(qMax(0, startTile.x()));
-    startTile.setY(qMax(0, startTile.y()));
+    if (!map()->infinite()) {
+        startTile.setX(qMax(0, startTile.x()));
+        startTile.setY(qMax(0, startTile.y()));
+    }
 
     startPos = tileToScreenCoords(startTile).toPoint();
 
@@ -170,14 +172,14 @@ void HexagonalRenderer::drawGrid(QPainter *painter, const QRectF &exposed,
         if (p.doStaggerX(startTile.x()))
             startPos.ry() -= p.rowHeight;
 
-        for (; startPos.x() <= rect.right() && startTile.x() < map()->width(); startTile.rx()++) {
+        for (; startPos.x() <= rect.right() && (startTile.x() < map()->width() || map()->infinite()); startTile.rx()++) {
             QPoint rowTile = startTile;
             QPoint rowPos = startPos;
 
             if (p.doStaggerX(startTile.x()))
                 rowPos.ry() += p.rowHeight;
 
-            for (; rowPos.y() <= rect.bottom() && rowTile.y() < map()->height(); rowTile.ry()++) {
+            for (; rowPos.y() <= rect.bottom() && (rowTile.y() < map()->height() || map()->infinite()); rowTile.ry()++) {
                 lines.append(QLine(rowPos + oct[1], rowPos + oct[2]));
                 lines.append(QLine(rowPos + oct[2], rowPos + oct[3]));
                 lines.append(QLine(rowPos + oct[3], rowPos + oct[4]));
@@ -208,14 +210,14 @@ void HexagonalRenderer::drawGrid(QPainter *painter, const QRectF &exposed,
         if (p.doStaggerY(startTile.y()))
             startPos.rx() -= p.columnWidth;
 
-        for (; startPos.y() <= rect.bottom() && startTile.y() < map()->height(); startTile.ry()++) {
+        for (; startPos.y() <= rect.bottom() && (startTile.y() < map()->height() || map()->infinite()); startTile.ry()++) {
             QPoint rowTile = startTile;
             QPoint rowPos = startPos;
 
             if (p.doStaggerY(startTile.y()))
                 rowPos.rx() += p.columnWidth;
 
-            for (; rowPos.x() <= rect.right() && rowTile.x() < map()->width(); rowTile.rx()++) {
+            for (; rowPos.x() <= rect.right() && (rowTile.x() < map()->width() || map()->infinite()); rowTile.rx()++) {
                 lines.append(QLine(rowPos + oct[0], rowPos + oct[1]));
                 lines.append(QLine(rowPos + oct[1], rowPos + oct[2]));
                 lines.append(QLine(rowPos + oct[3], rowPos + oct[4]));
@@ -288,27 +290,27 @@ void HexagonalRenderer::drawTileLayer(QPainter *painter,
     CellRenderer renderer(painter, CellRenderer::HexagonalCells);
 
     if (p.staggerX) {
-        startTile.setX(qMax(-1, startTile.x()));
-        startTile.setY(qMax(-1, startTile.y()));
+        if (!map()->infinite()) {
+            startTile.setX(qMax(-1, startTile.x()));
+            startTile.setY(qMax(-1, startTile.y()));
+        }
 
         startPos = tileToScreenCoords(startTile + layer->position()).toPoint();
         startPos.ry() += p.tileHeight;
 
         bool staggeredRow = p.doStaggerX(startTile.x() + layer->x());
 
-        for (; startPos.y() < rect.bottom() && startTile.y() < layer->height();) {
+        for (; startPos.y() < rect.bottom() && (startTile.y() < layer->height() || map()->infinite());) {
             QPoint rowTile = startTile;
             QPoint rowPos = startPos;
 
-            for (; rowPos.x() < rect.right() && rowTile.x() < layer->width(); rowTile.rx() += 2) {
-                if (layer->contains(rowTile)) {
-                    const Cell &cell = layer->cellAt(rowTile);
+            for (; rowPos.x() < rect.right() && (rowTile.x() < layer->width() || map()->infinite()); rowTile.rx() += 2) {
+                const Cell &cell = layer->cellAt(rowTile);
 
-                    if (!cell.isEmpty()) {
-                        Tile *tile = cell.tile();
-                        QSize size = tile ? tile->size() : map()->tileSize();
-                        renderer.render(cell, rowPos, size, CellRenderer::BottomLeft);
-                    }
+                if (!cell.isEmpty()) {
+                    Tile *tile = cell.tile();
+                    QSize size = tile ? tile->size() : map()->tileSize();
+                    renderer.render(cell, rowPos, size, CellRenderer::BottomLeft);
                 }
 
                 rowPos.rx() += p.tileWidth + p.sideLengthX;
@@ -328,8 +330,10 @@ void HexagonalRenderer::drawTileLayer(QPainter *painter,
             startPos.ry() += p.rowHeight;
         }
     } else {
-        startTile.setX(qMax(0, startTile.x()));
-        startTile.setY(qMax(0, startTile.y()));
+        if (!map()->infinite()) {
+            startTile.setX(qMax(0, startTile.x()));
+            startTile.setY(qMax(0, startTile.y()));
+        }
 
         startPos = tileToScreenCoords(startTile + layer->position()).toPoint();
         startPos.ry() += p.tileHeight;
@@ -338,14 +342,14 @@ void HexagonalRenderer::drawTileLayer(QPainter *painter,
         if (p.doStaggerY(startTile.y() + layer->y()))
             startPos.rx() -= p.columnWidth;
 
-        for (; startPos.y() < rect.bottom() && startTile.y() < layer->height(); startTile.ry()++) {
+        for (; startPos.y() < rect.bottom() && (startTile.y() < layer->height() || map()->infinite()); startTile.ry()++) {
             QPoint rowTile = startTile;
             QPoint rowPos = startPos;
 
             if (p.doStaggerY(startTile.y() + layer->y()))
                 rowPos.rx() += p.columnWidth;
 
-            for (; rowPos.x() < rect.right() && rowTile.x() < layer->width(); rowTile.rx()++) {
+            for (; rowPos.x() < rect.right() && (rowTile.x() < layer->width() || map()->infinite()); rowTile.rx()++) {
                 const Cell &cell = layer->cellAt(rowTile);
 
                 if (!cell.isEmpty()) {
--- a/src/libtiled/hexagonalrenderer.cpp
+++ b/src/libtiled/hexagonalrenderer.cpp
@@ -67,23 +67,52 @@ QRect HexagonalRenderer::mapBoundingRect() const
 {
     const RenderParams p(map());
 
+    QRect mapBounds;
+
+    LayerIterator iterator(map());
+    while (Layer *layer = iterator.next()) {
+        if (TileLayer *tileLayer = dynamic_cast<TileLayer*>(layer))
+            mapBounds = mapBounds.united(tileLayer->bounds());
+    }
+
+    if (mapBounds.size() == QSize(0, 0))
+        mapBounds.setSize(QSize(1, 1));
+
     // The map size is the same regardless of which indexes are shifted.
     if (p.staggerX) {
-        QSize size(map()->width() * p.columnWidth + p.sideOffsetX,
-                   map()->height() * (p.tileHeight + p.sideLengthY));
+        if (!map()->infinite()) {
+            QSize size(map()->width() * p.columnWidth + p.sideOffsetX,
+                       map()->height() * (p.tileHeight + p.sideLengthY));
 
-        if (map()->width() > 1)
-            size.rheight() += p.rowHeight;
+            if (map()->width() > 1)
+                size.rheight() += p.rowHeight;
 
-        return QRect(0, 0, size.width(), size.height());
+            return QRect(0, 0, size.width(), size.height());
+        }
+
+        QSize size(mapBounds.width() * p.columnWidth + p.sideOffsetX,
+                   mapBounds.height() * (p.tileHeight + p.sideLengthY));
+        QPoint origin(mapBounds.x() * p.columnWidth + p.sideOffsetX,
+                      mapBounds.y() * (p.tileHeight + p.sideLengthY));
+
+        return QRect(origin, size);
     } else {
-        QSize size(map()->width() * (p.tileWidth + p.sideLengthX),
-                   map()->height() * p.rowHeight + p.sideOffsetY);
+        if (!map()->infinite()) {
+            QSize size(map()->width() * (p.tileWidth + p.sideLengthX),
+                       map()->height() * p.rowHeight + p.sideOffsetY);
 
-        if (map()->height() > 1)
-            size.rwidth() += p.columnWidth;
+            if (map()->height() > 1)
+                size.rwidth() += p.columnWidth;
 
-        return QRect(0, 0, size.width(), size.height());
+            return QRect(0, 0, size.width(), size.height());
+        }
+
+        QSize size(mapBounds.width() * (p.tileWidth + p.sideLengthX),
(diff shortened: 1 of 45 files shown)


bjorn:
I think this will not be good for performance. Hence I was suggesting, to keep track of a bounding `QRect` and only update it based on which chunks have been allocated. There is no need to keep track of the exact region nor do we need it to be accurate down to the tile.

ketanhwr:
Alright, so the `region()` function should be changed back to what it was prevously?

ketanhwr:
But why do you think that this won't be good for performance? In fact I think it's better now. Because earlier whenever the `region()` function was called, it traversed through all the cells but now it directly returns a `QRegion` and updates it when `setCell` is called.

bjorn:
> But why do you think that this won't be good for performance? In fact I think it's better now. Because earlier whenever the region() function was called, it traversed through all the cells but now it directly returns a QRegion and updates it when setCell is called.

Yes, but when does the `region` function get called? I think this only happens for the tile layer used for the active brush. Maintaining the region for all tile layers of a map is largely a wasted effort, and it really isn't cheap to add/remove many 1x1 QRects to a QRegion (which is why the `TileLayer::region(...)` (now `Chunk::region`) function uses a special optimization to avoid too many operations on the QRegion).

ketanhwr:
So instead of maintaining a `mRegion`, you're suggesting that I should maintain a `mBounds`, right?

bjorn:
> So instead of maintaining a mRegion, you're suggesting that I should maintain a mBounds, right?

Yes. In addition, I was suggesting that you update `mBounds` with chunk-granularity instead of tile-granularity.

src/libtiled/tilelayer.h
```diff
@@ -568,6 +571,19 @@ inline void TileLayer::setSize(const QSize &size)
...
 }
 
+inline QRect TileLayer::bounds() const
+{
+    QRect bound = mBounds.boundingRect();
+
+    int x = std::min(0, bound.x());
+    int y = std::min(0, bound.y());
+    int width = std::max(mWidth, bound.width());
+    int height = std::max(mHeight, bound.height());
```
bjorn: Why is this? When a layer is empty, by all means its bounds should be empty as well. That is actually a nice optimization for the code using the `bounds` method.

src/libtiled/tilelayer.h
```diff
@@ -568,6 +571,19 @@ inline void TileLayer::setSize(const QSize &size)
...
+
+    return QRect(x, y, width, height).translated(mX, mY);
```
bjorn: Since in your only usage of `bounds` so far, you're translating the rectangle back to local coordinates, I wonder if it may be better to just return local coordinates instead?

src/libtiled/tilelayer.h
```diff
@@ -513,6 +515,7 @@ class TILEDSHARED_EXPORT TileLayer : public Layer
     int mHeight;
     Cell mEmptyCell;
     QHash<QPoint, Chunk> mChunks;
+    QRegion mBounds;
```
bjorn: Since we only care about the bounding rectangle, why is this a `QRegion` instead of a `QRect`? I don't currently see a use-case for using the region.

ketanhwr: I previously tried with `QRect`, but we cannot do arithmetic operations on `QRect`. Although we can convert `QRect` to `QRegion`, then add a chunk and then use `QRegion::boundingRect` but I guess that would have the same complexity.

src/libtiled/tilelayer.h
ketanhwr: I tried using `mBounds.boundintRect()` previously, but that wasn't working for me.

src/libtiled/tilelayer.h
ketanhwr: I didn't get this one?

src/libtiled/tilelayer.h
bjorn: You only need to extend the rectangle in some cases. It's easy to do, but then, Qt even has `QRect::united`. Not going through `QRegion` will be a lot more efficient.

src/libtiled/tilelayer.h
bjorn: I guess the must have been a problem elsewhere. Please try to take it out again and find out what's going wrong.

src/libtiled/tilelayer.h
bjorn: I was talking about the `.translated(mX, mY)` part. Because it's there, you're doing `layer->bounds().translated(-(layer->x()), -(layer->y()))` in the renderer. You could leave out both calls to `translate` for the same result.

ketanhwr: But for `BrushItem` it draws the selection on the basis of `x()` and `y()`.

ketanhwr: So if both the `translate` are removed, then `BrushItem` won't work.

bjorn: I don't see `TileLayer::bounds` being used by `BrushItem`. If you do use it there and it wants the bounds to be translated, then yes you should either leave this in there or add a translation in `BrushItem`.

src/libtiled/tilelayer.cpp
```diff
@@ -170,8 +170,15 @@ void Tiled::TileLayer::setCell(int x, int y, const Cell &cell)
...
-    if (cell.isEmpty() && !findChunk(x, y))
-        return;
+    if (!findChunk(x, y)) {
+        if (cell.isEmpty())
+            return;
+        else
+            mBounds = mBounds.united(QRect(x - (x & CHUNK_MASK),
+                                           y - (y & CHUNK_MASK),
+                                           CHUNK_SIZE,
+                                           CHUNK_SIZE));
```
Ablu: if the line in the else is multiple lines long please use `{}` around it. And since the `else` then has it, the `if`-part also should have it

src/libtiled/tilelayer.h
```diff
@@ -568,6 +571,16 @@ inline void TileLayer::setSize(const QSize &size)
     mHeight = size.height();
 }
 
+inline QRect TileLayer::bounds() const
+{
+    int x = std::min(0, mBounds.x());
+    int y = std::min(0, mBounds.y());
+    // int width = std::max(mWidth, mBounds.width());
+    // int height = std::max(mHeight, mBounds.height());
```
Ablu: I guess you want to remove those?

Ablu:
you broke a test apparently... Due to this the travis test failed.

src/libtiled/tilelayer.cpp
```diff
@@ -400,6 +410,7 @@ void TileLayer::rotate(RotateDirection direction)
     mWidth = newWidth;
     mHeight = newHeight;
     mChunks = newLayer->mChunks;
+    updateBounds();
```
bjorn: Why not `mBounds = newLayer->mBounds;`? (and of course, passing `mX` and `mY` to constructor when creating `newLayer`. This goes for all these functions creating a new layer.

src/libtiled/tilelayer.h
```diff
@@ -568,6 +571,16 @@ inline void TileLayer::setSize(const QSize &size)
     mHeight = size.height();
 }
 
+inline QRect TileLayer::bounds() const
+{
+    int x = std::min(0, mBounds.x());
+    int y = std::min(0, mBounds.y());
+    int width = std::max(mWidth, mBounds.width());
+    int height = std::max(mHeight, mBounds.height());
```
bjorn: Please remove this and document where it is causing issues, so that we can find a way to fix that.

src/libtiled/orthogonalrenderer.cpp
```diff
@@ -225,10 +225,11 @@ void OrthogonalRenderer::drawTileLayer(QPainter *painter,
     const QPointF layerPos(layer->x() * tileWidth,
                            layer->y() * tileHeight);
 
-    int startX = 0;
-    int startY = 0;
-    int endX = layer->width() - 1;
-    int endY = layer->height() - 1;
+    QRect bounds = layer->bounds().translated(-(layer->x()), -(layer->y()));
```
bjorn: Can be written shorter as `.translated(-layer->position())`

src/libtiled/tilelayer.h
ketanhwr:
```
QRegion TilePainter::paintableRegion(const QRegion &region) const
{
    const QRegion bounds = QRegion(mTileLayer->bounds());
    QRegion intersection = bounds.intersected(region);

    const QRegion &selection = mMapDocument->selectedArea();
    if (!selection.isEmpty())
        intersection &= selection;

    return intersection;
}
```

Due to this, if we directly return `mBounds` then we won't be able to paint the tile layer.

src/libtiled/tilelayer.h
```diff
@@ -349,7 +349,9 @@ class TILEDSHARED_EXPORT TileLayer : public Layer
     /**
      * Returns the bounds of this layer.
      */
-    QRect bounds() const { return QRect(mX, mY, mWidth, mHeight); }
+    QRect bounds() const { return mBounds.translated(mX, mY); }
```
bjorn: I suggested to use `bounds` in the map renderers, but due to an assert this will currently only work if this function is implemented as:
```c++
QRect bounds() const { return mBounds.translated(mX, mY).intersected(rect()); }
```
That is because `mBounds` may extend beyond the limits imposed by the layer size due to growing it chunk-by-chunk. I think that is a fine implementation for now, and the intersection can be taken out again when we actually allow layers to grow (and take out the assert in `cellAt`).

ketanhwr:
I'm not able to understand why the build fails :confused:

src/libtiled/orthogonalrenderer.cpp
```diff
@@ -225,10 +270,11 @@ void OrthogonalRenderer::drawTileLayer(QPainter *painter,
...
 
-    int startX = 0;
-    int startY = 0;
-    int endX = layer->width() - 1;
-    int endY = layer->height() - 1;
+    QRect bounds = layer->bounds().translated(-layer->position());
+    int startX = bounds.left();
+    int startY = bounds.top();
+    int endX = bounds.right();
+    int endY = bounds.bottom();
```
bjorn: I believe this should still have the -1:
```c++
int endX = bounds.right() - 1;
int endY = bounds.bottom() - 1;
```
Because the `right` and `bottom` of a QRect are inclusive.

src/tiled/mapscene.cpp
```diff
@@ -305,7 +306,11 @@ void MapScene::updateSceneRect()
                      margins.bottom());
 
     setSceneRect(sceneRect);
-    mDarkRectangle->setRect(sceneRect);
+
+    if (mMapDocument->map()->infinite())
+        mDarkRectangle->setRect(QRect());
```
bjorn: We'll need to find some way to still darken the view, for the "Highlight Current Layer" feature to work on infinite maps.

src/tiled/newmapdialog.ui
```diff
@@ -234,6 +234,13 @@
       <item row="2" column="1">
        <widget class="QComboBox" name="renderOrder"/>
       </item>
+      <item row="3" column="1">
+       <widget class="QCheckBox" name="mapInfinite">
+        <property name="text">
+         <string>Automatically Resize Map</string>
```
bjorn: Please see if this checkbox fits in the "Map Size" box, and disable the width/height inputs and hide the pixel size display for infinite maps (I think the map size should just be 0,0).

src/libtiled/orthogonalrenderer.cpp
ketanhwr: It was previously iterating upto `tileLayer->width() - 1`. So if `- 1` is added, then it doesn't render the last row/column.

src/tiled/stampbrush.cpp
```diff
@@ -271,8 +271,10 @@ void StampBrush::endCapture()
 
     // Intersect with the layer and translate to layer coordinates
     QRect captured = capturedArea();
-    captured &= QRect(tileLayer->x(), tileLayer->y(),
-                      tileLayer->width(), tileLayer->height());
+
+    if (!tileLayer->map()->infinite())
+        captured &= QRect(tileLayer->x(), tileLayer->y(),
+                          tileLayer->width(), tileLayer->height());
```
bjorn: Do we need this intersection at all?

src/libtiled/orthogonalrenderer.cpp
bjorn: Ah, right, precisely because QRect is inclusive, we should indeed not do -1.

src/tiled/offsetmapdialog.cpp
```diff
@@ -80,6 +80,30 @@ QRect OffsetMapDialog::affectedBoundingRect() const
     switch (boundsSelection()) {
     case WholeMap:
         boundingRect = QRect(QPoint(0, 0), mMapDocument->map()->size());
+
+        if (mMapDocument->map()->infinite()) {
+            LayerIterator iterator(mMapDocument->map());
+
+            switch (layerSelection()) {
+            case AllVisibleLayers:
```
bjorn: I'm not sure if taking into account the bounds of the layers is any use at all here. It is at least totally unpredictable from the user's point of view, which makes it basically useless. Probably the "Wrap" checkboxes should be disabled for "Whole Map" when the map is infinite, and the bounds can just be the united bounds of all layers in all cases, adjusted by the offset to make sure no tiles are lost.

src/tiled/terrainbrush.cpp
```diff
@@ -231,7 +231,9 @@ void TerrainBrush::doPaint(bool mergeable)
     if (!tileLayer->isUnlocked())
         return;
 
-    if (!tileLayer->bounds().intersects(stamp->bounds()))
+    QRect layerRect = (mapDocument()->map()->infinite()) ? tileLayer->bounds() : tileLayer->rect();
+
+    if (!layerRect.intersects(stamp->bounds()))
```
bjorn: You should never to this intersects check for infinite maps.

src/tiled/thumbnailrenderer.cpp
```diff
@@ -98,7 +98,7 @@ static QRectF cellRect(const MapRenderer &renderer,
 static QRect computeMapRect(const MapRenderer &renderer)
 {
     // Start with the basic map size
-    QRectF rect(QPointF(0, 0), renderer.mapSize());
+    QRectF rect(QPointF(0, 0), renderer.mapBoundingRect().size());
```
bjorn: I think this should just be initialized to `renderer.mapBoundingRect()`, not just its size.

src/tiled/tilepainter.cpp
```diff
@@ -197,12 +199,26 @@ static QRegion fillRegion(const TileLayer *layer, QPoint fillOrigin,
     if (!layer->contains(fillOrigin))
```
bjorn: This check should be disabled for infinite maps (once `contains` is back to its original form).

src/tiled/tilepainter.cpp
```diff
@@ -197,12 +199,26 @@ static QRegion fillRegion(const TileLayer *layer, QPoint fillOrigin,
...
+
     // Cache cell that we will match other cells against
     const Cell matchCell = layer->cellAt(fillOrigin);
 
     // Grab layer dimensions for later use.
-    const int layerWidth = layer->width();
-    const int layerHeight = layer->height();
+    QRect bounds = layer->bounds().translated(-layer->position());
+    if (!layer->map()->infinite())
+        bounds = layer->rect();
```
bjorn: Translated by `-layer->position()` as well, right?

src/tmxrasterizer/tmxrasterizer.cpp
```diff
@@ -101,7 +101,7 @@ int TmxRasterizer::render(const QString &mapFileName,
         break;
     }
 
-    QSize mapSize = renderer->mapSize();
+    QSize mapSize = renderer->mapBoundingRect().size();
```
bjorn: Will need to store the bounding rect and translate the painter based on the `topLeft` of it.

tests/staggeredrenderer/test_staggeredrenderer.cpp
```diff
@@ -53,10 +53,10 @@ void test_StaggeredRenderer::mapSize()
     StaggeredRenderer renderer(mMap);
 
     TileLayer *tileLayer = static_cast<TileLayer*>(mMap->layerAt(0));
-    QSize mapSize = renderer.mapSize();
+    QSize mapSize = renderer.mapBoundingRect().size();
```
bjorn: To make the test a little more complete, just take the bounding rect and compare it to a `QRect(0, 0, 10 * 64 + 32, 10 * 16 + 16)`.

src/libtiled/tilelayer.cpp
```diff
@@ -191,7 +205,12 @@ void Tiled::TileLayer::setCell(int x, int y, const Cell &cell)
 
 TileLayer *TileLayer::copy(const QRegion &region) const
 {
-    const QRegion area = region.intersected(QRect(0, 0, width(), height()));
+    QRegion area = region.intersected(QRect(0, 0, width(), height()));
+    if (map()) {
+        if (map()->infinite())
```
ketanhwr: I'm not entirely sure if that will be a good idea.

src/tiled/mainwindow.cpp
```diff
@@ -1178,8 +1178,23 @@ void MainWindow::resizeMap()
...
+        if (TileLayer *tileLayer = dynamic_cast<TileLayer*>(layer)) {
+            QRect bounds = tileLayer->bounds();
+            QSize layerSize(bounds.width(), bounds.height());
+
+            mapSize = mapSize.expandedTo(layerSize);
+        }
+    }
+
+    if (mapSize == QSize(0, 0) || !map->infinite())
+        mapSize = map->size();
```
ketanhwr: Oh! So this was the reason resizing wasn't working :man_facepalming:

src/tiled/thumbnailrenderer.cpp
ketanhwr: How exactly do I test the Thumbnail Renderer?

Ablu: the thumbnail renderer renders the tiles for the tile stamp dock. So you can simply add some tiles there to test it.

Ablu:
Travis seems to have a compiling issue:

~~~
test_staggeredrenderer.cpp:56:11: error: no viable conversion from 'QRect' to
      'QSize'
    QSize mapBoundingRect = renderer.mapBoundingRect();
          ^                 ~~~~~~~~~~~~~~~~~~~~~~~~~~
~~~

src/tiled/thumbnailrenderer.cpp
bjorn: `ThumbnailRenderer` should really be unified with `MiniMapRenderer`, but that's something for another time...

src/libtiled/orthogonalrenderer.cpp
```diff
@@ -188,12 +206,17 @@ void OrthogonalRenderer::drawGrid(QPainter *painter, const QRectF &rect,
     if (tileWidth <= 0 || tileHeight <= 0)
         return;
 
-    const int startX = qMax(0, (int) (rect.x() / tileWidth) * tileWidth);
-    const int startY = qMax(0, (int) (rect.y() / tileHeight) * tileHeight);
-    const int endX = qMin(qCeil(rect.right()),
-                          map()->width() * tileWidth + 1);
-    const int endY = qMin(qCeil(rect.bottom()),
-                          map()->height() * tileHeight + 1);
+    int startX = ((int) (rect.x() / tileWidth - 1)) * tileWidth;
```
bjorn: I guess the -1 is to make it work in case of negative coordinates, but it means an extra column of tiles will be rendered in positive coordinate space. A real fix would be to use `qFloor(rect.x() / tileWidth) * tileWidth`.

src/libtiled/orthogonalrenderer.cpp
```diff
@@ -188,12 +206,17 @@ void OrthogonalRenderer::drawGrid(QPainter *painter, const QRectF &rect,
...
+    int startY = ((int) (rect.y() / tileHeight - 1)) * tileHeight;
+    int endX = qCeil(rect.right());
+    int endY = qCeil(rect.bottom());
+
+    if (!map()->infinite()) {
+        startX = qMax(0, (int) (rect.x() / tileWidth) * tileWidth);
```
bjorn: Here you should reuse the value of `startX`:
```c++
startX = qMax(0, startX);
```

tests/staggeredrenderer/test_staggeredrenderer.cpp
```diff
@@ -52,12 +52,9 @@ void test_StaggeredRenderer::mapSize()
 {
     StaggeredRenderer renderer(mMap);
 
-    TileLayer *tileLayer = static_cast<TileLayer*>(mMap->layerAt(0));
-    QSize mapSize = renderer.mapSize();
+    QRect mapBoundingRect = renderer.mapBoundingRect();
 
-    QCOMPARE(mapSize, QSize(10 * 64 + 32, 10 * 16 + 16));
-    QCOMPARE(renderer.boundingRect(tileLayer->bounds()),
-             QRect(QPoint(), mapSize));
```
bjorn: Why are you removing the test of `StaggeredRenderer::boundingRect`? It should be possible to just change it to:
```c++
QCOMPARE(renderer.boundingRect(tileLayer->rect()), mapBoundingRect);
```

src/tiled/terrainbrush.cpp
```diff
@@ -204,7 +204,7 @@ void TerrainBrush::capture()
 
     const QPoint position = tilePosition() - tileLayer->position();
 
-    if (!tileLayer->contains(position))
+    if (!tileLayer->contains(position) && !mapDocument()->map()->infinite())
```
bjorn: This check is no longer necessary. I think capturing outside of the map can simply capture "no terrain", also for finite maps.

src/tiled/terrainbrush.cpp
```diff
@@ -373,7 +373,7 @@ void TerrainBrush::updateBrush(QPoint cursorPos, const QVector<QPoint> *list)
     }
 
     // if the cursor is outside of the map, bail out
-    if (!currentLayer->contains(cursorPos)) {
+    if (!currentLayer->contains(cursorPos) && !mapDocument()->map()->infinite()) {
```
bjorn: Contains check should not be needed anymore even for finite maps.

src/tiled/terrainbrush.cpp
```diff
@@ -395,7 +395,7 @@ void TerrainBrush::updateBrush(QPoint cursorPos, const QVector<QPoint> *list)
         transitionList.reserve(list->size());
         for (QPoint p : *list) {
             p -= layerPosition;
-            if (currentLayer->contains(p))
+            if (currentLayer->contains(p) || mapDocument()->map()->infinite())
```
bjorn: Contains check should not be needed anymore even for finite maps.

src/tiled/tilepainter.cpp
```diff
@@ -68,7 +70,7 @@ Cell TilePainter::cellAt(int x, int y) const
     const int layerX = x - mTileLayer->x();
     const int layerY = y - mTileLayer->y();
 
-    if (!mTileLayer->contains(layerX, layerY))
+    if (!mTileLayer->contains(layerX, layerY) && !mMapDocument->map()->infinite())
```
bjorn: Check is not necessary at all anymore. If the layer does not contain this position, it will anyway return an empty cell.

src/tiled/tilepainter.cpp
```diff
@@ -194,15 +196,29 @@ static QRegion fillRegion(const TileLayer *layer, QPoint fillOrigin,
     QRegion fillRegion;
 
     // Silently quit if parameters are unsatisfactory
-    if (!layer->contains(fillOrigin))
+    if (!layer->contains(fillOrigin) && !layer->map()->infinite())
+        return fillRegion;
+
+    if (layer->map()->infinite() && !layer->bounds().contains(fillOrigin))
```
bjorn: Please combine the above checks with getting the bounds below:
```c++
QRect bounds = layer->map()->infinite() ? layer->bounds() : layer->rect();
if (!bounds.contains(fillOrigin))
    return fillRegion;
bounds.translate(-layer->position());
```

src/tmxrasterizer/tmxrasterizer.cpp
```diff
@@ -101,7 +101,8 @@ int TmxRasterizer::render(const QString &mapFileName,
...
+    QPoint mapOffset = renderer->mapBoundingRect().topLeft();
```
bjorn: Please don't call `mapBoundingRect` twice.

src/libtiled/hexagonalrenderer.cpp
```diff
@@ -338,14 +342,14 @@ void HexagonalRenderer::drawTileLayer(QPainter *painter,
         if (p.doStaggerY(startTile.y() + layer->y()))
             startPos.rx() -= p.columnWidth;
 
-        for (; startPos.y() < rect.bottom() && startTile.y() < layer->height(); startTile.ry()++) {
+        for (; startPos.y() < rect.bottom() && (startTile.y() < layer->height() || map()->infinite()); startTile.ry()++) {
```
bjorn: Please still optimize this to use an `endY` that is either `layer->height()` or `layer->bounds()`, same for X.

src/libtiled/isometricrenderer.cpp
```diff
@@ -242,7 +268,7 @@ void IsometricRenderer::drawTileLayer(QPainter *painter,
         QPoint columnItr = rowItr;
 
         for (int x = startPos.x(); x < rect.right(); x += tileWidth) {
-            if (layer->contains(columnItr)) {
+            if (map()->infinite() || layer->contains(columnItr)) {
```
bjorn: You can drop both these checks and rely only on `!cell.isEmpty()` below.

src/libtiled/hexagonalrenderer.cpp
```diff
@@ -289,6 +318,9 @@ void HexagonalRenderer::drawTileLayer(QPainter *painter,
 
     CellRenderer renderer(painter, CellRenderer::HexagonalCells);
 
+    const int endX = map()->infinite() ? layer->bounds().width() : layer->width();
```
bjorn: I think the `endX` should be `layer->bounds().right() - layer->x() + 1` in case of infinite maps. Similar for Y.

src/libtiled/hexagonalrenderer.cpp
```diff
@@ -300,11 +332,11 @@ void HexagonalRenderer::drawTileLayer(QPainter *painter,
 
         bool staggeredRow = p.doStaggerX(startTile.x() + layer->x());
 
-        for (; startPos.y() < rect.bottom() && (startTile.y() < layer->height() || map()->infinite());) {
+        for (; startPos.y() < rect.bottom() && startTile.y() < endY;) {
             QPoint rowTile = startTile;
             QPoint rowPos = startPos;
 
-            for (; rowPos.x() < rect.right() && (rowTile.x() < layer->width() || map()->infinite()); rowTile.rx() += 2) {
+            for (; rowPos.x() < rect.right() && endX; rowTile.rx() += 2) {
```
bjorn: :-(

src/libtiled/hexagonalrenderer.cpp
```diff
@@ -67,23 +67,52 @@ QRect HexagonalRenderer::mapBoundingRect() const
...
+    }
+
+    if (mapBounds.size() == QSize(0, 0))
+        mapBounds.setSize(QSize(1, 1));
+
     // The map size is the same regardless of which indexes are shifted.
     if (p.staggerX) {
-        QSize size(map()->width() * p.columnWidth + p.sideOffsetX,
-                   map()->height() * (p.tileHeight + p.sideLengthY));
+        if (!map()->infinite()) {
```
bjorn: Please consider that infinite and finite maps are not so different that you couldn't use the same code for both, just using the bounds:
```c++
QRect mapBounds;

if (map()->inifinite()) {
    // unite with the layer bounds
} else {
    mapBounds = QRect(0, 0, map()->width(), map()->height());
}

// Calculate bounding rect based on bounds (no need for checking infinite anymore)
```

src/libtiled/hexagonalrenderer.cpp
```diff
@@ -67,23 +67,52 @@ QRect HexagonalRenderer::mapBoundingRect() const
...
+            QSize size(map()->width() * p.columnWidth + p.sideOffsetX,
+                       map()->height() * (p.tileHeight + p.sideLengthY));
+
+            if (map()->width() > 1)
+                size.rheight() += p.rowHeight;
```
bjorn: I believe this line is also needed for infinite maps. `rowHeight` needs to be added because of the staggering. Same for `columnWidth` below.

src/libtiled/hexagonalrenderer.cpp
```diff
@@ -67,23 +67,52 @@ QRect HexagonalRenderer::mapBoundingRect() const
...
+    QRect mapBounds;
+
+    LayerIterator iterator(map());
+    while (Layer *layer = iterator.next()) {
+        if (TileLayer *tileLayer = dynamic_cast<TileLayer*>(layer))
+            mapBounds = mapBounds.united(tileLayer->bounds());
+    }
+
+    if (mapBounds.size() == QSize(0, 0))
+        mapBounds.setSize(QSize(1, 1));
```
bjorn: Why is this?

ketanhwr: We discussed this before. If the size is set to (0, 0) then after creating a new map, the 'weird' scrolling problem starts.

src/libtiled/hexagonalrenderer.cpp
ketanhwr: :man_facepalming: :man_facepalming: :man_facepalming:

src/libtiled/hexagonalrenderer.cpp
ketanhwr: But not for `layer->width()` ?

bjorn: What about `layer->width()`? My point is that `endX` should be the right-most X coordinate (+1 because it's one beyond that), like it is for finite maps. But you just took `bounds().width()`, which does not take into account that `bounds().x()` isn't always 0.

src/libtiled/hexagonalrenderer.cpp
```diff
@@ -339,14 +365,14 @@ void HexagonalRenderer::drawTileLayer(QPainter *painter,
         if (p.doStaggerY(startTile.y() + layer->y()))
             startPos.rx() -= p.columnWidth;
 
-        for (; startPos.y() < rect.bottom() && startTile.y() < layer->height(); startTile.ry()++) {
+        for (; startPos.y() < rect.bottom() && endY; startTile.ry()++) {
```
bjorn: Please fix all cases.

src/libtiled/hexagonalrenderer.cpp
ketanhwr: so instead of `layer->x()`, it should be `layer->bounds().x()`, right?

src/libtiled/hexagonalrenderer.cpp
bjorn: Could we work around that problem in the `MapScene`? It's not related to the renderers.

src/libtiled/hexagonalrenderer.cpp
bjorn: No, the `- layer->x()` part is to change the bounds to local coordinates, since the `bounds()` include the layer's position. Subtracting `layer->bounds().x()` would defeat the point of using `layer->bounds().right()`.

PR #1652, opened 2017-07-14
Creating template instances

thabetx:
This pull request is part of the Reusable Object Templates project.

Added a template instance creation tool similar to the insert tile object tool. Also enabled dragging and dropping from the template dock into the map.

Right now, the functionality of the project can be tested without the need of manually editing files.


src/libtiled/map.cpp
```diff
@@ -298,6 +298,18 @@ bool Map::addTemplateGroup(TemplateGroup *templateGroup)
...
 
+void Map::addTemplateGroupReference(const TemplateGroup *templateGroup)
+{
+//    mTemplateGroupReferences[templateGroup]++;
+    mTemplateGroupReferences.insert(templateGroup, mTemplateGroupReferences.value(templateGroup) + 1);
+}
+
+void Map::removeTemplateGroupReference(const TemplateGroup *templateGroup)
+{
+//    mTemplateGroupReferences[templateGroup]--;
```
bjorn: Did this not work for some reason?

Btw, while it makes sense to automatically remove template group references, this is only relevant when saving the map. So rather than keeping reference counts, the map writer could also just iterate over all the objects to determine the list of template group references that need to be written. I would prefer that approach.

src/tiled/createobjecttool.h
```diff
@@ -74,6 +76,7 @@ public slots:
     MapObjectItem *mNewMapObjectItem;
     MapObjectItem *mOverlayPolygonItem;
     Tile *mTile;
+    ObjectTemplate *mObjectTemplate;
```
bjorn: Why is this a member of `CreateObjectTool` rather than `CreateTemplateTool`?

The same question could be asked about `mTile` of course. That one is there for historical reasons and can be moved to `CreateTileObjectTool` since change 33e048b2ddc9722fbf491562fb9bc8883a34f78c.

src/libtiled/map.cpp
thabetx: It worked fine but I wasn't sure which style to use. I will try the suggested approach it should be simpler.

src/tiled/templatesdock.h
```diff
@@ -43,6 +45,8 @@ class TemplatesDock : public QDockWidget
     TemplatesDock(QWidget *parent = nullptr);
     ~TemplatesDock();
 
+    TemplatesView *templatesView() const;
```
thabetx: In that case the view emits the change then the dock emits it again, right?

bjorn: Yes, you can directly connect one signal to another signal to achieve this.

src/tiled/createtemplatetool.cpp
```diff
@@ -0,0 +1,98 @@
...
+    setIcon(icon);
+    Utils::setThemeIcon(this, "insert-image");
+    languageChanged();
+}
+
+void CreateTemplateTool::mouseMovedWhileCreatingObject(const QPointF &pos, Qt::KeyboardModifiers modifiers)
+{
+    const MapRenderer *renderer = mapDocument()->renderer();
+
+    const QSize imgSize = mNewMapObjectItem->mapObject()->cell().tile()->size();
```
thabetx: I have a tendency to test only tile objects and forget about the others.

I tried using the bounding rectangle dimensions to calculate the shift instead of the tile, but when clicking, each type has a different starting point, the tile objects uses the bottom left, ellipses, rectangles and text objects will use the top left, while a polyline or a polygon will use their first point.

The easy solution would be just use `pos` without any shifting. but I think we can handle them to always be centered using the bounding rectangle but with different shift for each type. Drag and drop uses the default starting point so it might be changed as well to be centered.

bjorn: I think it will be fine for now to not bother with the centering at all. On the upside, it will give a clear indication to the user as to where the object is actually positioned, which will be beneficial also when this anchor point eventually becomes configurable.

thabetx:
The warnings were not relevant to the pull request so I left them. I will fix them in the next PR.

thabetx:
I think I will just add them to suppress the warnings without adding any functionality as the new object types are not used in the property browser.

bjorn:
> I think I will just add them to suppress the warnings without adding any functionality as the new object types are not used in the property browser.

Sounds reasonable.

PR #1658, opened 2017-07-22
Show red highlight over brush item on locked layer

ketanhwr:
This gives a clear sign to the user in case of locked layer. Also helps when the current layer is not locked but some parent layer is locked because it's not very intuitive that we won't be able to paint on the current layer.


bjorn:
I think it's a good idea. Regarding implementation, what about implementing it like:
```c++
if (!mMapDocument->currentLayer()->isUnlocked())
    mapRegion = QRegion();
```
It's a bit shorter. :-)

PR #1663, opened 2017-07-25
Editing templates

thabetx:
This PR is part of the object templates project.

The goal is to be able to edit the templates and reflect those edits in the map.

A new view is embedded into the dock to show the template and allow graphical editing, also the properties can be edited through the properties dock.

### DIFF
--- a/src/tiled/mapdocument.cpp
+++ b/src/tiled/mapdocument.cpp
@@ -924,6 +924,20 @@ void MapDocument::onLayerRemoved(Layer *layer)
     emit layerRemoved(layer);
 }
 
+void MapDocument::updateTemplateInstances(const MapObject *mapObject)
+{
+    QList<MapObject*> objectList;
+    for (ObjectGroup *group : mMap->objectGroups()) {
+        for (auto object : group->objects()) {
+            if (object->isTemplateInstance() && object->templateObject() == mapObject) {
+                object->syncWithTemplate();
+                objectList.append(object);
+            }
+        }
+    }
+    emit objectsChanged(objectList);
+}
+
 void MapDocument::deselectObjects(const QList<MapObject *> &objects)
 {
     // Unset the current object when it was part of this list of objects
--- a/src/tiled/mapeditor.cpp
+++ b/src/tiled/mapeditor.cpp
@@ -243,7 +243,8 @@ MapEditor::MapEditor(QObject *parent)
 
     connect(mWidgetStack, &QStackedWidget::currentChanged, this, &MapEditor::currentWidgetChanged);
     connect(mToolManager, &ToolManager::statusInfoChanged, this, &MapEditor::updateStatusInfoLabel);
-    connect(mTilesetDock, &TilesetDock::currentTileChanged, tileObjectsTool, &CreateObjectTool::setTile);
+    connect(mTilesetDock, &TilesetDock::currentTileChanged, mToolManager, &ToolManager::setTile);
+    connect(mTilesetDock, &TilesetDock::currentTileChanged, mTemplatesDock, &TemplatesDock::setTile);
     connect(mTilesetDock, &TilesetDock::stampCaptured, this, &MapEditor::setStamp);
     connect(mTilesetDock, &TilesetDock::localFilesDropped, this, &MapEditor::filesDroppedOnTilesetDock);
     connect(mTemplatesDock, &TemplatesDock::currentTemplateChanged, templatesTool, &CreateTemplateTool::setTemplate);
--- a/src/tiled/mapeditor.cpp
+++ b/src/tiled/mapeditor.cpp
@@ -403,9 +403,6 @@ void MapEditor::setCurrentDocument(Document *document)
     mMiniMapDock->setMapDocument(mapDocument);
 
     if (mapDocument) {
-        connect(mapDocument, &MapDocument::currentObjectChanged,
-                this, [this, mapDocument](){ mPropertiesDock->setDocument(mapDocument); });
-
         connect(mapDocument, &MapDocument::currentLayerChanged,
                 this, &MapEditor::updateLayerComboIndex);
 //        connect(mapDocument, SIGNAL(selectedAreaChanged(QRegion,QRegion)),
@@ -417,6 +414,12 @@ void MapEditor::setCurrentDocument(Document *document)
             mZoomable = mapView->zoomable();
             mZoomable->setComboBox(mZoomComboBox);
         }
+
+        connect(mCurrentMapDocument, &MapDocument::currentObjectChanged,
+                this, [this, mapDocument](){ mPropertiesDock->setDocument(mapDocument); });
+
+        connect(mapView, &MapView::focused,
+                this, [this, mapDocument](){ mPropertiesDock->setDocument(mapDocument); });
 
         mReversingProxyModel->setSourceModel(mapDocument->layerModel());
     } else {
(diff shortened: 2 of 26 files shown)


thabetx:
Sometimes after editing a template and closing Tiled, the visibility of the object inside the template is set to false, so upon opening a map again all instances of that template will be invisible. I don't know yet why this happens but I will investigate it.

thabetx:
I fixed the invisible template issue by disconnecting the applyChanges signal in the templates dock destructor.

src/tiled/templatesdock.cpp
```diff
@@ -180,4 +349,6 @@ void TemplatesView::onPressed(const QModelIndex &index)
 
     if (ObjectTemplate *objectTemplate = model->toObjectTemplate(index))
         emit currentTemplateChanged(objectTemplate);
+    else
+        emit currentTemplateChanged(nullptr);
```
bjorn: The condition is no longer needed now.

src/tiled/mapeditor.cpp
```diff
@@ -314,6 +316,10 @@ void MapEditor::addDocument(Document *document)
         if (Layer *layer = layerAtGlobalIndex(mapDocument->map(), layerIndex))
             mapDocument->setCurrentLayer(layer);
     }
+
+    // Update all templates instances in this MapDocument when a template is edited
+    // Maybe it can be optimized
```
bjorn: Maybe!? :-)

It'd be cool to get it to the point where it is fast enough that we can apply rotation and size changes live instead of connecting to the index change of the undo history.

src/tiled/objecttemplatemodel.cpp
```diff
@@ -206,6 +206,12 @@ ObjectTemplate *ObjectTemplateModel::toObjectTemplate(const QModelIndex &index)
     return nullptr;
 }
 
+void ObjectTemplateModel::save()
+{
+    for (auto document : mTemplateDocuments)
```
bjorn: Please save only the edited template document.

src/tiled/templatesdock.cpp
```diff
@@ -78,20 +125,46 @@ TemplatesDock::TemplatesDock(QWidget *parent):
...
+    setFocusPolicy(Qt::ClickFocus);
+    mTemplatesView->setFocusProxy(this);
+    mMapView->setFocusProxy(this);
 }
 
 TemplatesDock::~TemplatesDock()
 {
+    mMapScene->disableSelectedTool();
+
+    disconnect(mDummyMapDocument->undoStack(), &QUndoStack::indexChanged,
```
bjorn: You should only try this when `mDummyMapDocument` isn't null. On the other hand, this should also not be needed at all, because the deletion of the dummy document should delete the undo stack, which should disconnect this signal.

src/tiled/templatesdock.cpp
thabetx: It sends a nullptr when a template group is selected so the view gets cleared.

src/tiled/templatesdock.cpp
thabetx: I agree with disconnecting when it's not null.

I'm not sure why but without disconnecting, the tile of an object will go invisible after closing tiled.

src/tiled/templatesdock.cpp
bjorn: Yes, I just said the condition is no longer needed, because it will do that also when you just write:
```c++
ObjectTemplate *objectTemplate = model->toObjectTemplate(index);
emit currentTemplateChanged(objectTemplate);
```

thabetx: Alright, for some reason I thought you meant removing the else branch.

thabetx:
Changing the tile of an object is restricted to the map view. To allow editing the tile of an object, I'm planning to move the mTile member from the createObjectTool to the AbstractObjectTool, so it can be integrated into the right click menu when clicking on an object.

src/tiled/mapeditor.cpp
thabetx: Hmm, I had the feeling that live updating might be slow. I tried it anyway but found that the object itself wasn't updated until the undo index is changed, i.e. the object is not saved mid-edit, so calling sync for the template instances didn't do anything. So it might require some other way of syncing, I think we can leave it for now.

bjorn: That would be because the changing object is a clone of the template object, so indeed the template object needs to be synchronized to its clone each time before the instances are synced with the template object. But sure, we can wait with this change and making it would definitely also require optimization to how the instances are updated.

src/tiled/templatesdock.cpp
bjorn: Can you provide a bit more details about how to reproduce this issue? I've fiddled a bit after commenting out the line but didn't manage to run into the problem yet.

src/tiled/mapdocument.cpp
```diff
@@ -924,6 +924,18 @@ void MapDocument::onLayerRemoved(Layer *layer)
     emit layerRemoved(layer);
 }
 
+void MapDocument::updateAllTemplateInstances()
+{
+    QList<MapObject*> objectList;
+    for (ObjectGroup *group : mMap->objectGroups()) {
+        for (auto object : group->objects()) {
+            object->syncWithTemplate();
+            objectList.append(object);
```
thabetx: Alright I will only update the instances of the changed template instead instead.

src/tiled/templatesdock.cpp
```diff
@@ -175,6 +252,110 @@ void TemplatesDock::openTemplateGroup()
...
+
+        delete previousDocument;
+    }
+}
+
+void TemplatesDock::undo()
+{
+    if (mDummyMapDocument) {
+        mDummyMapDocument->undoStack()->undo();
+        emit mDummyMapDocument->selectedObjectsChanged();
```
thabetx: It updates the selection rectangle and resize handles after undoing.

src/tiled/mapeditor.cpp
```diff
@@ -386,6 +392,9 @@ void MapEditor::setCurrentDocument(Document *document)
             mZoomable->setComboBox(mZoomComboBox);
         }
 
+        connect(mapView, &MapView::focused,
+                this, [this, mapDocument](){ mPropertiesDock->setDocument(mapDocument); });
```
thabetx: Alright, I missed that. I added the currentObjectChanged, it works fine, but it won't work if the clicked object is the same, as the signal won't be emitted.

So I can connect all docks' focus signals to trigger setDocument which might not be that good.

thabetx:
I updated the PR, now the editor view size can be changed, keyboard navigation works and the template is deselected when focus is lost.

There was a bug that was solved by combining all the slots connected to the indexChanged signal into the apply changes function, so disconnecting disconnect everything correctly.

I made a separate commit for the focus changes, it was kind of tricky to make it behave as expected.

bjorn:
I've cherry-pick three of the commits into `wip/templates`, then merged `wip/templates` into `master` and then changed the target branch of this pull request to `master`. Please rebase the PR onto `master`, where you can skip these commits:

* Suppress PropertyBrowser warnings
* Update the embedded template groups code
* Display template properties when selecting instances

While testing this branch (before doing the above), I noticed the following issues:

* The "S" and "O" shortcuts are ambiguous when the Templates view is open, because it features duplicates of those tools.
* I could not get the "Insert Template" tool to work.

PR #1666, opened 2017-07-27
Templates dock enhancements

thabetx:
This PR adds some convenience features to the templates dock and template saving.

### DIFF
--- a/src/tiled/mapdocument.h
+++ b/src/tiled/mapdocument.h
@@ -76,7 +76,7 @@ public:
     ~MapDocument();
 
     bool save(const QString &fileName, QString *error = nullptr) override;
-    void saveSelectedObject(QString name, int groupIndex);
+    void saveSelectedObject(const QString &name, int groupIndex);
 
     /**
      * Loads a map and returns a MapDocument instance on success. Returns null
--- a/src/tiled/newtemplatedialog.cpp
+++ b/src/tiled/newtemplatedialog.cpp
@@ -27,13 +27,17 @@
 using namespace Tiled;
 using namespace Tiled::Internal;
 
-NewTemplateDialog::NewTemplateDialog(QList<QString> groupNames, QWidget *parent) :
+NewTemplateDialog::NewTemplateDialog(const QList<QString> &groupNames, const QString &objectName, QWidget *parent) :
     QDialog(parent),
     mUi(new Ui::NewTemplateDialog)
 {
     mUi->setupUi(this);
+
+    mUi->templateName->setText(objectName);
+
     mUi->groupsComboBox->addItems(groupNames);
-    connect(mUi->templateName, SIGNAL(textChanged(QString)), SLOT(updateOkButton()));
+    connect(mUi->templateName, &QLineEdit::textChanged,
+            this, &NewTemplateDialog::updateOkButton);
     updateOkButton();
 }
 
--- a/src/tiled/newtemplatedialog.cpp
+++ b/src/tiled/newtemplatedialog.cpp
@@ -20,8 +20,14 @@
  */
 
 #include "newtemplatedialog.h"
+#include "objecttemplatemodel.h"
+#include "preferences.h"
+#include "templatemanager.h"
+#include "tmxmapformat.h"
 #include "ui_newtemplatedialog.h"
 
+#include <QFileDialog>
+#include <QMessageBox>
 #include <QPushButton>
 
 using namespace Tiled;
@@ -34,6 +40,9 @@ NewTemplateDialog::NewTemplateDialog(const QList<QString> &groupNames, const QSt
     mUi->setupUi(this);
 
     mUi->templateName->setText(objectName);
+
+    connect(mUi->createGroupButton, &QPushButton::pressed,
+            this, &NewTemplateDialog::newTemplateGroup);
 
     mUi->groupsComboBox->addItems(groupNames);
     connect(mUi->templateName, &QLineEdit::textChanged,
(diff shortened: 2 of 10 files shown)


src/tiled/newtemplatedialog.cpp
```diff
@@ -27,13 +27,20 @@
...
 
-NewTemplateDialog::NewTemplateDialog(QList<QString> groupNames, QWidget *parent) :
+NewTemplateDialog::NewTemplateDialog(const QList<QString> &groupNames, const QString &objectName, QWidget *parent) :
     QDialog(parent),
     mUi(new Ui::NewTemplateDialog)
 {
     mUi->setupUi(this);
+
+    mObjectName = objectName;
+    mUi->templateNameCheckBox->setEnabled(!objectName.isEmpty());
```
bjorn: Why did you choose to have a checkbox? I think it will be enough to auto-fill the field with the name of the object.

src/tiled/mapdocument.h
```diff
@@ -76,7 +76,7 @@ class MapDocument : public Document
     ~MapDocument();
 
     bool save(const QString &fileName, QString *error = nullptr) override;
-    void saveSelectedObject(QString name, int groupIndex);
+    void saveSelectedObject(const QString &name, const int groupIndex);
```
bjorn: Please don't put `const` on parameters that are passed by value, like `groupIndex`. It has no relevance in regards to the signature of the function.

src/tiled/newtemplatedialog.cpp
```diff
@@ -73,9 +82,50 @@ void NewTemplateDialog::updateTemplateName()
 void NewTemplateDialog::updateOkButton()
 {
     QPushButton *okButton = mUi->buttonBox->button(QDialogButtonBox::Ok);
-    QString templateName = mUi->templateName->text();
 
+    bool noTemplateName = mUi->templateName->text().isEmpty();
     int index = mUi->groupsComboBox->currentIndex();
 
-    okButton->setEnabled(!templateName.isEmpty() && index != -1);
+    okButton->setDisabled(noTemplateName || index==-1);
```
bjorn: Coding style: Missing space around `==`

src/tiled/newtemplatedialog.cpp
```diff
@@ -73,9 +82,50 @@ void NewTemplateDialog::updateTemplateName()
...
+/**
+ * This code is duplicated from the templates dock.
+ */
+void NewTemplateDialog::newTemplateGroup()
+{
+    QString filter = TtxTemplateGroupFormat::instance()->nameFilter();
+
+    Preferences *prefs = Preferences::instance();
+    QString suggestedFileName = prefs->lastPath(Preferences::TemplateDocumentsFile);
+    suggestedFileName += QLatin1String("/untitled.ttx");
```
bjorn: This should be a translatable string.

src/tiled/newtemplatedialog.cpp
```diff
@@ -73,9 +82,50 @@ void NewTemplateDialog::updateTemplateName()
...
+    }
+
+    auto model = ObjectTemplateModel::instance();
+    model->addNewDocument(templateGroupDocument.take());
+
+    prefs->setLastPath(Preferences::TemplateDocumentsFile,
+                       QFileInfo(fileName).path());
+
+    mUi->groupsComboBox->addItem(name);
+    mUi->groupsComboBox->setCurrentIndex(mUi->groupsComboBox->count() - 1);
```
bjorn: After this it should call `updateOkButton` (or connect the index change with this function).

src/tiled/newtemplatedialog.ui
```diff
@@ -59,53 +59,25 @@
...
+       <widget class="QPushButton" name="createGroupButton">
         <property name="text">
-         <string>Save to a new group</string>
-        </property>
-        <property name="checkable">
-         <bool>true</bool>
-        </property>
-        <property name="checked">
-         <bool>false</bool>
+         <string>Create new Group</string>
```
bjorn: I think this button should probably use the same "new file" icon as used in the templates dock. This could be its tool tip.

src/tiled/newtemplatedialog.h
```diff
@@ -37,15 +37,17 @@ class NewTemplateDialog : public QDialog
     Q_OBJECT
 
 public:
-    explicit NewTemplateDialog(QList<QString> groupNames, QWidget *parent = 0);
+    explicit NewTemplateDialog(const QList<QString> &groupNames, const QString &objectName, QWidget *parent = 0);
```
bjorn: Please use `nullptr` instead of 0.

src/tiled/newtemplatedialog.cpp
```diff
@@ -20,20 +20,36 @@
...
+
+    mObjectName = objectName;
+    mUi->templateNameCheckBox->setEnabled(!objectName.isEmpty());
+    connect(mUi->templateNameCheckBox, &QCheckBox::toggled,
+            this, &NewTemplateDialog::updateTemplateName);
+
+    connect(mUi->createGroupButton, &QPushButton::pressed,
+            this, &NewTemplateDialog::newTemplateGroup);
+
     mUi->groupsComboBox->addItems(groupNames);
```
bjorn: Note that you can just do this:
```c++
    auto model = ObjectTemplateModel::instance();
    mUi->groupsComboBox->setModel(model);
```
This way, there is also no need to update the list of items after adding a new template group.

src/tiled/newtemplatedialog.cpp
thabetx: Auto-filling is a nice idea and I won't need the update name function.

src/tiled/objecttemplatemodel.cpp
```diff
@@ -153,6 +153,24 @@ bool ObjectTemplateModel::addNewDocument(TemplateGroupDocument *document)
     return true;
 }
 
+bool ObjectTemplateModel::addDocument(TemplateGroupDocument *document)
+{
+    if (!document)
```
thabetx: Why is that? This will only handle the null document in debug mode, right?

bjorn: Yes, because it is simply invalid to try adding null document, so it's never expected to happen. The assert would document that and in addition checks for it in debug mode.

thabetx: Okay, my understanding now is that we don't need to handle invalid scenarios that won't happen and the Q_Assert helps ensure that while coding.

bjorn: Exactly.

src/tiled/newtemplatedialog.cpp
thabetx: This is better, but I had to add a condition to the model that the row is >= 0 so that the -1 initial index doesn't crash.

src/tiled/newtemplatedialog.cpp
thabetx: The new Qt signal-slot syntax for this signal is a little longer. Calling updateOkButton would be enough in all cases.

src/tiled/newtemplatedialog.ui
thabetx: Nice solution, I didn't like how the button was very wide.

thabetx:
I updated the PR. I was thinking about adding delete as well but I will work on it while adding detach.

bjorn:
Cool!

PR #1667, opened 2017-07-30
Add JSON support for reusable object templates

thabetx:
This PR adds JSON support for the reusable object templates, Tiled should be able to:
- [x] Write template instances.
- [x] Read template instances.
- [x] Write template groups.
- [x] Read template groups.

The groups file is used internally by Tiled, so I'm not sure if it's necessary to save it in JSON.

### DIFF
--- a/src/libtiled/maptovariantconverter.cpp
+++ b/src/libtiled/maptovariantconverter.cpp
@@ -124,6 +124,38 @@ QVariant MapToVariantConverter::toVariant(const Tileset &tileset,
     return toVariant(tileset, 0);
 }
 
+QVariant MapToVariantConverter::toVariant(const TemplateGroup &templateGroup,
+                                          const QDir &directory)
+{
+    mMapDir = directory;
+    QVariantMap templateGroupVariant;
+
+    templateGroupVariant[QLatin1String("type")] = QLatin1String("templategroup");
+    templateGroupVariant[QLatin1String("name")] = templateGroup.name();
+    templateGroupVariant[QLatin1String("nexttemplateid")] = templateGroup.nextTemplateId();
+
+    QVariantList tilesetVariants;
+
+    mGidMapper.clear();
+    unsigned firstGid = 1;
+    for (const SharedTileset &tileset : templateGroup.tilesets()) {
+        tilesetVariants << toVariant(*tileset, firstGid);
+        mGidMapper.insert(firstGid, tileset.data());
+        firstGid += tileset->nextTileId();
+    }
+
+    templateGroupVariant[QLatin1String("tilesets")] = tilesetVariants;
+
+    QVariantList templateVariants;
+
+    for (const ObjectTemplate *objectTemplate: templateGroup.templates())
+        templateVariants << toVariant(*objectTemplate);
+
+    templateGroupVariant[QLatin1String("templates")] = templateVariants;
+
+    return templateGroupVariant;
+}
+
 QVariant MapToVariantConverter::toVariant(const Tileset &tileset,
                                           int firstGid) const
 {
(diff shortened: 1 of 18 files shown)


bjorn:
> The groups file is used internally by Tiled, so I'm not sure if it's necessary to save it in JSON.

Yeah, I initially thought the same of the `objecttypes.xml` file, but then people who used the JSON format for maps/tilesets had a reason to parse that file as well and asked for it in JSON format. I guess we'll run into a similar situation for the templates, since game engines will need to load them to understand what the objects are.

Sometimes it's nice to give people options but it also becomes quite a burden sometimes and it causes complexity that often also affects people not using those options. :-/

src/libtiled/maptovariantconverter.cpp
```diff
@@ -263,15 +270,41 @@ QVariant MapToVariantConverter::toVariant(const Tileset &tileset,
     return tilesetVariant;
 }
 
-QVariant MapToVariantConverter::toVariant(const TemplateGroup &templateGroup, int firstTid) const
+QVariant MapToVariantConverter::toVariant(const TemplateGroup &templateGroup, int firstTid)
 {
     QVariantMap templateGroupVariant;
 
-    templateGroupVariant[QLatin1String("firsttid")] = firstTid;
+    if (firstTid > 0) {
```
bjorn: Since you have essentially two entirely different implementations based on whether a `firstTid` is passed or not, I think it would be better to have two separate overloads for this. In fact the version that doesn't take a `firstTid` can be inlined in `MapToVariantConverter::toVariant(const TemplateGroup &templateGroup, QDir &directory)`.

This is different for the function for tilesets, since those can be embedded.

src/libtiled/maptovariantconverter.cpp
```diff
@@ -263,15 +270,41 @@ QVariant MapToVariantConverter::toVariant(const Tileset &tileset,
...
+        templateGroupVariant[QLatin1String("type")] = QLatin1String("templategroup");
```
bjorn: Actually the `type` should be set when no `firstTid` is passed, and doesn't need to be set when it is.

thabetx:
Okay, the groups file contains just a list of file paths for template groups to be loaded on startup, so it should be simple to use JSON, although I'm not entirely convinced that it might be useful to read them as the path of the group will be also included in the map :).

Either way, I'm not sure how to integrate both formats in the UI. Curently the groups XML file is updated automatically with a new path whenever a new group is added or created with no action required from the user. I think it could be done with an export option offered by a button in the dock that offers saving in any of the two formats. What do you think?

bjorn:
> Okay, the groups file contains just a list of file paths for template groups to be loaded on startup, so it should be simple to use JSON, although I'm not entirely convinced that it might be useful to read them as the path of the group will be also included in the map :).

Ah, right, this file is really not very useful and can just be maintained by Tiled in whatever format it wants, so XML is fine.

thabetx:
Okay :smile: I will support the the template group file for sure.

thabetx:
Sorry for not being clear in the first description.

thabetx:
I think this PR is now ready. Now all actions are supported on maps and template groups.

thabetx:
XML format was hard coded as it was the only format, now the enhancements PR should be updated to allow opening JSON template groups as well, this could be done when both PRs are merged.

src/libtiled/varianttomapconverter.cpp
```diff
@@ -334,20 +342,51 @@ TemplateGroup *VariantToMapConverter::toTemplateGroup(const QVariant &variant)
...
+        templateGroup = TemplateManager::instance()->loadTemplateGroup(source, &error);
+
+        if (!templateGroup) { // Couldn't open the file
+            templateGroup = new TemplateGroup();
+            templateGroup->setFileName(source);
+            templateGroup->setLoaded(false);
+        }
     }
 
+    if (templateGroup)
```
bjorn: This check is superfluous since the above code always makes sure there is a template group.

src/plugins/json/jsonplugin.cpp
```diff
@@ -326,8 +326,35 @@ JsonTemplateGroupFormat::JsonTemplateGroupFormat(QObject *parent)
 
 Tiled::TemplateGroup *JsonTemplateGroupFormat::read(const QString &fileName)
 {
-    Q_UNUSED(fileName);
-    return nullptr;
+    QFile file(fileName);
+
+    if (!file.open(QIODevice::ReadOnly | QIODevice::Text)) {
+        mError = tr("Could not open file for reading.");
+        return new Tiled::TemplateGroup();
```
bjorn: Why not return `nullptr`? (and on the line below)

thabetx: Yeah that's wrong, I used tileset JSON reader as reference, where the returned shared pointer is actually a nullptr.

thabetx:
I applied the last two review comments,The hardest things while doing this PR was connecting everything together, like finding out that format instances should be added to the plugin manager. at first I thought it would pick them up automatically :).

bjorn:
Can you rebase this change onto the latest `wip/templates` now that #1666 is merged?

> The hardest things while doing this PR was connecting everything together, like finding out that format instances should be added to the plugin manager. at first I thought it would pick them up automatically :).

Ah yeah, this is where some documentation of the overall design could help...

thabetx:
I updated the dialogs to support JSON. The new template group is duplicated from the templates dock in the new template dialog as they both offer creating new groups. Is there a way to reuse the newTemplateGroup dialog directly from the templates dock without duplication?

bjorn:
> The new template group is duplicated from the templates dock in the new template dialog as they both offer creating new groups. Is there a way to reuse the newTemplateGroup dialog directly from the templates dock without duplication?

Maybe you could create a public static helper function in the `NewTemplateDialog` that takes the needed parameters, so that you can call this from the `TemplatesDock` as well. For the rest the PR looks good so I'll merge it now.

PR #1668, opened 2017-07-30
Support maps with non-consecutive tile IDs in libtiled-java

stephaneseng:
This PR causes `TileSet` to store tiles in a `Map` instead of a `List` in order to properly load tiles when there are gaps in the tile IDs sequence.

## Problem

When the following `map` is read with `TMXMapReader.readMap()`:

```
<?xml version="1.0" encoding="UTF-8"?>
<map version="1.0" tiledversion="1.0.2" orientation="orthogonal" renderorder="right-down" width="3" height="1" tilewidth="32" tileheight="32" nextobjectid="1">
 <tileset firstgid="1" name="embedded_image_collection" tilewidth="32" tileheight="32" tilecount="2" columns="0">
  <grid orientation="orthogonal" width="1" height="1"/>
  <tile id="0">
   <image width="32" height="32" source="embedded_image_collection_1.png"/>
  </tile>
  <tile id="2">
   <image width="32" height="32" source="embedded_image_collection_2.png"/>
  </tile>
 </tileset>
 <layer name="Tile Layer 1" width="3" height="1">
  <data encoding="csv">
1,0,3
</data>
 </layer>
</map>
```

Calling `((TileLayer) map.getLayer(0)).getTileAt(2, 0)` returns `null` instead of returning the tile with the source `embedded_image_collection_2.png`.

## Explanations

This problem is caused by the fact that there is a gap in the tile IDs sequence, which is a common situation since Tiled 0.15 (see http://doc.mapeditor.org/reference/tmx-changelog/#tiled-015).

Indeed, when `TMXMapReader.getTileForTileGID()` (called by `TMXMapReader.setTileAtFromTileId()`, called by `TMXMapReader.readLayer()`) tries to retreive the tile with the global ID of `3`:

* `TileSet.getTile(3 - 1)` is called
* But because the current implementation of `TileSet` stores tiles in a `List`, the two loaded tiles are indexed at positions `0` and `1` but `TileSet.getTile(3 - 1)` tries to retreive the element indexed at position `2` which doesn't exist
* This wouldn't have been a problem if there wasn't a gap between tile IDs because `TileSet.getTile(2 - 1)` would have been called instead but this is not the case anymore

## Proposal

By storing tiles in a `Map` indexed by tiles IDs, `TileSet.getTile(3 - 1)` will return the properly indexed tile at key `2`.


bjorn:
Looks good, but maybe @mikepthomas can chime in about possible conflicts with #1637. This PR seems relevant in addition to #1637.

Note that this little piece of code in `addTile` would also need adjustment:
```java
         if (t.getId() < 0) {
             t.setId(tiles.size());
         }
```
Instead of using `tiled.size()`, it should allocate the next free tile ID, which is also stored in the TMX file. Though this is not related to loading and fixing it will probably cause more conflicts with the other PR, so it's something to keep in mind for later.

Thanks for the detailed explanation of the PR!

bjorn:
@stephaneseng Could you rebase this change onto latest `master` now that #1637 is merged?

stephaneseng:
First of all, sorry for the very long delay on this PR @bjorn.

Unfortunately, I couldn't get xjc to generate a `Map<Integer, Tile>` instead of a `List<Tile>` in the generated `TileSetData` class.
I've tried to use the `<jxc:javaType />` tag in the bindings file without success (related links found: https://stackoverflow.com/a/1889584, https://github.com/javaee/jaxb-v2/issues/209).

I've also tried to manually declare the `Map<Integer, String> tiles` field with an `@XmlJavaTypeAdapter` annnotation in the `TileSet` class, in order to hide the generated `List<Tile>` tiles field, using the example in the JavaDoc : https://docs.oracle.com/javase/7/docs/api/javax/xml/bind/annotation/adapters/XmlAdapter.html.
But I couldn't make it to work either, it seems that the absence of a `<tiles>` tag wrapping the sequence of `<tile>` tag is problematic (related link found: https://stackoverflow.com/a/35186091).

Therefore I have updated this PR with the only working solution I could come up with which doesn't alter the generated code.
It uses a new `TreeMap<Integer, Tile> tiles` field in the `TileSet` class whose values are "synchronized" on marshall and unmarshal operations with the generated `List<Tile> internalTiles` field of `TileSetData`.
I've also updated the `addTile()` and `getMaxTileId()` functions accordingly, thanks to your first comment.

bjorn:
Nice! I'm sorry to hear this was such a hassle, but it seems like you found a good solution!

PR #1670, opened 2017-07-31
WangTiles: WangColor now have properties including name, color, image, and probability...

Bdtrotte:
Also to be in this pr:

- Right click to choose a wangcolor from the map
- Add/remove buttons for wangColors
- Any other fixes/features before 1.1


Bdtrotte:
For assignment purposes in regards to staggered maps, one can think of the assignment being rotated 45 degrees clock wise. At some point this should be reflected graphically. The wang brush now supports staggered maps, (other tools to come shortly), except when working with edges. Because edges are very dependent on which edge they are closest to, the mouse moved function must be changed to take staggered into account. This is a fairly big effort at the moment, and will be saved for later.

bjorn:
> For assignment purposes in regards to staggered maps, one can think of the assignment being rotated 45 degrees clock wise. At some point this should be reflected graphically.

Did you notice that the overlay used for terrain assignment is already able to handle isometric tilesets? Check out the usages of `tilesetGridTransform`.

Bdtrotte:
> Did you notice that the overlay used for terrain assignment is already able to handle isometric tilesets? Check out the usages of tilesetGridTransform.

I did not! Transforms will be very useful in supporting edges as well, thanks for pointing that out

Bdtrotte:
I have not tested the results of the change of the last commit at length. Map rendering and such seems fine. The terrain brush may behave a little out of sorts on staggered maps, though should likely be updated to properly follow the corner the mouse is closest to (which I'm pretty sure it did not do prior in staggered maps.) In fact this could probably be done in the AbstractTileTool for BetweenTiles mode.

src/tiled/wangfiller.cpp
```diff
@@ -84,20 +131,25 @@ Cell WangFiller::findFittingCell(const TileLayer &back,
                     break;
                 }
             }
+            delete adjacentPoints;
```
bjorn: Since `getSurroundingPoints` uses `malloc`, this should be `free`.

src/libtiled/wangset.cpp
```diff
@@ -301,6 +319,132 @@ WangSet::WangSet(Tileset *tileset,
     mCornerColors(cornerColors),
     mUniqueFullWangIdCount(0)
 {
+    Q_ASSERT(mEdgeColors <= 15 && mCornerColors <= 15);
+
+    if (mEdgeColors > 1) {
```
Bdtrotte: There is an empty element at index zero to make indexing a bit cleaner later on, and having just one color means no WangColors should actually be stored. I suppose the condition could be removed, and it would just default to storing the empty color, but is there any value in this?

src/libtiled/wangset.cpp
```diff
@@ -301,6 +319,132 @@ WangSet::WangSet(Tileset *tileset,
...
+void WangSet::setEdgeColors(int n)
+{
+    Q_ASSERT(n <= 15);
+
+    if (n == mEdgeColors)
+        return;
+
+    if (n == 1) {
+        mEdgeColors = 1;
+        mEdges.clear();
```
Bdtrotte: The wangSet does, bit of an oversight on my part!

Bdtrotte: Ill use `QSharedPointer` I think

src/libtiled/wangset.cpp
```diff
@@ -301,6 +319,132 @@ WangSet::WangSet(Tileset *tileset,
...
+                                          false,
+                                          QString(),
+                                          defaultWangColors[mCorners.size() - 1],
+                                          -1));
+    }
+
+    mCornerColors = n;
+}
+
+void WangSet::insertWangColor(WangColor *wangColor)
```
Bdtrotte: WangColor contains the index in the mColor variable

src/libtiled/wangset.cpp
```diff
@@ -301,6 +319,132 @@ WangSet::WangSet(Tileset *tileset,
...
+    } else {
+        while (mCorners.size() != n + 1)
+            mCorners.append(new WangColor(mCorners.size(),
+                                          false,
+                                          QString(),
+                                          defaultWangColors[mCorners.size() - 1],
+                                          -1));
+    }
+
+    mCornerColors = n;
```
Bdtrotte: I have made this change, as well as a couple other name changes:
mEdgeColors -> [REMOVED]
mCornerColors -> [REMOVED]
mEdges -> mEdgeColors
mCorners -> mCornerColors
edgeColors() -> edgeColorCount()
cornerColors() -> cornerColorCount()

src/tiled/stampbrush.cpp
```diff
@@ -442,7 +444,9 @@ void StampBrush::drawPreviewLayer(const QVector<QPoint> &list)
             Cell cell = mWangFiller->findFittingCell(*tileLayer,
                                                      *preview.data(),
                                                      paintedRegion,
-                                                     p);
+                                                     p,
+                                                     dynamic_cast<StaggeredRenderer*>(mapDocument()->renderer()),
```
Bdtrotte: `findFittingCell(..)` uses the staggeredRenderer

src/tiled/wangdock.cpp
```diff
@@ -354,6 +426,16 @@ void WangDock::wangSetChanged()
 
     refreshCurrentWangColor();
     refreshCurrentWangId();
+
+    if (mCurrentWangSet->edgeColorCount() == 15)
+        mAddEdgeColor->setEnabled(false);
+    else
+        mAddEdgeColor->setEnabled(true);
```
bjorn: Please use:
```c++
mAddEdgeColor->setEnabled(mCurrentWangSet->edgeColorCount() < 15);
mAddCornerColor->setEnabled(mCurrentWangSet->cornerColorCount() < 15);
```

src/tiled/wangdock.cpp
```diff
@@ -394,12 +518,23 @@ void WangDock::setCurrentWangSet(WangSet *wangSet)
                 setColorView();
         }
 
+        if (mCurrentWangSet->edgeColorCount() == 15)
+            mAddEdgeColor->setEnabled(false);
+        else
+            mAddEdgeColor->setEnabled(true);
```
bjorn: Again? Please have one place to update enabled state of these actions.

src/tiled/wangdock.cpp
```diff
@@ -473,33 +629,29 @@ void WangDock::onCurrentWangIdChanged(WangId wangId)
 
 void WangDock::setTemplateView()
 {
-    if (!mTemplateAndColorView->isVisible()) {
-        mSwitchTemplateViewButton->setEnabled(true);
-        mTemplateAndColorView->setVisible(true);
+    if (!mTemplateAndColorWidget->isVisible()) {
```
bjorn: Seems like a superfluous check.

src/tiled/wangdock.cpp
```diff
@@ -473,33 +629,29 @@ void WangDock::onCurrentWangIdChanged(WangId wangId)
...
+    mTemplateAndColorView->setCurrentIndex(0);
     retranslateUi();
 }
 
 void WangDock::setColorView()
 {
-    if (!mTemplateAndColorView->isVisible()) {
-        mSwitchTemplateViewButton->setEnabled(true);
-        mTemplateAndColorView->setVisible(true);
+    if (!mTemplateAndColorWidget->isVisible()) {
```
bjorn: Can also be removed, right?

src/tiled/wangdock.cpp
```diff
@@ -473,33 +629,29 @@ void WangDock::onCurrentWangIdChanged(WangId wangId)
...
-    mTemplateAndColorView->setCurrentWidget(mWangColorView);
+    mTemplateAndColorView->setCurrentIndex(1);
 
     retranslateUi();
 }
 
 void WangDock::hideTemplateColorView()
 {
-    if (!mTemplateAndColorView->isVisible())
+    if (!mTemplateAndColorWidget->isVisible())
```
bjorn: Should also be superfluous now.

src/tiled/wangfiller.cpp
```diff
@@ -47,26 +50,68 @@ void WangFiller::setWangSet(WangSet *wangSet)
     mWangSet = wangSet;
 }
 
+static QPoint *getSurroundingPoints(QPoint point, StaggeredRenderer *staggeredRenderer, Map::StaggerAxis staggerAxis)
+{
+    QPoint *adjacentPoints = (QPoint*)malloc(sizeof(QPoint) * 8);
```
bjorn: There's still the problem with `malloc` / `delete` mismatch. Actually, I suggested using a static array but I think it's not going to work because the function is called multiple times. But you could allocate these arrays on the stack and pass them into the function:
```c++
static void getSurroundingPoints(QPoint point, StaggeredRenderer *staggeredRenderer, Map::StaggerAxis staggerAxis, QPoint *points)

QPoint adjacentPoints[8];
getSurroundingPoints(point, staggeredRenderer, staggerAxis, adjacentPoints);
```

src/tiled/wangfiller.h
```diff
@@ -52,15 +55,19 @@ class WangFiller
     Cell findFittingCell(const TileLayer &back,
                          const TileLayer &front,
                          const QRegion &fillRegion,
-                         QPoint point) const;
+                         QPoint point,
+                         StaggeredRenderer *staggeredRenderer = nullptr,
+                         Map::StaggerAxis staggerAxis = Map::StaggerX) const;
```
bjorn: Instead of adding these parameters to each `WangFiller` method, passing them around on each call, consider adding them to the `WangFiller` constructor. And to avoid having to keep a `WangFiller` instance up-to-date, just allocate it on the stack when you're going to call its methods. This is similar to how `TilePainter` is used.

tests/staggeredrenderer/test_staggeredrenderer.cpp
```diff
@@ -89,11 +89,11 @@ void test_StaggeredRenderer::screenToTileCoords_data()
     QTest::addColumn<QPointF>("screenCoords");
     QTest::addColumn<QPointF>("tileCoords");
 
-    QTest::newRow("10,16") << QPointF(10, 16) << QPointF(0, 0);
-    QTest::newRow("5,5")   << QPointF(5, 5)   << QPointF(-1, -1);
+    QTest::newRow("10,16") << QPointF(10, 16) << QPointF(0, 1);
```
bjorn: I think these values are very hard to understand, since they rely on rounding the internal fraction in a coordinate space where rounding makes no sense at all. Please keep these values the same, and change the `test_StaggeredRenderer::screenToTileCoords` to truncate instead of round.

src/libtiled/wangset.cpp
```diff
@@ -317,34 +315,14 @@ WangSet::WangSet(Tileset *tileset,
...
-
 }
 
 int WangSet::edgeColorCount() const
 {
     int size = mEdgeColors.size();
 
     if (size > 1)
-        return size - 1;
+        return size;
```
bjorn: Let's shorten this to:
```c++
return qMax(1, mEdgeColors.size());
```

src/libtiled/mapreader.cpp
```diff
@@ -614,22 +612,18 @@ void MapReaderPrivate::readTilesetWangSets(Tileset &tileset)
                 } else if (xml.name() == QLatin1String("wangedgecolor")
                            || xml.name() == QLatin1String("wangcornercolor")) {
                     const QXmlStreamAttributes wangColorAtts = xml.attributes();
-                    QString name = wangColorAtts.value(QLatin1String("name")).toString();
-                    int index = wangColorAtts.value(QLatin1String("index")).toInt();
+                    QString name = wangColorAtts.value(QLatin1String("name")).toString();\
```
bjorn: Stray `\`.

src/tiled/stampbrush.cpp
```diff
@@ -441,12 +441,14 @@ void StampBrush::drawPreviewLayer(const QVector<QPoint> &list)
...
         for (const QPoint p : list) {
-            Cell cell = mWangFiller->findFittingCell(*tileLayer,
-                                                     *preview.data(),
-                                                     paintedRegion,
-                                                     p,
-                                                     dynamic_cast<StaggeredRenderer*>(mapDocument()->renderer()),
-                                                     mapDocument()->map()->staggerAxis());
+            WangFiller wangFiller(mWangSet,
+                                  dynamic_cast<StaggeredRenderer *>(mapDocument()->renderer()),
+                                  mapDocument()->map()->staggerAxis());
```
bjorn: Please move the `wangFiller` out of the loop.

PR #1512, opened 2017-03-26
Shortcut Handler Class and Basic Functional Shorcut Editor

laspencer91:
This impliments the Singleton shortcut handler class, that stores any desired actions. These actions can be added either by a list of QActions or by adding them one at a time. 

In addition I have added basic shortcut editing functionality, so that you can see how everything interacts together. A few of the actions are duplicates. Their attributes are duplicates, but they are different instances of QActions, so their may be duplicate actions being added somewhere in the code. (Not in this submitted code, but something previous). I am looking further into this next. You can see what I mean when you open the preference options.

PR #1672, opened 2017-08-02
Added keyboard shortcut for Save All, Ctrl+Shift+S

ttencate:
There doesn't seem to be a strong standard for which shortcut to use. I
chose Ctrl+Shift+S because it's used in Visual Studio and Qt Creator,
and consistent with Ctrl+Shift+W for Close All.

Some applications, like GEdit, use Ctrl+Shift+S for Save As instead, but
I think that's a much rarer operation which doesn't warrant the visual
clutter of a keyboard shortcut in the menu.

### DIFF
--- a/src/tiled/mainwindow.cpp
+++ b/src/tiled/mainwindow.cpp
@@ -163,7 +163,6 @@ MainWindow::MainWindow(QWidget *parent, Qt::WindowFlags flags)
     mUi->actionNewMap->setShortcuts(QKeySequence::New);
     mUi->actionOpen->setShortcuts(QKeySequence::Open);
     mUi->actionSave->setShortcuts(QKeySequence::Save);
-    mUi->actionSaveAs->setShortcuts(QKeySequence::SaveAs);
     mUi->actionClose->setShortcuts(QKeySequence::Close);
     mUi->actionQuit->setShortcuts(QKeySequence::Quit);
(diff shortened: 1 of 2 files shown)


bjorn:
Actually in Tiled, Ctrl+Shift+S has been "Save As" since many years on some systems, so the shortcut would have be removed there for this to work. This is done via the following line, so actually it [depends on the system](http://doc.qt.io/qt-5/qkeysequence.html#standard-shortcuts) which shortcut is used:

```c++
mUi->actionSaveAs->setShortcuts(QKeySequence::SaveAs);
```

Ctrl+Shift+S is also "Save As" in GIMP, Inkscape and Sublime Text, but I agree the action is actually more rare than Save All and there is something to say for consistency as well.

bjorn:
I'm worried about those few people who may have gotten used to using Ctrl+Shift+S for saving their map as a different file. When they will press their shortcut in the next version of Tiled, it will do exactly what they did not want to do. :-(

ttencate:
Good point. I'll take a page from GEdit's book and use Ctrl+Shift+L instead (in a fresh PR). How's that?

bjorn:
> Good point. I'll take a page from GEdit's book and use Ctrl+Shift+L instead (in a fresh PR). How's that?

Sure, let's have it. Though I think really what we're looking for is #1512.

ttencate:
Closing in favour of #1673.

Issue #1555, opened 2017-04-25
When using the Edit Polygons tool, Delete key should delete selected nodes


bjorn:
Currently the Delete key will delete all selected objects as usual. That may still be a valid fallback when there are no selected polygon nodes.

Issue #404, opened 2013-03-29
Allow to close a poly line


wolfscaptain:
If an end point of a poly line is next to the starting point, it should snap to it to form a closed shape.

In addition, allow to join the starting and ending points of an already constructed poly line.

bjorn:
In the current `master` version it is possible to close a polyline to make a polygon, either while creating it or while extending it (change aa8d7317c0e89da0a37fe020102db4dac361ee93). These improvements are scheduled to be released as part of Tiled 1.2, and are already available in the development snapshots.

Issue #157, opened 2012-02-13
Add new point to Polyline or Poligon object.


NinjaFighter:
I cannot figure how to add points to polyline or poligon objects. (is it possible?)
Could be great if I could do a left click, over the line, and selecting an "Add a New Point" option.
Nowadays, I open the TMX file with notepad++ and add points manually when I need that.

bjorn:
Better late than never (in truth, somebody stepped up to sponsor this improvement!): Double-clicking a segment now insert a new point at the clicked location, and you can immediately drag the new point to where you want it. I think it's a great improvement in usability.

The old "Split Segments" menu action remains available, mainly since it allows quickly splitting many segments, if you wanted that.

> I think could be great, if I can select two points only clicking the line, but it is a fairly trivial detail.

That's also implemented now. :-)

PR #1674, opened 2017-08-03
Polygon Editing Tools Improvements

ketanhwr:
The opacity for preview edge is (1/3)rd of the fixed edges. Here's a screenshot:

screenshot from 2017-08-03 20-30-28

### DIFF
--- a/src/tiled/editpolygontool.cpp
+++ b/src/tiled/editpolygontool.cpp
@@ -39,6 +39,7 @@
 #include <QApplication>
 #include <QGraphicsItem>
 #include <QGraphicsView>
+#include <QKeyEvent>
 #include <QMenu>
 #include <QPainter>
 #include <QPalette>
@@ -181,6 +182,20 @@ void EditPolygonTool::deactivate(MapScene *scene)
     AbstractObjectTool::deactivate(scene);
 }
 
+void EditPolygonTool::keyPressed(QKeyEvent *event)
+{
+    switch (event->key()) {
+    case Qt::Key_Escape:
+        if (mMode == Extending) {
+            cancelAddingNode();
+            return;
+        }
+        break;
+    }
+
+    AbstractObjectTool::keyPressed(event);
+}
+
 void EditPolygonTool::mouseEntered()
 {
 }
@@ -208,6 +223,9 @@ void EditPolygonTool::mouseMoved(const QPointF &pos,
     case Moving:
         updateMovingItems(pos, modifiers);
         break;
+    case Extending:
+        addNode(pos, modifiers);
+        break;
     case NoMode:
         break;
     }
@@ -233,11 +251,16 @@ static QTransform viewTransform(QGraphicsSceneMouseEvent *event)
 
 void EditPolygonTool::mousePressed(QGraphicsSceneMouseEvent *event)
 {
-    if (mMode != NoMode) // Ignore additional presses during select/move
+    if (mMode != NoMode && mMode != Extending) // Ignore additional presses during select/move
         return;
 
     switch (event->button()) {
     case Qt::LeftButton: {
+        if (mMode == Extending) {
+            finishAddingNode();
+            break;
+        }
+
         mMousePressed = true;
         mStart = event->scenePos();
         mScreenStart = event->screenPos();
@@ -260,6 +283,11 @@ void EditPolygonTool::mousePressed(QGraphicsSceneMouseEvent *event)
         break;
     }
     case Qt::RightButton: {
+        if (mMode == Extending) {
+            cancelAddingNode();
+            break;
+        }
+
         const QList<QGraphicsItem *> items = mapScene()->items(event->scenePos(),
                                                                Qt::IntersectsItemShape,
                                                                Qt::DescendingOrder,
@@ -330,6 +358,8 @@ void EditPolygonTool::mouseReleased(QGraphicsSceneMouseEvent *event)
         break;
     case Moving:
         finishMoving(event->scenePos());
+        break;
+    case Extending:
         break;
     }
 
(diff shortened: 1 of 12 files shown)


ketanhwr:
I've also added an option to extend a polyline (#157). Here's a gif for demonstration:

extend-poygon

src/tiled/editpolygontool.cpp
```diff
@@ -618,6 +664,15 @@ void EditPolygonTool::showHandleContextMenu(PointHandle *clickedHandle,
     connect(joinNodesAction, SIGNAL(triggered()), SLOT(joinNodes()));
     connect(splitSegmentsAction, SIGNAL(triggered()), SLOT(splitSegments()));
 
+    const auto firstHandle = *mSelectedHandles.begin();
+    MapObject *mapObject = firstHandle->mapObjectItem()->mapObject();
+
+    if (mapObject->shape() == MapObject::Polyline) {
+        QAction *extendPolyline = menu.addAction(tr("Extend Polyline"));
+        extendPolyline->setEnabled(n == 1 && firstHandle->pointIndex() == mapObject->polygon().size() - 1);
```
Ablu: I found this boolean expression a bit hard to read... Maybe passing it into a variable could help here (would also shorten the line a bit). But that might be personal taste...

ketanhwr:
Now, the user can extend a polyline from both the first point and the end point. Also, the user can use right click to cancel extending the polyline.

ketanhwr:
While 'Extending' a polyline, the user can now choose to close it as well (Basically convert it into a polygon).

ketanhwr:
When the user selects two nodes on a polygon which are side-by-side, then the user can delete that edge which will convert that polygon into a polyline.

ketanhwr:
Conversion between Polygon/Polyline is shown in the gif here:

peek 2017-08-05 17-49

ketanhwr:
My previous commit closes #1555. Delete key now deletes the selected nodes. If no nodes are selected, then the object is deleted which was previously the standard action.

Ablu:
Looks nice already

Maybe ESC should cancel the extension process too (not only right click).
And I find it a bit weird that during closing a polyline, the polygon preview is displayed without snapping the previewed point to the starting point.

src/tiled/editpolygontool.cpp
```diff
@@ -566,6 +583,91 @@ void EditPolygonTool::updateMovingItems(const QPointF &pos,
...
+
+    const auto &selectedHandle = *mSelectedHandles.begin();
+    const MapObjectItem *item = selectedHandle->mapObjectItem();
+    MapObject *mapObject = item->mapObject();
+
+    pixelCoords -= mapObject->position();
+
+    QPolygonF polygon = mapObject->polygon();
+
+    if ((polygon.last() - polygon.first()).manhattanLength() < QApplication::startDragDistance() * 1.0)
```
Ablu: Maybe introduce a variable for the left side of the `<` here, to keep the line shorter?

src/tiled/editpolygontool.cpp
```diff
@@ -566,6 +583,91 @@ void EditPolygonTool::updateMovingItems(const QPointF &pos,
...
+    const auto &selectedHandle = *mSelectedHandles.begin();
+    const MapObjectItem *item = selectedHandle->mapObjectItem();
+    MapObject *mapObject = item->mapObject();
+
+    mapObject->setComplete(true);
+
+    QPolygonF polygon = mapObject->polygon();
+
+    if ((polygon.last() - polygon.first()).manhattanLength() < QApplication::startDragDistance() * 1.0) {
+
```
Ablu: unneeded empty line

src/tiled/editpolygontool.cpp
```diff
@@ -618,6 +720,35 @@ void EditPolygonTool::showHandleContextMenu(PointHandle *clickedHandle,
     connect(joinNodesAction, SIGNAL(triggered()), SLOT(joinNodes()));
     connect(splitSegmentsAction, SIGNAL(triggered()), SLOT(splitSegments()));
 
+    const auto firstHandle = *mSelectedHandles.begin();
+    MapObject *mapObject = firstHandle->mapObjectItem()->mapObject();
+
+    if (mapObject->shape() == MapObject::Polyline) {
+        QAction *extendPolyline = menu.addAction(tr("Extend Polyline"));
+
+        bool indexCheck = (firstHandle->pointIndex() == mapObject->polygon().size() - 1) ||
```
Ablu: I find the name `indexCheck` a bit weird... Maybe `handleCanBeExtended`?

src/tiled/mapeditor.cpp
```diff
@@ -479,7 +480,10 @@ void MapEditor::performStandardAction(StandardAction action)
         paste(ClipboardManager::PasteInPlace);
         break;
     case DeleteAction:
-        MapDocumentActionHandler::instance()->delete_();
+        if (mEditPolygonTool->countSelectedHandles())
```
Ablu: I personally do not like these implicit > 0 checks on integer values in this context. I usually prefer to write `> 0` here. But not sure what @bjorn. Thinks here...

bjorn:
In general, I think this pull request is too much of a collection of different issues. It would really help if separate pull requests were opened for things that aren't directly related to each other.

ketanhwr:
> In general, I think this pull request is too much of a collection of different issues. It would really help if separate pull requests were opened for things that aren't directly related to each other.

So should I close this PR, and open separate PRs for different issues? :|

bjorn:
> So should I close this PR, and open separate PRs for different issues? :|

I'd say keep this one for "reduced opacity of last edge" and "add option to extend polyline" (including from first node and right-click to cancel). And create separate PRs for:

* Added action to delete polygon edge
* Added option to close a polyline
* Delete key now deletes selected nodes

Unless inter-dependent code makes this very difficult.

src/tiled/editpolygontool.cpp
```diff
@@ -566,6 +598,92 @@ void EditPolygonTool::updateMovingItems(const QPointF &pos,
...
+
+    const auto &selectedHandle = *mSelectedHandles.begin();
+    const MapObjectItem *item = selectedHandle->mapObjectItem();
+    MapObject *mapObject = item->mapObject();
+
+    mapObject->setComplete(true);
+
+    QPolygonF polygon = mapObject->polygon();
+
+    if (mapObject->shape() == MapObject::Polygon)
```
bjorn: There is no point in this check, is there?

ketanhwr:
> Also, if we snap to the other end of the polyline to create a polygon, then I'd expect this snapping to be also available when creating a new polyline. Essentially implementing both parts of issue #404.

So basically, merge both the tools into a single tool ?

PR #1683, opened 2017-08-11
Allows extending of the polyline

ketanhwr:
#157 

TODOs:
- [x] Don't stop extending after one point. Keep extending until right-click or escape key is encountered.


src/tiled/editpolygontool.cpp
```diff
@@ -566,6 +596,59 @@ void EditPolygonTool::updateMovingItems(const QPointF &pos,
     }
 }
 
+void EditPolygonTool::addNode(const QPointF &pos,
```
bjorn: The name `addNode` does not seem to reflect what the function is doing. I think `updateAddingNode` would be better, but it still leaves two kinds of terminology for the same thing. Since you're already using the mode `Extending` and have `mExtendingFirst`, I would suggest to rename all the other functions as follows:
* `updateExtending`
* `finishExtending`
* `cancelExtending`

src/tiled/editpolygontool.h
```diff
@@ -83,6 +86,13 @@ private slots:
     void startMoving();
     void updateMovingItems(const QPointF &pos,
                            Qt::KeyboardModifiers modifiers);
+
+    void addNode(const QPointF &pos, Qt::KeyboardModifiers modifiers);
+
+    void finishAddingNode();
+
```
bjorn: Coding style: I'd suggest to leave out these empty lines so that the related functions are grouped together.

src/tiled/editpolygontool.h
```diff
@@ -98,6 +108,7 @@ private slots:
     QPointF mStart;
     QPoint mScreenStart;
     Qt::KeyboardModifiers mModifiers;
+    bool mExtendingFirst;
```
bjorn: I know this variable isn't expected to be used before being set, but please anyway initialize the variable.

Ablu:
Hm. Could you rebase this instead of merging the master? For example you reordered the WangTool declaration as it is.

ketanhwr:
Wait, I'll rebase this.

src/tiled/createmultipointobjecttool.cpp
```diff
@@ -75,15 +82,24 @@ void CreateMultipointObjectTool::mousePressedWhileCreatingObject(QGraphicsSceneM
         QPolygonF next = mOverlayPolygonObject->polygon();
 
         // If the last position is still the same, ignore the click
-        if (next.last() == current.last())
+        if (next.last() == current.last() && !mExtending)
```
bjorn: Why is this check irrelevant while extending? I think it was introduced to avoid inserting overlapping points in case the user double-clicks. It seems to me like it would be useful to check the first or last point, depending on `mExtendingFirst`.

src/tiled/createobjecttool.h
```diff
@@ -66,6 +69,8 @@ class CreateObjectTool : public AbstractObjectTool
     ObjectGroupItem *mObjectGroupItem;
     MapObjectItem *mNewMapObjectItem;
     MapObjectItem *mOverlayPolygonItem;
+    bool mExtending;
+    bool mExtendingFirst;
```
bjorn: You can't in general extend an object so I'd prefer these variables to be in the `CreateMultipointObjectTool`. At least `mExtendingFirst` could easily be moved there since it isn't used in the base class. Regarding `mExtending`, are there any functions that couldn't be just overridden?

src/tiled/mapeditor.cpp
```diff
@@ -673,6 +677,17 @@ void MapEditor::setWangFill(bool value)
     mBucketFillTool->setWangFill(value);
 }
 
+void MapEditor::extend(MapObjectItem *mapObjectItem, bool extendingFirst)
+{
+    setSelectedTool(mPolylineObjectsTool);
```
bjorn: Using this instead of `mToolManager->selectTool(mPolylineObjectsTool)` makes the tool switch without this being visible on the "Tools" tool bar. Was that a conscious decision? I think it wouldn't hurt for the user to see the active tool changing. Same for `extendingFinished`.

src/tiled/createobjecttool.h
ketanhwr: I tried overriding the functions `cancelNewMapObject()` and `finishNewMapObject()`, but apparently, the parent function gets called only and not the child one.

bjorn: That makes no sense, so I quickly searched where these functions are being called.

Turns out it's because the `CreatePolylineObjectTool` is explicitly calling the functions of the `CreateObjectTool` class here:
```c++
void CreatePolylineObjectTool::finishNewMapObject()
{
    if (mNewMapObjectItem->mapObject()->polygon().size() >= 2)
        CreateObjectTool::finishNewMapObject();
    else
        CreateObjectTool::cancelNewMapObject();
}
```
I see no reason why it should be doing that. There are no existing overrides that would need to be skipped. So please just remove the `CreateObjectTool::` part there.

ketanhwr: Oh, right :man_facepalming:

ketanhwr: I removed `CreateObjectTool::` but added `CreateMultipointObjectTool::`, because without any `className::`, their was some runtime error (the window wasn't responding at all). I wasn't able to understand why this happened.

bjorn: Well, I missed it too, but this is obviously because we're in `CreatePolylineObjectTool::finishNewMapObject` there, so we should at least avoid calling that exact function recursively.

ketanhwr: Oh, right :man_facepalming:

ketanhwr:
After we start extending the polyline, since we've shifted to the Create Polyline Tool, the color changes back to blue. Is this fine?

bjorn:
> After we start extending the polyline, since we've shifted to the Create Polyline Tool, the color changes back to blue. Is this fine?

I hoped it may be fine, but while checking it out it felt strange and it'd probably be better for just the new segment to be blue.

I've pushed two bugfixes for things I noticed while testing.

In addition to not coloring the whole line blue, I think we need to find a way to have the new segment show up immediately. Currently, you have to move your mouse at least once after clicking "Extend Polyline" to make the new segment show up.

ketanhwr:
So probably, when the user clicks on the first/last node, the extending should begin?

bjorn:
> So probably, when the user clicks on the first/last node, the extending should begin?

Well, then it would work mostly like Vectr, right? But if it's just activated on click, then we can't switch tools because we'd need it to not interfere with just trying to select some nodes.

Or, it would need to work like Inkscape, where the nodes to click for extending the polyline are only visible in the "Insert Polyline" tool.

Or we find some way to place the new node based on the current mouse position before we get the next mouse move event.

ketanhwr:
@bjorn what exactly was wrong here? I'd like to get this one merged.

bjorn:
@ketanhwr The main issues I seem to have ran into was that you need to move the mouse for the line to show up when extending the polyline, and that the whole thing turns blue rather than just the new line.

For the overal implementation, I'd have to take a detailed look again and consider alternatives, but right now I'm trying to finish some stuff people have been paying for. Either this week or the next I'm also going to look at making some improvements to the polygon editing in general.

ketanhwr:
Alright :slightly_smiling_face:

bjorn:
Hmm, I realized another problem with this implementation. Since it goes through the `MapEditor`, the "Extend Polyline" action does nothing when used in the Tile Collision Editor, unless we duplicate the same connection setup in the `TileCollisionDock`.

I'm considering a rather big rewrite of the whole polygon editing functionality, or at least to re-think it all from scratch based on the wishlist of functionality, because of the difficulties with putting such small functions on top of the current implementation.

ketanhwr:
> I'm considering a rather big rewrite of the whole polygon editing functionality, or at least to re-think it all from scratch based on the wishlist of functionality, because of the difficulties with putting such small functions on top of the current implementation.

Interested in helping with this.

bjorn:
A few more fixes later, I think we can merge this, but I'll still be looking at redoing the general setup.

> Interested in helping with this.

That is appreciated, but I'm under some time pressure to improve this aspect of Tiled and I don't currently see how we can do this together.

PR #1685, opened 2017-08-13
Adds option to remove an edge from polygon

ketanhwr:
If two consecutive nodes are selected, then you can delete the edge between them to convert a polygon to polyline.

### DIFF
--- a/src/tiled/editpolygontool.cpp
+++ b/src/tiled/editpolygontool.cpp
@@ -44,6 +44,8 @@
 #include <QPalette>
 #include <QUndoStack>
 
+#include <cstdlib>
+
 using namespace Tiled;
 using namespace Tiled::Internal;
 
--- a/src/tiled/editpolygontool.cpp
+++ b/src/tiled/editpolygontool.cpp
@@ -621,14 +621,15 @@ void EditPolygonTool::showHandleContextMenu(PointHandle *clickedHandle,
     connect(splitSegmentsAction, SIGNAL(triggered()), SLOT(splitSegments()));
 
     const auto firstHandle = *mSelectedHandles.begin();
-    const auto secondHandle = *(mSelectedHandles.begin() + 1);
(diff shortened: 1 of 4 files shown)


src/tiled/editpolygontool.cpp
```diff
@@ -870,3 +890,38 @@ void EditPolygonTool::splitSegments()
...
+        for (int i = maxIndex; i < polygon.size(); ++i)
+            newPolygon[i - maxIndex] = polygon[i];
+
+        for (int i = 0; i < maxIndex; ++i)
+            newPolygon[polygon.size() - maxIndex + i] = polygon[i];
+    }
+
+    setSelectedHandles(QSet<PointHandle*>());
+
+    mapDocument()->mapObjectModel()->setObjectPolygon(mapObject, newPolygon);
```
bjorn: Instead of setting it here, you could use the `ChangePolygon` constructor that takes both the old and the new polygon.

bjorn:
I pushed a small change that was more in the direction where I was trying to hint in my last review.

In general, I think users will wonder:

* Why this action does not work for splitting polylines (of course, that would create multiple objects, but why not)
* Why it doesn't work on possibly multiple selected edges, like the "Split Segments" action
* And just now, I realized we now have two terms for the same thing "edge" and "segment". I guess this action should be called "Delete Segment".

ketanhwr:
> Why it doesn't work on possibly multiple selected edges, like the "Split Segments" action.

The first and second point mainly refer to the same point. Personally, I don't think such an action would be common. 

> And just now, I realized we now have two terms for the same thing "edge" and "segment". I guess this action should be called "Delete Segment".

Right, I'll rename it.

bjorn:
> The first and second point mainly refer to the same point. Personally, I don't think such an action would be common.

I think it's quite normal to spend most of the time working on stuff that will be rarely used, but let's have this action as-is then and spend time on other things then. :-)

PR #1688, opened 2017-08-16
WangTiles now supports infinite maps

Bdtrotte:
Fixed tests as well

### DIFF
--- a/src/tiled/wangbrush.cpp
+++ b/src/tiled/wangbrush.cpp
@@ -365,9 +365,6 @@ void WangBrush::captureHoverColor()
 
     QPoint mousePoint = mPaintPoint - tileLayer->position();
 
-    if (!tileLayer->contains(mousePoint) && !mapDocument()->map()->infinite())
-        return;
-
     const Cell &cell = tileLayer->cellAt(mousePoint);
 
     if (WangId wangId = mWangSet->wangIdOfCell(cell)) {
@@ -544,7 +541,7 @@ void WangBrush::updateBrush()
                 continue;
 
             QPoint p = adjacentPositions[i];
-            if ((!currentLayer->contains(p) && !mapDocument()->map()->infinite()) || currentLayer->cellAt(p).isEmpty())
+            if (currentLayer->cellAt(p).isEmpty())
                 continue;
 
             WangId wangId = mWangSet->wangIdOfCell(currentLayer->cellAt(p));
@@ -622,7 +619,7 @@ void WangBrush::updateBrush()
             for (int i = 0; i < 4; ++i) {
                 QPoint p = adjacentPoints[i];
 
-                if ((!currentLayer->contains(p) && !mapDocument()->map()->infinite()) || currentLayer->cellAt(p).isEmpty())
+                if (currentLayer->cellAt(p).isEmpty())
                     continue;
 
                 WangId wangId = mWangSet->wangIdOfCell(currentLayer->cellAt(p));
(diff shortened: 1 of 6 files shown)


src/tiled/wangbrush.cpp
```diff
@@ -544,7 +544,7 @@ void WangBrush::updateBrush()
                 continue;
 
             QPoint p = adjacentPositions[i];
-            if (!currentLayer->contains(p) || currentLayer->cellAt(p).isEmpty())
+            if ((!currentLayer->contains(p) && !mapDocument()->map()->infinite()) || currentLayer->cellAt(p).isEmpty())
```
bjorn: You can drop the `(!currentLayer->contains(p) && !mapDocument()->map()->infinite()) ||` part now, since just `currentLayer->cellAt(p).isEmpty()` will work in both cases.

src/tiled/wangbrush.cpp
```diff
@@ -622,7 +622,7 @@ void WangBrush::updateBrush()
             for (int i = 0; i < 4; ++i) {
                 QPoint p = adjacentPoints[i];
 
-                if (!currentLayer->contains(p) || currentLayer->cellAt(p).isEmpty())
+                if ((!currentLayer->contains(p) && !mapDocument()->map()->infinite()) || currentLayer->cellAt(p).isEmpty())
```
bjorn: Can also be just `if (currentLayer->cellAt(p).isEmpty())`.

src/tiled/wangbrush.cpp
```diff
@@ -681,7 +681,7 @@ void WangBrush::updateBrush()
                 dirPoint = mPaintPoint + aroundTilePoints[mEdgeDir*2];
             }
 
-            if (currentLayer->contains(mPaintPoint)) {
+            if (mapDocument()->map()->infinite() || currentLayer->contains(mPaintPoint)) {
```
bjorn: Since the code nested in this condition actually only works on non-empty cells, the above condition can be dropped entirely. Actually, you can also drop the `!currentLayer->cellAt(mPaintPoint).isEmpty()` check, since in that case the `wangId` is always 0 anyway.

Same for the below code dealing with `dirPoint`.

tests/staggeredrenderer/test_staggeredrenderer.cpp
```diff
@@ -102,8 +102,10 @@ void test_StaggeredRenderer::screenToTileCoords()
     QFETCH(QPointF, tileCoords);
 
     StaggeredRenderer renderer(mMap);
-    QCOMPARE(renderer.screenToTileCoords(screenCoords).toPoint(), QPoint(qFloor(tileCoords.x()),
-                                                                         qFloor(tileCoords.y())));
+
+    QPointF point = renderer.screenToTileCoords(screenCoords);
+
+    QCOMPARE(QPoint(qFloor(point.x()), qFloor(point.y())), tileCoords.toPoint());
```
bjorn: Whoops, should have noticed that. Thanks for fixing!

src/tiled/wangfiller.cpp
```diff
@@ -285,7 +287,7 @@ WangId WangFiller::wangIdFromSurroundings(const TileLayer &back,
     getSurroundingPoints(point, mStaggeredRenderer, mStaggerAxis, adjacentPoints);
 
     for (int i = 0; i < 8; ++i) {
-        if (!fillRegion.contains(adjacentPoints[i]) && back.contains(adjacentPoints[i]))
+        if (!fillRegion.contains(adjacentPoints[i]) && (mIsInfinite || back.contains(adjacentPoints[i])))
```
bjorn: Since `cellAt` will just return an empty cell and no longer asserts when the position is outside of the layer, the ` && (mIsInfinite || back.contains(adjacentPoints[i]))` part can be left out entirely.

src/tiled/wangfiller.cpp
```diff
@@ -248,9 +250,9 @@ const Cell &WangFiller::getCell(const TileLayer &back,
                                 const QRegion &fillRegion,
                                 QPoint point) const
 {
-    if (!fillRegion.contains(point) && back.contains(point)) {
+    if (!fillRegion.contains(point) && (mIsInfinite || back.contains(point))) {
         return back.cellAt(point);
-    } else if (front.contains(point.x() - front.x(), point.y() - front.y())) {
+    } else if ((mIsInfinite || front.contains(point.x() - front.x(), point.y() - front.y()))) {
```
bjorn: This check can actually be dropped entirely, so we can change this to `else` and drop the current `else` part.

src/tiled/wangfiller.cpp
```diff
@@ -248,9 +250,9 @@ const Cell &WangFiller::getCell(const TileLayer &back,
                                 const QRegion &fillRegion,
                                 QPoint point) const
 {
-    if (!fillRegion.contains(point) && back.contains(point)) {
+    if (!fillRegion.contains(point) && (mIsInfinite || back.contains(point))) {
```
bjorn: Outside the fill region I guess we should always take the tile from `back`? Cause in that case we can drop the rest of the condition.

PR #1689, opened 2017-08-18
Shape Fill Tool

Bdtrotte:
Allows users to click and drag to fill an area with a selected fill method and polygon.


Bdtrotte:
Still needs some graphics updates. Generally at this stage I'm looking for feed back on the general flow of the tool.

bjorn:
Please don't merge `master` into your branch, but rather rebase your branch on top of `master`.

bjorn:
I've tried this out and while I think it's interesting, I am having trouble thinking of the use-case for the custom polygon shape filling. Where did you get this idea from? I personally think supporting rectangle and circle shapes should be enough. It's a bit unfortunate since I can see a huge amount of effort has been put into it, but I had no idea you were planning to implement this.

I found it a bit surprising that the shapes are stored in the map file. What was the thought behind this? I would probably put them in the Tiled settings.

The ellipse seems to be of somewhat low quality, or at the very least it isn't symmetrical (I wonder how `QRegion::Ellipse` is implemented...). Check out the `pointsOnEllipse` function in `geometry.cpp` for a function generating a nicely shaped ellipse, though it returns a list of points. It could probably be copied and adapted to return a region.

For the rectangular filling mode, I think it would be interesting to add another option for doing a "border fill". Often when painting walls, carpets or other rectangular stuff with tiles, you have 4 corner tiles and 4 edge tiles and then one or more tiles to fill the center. It would be cool have the option to tile the stamp this way.

src/tiled/shapefilltool.cpp
```diff
@@ -219,14 +220,17 @@ void ShapeFillTool::updateFillOverlay()
     int left = std::min(mStartCorner.x(), mStartCorner.x() + width);
     int top = std::min(mStartCorner.y(), mStartCorner.y() + height);
 
+    QRect boundingRect(left, top, abs(width), abs(height));
+
     switch (mCurrentShape) {
     case Rect:
-        mFillRegion = QRegion(left, top, abs(width), abs(height),
-                              QRegion::Rectangle);
+        mFillRegion = QRegion(boundingRect, QRegion::Rectangle);
```
bjorn: Since rectangle is the default and the constructor taking a rectangle implicit, this could be shortened to:
```c++
mFillRegion = boundingRect;
```

src/tiled/geometry.cpp
```diff
@@ -97,6 +97,71 @@ QVector<QPoint> pointsOnEllipse(int x0, int y0, int x1, int y1)
...
+    x = radiusX;
+    y = 0;
+    xChange = radiusY * radiusY * (1 - 2 * radiusX);
+    yChange = radiusX * radiusX;
+    ellipseError = 0;
+    stoppingX = twoYSquare*radiusX;
+    stoppingY = 0;
+    while (stoppingX >= stoppingY) {
+        ret += QRect(x, -y, 1, y * 2);
+        ret += QRect(-x, -y, 1, y * 2);
```
bjorn: I think QRegion stores the list of rectangles spanning horizontally, so it may be more efficient do add horizontal rects than vertical rects.

bjorn:
I think the circle drawing is still acting strange, but I couldn't get it to play nice after playing around with it for a while. In the end I had this:
```c++
    case Circle: {
        int radiusX = boundingRect.width() / 2;
        int radiusY = boundingRect.height() / 2;
        mFillRegion = ellipseRegion(boundingRect.left() + radiusX,
                                    boundingRect.top() + radiusY,
                                    boundingRect.left() + radiusX * 2,
                                    boundingRect.top() + radiusY * 2);
        break;
    }
```
Which at least stopped it from wiggling on the left and top sides, but the weird thing is that this would refuse to draw square circles, though in fact the Stamp Brush has no problems drawing square circles. Maybe you have an idea what's going on?

Bdtrotte:
Strange... When I added in the code it still draws square circles. Also I think the wiggling has just moved to the right and bottom sides (which is noticeable when dragging to the left or above the first corner). We could just make the center point the point where you click so it acts like the stamp tool.

```
mFillRegion = ellipseRegion(mStartCorner.x(),
                            mStartCorner.y(),
                            mStartCorner.x() + width,
                            mStartCorner.y() + height);
```

bjorn:
> We could just make the center point the point where you click so it acts like the stamp tool.

I think that's a reasonable solution. I've pushed that change and rebased on latest master, to resolve the conflict in the `tiled.qrc` file.

PR #1690, opened 2017-08-20
Made Wang Pattern and color view into a tab widget

Bdtrotte:
As well as removed the unused `mIsInfinite` from the `WangFiller`


bjorn:
I've just added `mTemplateAndColorView->setDocumentMode(true);` for the styling. Thanks!

PR #1692, opened 2017-08-20
Fix automap

ketanhwr:
This patch aims to fix the automap issue in infinite-maps.

### DIFF
--- a/src/libtiled/tilelayer.cpp
+++ b/src/libtiled/tilelayer.cpp
@@ -197,20 +197,17 @@ void Tiled::TileLayer::setCell(int x, int y, const Cell &cell)
 
 TileLayer *TileLayer::copy(const QRegion &region) const
 {
-    const QRect bounds = region.boundingRect();
     const QRect areaBounds = region.boundingRect();
-    const int offsetX = qMax(0, areaBounds.x() - bounds.x());
-    const int offsetY = qMax(0, areaBounds.y() - bounds.y());
 
     TileLayer *copied = new TileLayer(QString(),
                                       0, 0,
-                                      bounds.width(), bounds.height());
+                                      areaBounds.width(), areaBounds.height());
 
     for (const QRect &rect : region.rects())
         for (int x = rect.left(); x <= rect.right(); ++x)
             for (int y = rect.top(); y <= rect.bottom(); ++y)
-                copied->setCell(x - areaBounds.x() + offsetX,
-                                y - areaBounds.y() + offsetY,
+                copied->setCell(x - areaBounds.x(),
+                                y - areaBounds.y(),
                                 cellAt(x, y));
 
     return copied;
@@ -236,7 +233,10 @@ void TileLayer::setCells(int x, int y, TileLayer *layer,
                          const QRegion &mask)
 {
     // Determine the overlapping area
-    QRegion area = QRect(x, y, layer->width(), layer->height());
+    int width = qMax(layer->width(), layer->bounds().width());
+    int height = qMax(layer->height(), layer->bounds().height());
+
+    QRegion area = QRect(x, y, width, height);
 
     if (!mask.isEmpty())
         area &= mask;
--- a/src/libtiled/tilelayer.cpp
+++ b/src/libtiled/tilelayer.cpp
@@ -635,14 +635,21 @@ Layer *TileLayer::mergedWith(Layer *other) const
     return merged;
 }
 
-QRegion TileLayer::computeDiffRegion(const TileLayer *other) const
+QRegion TileLayer::computeDiffRegion(const TileLayer *other, const QRect rect) const
 {
     QRegion ret;
 
     const int dx = other->x() - mX;
     const int dy = other->y() - mY;
-    QRect r = QRect(0, 0, width(), height());
-    r &= QRect(dx, dy, other->width(), other->height());
+
+    QRect r;
+
+    if (rect.isEmpty()) {
+        QRect r = QRect(0, 0, width(), height());
+        r &= QRect(dx, dy, other->width(), other->height());
+    } else {
+        r = rect;
+    }
 
     for (int y = r.top(); y <= r.bottom(); ++y) {
         for (int x = r.left(); x <= r.right(); ++x) {
--- a/src/libtiled/tilelayer.cpp
+++ b/src/libtiled/tilelayer.cpp
@@ -635,21 +635,14 @@ Layer *TileLayer::mergedWith(Layer *other) const
     return merged;
 }
 
-QRegion TileLayer::computeDiffRegion(const TileLayer *other, const QRect rect) const
+QRegion TileLayer::computeDiffRegion(const TileLayer *other) const
 {
     QRegion ret;
 
     const int dx = other->x() - mX;
     const int dy = other->y() - mY;
 
-    QRect r;
-
-    if (rect.isEmpty()) {
-        QRect r = QRect(0, 0, width(), height());
-        r &= QRect(dx, dy, other->width(), other->height());
-    } else {
-        r = rect;
-    }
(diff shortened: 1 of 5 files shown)


src/tiled/automapper.cpp
```diff
@@ -809,11 +809,18 @@ void AutoMapper::copyTileRegion(const TileLayer *srcLayer, int srcX, int srcY,
...
+    int startX = dstX;
+    int startY = dstY;
 
-    const int endX = qMin(dstX + width, dstLayer->width());
-    const int endY = qMin(dstY + height, dstLayer->height());
+    int endX = dstX + width;
+    int endY = dstY + height;
+
+    if (!mMapWork->infinite()) {
+        startX = qMax(dstX, startX);
```
bjorn: `dstX` and `startX` are the same at this point. You meant `0` instead of `dstX`.

src/libtiled/tilelayer.cpp
```diff
@@ -635,14 +635,21 @@ Layer *TileLayer::mergedWith(Layer *other) const
...
-    r &= QRect(dx, dy, other->width(), other->height());
+
+    QRect r;
+
+    if (rect.isEmpty()) {
+        QRect r = QRect(0, 0, width(), height());
+        r &= QRect(dx, dy, other->width(), other->height());
+    } else {
+        r = rect;
+    }
```
bjorn: Why do you need the `rect` parameter? As far as I can see, `r` just needs to cover both layers and it should be in local coordinates, so this should do it:
```c++
const QRect r = bounds().united(other->bounds()).translated(-position());
```

ketanhwr:
No, it doesn't work if the automap rules are infinite as well, right now.

src/tiled/automapper.cpp
```diff
@@ -691,7 +691,7 @@ static bool compareLayerTo(const TileLayer *setLayer,
                 bool matchListNo  = false;
 
 
-                if (!setLayer->contains(x + offset.x(), y + offset.y())) {
+                if (!(setLayer->contains(x + offset.x(), y + offset.y()) || setLayer->map()->infinite())) {
```
bjorn: Intuitively, this check and the special handling inside should actually not be necessary at all. If the `setLayer` does not contain the given location, then `c1` will just be an empty cell, which I think should be treated the same as a cell outside of the layer boundaries. Can you check this?

ketanhwr: No no, it needs to have that check. I found it strange too, but it doesn't work if this check is removed.

bjorn: Can you give an example of where it goes wrong?

ketanhwr: This is what happens when I remove the check:
screenshot from 2017-08-21 17-10-38
This is what happens when I keep the check (this obviously is the expected behaviour): 
screenshot from 2017-08-21 17-11-05

bjorn: What part did you try removing exactly? I suggested removing this entire bit of code:
```c++
                if (!(setLayer->contains(x + offset.x(), y + offset.y()) || setLayer->map()->infinite())) {
                    foreach (const TileLayer *comparedTileLayer, listYes) {
                        if (!comparedTileLayer->contains(x, y))
                            return false;

                        const Cell &c2 = comparedTileLayer->cellAt(x, y);
                        if (!c2.isEmpty())
                            return false;
                    }
                    continue;
                }
```
It's not clear to me what is happening from those screenshots.

ketanhwr: Aah, I thought you meant `setLayer->map()->infinite()`

bjorn:
Reminder to self, or when you have time: need to check this change against issue #1224.

bjorn:
Alright, I checked this against issue #1224 and unfortunately it re-introduces that issue. However, I think the consistency is more important than getting that particular example to work. The example in issue #1224 could be adjusted to have separate input and output regions in order to make the rule work at the border.

However, I found an issue with the automapping in that even after this patch it doesn't work with infinite maps outside of the already used bounds. Please open this [example.zip](https://github.com/bjorn/tiled/files/1250236/example.zip) and try to place a rock somewhere. You should see that the first time it works, but then when you paint rocks in many places, it only places the flowers within the chunks that were allocated the first time the automapping was activated.

ketanhwr:
Oh :confused: 

I'll try to fix this.

ketanhwr:
What I've observed is when we apply automap and then paint outside the bounds and try automap again it doesn't work. Then if we undo the first automap and then paint outside the bounds and then try automap again, it works. What exactly is stored after calling 'Automap' for the first time ?

ketanhwr:
Okay, so after spending the past 3 hours into this, I can say that I've wasted the whole time :slightly_frowning_face: 

So actually, automapping **works** even if we paint out-of-bounds. If you test after my previous commit, you cannot _visually see_ the automapped tile layer. If you save it and then restart Tiled, the automapped tile layer is clearly visible :man_facepalming: 

I have no idea why this is happening. Here's a gif of the same.

peek 2017-08-26 20-20

:cry:

src/libtiled/tilelayer.cpp
```diff
@@ -236,7 +233,10 @@ void TileLayer::setCells(int x, int y, TileLayer *layer,
                          const QRegion &mask)
 {
     // Determine the overlapping area
-    QRegion area = QRect(x, y, layer->width(), layer->height());
+    int width = qMax(layer->width(), layer->bounds().width());
+    int height = qMax(layer->height(), layer->bounds().height());
```
bjorn: I think this makes little sense, please revert it back to what it was.

ketanhwr: But why? Suppose a layer was created with size (1, 1) and then some cells were set out-of-bounds. If this layer is then used in `setCells` for some other layer, then technically, these 'out-of-bounds' cells should also be set in this layer. Don't you think?

ketanhwr:
Okay, automap is fixed now. But still some problem persists. When we undo, we need to zoom in/out or move around to see the effect of undo. This is because when we undo, the bounds of Tile Layer are not changed. But I don't think it's that big of a problem.

src/libtiled/tilelayer.cpp
bjorn: Well, the part that does not make sense, is to take the width/height of the bounds but still copy starting from 0,0.

bjorn:
Right, so this signal should actually be made part of the undo command. Would that be hard to do?

ketanhwr:
> This is because when we undo, the bounds of Tile Layer are not changed.

This is primarily the reason, the signal cannot emit. Since the bounds do not change after removing cells, we cannot actually check if it is necessary to emit this signal.

src/libtiled/tilelayer.cpp
ketanhwr: So I guess we need to unite `QRect(0, 0, layer->width(), layer->height()` with `layer->bounds()`.

ketanhwr: And adjust the offset accordingly.

bjorn:
> This is primarily the reason, the signal cannot emit. Since the bounds do not change after removing cells, we cannot actually check if it is necessary to emit this signal.

If the bounds do not change, it is not necessary to emit the signal. Then there must be some other thing missing to trigger an update.

src/libtiled/tilelayer.cpp
bjorn: No, if you use the bounds then the `rect()` of the layer is completely irrelevant.

ketanhwr: I got confused by something else :confused:

ketanhwr:
Done! :slightly_smiling_face:

bjorn:
Nice, thanks!

PR #1693, opened 2017-08-21
Option to split polyline

ketanhwr:
If two consecutive nodes are selected in a polyline and then the segment between them can be deleted to split polyline into two polylines.

### DIFF
--- a/src/tiled/changepolygon.cpp
+++ b/src/tiled/changepolygon.cpp
@@ -76,3 +76,50 @@ void TogglePolygonPolyline::toggle()
 {
     mMapObject->setShape((mMapObject->shape() == MapObject::Polygon) ? MapObject::Polyline : MapObject::Polygon);
 }
+
+SplitPolyline::SplitPolyline(MapDocument *mapDocument,
+                             MapObject *mapObject,
+                             int index)
+    : mMapDocument(mapDocument)
+    , mFirstPolyline(mapObject)
+    , mEdgeIndex(index)
+    , mObjectIndex(-1)
+{
+    mSecondPolyline = mFirstPolyline->clone();
+    mSecondPolyline->resetId();
+
+    setText(QCoreApplication::translate("Undo Commands", "Split Polyline"));
+}
+
+void SplitPolyline::undo()
+{
+    mObjectIndex = mMapDocument->mapObjectModel()->removeObject(mFirstPolyline->objectGroup(),
+                                                                mSecondPolyline);
+
+    QPolygonF firstPolygon = mFirstPolyline->polygon();
+    QPolygonF secondPolygon = mSecondPolyline->polygon();
+
+    firstPolygon.append(secondPolygon);
+    secondPolygon = firstPolygon;
+    mSecondPolyline->setPolygon(secondPolygon);
+
+    mMapDocument->mapObjectModel()->setObjectPolygon(mFirstPolyline, firstPolygon);
+    mFirstPolyline->setPropertyChanged(MapObject::ShapeProperty);
+}
+
+void SplitPolyline::redo()
+{
+    mMapDocument->mapObjectModel()->insertObject(mFirstPolyline->objectGroup(), mObjectIndex,
+                                                 mSecondPolyline);
+
+    QPolygonF firstPolygon = mFirstPolyline->polygon();
+    QPolygonF secondPolygon = mSecondPolyline->polygon();
+
+    firstPolygon.erase(firstPolygon.begin() + mEdgeIndex + 1, firstPolygon.end());
+    secondPolygon.erase(secondPolygon.begin(), secondPolygon.begin() + mEdgeIndex + 1);
+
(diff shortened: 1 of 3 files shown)


src/tiled/changepolygon.cpp
```diff
@@ -76,3 +76,50 @@ void TogglePolygonPolyline::toggle()
...
+
+    QPolygonF firstPolygon = mFirstPolyline->polygon();
+    QPolygonF secondPolygon = mSecondPolyline->polygon();
+
+    firstPolygon.append(secondPolygon);
+    secondPolygon = firstPolygon;
+    mSecondPolyline->setPolygon(secondPolygon);
+
+    mMapDocument->mapObjectModel()->setObjectPolygon(mFirstPolyline, firstPolygon);
+    mFirstPolyline->setPropertyChanged(MapObject::ShapeProperty);
```
bjorn: Since this is the undo, whether the property has been changed or not should be set back to the state it had when `redo` was called.

src/tiled/changepolygon.cpp
```diff
@@ -76,3 +76,50 @@ void TogglePolygonPolyline::toggle()
...
+void SplitPolyline::undo()
+{
+    mObjectIndex = mMapDocument->mapObjectModel()->removeObject(mFirstPolyline->objectGroup(),
+                                                                mSecondPolyline);
+
+    QPolygonF firstPolygon = mFirstPolyline->polygon();
+    QPolygonF secondPolygon = mSecondPolyline->polygon();
+
+    firstPolygon.append(secondPolygon);
+    secondPolygon = firstPolygon;
```
bjorn: Please skip this assignment and just do `mSecondPolyline->setPolygon(firstPolygon);`.

src/tiled/changepolygon.cpp
```diff
@@ -76,3 +76,50 @@ void TogglePolygonPolyline::toggle()
...
     mMapObject->setShape((mMapObject->shape() == MapObject::Polygon) ? MapObject::Polyline : MapObject::Polygon);
 }
+
+SplitPolyline::SplitPolyline(MapDocument *mapDocument,
+                             MapObject *mapObject,
+                             int index)
+    : mMapDocument(mapDocument)
+    , mFirstPolyline(mapObject)
+    , mEdgeIndex(index)
+    , mObjectIndex(-1)
```
bjorn: Rather than putting the new polyline as the last object in the group, it would be nice to use the index of the polyline being split + 1.

src/tiled/changepolygon.cpp
```diff
@@ -76,3 +76,50 @@ void TogglePolygonPolyline::toggle()
...
+
+void SplitPolyline::undo()
+{
+    mObjectIndex = mMapDocument->mapObjectModel()->removeObject(mFirstPolyline->objectGroup(),
+                                                                mSecondPolyline);
+
+    QPolygonF firstPolygon = mFirstPolyline->polygon();
+    QPolygonF secondPolygon = mSecondPolyline->polygon();
+
+    firstPolygon.append(secondPolygon);
```
Ablu: Clang complains here:
~~~
changepolygon.cpp:102:25: error: no viable conversion from 'QPolygonF' to

      'const QPointF'

    firstPolygon.append(secondPolygon);

                        ^~~~~~~~~~~~~

/opt/qt54/include/QtCore/qpoint.h:205:21: note: candidate constructor (the

      implicit copy constructor) not viable: no known conversion from

      'QPolygonF' to 'const QPointF &' for 1st argument

class Q_CORE_EXPORT QPointF

                    ^

/opt/qt54/include/QtCore/qpoint.h:205:21: note: candidate constructor (the

      implicit move constructor) not viable: no known conversion from

      'QPolygonF' to 'QPointF &&' for 1st argument

class Q_CORE_EXPORT QPointF

                    ^

/opt/qt54/include/QtCore/qpoint.h:270:34: note: candidate constructor not

      viable: no known conversion from 'QPolygonF' to 'const QPoint &' for 1st

      argument

Q_DECL_CONSTEXPR inline QPointF::QPointF(const QPoint &p) : xp(p.x()), y...

                                 ^

/opt/qt54/include/QtGui/qpolygon.h:135:5: note: candidate function

    operator QVariant() const;

    ^

/opt/qt54/include/QtCore/qvector.h:130:26: note: passing argument to parameter

      't' here

    void append(const T &t);

                         ^

1 error generated.
~~~

ketanhwr: Yup I saw this but couldn't understand why it was happening :confused:

bjorn: Because it's compiling against Qt 5.4, which does not have a `QVector::append` overload that will append a vector. You can use the `+=` operator instead:
```c++
firstPolygon += secondPolygon;
```

bjorn:
When you delete the first or the last segment of a polyline, it creates an additional empty polygon object. It should in this case just delete the segment.

PR #1696, opened 2017-08-22
New chunk based format for infinite maps

ketanhwr:
I've removed `startx` and `starty` attributes from Tile Layer. Now, for infinite maps, each 16x16 chunk will be stored along with its `startx` and `starty` coordinates.

### DIFF
--- a/src/libtiled/gidmapper.cpp
+++ b/src/libtiled/gidmapper.cpp
@@ -30,6 +30,7 @@
 
 #include "compression.h"
 #include "tile.h"
+#include "tiled.h"
 #include "tileset.h"
 
 using namespace Tiled;
@@ -152,18 +153,10 @@ QByteArray GidMapper::encodeLayerData(const TileLayer &tileLayer,
     Q_ASSERT(format != Map::XML);
     Q_ASSERT(format != Map::CSV);
 
-    int startX = 0;
-    int startY = 0;
-    int endX = tileLayer.width() - 1;
-    int endY = tileLayer.height() - 1;
-
-    if (tileLayer.map()->infinite()) {
-        QRect bounds = tileLayer.bounds().translated(-tileLayer.position());
-        startX = bounds.left();
-        startY = bounds.top();
-        endX = bounds.right();
-        endY = bounds.bottom();
-    }
+    const int startX = 0;
+    const int startY = 0;
+    const int endX = tileLayer.width() - 1;
+    const int endY = tileLayer.height() - 1;
 
     QByteArray tileData;
     tileData.reserve((endX - startX + 1) * (endY - startY + 1) * 4);
@@ -186,10 +179,44 @@ QByteArray GidMapper::encodeLayerData(const TileLayer &tileLayer,
     return tileData.toBase64();
 }
 
+QByteArray GidMapper::encodeChunkData(const TileLayer &tileLayer,
+                                      int chunkStartX,
+                                      int chunkStartY,
+                                      Map::LayerDataFormat format) const
+{
+    Q_ASSERT(format != Map::XML);
+    Q_ASSERT(format != Map::CSV);
+
+    int startX = chunkStartX;
+    int startY = chunkStartY;
+    int endX = chunkStartX + CHUNK_SIZE - 1;
+    int endY = chunkStartY + CHUNK_SIZE - 1;
+
+
+    QByteArray chunkData;
+    chunkData.reserve((endX - startX + 1) * (endY - startY + 1) * 4);
+
+    for (int y = startY; y <= endY; ++y) {
+        for (int x = startX; x <= endX; ++x) {
+            const unsigned gid = cellToGid(tileLayer.cellAt(x, y));
+            chunkData.append((char) (gid));
+            chunkData.append((char) (gid >> 8));
+            chunkData.append((char) (gid >> 16));
+            chunkData.append((char) (gid >> 24));
+        }
+    }
+
+    if (format == Map::Base64Gzip)
+        chunkData = compress(chunkData, Gzip);
+    else if (format == Map::Base64Zlib)
+        chunkData = compress(chunkData, Zlib);
+
+    return chunkData.toBase64();
+}
+
 GidMapper::DecodeError GidMapper::decodeLayerData(TileLayer &tileLayer,
                                                   const QByteArray &layerData,
-                                                  Map::LayerDataFormat format,
-                                                  int startX, int startY) const
+                                                  Map::LayerDataFormat format) const
 {
     Q_ASSERT(format != Map::XML);
     Q_ASSERT(format != Map::CSV);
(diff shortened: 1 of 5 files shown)


src/libtiled/mapreader.cpp
```diff
@@ -825,42 +821,96 @@ void MapReaderPrivate::readTileLayerData(TileLayer &tileLayer,
...
+                                                      startX,
+                                                      startY);
+                            } else if (encoding == QLatin1String("csv")) {
+                                decodeCSVLayerData(tileLayer, xml.text(), startX, startY);
+                            }
+                        }
+                    }
                 }
+            }
+        }
```
Ablu: I think it would make sense to move this (and the following else branch) into a seperate function to make the code a bit easier to read.

That many braces closing in the same place is usually a good indicator that the code could be moved into a function...

src/libtiled/mapreader.cpp
```diff
@@ -825,42 +821,96 @@ void MapReaderPrivate::readTileLayerData(TileLayer &tileLayer,
...
+                } else {
+                    readUnknownElement();
+                }
+            } else if (xml.isCharacters() && !xml.isWhitespace()) {
+                if (encoding == QLatin1String("base64")) {
+                    decodeBinaryLayerData(tileLayer,
+                                          xml.text().toLatin1(),
+                                          layerDataFormat,
+                                          0,
+                                          0);
```
Ablu: I guess the 0 does not need to be in an extra line now ;)

src/libtiled/mapwriter.cpp
```diff
@@ -648,41 +649,105 @@ void MapWriterPrivate::writeTileLayer(QXmlStreamWriter &w,
...
+
+        for (int y = startY; y < endY; ++y) {
+            for (int x = startX; x < endX; ++x) {
+                if (tileLayer.findChunk(chunkStartX, chunkStartY)) {
+                    w.writeStartElement(QLatin1String("chunk"));
+                    w.writeAttribute(QLatin1String("startx"), QString::number(chunkStartX));
+                    w.writeAttribute(QLatin1String("starty"), QString::number(chunkStartY));
+
+                    if (mLayerDataFormat == Map::XML) {
+                        for (int y1 = 0; y1 < CHUNK_SIZE; y1++) {
```
Ablu: Maybe use `cellY` here (or `chunkY` before)? I find adding numbers to variables to prevent name clashes a bit confusing often.

ketanhwr:
> Generally I wonder whether it would maybe also make sense to store non-infinite maps in the chunk based format... In some cases this could help to speed up the processing of the file (and file size if no compression is used...).

Although this would be quite helpful, but this would break all the current maps :confused:

Ablu:
> Although this would be quite helpful, but this would break all the current maps

Sure... It would have to be configurable somehow... But I am not sure how such a configuration could look like...

src/libtiled/gidmapper.cpp
```diff
@@ -186,16 +179,96 @@ QByteArray GidMapper::encodeLayerData(const TileLayer &tileLayer,
     return tileData.toBase64();
 }
 
+QByteArray GidMapper::encodeChunkData(const TileLayer &tileLayer,
```
bjorn: Rather than duplicating this whole function, please just add a `QRect` parameter. If the provided rect is empty it could probably default to `QRect(0, 0, tileLayer.width(), tileLayer.height())`.

src/libtiled/gidmapper.cpp
```diff
@@ -186,16 +179,96 @@ QByteArray GidMapper::encodeLayerData(const TileLayer &tileLayer,
...
+        chunkData = compress(chunkData, Gzip);
+    else if (format == Map::Base64Zlib)
+        chunkData = compress(chunkData, Zlib);
+
+    return chunkData.toBase64();
+}
+
 GidMapper::DecodeError GidMapper::decodeLayerData(TileLayer &tileLayer,
                                                   const QByteArray &layerData,
+                                                  Map::LayerDataFormat format) const
```
bjorn: Similarly, this function could take a `QRect` to specify the target area of the decoded layer data.

src/libtiled/mapreader.cpp
bjorn: Yes, the reading of the inner part of either `data` itself or each of its `chunk` elements should be done in a separate function rather than duplicating the whole code. Property this function again needs to take a target `QRect` so that it knows where to set the tiles.

src/libtiled/mapreader.cpp
```diff
@@ -897,23 +952,30 @@ void MapReaderPrivate::decodeCSVLayerData(TileLayer &tileLayer,
     QString trimText = text.trimmed().toString();
     QStringList tiles = trimText.split(QLatin1Char(','));
 
-    if (tiles.length() != tileLayer.width() * tileLayer.height()) {
+    int lengthCheck = (mMap->infinite()) ? CHUNK_SIZE * CHUNK_SIZE :
```
bjorn: `decodeCSVLayerData` should take a `QRect` parameter specifying the size and position of the tiles it is decoding.

src/libtiled/mapwriter.cpp
bjorn: Yes, but actually this whole code should not be duplicated in the first place. It should be moved into a function taking the QXmlStreamWriter, the tile layer and a `QRect` specifying which part of the layer to store. That function can then be called either once for the whole tile layer, or repeatedly for each chunk, depending on whether the map is infinite.

src/libtiled/mapwriter.cpp
```diff
@@ -715,43 +681,55 @@ void MapWriterPrivate::writeTileLayer(QXmlStreamWriter &w,
...
-    w.writeEndElement(); // </layer>
+        w.writeCharacters(QLatin1String("\n"));
+        w.writeCharacters(chunkData);
+    } else {
+        QByteArray chunkData = mGidMapper.encodeLayerData(tileLayer,
+                                                          mLayerDataFormat,
+                                                          QRect(bounds.x(),
+                                                                bounds.y(),
+                                                                bounds.width(),
+                                                                bounds.height()));
```
bjorn: Just pass `bounds`?

src/libtiled/mapreader.cpp
```diff
@@ -825,6 +823,30 @@ void MapReaderPrivate::readTileLayerData(TileLayer &tileLayer,
...
 
+    if (mMap->infinite()) {
+        while (xml.readNext() != QXmlStreamReader::Invalid) {
+            if (xml.isEndElement()) {
+                break;
+            } else if (xml.isStartElement()) {
+                if (xml.name() == QLatin1String("chunk")) {
+                    const QXmlStreamAttributes atts = xml.attributes();
+                    int startX = atts.value(QLatin1String("startx")).toInt();
+                    int startY = atts.value(QLatin1String("starty")).toInt();
```
bjorn: Please have attributes `x`, `y`, `width` and `height` for `chunk` elements. The file format should not rely on a hard-coded internal chunk size. If you write it in the file it is much more flexible.

src/libtiled/mapwriter.cpp
ketanhwr: :man_facepalming:

src/libtiled/mapwriter.cpp
```diff
@@ -648,45 +650,83 @@ void MapWriterPrivate::writeTileLayer(QXmlStreamWriter &w,
...
+                }
+
+                chunkStartX += CHUNK_SIZE;
+            }
+
+            chunkStartX = resetX;
+            chunkStartY += CHUNK_SIZE;
+        }
+    } else {
+        writeTileLayerData(w, tileLayer, QRect(0, 0, endX + 1, endY + 1));
```
bjorn: Judging by the code above, `endX + 1` and `endY + 1` can't possibly be the right values here, and should rather be `tileLayer.width()` and `tileLayer.height()`.

src/libtiled/mapreader.cpp
```diff
@@ -825,26 +823,52 @@ void MapReaderPrivate::readTileLayerData(TileLayer &tileLayer,
 
     mMap->setLayerDataFormat(layerDataFormat);
 
-    int x = 0;
-    int y = 0;
+    if (mMap->infinite()) {
```
bjorn: It's a pity the reading code relies on checking whether the map it's reading is infinite or not. I believe this really shouldn't be necessary, and not doing so would help with forward-compatibility when eventually we may support chunked storage for finite maps. Can you think of a way to handle the chunks without this check?

src/libtiled/mapwriter.cpp
ketanhwr:
```
int startX = 0;
int startY = 0;
int endX = tileLayer.width() - 1;
int endY = tileLayer.height() - 1;
```

`endX + 1` and `endY + 1` are same as `tileLayer.width()` and `tileLayer.height()`.

src/libtiled/mapreader.cpp
ketanhwr: I'm not able to think up of a solution. I don't see how we could remove this check, atleast as of now.

src/libtiled/mapwriter.cpp
bjorn: Ah, that's a little confusing. Please just use `tileLayer.width()` and `tileLayer.height()` here and move those variables local to the other branch.

bjorn:
Alright, I'll have a look later at trying to remove the infinite check on reading, but otherwise it looks good, thanks!

PR #1698, opened 2017-08-22
Various improvements for templates

thabetx:
This pull request contains various minor improvements to templates

- [x] Enable detaching template instances from the template.
- [x] Disable some tools when using templates.
- [x] Create Template instance indicator in the properties dock.
- [x] Convert an object to a template instance after saving it as a template.
- [x] Removed code duplication for creating new template group.

I'm not sure about these two
- [x] Select all instances in the current map when a template is selected.
- [x] Change the format from ttx to something else like ttg.

### DIFF
--- a/src/tiled/abstractobjecttool.cpp
+++ b/src/tiled/abstractobjecttool.cpp
@@ -333,10 +333,18 @@ void AbstractObjectTool::showContextMenu(MapObjectItem *clickedObjectItem,
     }
 
     if (selectedObjects.size() == 1) {
-        // Saving objects with embedded tilesets is disabled
-        auto cell = selectedObjects.first()->cell();
-        if (cell.isEmpty() || cell.tileset()->isExternal())
-            menu.addAction(tr("Save As Template"), this, SLOT(saveSelectedObject()));
+        MapObject *currentObject = selectedObjects.first();
+        if (!(currentObject->isTemplateBase() || currentObject->isTemplateInstance())) {
+            const Cell cell = selectedObjects.first()->cell();
+            // Saving objects with embedded tilesets is disabled
+            if (cell.isEmpty() || cell.tileset()->isExternal())
+                menu.addAction(tr("Save As Template"), this, SLOT(saveSelectedObject()));
+        }
+
+        if (currentObject->isTemplateBase()) { // Hide this operations for template base
+            duplicateAction->setVisible(false);
+            removeAction->setVisible(false);
+        }
     }
 
     bool anyIsTemplateInstance = std::any_of(selectedObjects.begin(),
--- a/src/tiled/abstractobjecttool.cpp
+++ b/src/tiled/abstractobjecttool.cpp
@@ -209,6 +209,37 @@ void AbstractObjectTool::saveSelectedObject()
         mapDocument()->saveSelectedObject(name, groupIndex);
 }
 
+void AbstractObjectTool::detachSelectedObjects()
+{
+    MapDocument *currentMapDocument = mapDocument();
+    QList<MapObject *> templateInstances;
+
+    /**
+     * Stores the unique tilesets used by the templates
+     * to avoid creating multiple undo commands for the same tileset
+     */
+    QSet<SharedTileset> sharedTilesets;
+
+    for (MapObject *object : mapDocument()->selectedObjects()) {
+        if (object->templateObject()) {
+            templateInstances.append(object);
+
+            if (Tile *tile = object->cell().tile())
+                sharedTilesets.insert(tile->tileset()->sharedPointer());
+        }
+    }
+
+    auto changeMapObjectCommand = new DetachObjects(currentMapDocument, templateInstances);
+
+    // Add any missing tileset used by the templates to the map map before detaching
+    for (SharedTileset sharedTileset : sharedTilesets) {
+        if (!currentMapDocument->map()->tilesets().contains(sharedTileset))
+            new AddTileset(currentMapDocument, sharedTileset, changeMapObjectCommand);
+    }
+
+    currentMapDocument->undoStack()->push(changeMapObjectCommand);
+}
+
 void AbstractObjectTool::changeTile()
 {
     QList<MapObject*> tileObjects;
@@ -307,6 +338,13 @@ void AbstractObjectTool::showContextMenu(MapObjectItem *clickedObjectItem,
         if (cell.isEmpty() || cell.tileset()->isExternal())
             menu.addAction(tr("Save As Template"), this, SLOT(saveSelectedObject()));
     }
+
+    bool anyIsTemplateInstance = std::any_of(selectedObjects.begin(),
+                                             selectedObjects.end(),
+                                             [](MapObject *object) { return object->isTemplateInstance(); });
+
+    if (anyIsTemplateInstance)
+        menu.addAction(tr("Detach"), this, SLOT(detachSelectedObjects()));
 
     menu.addSeparator();
     menu.addAction(tr("Flip Horizontally"), this, SLOT(flipHorizontally()), QKeySequence(tr("X")));
--- a/src/tiled/changemapobject.cpp
+++ b/src/tiled/changemapobject.cpp
@@ -158,3 +158,48 @@ void ChangeMapObjectsTile::changeTiles()
 
     emit mMapDocument->mapObjectModel()->objectsChanged(mMapObjects);
 }
+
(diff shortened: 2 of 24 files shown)


bjorn:
> I'm not sure about these two
> 
> * Select all instances in the current map when a template is selected.
> * Change the format from ttx to something else like ttg.

What about adding a context menu item to the template items with the action to select all instances?

Regarding the extension, what about "tgx"? I think this is what I had in mind with "tmx" (tile map XML), "tsx" (tile set XML) and now "template group XML".

src/tiled/abstractobjecttool.cpp
```diff
@@ -302,12 +307,27 @@ void AbstractObjectTool::showContextMenu(MapObjectItem *clickedObjectItem,
     }
 
     if (selectedObjects.size() == 1) {
-        // Saving objects with embedded tilesets is disabled
-        auto cell = selectedObjects.first()->cell();
-        if (cell.isEmpty() || cell.tileset()->isExternal())
-            menu.addAction(tr("Save As Template"), this, SLOT(saveSelectedObject()));
+        MapObject *currentObject = selectedObjects.first();
+        if (!(currentObject->isTemplateBase() || currentObject->isTemplateInstance())) {
```
bjorn: I'm not sure if it makes sense to hide the "Save As Template" action for template instances, given that you could instantiate a template, override some properties, and then want to save the result as another template.

On the other hand, the user may expect one template to derive from another template then, which might be interesting but also a major pain to add support for currently. So maybe it's good to avoid confusion and force a Detach first.

src/tiled/changemapobject.cpp
```diff
@@ -158,3 +158,35 @@ void ChangeMapObjectsTile::changeTiles()
...
+
+void DetachObject::redo()
+{
+    // Merge the instance properties into the template properties
+    Properties newProperties = mMapObject->templateObject()->properties();
+    newProperties.merge(mMapObject->properties());
+    mMapObject->setProperties(newProperties);
+
+    mMapObject->setTemplateRef({nullptr, 0});
+    mMapObject->syncWithTemplate();
```
bjorn: Why do a `syncWithTemplate` call after setting the template reference to 0? Did you mean to do that before, just in case? I guess it can just be removed.

src/tiled/changemapobject.h
```diff
@@ -111,5 +111,26 @@ class ChangeMapObjectsTile : public QUndoCommand
...
 };
 
+class DetachObject : public QUndoCommand
+{
+public:
+    /**
+     * Creates an undo command that detaches the given template instance from the template.
+     */
+    DetachObject(MapDocument *mapDocument,
+                 MapObject *mapObject);
```
bjorn: For efficiency reasons, please make it a `DetachObjects` undo command that takes a list of objects.

src/tiled/changemapobject.cpp
```diff
@@ -158,3 +158,35 @@ void ChangeMapObjectsTile::changeTiles()
...
+    , mMapObject(mapObject)
+{
+    mTemplateRef = mMapObject->templateRef();
+    mProperties = mMapObject->properties();
+}
+
+void DetachObject::redo()
+{
+    // Merge the instance properties into the template properties
+    Properties newProperties = mMapObject->templateObject()->properties();
```
bjorn: This is going to crash if you select both template instances and regular objects and choose "Detach". Either there should be a check here or the `detachObjects` or its caller should filter the objects to make sure they are all template instances (and in fact it should check the return value of `templateObject()` because it may be a broken template reference so `isTemplateInstance` won't be enough).

src/tiled/abstractobjecttool.cpp
thabetx: Yeah I had similar thoughts, Assuming that we don't want recursive templates, saving would internally detach then save as a template.

src/tiled/changemapobject.cpp
thabetx: I'm not sure why I added it but it does nothing.

src/tiled/changemapobject.h
thabetx: Okay, I think bulking the commands together would be more efficient. I did it on individual objects to avoid using lists inside the undo command.

thabetx:
I found out that after detaching, the tileset of the template should be added to the map document, so I will try fixing it.

thabetx:
I updated the detaching commit. Detaching now works on the objects as a whole and adds any missing tilesets to the map.

src/tiled/abstractobjecttool.cpp
```diff
@@ -209,6 +209,42 @@ void AbstractObjectTool::saveSelectedObject()
...
+    for (MapObject *object : mapDocument()->selectedObjects()) {
+        if (object->templateObject()) {
+            templateInstances.append(object);
+
+            if (Tile *tile = object->cell().tile())
+                sharedTilesets.insert(tile->tileset()->sharedPointer());
+        }
+    }
+
+    auto changeMapObjectCommand = new  QUndoCommand(
```
bjorn: One space too much after `new`.

src/tiled/abstractobjecttool.cpp
```diff
@@ -209,6 +209,42 @@ void AbstractObjectTool::saveSelectedObject()
...
+                                            "Detach %n Template Instance(s)",
+                                            nullptr, templateInstances.size()));
+
+    // Add any missing tileset used by the templates to the map map before detaching
+    for (SharedTileset sharedTileset : sharedTilesets) {
+        if (!currentMapDocument->map()->tilesets().contains(sharedTileset))
+            new AddTileset(currentMapDocument, sharedTileset, changeMapObjectCommand);
+    }
+
+    new DetachObject(currentMapDocument, templateInstances, changeMapObjectCommand);
```
bjorn: Please rename this class to `DetachObjects` now that it handles multiple.

Note that you could add support for child undo commands to `DetachObjects` by calling the parent undo/redo methods as well, in which case you can make the `AddTileset` commands children and don't need the separate `changeMapObjectCommand`.

src/tiled/changemapobject.h
```diff
@@ -111,5 +111,27 @@ class ChangeMapObjectsTile : public QUndoCommand
...
+    DetachObject(MapDocument *mapDocument,
+                 QList<MapObject *> mapObjects,
+                 QUndoCommand *parent = nullptr);
+
+    void redo() override;
+    void undo() override;
+
+private:
+    void attach();
+    void detach();
```
bjorn: You can remove these declarations.

src/tiled/abstractobjecttool.cpp
thabetx: I'm not sure what do you mean so this might not be related, I created the `AddTileset` undo action before the `DetachObjects` action to make sure that any tile replacement works as expected, otherwise the `GidMapper` won't be able to to map the tile to the correct gid. The initial QUndo command contains the `AddTileset` Actions and eventually the `DetachObjects` action as well.

bjorn: The order of execution of the undo commands doesn't matter in this case, because they do not depend on each other for their state modifications and they can't be undone or redone separately (they form a single entry in the undo history).

See other created instances of `AddTileset`, like in `AbstractObjectTool::changeTile`, for what I meant with attaching it as a child to the main undo command (and `ChangeMapObjectsTile::undo/redo`). I just meant that doing it that way avoids the need to create an additional `QUndoCommand` instance.

thabetx: Hmm, I initially it was changeTile, specifically like this:
```
    auto changeMapObjectCommand = new DetachObjects(currentMapDocument,
                                                    templateInstances);

    for (SharedTileset sharedTileset : sharedTilesets) {
        if (!currentMapDocument->map()->tilesets().contains(sharedTileset))
            new AddTileset(currentMapDocument, sharedTileset, changeMapObjectCommand);
    }

    currentMapDocument->undoStack()->push(changeMapObjectCommand);
```

But the gid of the detached instance became 0, I suspected that the order mattered so added the empty udnoCommand.

bjorn: It's because it won't magically work, you'd need to change `DetachObjects` to call its superclass in the undo/redo functions, like `ChangeMapObjectsTile::undo/redo`.

thabetx: Alright, I've updated it. I thought I had to manually order them on creation.

bjorn:
Awesome!

PR #1701, opened 2017-08-23
Improved docs 

ketanhwr:
Adding and modifying docs for everything I've added in the past few months.


bjorn:
When I read the updated part about deletion of polygon nodes, I wondered if it was really a good idea to also allow entire object deletion when using the Edit Polygons Tool. I checked with Inkscape and they will only delete selected nodes when pressing this shortcut while using the "Edit paths by nodes" tool (unless you specially select the "Delete" action via the menu). Anyway I guess it doesn't matter much.

I'll merge this now in any case.

PR #1705, opened 2017-08-24
Double click splits a segment

ketanhwr:
Double clicking on a segment splits it. Single click selects the two nodes of that particular segment.

### DIFF
--- a/src/tiled/editpolygontool.cpp
+++ b/src/tiled/editpolygontool.cpp
@@ -42,6 +42,7 @@
 #include <QMenu>
 #include <QPainter>
 #include <QPalette>
+#include <QtCore/QtMath>
 #include <QUndoStack>
 
 #include <cstdlib>
@@ -138,6 +139,7 @@ EditPolygonTool::EditPolygonTool(QObject *parent)
     , mClickedHandle(nullptr)
     , mClickedObjectItem(nullptr)
     , mMode(NoMode)
+    , mSplit(false)
 {
 }
 
@@ -238,6 +240,8 @@ void EditPolygonTool::mousePressed(QGraphicsSceneMouseEvent *event)
     if (mMode != NoMode) // Ignore additional presses during select/move
         return;
 
+    mSplit = false;
+
     switch (event->button()) {
     case Qt::LeftButton: {
         mMousePressed = true;
@@ -259,6 +263,10 @@ void EditPolygonTool::mousePressed(QGraphicsSceneMouseEvent *event)
             }
         }
         mClickedHandle = first<PointHandle>(items);
+
+        if (!mClickedHandle)
+            selectEdge(mStart, event->modifiers());
+
         break;
     }
     case Qt::RightButton: {
@@ -337,6 +345,12 @@ void EditPolygonTool::mouseReleased(QGraphicsSceneMouseEvent *event)
 
     mMousePressed = false;
     mClickedHandle = nullptr;
+}
+
+void EditPolygonTool::mouseDoubleClicked(QGraphicsSceneMouseEvent*)
+{
+    if (mSplit)
+        splitSegments();
 }
 
 void EditPolygonTool::modifiersChanged(Qt::KeyboardModifiers modifiers)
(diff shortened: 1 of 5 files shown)


src/tiled/editpolygontool.cpp
```diff
@@ -925,3 +939,101 @@ void EditPolygonTool::deleteSegment()
...
+    qreal sideA = bc.length();
+    qreal sideB = ca.length();
+
+    qreal cosA = (sideB * sideB + sideC * sideC - sideA * sideA) / (2.0 * sideB * sideC);
+    qreal angleA = qAcos(cosA);
+
+    if (angleA < M_PI / 2.0)
+        return INT_MAX;
+
+    qreal x0 = a.x();
```
Ablu: why not `ax`? (or according to the naming suggestions I made before...)

src/tiled/editpolygontool.cpp
```diff
@@ -925,3 +939,101 @@ void EditPolygonTool::deleteSegment()
...
+            if (temp < minDistance) {
+                minDistance = temp;
+                index1 = i;
+                index2 = i + 1;
+                selectedItem = item;
+            }
+        }
+
+        if (object->shape() == MapObject::Polygon) {
+            qreal temp = helper(point, polygon.first(), polygon.last());
```
Ablu: `temp = helper`. If you see this code, what do you think does it do? `distance = distanceOfPointToLine` allows you to continue reading without checking the `helper` function.

Ablu:
See https://github.com/Ablu/tiled/commit/14fc32605013dcfe153198ea3330f088bdf274f5 for an alternative solution. Granted... it is very verbose, I guess one could meet somewhere in the middle. Also there is no real need for the square root to get the length, the squared length works perfectly fine too.

Otherwise I find it slightly confusing that it does not add the new point at the positionof the double click, but in the center. Also there are still two points selected after the split...

ketanhwr:
Regarding your solution, it works a bit weirdly. Clicking on one side of segment works and on the other side doesn't.

ketanhwr:
Because we won't expect the user to click on a point lying perfectly on the line.

Ablu:
Ah right... There is an issue with the unbounded check. it does not work with the normal like this. I will update it.

ketanhwr:
No I don't think that's the issue. Maybe because we need to check with both the directions of the normal vector. Currently you are checking with only one orientation of the normal

Ablu:
well... yes. Because the normal goes into the other direction it only results in an unbounded intersection.

Ablu:
see https://github.com/Ablu/tiled/commit/76fc754f6efa5fd528ea83463e895fc371b47426 for the fix. I need to check the bounds of the line again...

bjorn:
I tested this a bit and noticed a few problems:

* The split happens at the center, whereas I would expect it to be at the location where I double-clicked.

* Half of the time that I try to double-click, it does not work. I guess that's the problem you're still trying to find a good solution to.

* It seems I can select the two nodes at the ends of a segment by clicking the segment. However, sometimes they seem to get selected on press, but then they deselect again on release. This happens even when I'm not moving the mouse at all between press and release. Also, it should be possible to hold Shift to add to the selection.

* There should probably be a highlight of the segment when you hover it.

ketanhwr:
> The split happens at the center, whereas I would expect it to be at the location where I double-clicked.

Okay, this might be a bit tricky. I'll try it after the next three points are solved first.

> Half of the time that I try to double-click, it does not work. I guess that's the problem you're still trying to find a good solution to.

Highlighting will help here, I think.

> It seems I can select the two nodes at the ends of a segment by clicking the segment. However, sometimes they seem to get selected on press, but then they deselect again on release. This happens even when I'm not moving the mouse at all between press and release. Also, it should be possible to hold Shift to add to the selection.

This is something I'm trying to debug. Will fix soon.

> There should probably be a highlight of the segment when you hover it.

Highlight as in? The edge changes color when highlighted over it?

bjorn:
> Highlight as in? The edge changes color when highlighted over it?

Exactly. I think it's not the full solution to the second issue though, since sometimes I'm pretty sure I should be hitting the segment.

ketanhwr:
> However, sometimes they seem to get selected on press, but then they deselect again on release. 

This is something even I'm experiencing, and I have no idea why. Probably some function which is called on mouse release..I'll dig further.

bjorn:
I've decided to implement an alternative solution in change c7dae1ba21a7644a5f611355adf326f358fda9b7 and 3b99820e422c883edba573a2ecd1a88c55563dee.

The first change only implements a number of expected basic interactions with segments, like clicking them to select the handles on both sides, dragging them and right-clicking them. The second change added the splitting of the segment on double-click.

Issue #1686, opened 2017-08-14
Add an icon in the Tileset Editor for the Animation Editor like the Collision Editor has


anchpop:
It's kind of hard to find as-is I think. Please excuse my shitty ms paint mockup!

![image](https://user-images.githubusercontent.com/3711047/29258461-c9d164b4-8087-11e7-9f12-fa2e61879e5f.png)

ketanhwr:
Hmm, seems like a nice improvement. @bjorn should I add this?

bjorn:
@ketanhwr Alright, but please make sure to finish your existing projects first.

PR #1706, opened 2017-08-26
Adds icon for animation editor in the tileset editor

ketanhwr:
#1686


bjorn:
I don't really like the icon, but it can be improved later. :-)

The patch looks fine, thanks!

PR #1708, opened 2017-08-28
Adds infinite maps docs

ketanhwr:


bjorn:
Please wrap the lines, say to 78 characters.

The documentation for `<chunk>` is too short. It should mention "Can contain: tile", like other elements are documenting their possible child elements. Also, the element should be mentioned at `<data>`, at least as part of the "Can contain" list of elements.

Somewhere, probably as part of the `<chunk>` documentation, it should be mentioned that the contents is the same as for the `<data>` element, except that it only stores the data for the area specified with the attributes and that no `<chunk>` element can be nested in a `<chunk>`.

PR #1712, opened 2017-08-29
Enable resetting template instances

thabetx:
Added a context menu item to reset template instances.

### DIFF
--- a/src/tiled/abstractobjecttool.cpp
+++ b/src/tiled/abstractobjecttool.cpp
@@ -373,7 +373,7 @@ void AbstractObjectTool::showContextMenu(MapObjectItem *clickedObjectItem,
     if (anyIsTemplateInstance) {
         menu.addAction(tr("Detach"), this, SLOT(detachSelectedObjects()));
 
-        auto resetToTemplateAction = menu.addAction(tr("Reset Template instance(s)"), this, SLOT(resetInstances()));
+        auto resetToTemplateAction = menu.addAction(tr("Reset Template Instance(s)"), this, SLOT(resetInstances()));
         resetToTemplateAction->setEnabled(std::any_of(selectedObjects.begin(),
                                                       selectedObjects.end(),
                                                       isChangedTemplateInstance));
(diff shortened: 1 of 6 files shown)


src/tiled/abstractobjecttool.cpp
```diff
@@ -57,6 +57,12 @@ static bool isResizedTileObject(MapObject *mapObject)
     return false;
 }
 
+static bool isChangedTemplateInstance(MapObject *mapObject){
```
bjorn: Coding style: please put `{` on its own line for functions.

src/tiled/changemapobject.cpp
```diff
@@ -203,3 +203,63 @@ void DetachObjects::undo()
...
+                               QUndoCommand *parent)
+    : QUndoCommand(QCoreApplication::translate("Undo Commands",
+                                               "Reset %n Instnaces",
+                                               nullptr, mapObjects.size()), parent)
+    , mMapDocument(mapDocument)
+    , mMapObjects(mapObjects)
+{
+    for (const MapObject *object : mMapObjects) {
+        mProperties.append(object->properties());
+        mChangedProperties.append(object->changedProperties());
```
bjorn: `mChangedProperties` isn't actually used.

src/tiled/changemapobject.cpp
```diff
@@ -203,3 +203,63 @@ void DetachObjects::undo()
...
+
+    emit mMapDocument->objectsChanged(mMapObjects);
+
+    // This signal forces updating custom properties in the properties dock
+    emit mMapDocument->selectedObjectsChanged();
+}
+
+void ResetInstances::undo()
+{
+    // TODO: Is there a better way than saving clones and resetting like this?
```
bjorn: Maybe, but I think this is fine.

thabetx: Alright, I found out that I didn't update name or type, I will add them.

src/tiled/changemapobject.cpp
thabetx: Okay, even `mProperties` isn't used, I didn't think I will clone the objects when I created them.

src/tiled/changemapobject.cpp
thabetx: Also updating the `templateRef` was unnecessary.

thabetx:
I've updated the PR.

PR #1715, opened 2017-08-30
Add retranslateUi() to CommandManager

### LINKED ISSUE #1714: Commands sub-menu is not being immediately updated after language change
After changing language the sub-menu "Commands" is not being updated correctly.

IMG

ketanhwr:
Closes #1714

### DIFF
--- a/src/tiled/commandmanager.cpp
+++ b/src/tiled/commandmanager.cpp
@@ -40,7 +40,6 @@ CommandManager::CommandManager()
     , mLogger(new LoggingInterface())
 {
     updateActions();
-    retranslateUi();
 }
 
 CommandManager *CommandManager::instance()
@@ -119,6 +118,7 @@ void CommandManager::updateActions()
 
     mActions.append(mEditCommands);
 
+    retranslateUi();
     populateMenus();
 }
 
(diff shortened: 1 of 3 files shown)


src/tiled/commandmanager.cpp
```diff
@@ -109,7 +109,7 @@ void CommandManager::updateActions()
 
     mActions.append(mSeparator);
 
-    QAction *mEditCommands = new QAction(this);
+    mEditCommands = new QAction(this);
     mEditCommands->setIcon(
             QIcon(QLatin1String(":/images/24x24/system-run.png")));
     mEditCommands->setText(tr("Edit Commands..."));
```
bjorn: Please add a call to `retranslateUi` to the end of the constructor and remove this line, to avoid having the text twice.

src/tiled/commandmanager.cpp
```diff
@@ -40,6 +40,7 @@ CommandManager::CommandManager()
     , mLogger(new LoggingInterface())
 {
     updateActions();
+    retranslateUi();
```
bjorn: Erm, while I did say this line should be in the constructor, when I said that I didn't realize that the action was actually not created here, but in `CommandManager::updateActions`. As such, the call to `retranslateUi` should of course be in `updateActions`, otherwise the text isn't set when `updateActions` is called in response to the commands changing.

ketanhwr: :man_facepalming:

PR #1735, opened 2017-09-07
Adds a scope switch to the magic wand tool

EJlol:
Adds a scope switch to the magic wand tool. The user may choose between Contiguous and Global. With Contiguous all the tiles have to be
connected (the old magic wand tool), and with Global it will sample the whole layer (the old select same tile tool).

Since this makes the select same tile tool obsolete, I have removed it.

The reason I moved this to the tool options is because I also want to try to add this option to the bucket tool, and to keep things consistent. 

The icons still need to be done.


bjorn:
I'm a bit worried this will make the "select same tile" functionality somewhat hidden. Do you know of any other applications implementing this as a scope switch?

EJlol:
Yes, Paint.net has implemented it this way. I have never felt it as hidden there. Though I do need to say that it never was a seperate tool in the first place, and just got implemented there like this. I can Imagine old users will be confused, just like when the edit tileset collision button moved. Maybe something can be done as a transition kind  of period to stop making old users confused?

EJlol:
Also Photoshop seems to have it implemented it this way, though since I dont have Photoshop I cant be sure about that.

eishiya:
This PR is 5 years old, but this is a feature that's been mentioned a few times on the Discord.

Photoshop and Aseprite both have a "Contiguous" toggle (checkbox) for the bucket fill and magic wand in their toolbars. I think a checkbox is better than two named modes, especially since in Tiled, "Global" isn't *global* - it's limited to the current layer.
image
At least for those of us used to such image editors, this kind of checkbox on the Magic Wand Tool is more intuitive than a separate "Select Same Tile" tool xP

However, if these kinds of modes are added to Tiled, I think they need to be added to both the Magic Wand and the Fill Bucket. IME, the most common use case for Select Same Tile (=non-contiguous Magic Wand) is to then Shift+Bucket Fill that selection with other tiles. Having non-contiguous mode available on the Fill Bucket would save some clicks and key presses, and have a clearer preview.

PR #1739, opened 2017-09-13
Add dtiled to libraries doc section.

rcorre:
Didn't suggest this before as I thought it was just for TMX.


bjorn:
Thanks!

> Didn't suggest this before as I thought it was just for TMX.

Indeed, actually I recently also updated the introduction on that page to mention explicitly the JSON format, in response to our recent messages.

PR #1741, opened 2017-09-14
Update json-map-format.rst

ldurniat:
Update backgroundcolor parameter description


bjorn:
Thanks! There's a lot that needs to be updated about this document, but we need to start somewhere. :-)

PR #1752, opened 2017-09-21
Made files non-executable

Ablu:


bjorn:
Heh, I wonder how that happened. Thanks!

PR #1756, opened 2017-09-22
Update export.rst

TrunXGIT:
Not sure if the "Otherwise..." part is necessary but nice to know as a warning.

### DIFF
--- a/docs/manual/export.rst
+++ b/docs/manual/export.rst
@@ -62,7 +62,7 @@ Tile layers and tile objects (when no type is set) will export as "tile"
 elements. These support horizontal and vertical flipping, but no
 rotation. For tile objects, scaling is also supported.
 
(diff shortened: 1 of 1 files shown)


docs/manual/export.rst
```diff
@@ -62,6 +62,8 @@ Tile layers and tile objects (when no type is set) will export as "tile"
 elements. These support horizontal and vertical flipping, but no
 rotation. For tile objects, scaling is also supported.
 
+Tilesets in Tiled have to be named the same as the corresponding backgrounds in the GameMaker project. Otherwise GameMaker gets cought in an infinite error loop when loading the exported room.gmx file from Tiled.
```
Ablu: `caught` ;)

bjorn:
Actually, I wonder if it would be better to base the exported name on the _image file name_ of the tileset, rather than the user-chosen _name_ of the tileset. That way we should never run into problems, right?

Talking about this particular export code:
```c++
if (tileset->isCollection()) {
    bgName = QFileInfo(tile->imageSource().path()).baseName();
} else {
    bgName = tileset->name();
```
The non-collection tileset could be treated similarly to the tiles in an image collection.

PR #1769, opened 2017-10-05
fix python scripts for updated tiled interfaces

spiiin:
I added SharedTileset data() method, regenerated pythonbind.cpp (I think, change order of lines in python file caused change of many lines), and fixed scripts for this and other interfaces changes (I tested them all, but don't know, what SNES game used by the author or original plugin - they can have different tiles format, as I know, maybe I write my own plugin for NES format in future, I know it much better).


bjorn:
Yes, I think this is a good start. It would be great if we could have `Tiled.Tileset(...)` create a `SharedTileset` and have the members work on it directly, but for now this solution is the next best thing. Thanks!

PR #1774, opened 2017-10-07
Update tiled_ru.ts

pavel-tkachenko:


bjorn:
@oktonion Maybe you could help reviewing these changes?

oktonion:
@bjorn yep

translations/tiled_ru.ts
```diff
@@ -3265,12 +3265,12 @@ Please select specific format.</source>
     <message>
         <location line="+4"/>
         <source>Group Layer</source>
-        <translation>Сгруппировать слой</translation>
+        <translation>Группа слоев</translation>
```
oktonion: Hard to tell what is 'group' there - a noun or a verb, because in russian that will make difference. You could look it in the redactor in menu `Layer->New` and `Layer->Group`.
`Layer->New->Group Layer` should be a noun, but `Layer->Group->Group Layer` a verb.

oktonion: So based on functional of these menu options I assume that for Russian translation should be two different strings for each menu. It will be an equivalent for English 'Layer->New->a Layer of Group' and 'Layer->Group->perform Group to Layer'.
Maybe @bjorn could clarify what exactly the context is.

bjorn: Unfortunately both "AddGroupLayer" and "GroupLayers" actions have the same context and the same text, so it is currently not possible to have different translations for them. I'll fix that for Tiled 1.1 by using a disambiguation string. For this PR, just pick one.

oktonion: @bjorn So it doesn't matter actually, let it be the updated string. One menu option will be ok and other...confusing =)

bjorn:
> 'е- ё' is a holywar topic, try to avoid it.

Does this represent a change still to be made, or can I merge this?

oktonion:
@bjorn no, just for the commiter to keep in mind. You can merge it.

bjorn:
Cool, thanks for the review @oktonion! I hope you and @pavel-tkachenko will be ready once I do the string freeze for Tiled 1.1. :-)

oktonion:
@bjorn You're welcome. Looking forward for the release.

translations/tiled_ru.ts
```diff
@@ -370,7 +370,7 @@
     <message>
         <location filename="../src/plugins/gmx/gmxplugin.cpp" line="+82"/>
         <source>Could not open file for writing.</source>
-        <translation>Не удалось открыть файл для записи.</translation>
+        <translation>Не удается открыть файл для записи.</translation>
```
HellGhost: I think this is post action, and first version is correct

translations/tiled_ru.ts
```diff
@@ -1983,12 +1983,12 @@ Line %1, column %2</source>
...
         <location line="+11"/>
         <source>Can't parse file header!</source>
-        <translation>Не удается распарсить заголовок файла!</translation>
+        <translation>Не удается разобрать заголовок файла!</translation>
     </message>
     <message>
         <location line="+22"/>
         <source>Can't parse layer header!</source>
-        <translation>Не удается распарсить заголовок слоя!</translation>
+        <translation>Не удается разобрать заголовок слоя!</translation>
```
HellGhost: It would be greet "определить" instead of "разобрать"

translations/tiled_ru.ts
```diff
@@ -5080,7 +5080,7 @@ Please select specific format.</source>
     <message>
         <location line="+20"/>
         <source>Change Layer Opacity</source>
-        <translation>Изменить прозрачность слоя</translation>
+        <translation>Изменить непрозрачность слоя</translation>
```
HellGhost: Изменить прозрачность слоя better in this context

bjorn:
@HellGhost Since this pull request has already been merged, would you care to open a new PR with your suggested changes?

translations/tiled_ru.ts
pavel-tkachenko: In my opinion that's a moot point as well as "е-ё" topic.

translations/tiled_ru.ts
pavel-tkachenko: Now I think so too. My bad.

Issue #1679, opened 2017-08-07
Rearrange order of object types


Arcanorum:
It would be cool if we could rearrange the order of the list of object types.

![rearrange-object-types](https://user-images.githubusercontent.com/9785268/29008256-ccb8845e-7b0b-11e7-9630-61b4e6e0664b.PNG)

bjorn:
Maybe it makes more sense to show them in alphabetical order?

Arcanorum:
Situational/subjective. Perhaps being able to press on the "Type" column heading to arrange things alphabetically, or press again to switch between reverse-alphabetically, but still able to manually rearrange them if needed, otherwise someone might want to have types "grouped" next to each other in the list in a certain way that wouldn't follow alphabetical ordering.

I suppose if you did go with alphabetical ordering, we could just prefix each type with a character to adjust the order, if a bit cumbersome.

bjorn:
Usually it's a little strange to combine manual sorting with automated sorting. Since the order of the object types is not actually relevant and this dialog is generally not used very often, I'd say we just sort the list by name.

PR #1775, opened 2017-10-08
Sort object types alphabetically

agersant:
This is my attempt at fixing https://github.com/bjorn/tiled/issues/1679, please let me know if it can be done in a better way!

### DIFF
--- a/src/tiled/objecttypesmodel.cpp
+++ b/src/tiled/objecttypesmodel.cpp
@@ -97,15 +97,23 @@ bool ObjectTypesModel::setData(const QModelIndex &index,
                                int role)
 {
     if (role == Qt::EditRole && index.column() == 0) {
-        const ObjectTypes oldObjectTypes(mObjectTypes);
         const int oldRow = index.row();
 
-        mObjectTypes[oldRow].name = value.toString().trimmed();
+        ObjectType objectType = mObjectTypes.at(oldRow);
+        objectType.name = value.toString().trimmed();
+
+        auto nextObjectType = std::lower_bound(mObjectTypes.constBegin(),
+                                               mObjectTypes.constEnd(),
+                                               objectType,
+                                               objectTypeLessThan);
+
+        const int newRow = nextObjectType - mObjectTypes.constBegin();
+        // QVector::move works differently from beginMoveRows
+        const int moveToRow = newRow > oldRow ? newRow - 1 : newRow;
+
+        mObjectTypes[oldRow].name = objectType.name;
         emit dataChanged(index, index);
 
-        auto nextObjectType = std::lower_bound(oldObjectTypes.begin(), oldObjectTypes.end(), mObjectTypes[oldRow], objectTypeLessThan);
-        const int newRow = nextObjectType - oldObjectTypes.begin();
-        const int moveToRow = newRow > oldRow ? newRow - 1 : newRow;
         if (moveToRow != oldRow) {
             Q_ASSERT(newRow != oldRow);
             Q_ASSERT(newRow != oldRow + 1);
--- a/src/tiled/objecttypesmodel.cpp
+++ b/src/tiled/objecttypesmodel.cpp
@@ -101,13 +101,13 @@ bool ObjectTypesModel::setData(const QModelIndex &index,
         mObjectTypes[oldRow].name = value.toString().trimmed();
         emit dataChanged(index, index);
 
-		auto nextObjectType = std::lower_bound(mObjectTypes.begin(), mObjectTypes.end(), mObjectTypes[oldRow], objectTypeLessThan);
+        auto nextObjectType = std::lower_bound(mObjectTypes.begin(), mObjectTypes.end(), mObjectTypes[oldRow], objectTypeLessThan);
         const int newRow = nextObjectType - mObjectTypes.begin();
         if (newRow != oldRow) {
             beginMoveRows(QModelIndex(), oldRow, oldRow, QModelIndex(), newRow);
             mObjectTypes.move(oldRow, newRow > oldRow ? newRow - 1 : newRow);
             endMoveRows();
-		}
+        }
         return true;
     }
     return false;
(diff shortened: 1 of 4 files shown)


src/tiled/objecttypesmodel.cpp
```diff
@@ -106,6 +108,21 @@ Qt::ItemFlags ObjectTypesModel::flags(const QModelIndex &index) const
...
+static bool objectTypeLessThan(ObjectType const &a, ObjectType const &b)
+{
+    return a.name.toLower() < b.name.toLower();
+}
+
+void ObjectTypesModel::sort(int column, Qt::SortOrder order)
+{
+    if (column == 0 && order == Qt::AscendingOrder)
+    {
+       	emit layoutAboutToBeChanged();
```
bjorn: A tab sneaked in here.

src/tiled/objecttypesmodel.cpp
```diff
@@ -106,6 +108,21 @@ Qt::ItemFlags ObjectTypesModel::flags(const QModelIndex &index) const
     return f;
 }
 
+static bool objectTypeLessThan(ObjectType const &a, ObjectType const &b)
```
bjorn: Please change this to `const ObjectType &` to keep with the same style used in the rest of the code.

src/tiled/objecttypesmodel.cpp
```diff
@@ -106,6 +108,21 @@ Qt::ItemFlags ObjectTypesModel::flags(const QModelIndex &index) const
...
+{
+    return a.name.toLower() < b.name.toLower();
+}
+
+void ObjectTypesModel::sort(int column, Qt::SortOrder order)
+{
+    if (column == 0 && order == Qt::AscendingOrder)
+    {
```
bjorn: Coding style: please put the `{` on the same line as the `if`.

src/tiled/objecttypesmodel.cpp
```diff
@@ -141,4 +158,5 @@ void ObjectTypesModel::appendNewObjectType()
     beginInsertRows(QModelIndex(), mObjectTypes.size(), mObjectTypes.size());
     mObjectTypes.append(ObjectType());
     endInsertRows();
+    sort(0, Qt::AscendingOrder);
```
bjorn: The new object type will have an empty name. Rather than sorting afterwards, we could probably just prepend instead of append?

`ObjectTypesModel::appendNewObjectType` should be renamed to `ObjectTypesModel::addNewObjectType`, and I think it should return a `QModelIndex` pointing to the added object type (which can be assigned to `newIndex` in `ObjectTypesEditor::addObjectType`, to avoid making assumptions about where the new object is added there).

src/tiled/objecttypesmodel.cpp
```diff
@@ -27,6 +27,7 @@ void ObjectTypesModel::setObjectTypes(const ObjectTypes &objectTypes)
 {
     beginResetModel();
     mObjectTypes = objectTypes;
+    sort(0, Qt::AscendingOrder);
```
bjorn: I would probably just do this here instead:
```c++
qSort(mObjectTypes.begin(), mObjectTypes.end(), objectTypeLessThan);
```
Since there is no need for the additional signals about the layout.

src/tiled/objecttypesmodel.cpp
agersant: Great suggestion. I wasn't happy with this setup either. That's a good way to clean it up!

agersant:
Updated! Thanks for the rapid response and detailed review.

I had originally started with QSortFilterProxyModel too, but it ended up a bit wacky. Either the ObjectTypesEditor window needed to interact with two distinct models, or I would have had to make a QSortFilterProxyModel subclass with all the operations that the UI expects from ObjectTypesModel.

bjorn:
Hmm, in testing this I noticed a problem when you change the name of a type, in that when it causes the type to be moved, the selected row stays the same (so now a different type is selected) but the properties of the previously selected type are displayed.

Instead of doing a sort after a rename, we could find out if the index of the type needs to change and use `beginMoveRows` / `endMoveRows`. I think this would be ideal since it will likely move the selection along with it. Since the list can be assumed to be already sorted, you can use `std::lower_bound` to find the insertion position.

agersant:
Sorry for the tabs, I'll shift part of the blame on the text editor.

agersant:
Good catch! Took me longer to fix than I thought, QVector::move has a surprising behavior when moving an item to a higher index in the vector.

src/tiled/objecttypesmodel.cpp
```diff
@@ -91,8 +97,17 @@ bool ObjectTypesModel::setData(const QModelIndex &index,
                                int role)
 {
     if (role == Qt::EditRole && index.column() == 0) {
-        mObjectTypes[index.row()].name = value.toString().trimmed();
+        const int oldRow = index.row();
+        mObjectTypes[oldRow].name = value.toString().trimmed();
         emit dataChanged(index, index);
+
+        auto nextObjectType = std::lower_bound(mObjectTypes.begin(), mObjectTypes.end(), mObjectTypes[oldRow], objectTypeLessThan);
```
bjorn: Hmm, I think there may be problems here because due to the above modification, `mObjectTypes` may no longer be sorted, which could throw off`std::lower_bound`. Maybe just make a local copy first.

src/tiled/objecttypesmodel.cpp
```diff
@@ -91,8 +97,17 @@ bool ObjectTypesModel::setData(const QModelIndex &index,
...
+        const int newRow = nextObjectType - mObjectTypes.begin();
+        if (newRow != oldRow) {
+            beginMoveRows(QModelIndex(), oldRow, oldRow, QModelIndex(), newRow);
+            mObjectTypes.move(oldRow, newRow > oldRow ? newRow - 1 : newRow);
```
bjorn: It makes sense if you consider that `move(from,to)` is just a more efficient way of doing `insert(to,takeAt(from))`. After the operation, you expect the element's index to be `to`.

agersant:
Updated! I also got my development environment for Tiled in a much better state so I was able to test this more comprehensively (renames that cause no movement, move up, move down, etc.).

bjorn:
I was bothered a bit by the duplication of the vector, but since it will generally be small and never noticeable to the user I decided to just change that myself.

Thanks for another nice change!

PR #1417, opened 2016-12-14
Flare Code Changed

PeterJFerrarotto:
Changed code of flare plugin to allow for exportation of custom layer properties and the previously excluded map property of "Background Color".

I placed each section of the color - red, blue, green, and alpha - into its own respective key-value pair, as some computers utilize RGBA encoding, and others utilize RBGA encoding, so anybody trying to use these values shouldn't have a problem worrying about the order of the color values.

PR #1781, opened 2017-10-11
Fixed proposed changes to Flare plugin

dorkster:
#1417 appears to have been untouched for the passed 9 months or so. I've fixed the merge conflicts, as well as making a few style changes based on how we format data for Flare. The output we want is `background_color=R,G,B,A`, where RGBA are 8-bit integers.

### DIFF
--- a/src/plugins/flare/flareplugin.cpp
+++ b/src/plugins/flare/flareplugin.cpp
@@ -154,7 +154,7 @@ Tiled::Map *FlarePlugin::read(const QString &fileName)
                     map->setTileHeight(value.toInt());
                 else if (key == QLatin1String("orientation"))
                     map->setOrientation(orientationFromString(value));
-                else if (key == QLatin1String("backgroundColor")){
+                else if (key == QLatin1String("background_color")){
                     backgroundColor = enrichColorInformation(value);
                     map->setBackgroundColor(backgroundColor);
                 }
@@ -354,7 +354,7 @@ bool FlarePlugin::write(const Tiled::Map *map, const QString &fileName)
     out << "tilewidth=" << map->tileWidth() << "\n";
     out << "tileheight=" << map->tileHeight() << "\n";
     out << "orientation=" << orientationToString(map->orientation()) << "\n";
-    out << "backgroundColor=" << backgroundColor.name(QColor::HexArgb) << "\n";
+    out << "background_color=" << backgroundColor.name(QColor::HexArgb) << "\n";
 
     const QDir mapDir = QFileInfo(fileName).absoluteDir();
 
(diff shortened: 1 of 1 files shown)


dorkster:
I've made the requested fixes. And feel free to squash the commits.

src/plugins/flare/flareplugin.cpp
```diff
@@ -367,10 +394,14 @@ bool FlarePlugin::write(const Tiled::Map *map, const QString &fileName)
                     out << "," << w << "," << h << "\n";
 
                     // write all properties for this object
-                    Properties::const_iterator it = o->properties().constBegin();
-                    Properties::const_iterator it_end = o->properties().constEnd();
-                    for (; it != it_end; ++it) {
-                        out << it.key() << "=" << it.value().toString() << "\n";
+                    QVariantMap propsMap = o->properties();
+                    for (QVariantMap::const_iterator it = propsMap.constBegin(); it != propsMap.constEnd(); ++it) {
+                        if(it->userType() == filePathTypeId()){
```
bjorn: Sorry to be such a nitpick, but please also add the spaces here and adjust the `else` below as well.

dorkster: No worries. Just pushed a fix.

src/plugins/flare/flareplugin.cpp
```diff
@@ -367,10 +394,13 @@ bool FlarePlugin::write(const Tiled::Map *map, const QString &fileName)
...
-                    Properties::const_iterator it_end = o->properties().constEnd();
-                    for (; it != it_end; ++it) {
-                        out << it.key() << "=" << it.value().toString() << "\n";
+                    QVariantMap propsMap = o->properties();
+                    for (QVariantMap::const_iterator it = propsMap.constBegin(); it != propsMap.constEnd(); ++it) {
+                        if (it->userType() == filePathTypeId()) {
+                            out << it.key() << "=" << mapDir.relativeFilePath(toExportValue(it.value()).toString()) << "\n";
+                        } else {
+                            out << it.key() << "=" << toExportValue(it.value()).toString() << "\n";
+                        }
```
bjorn: On closer inspection, I believe this could be simplified to just:
```c++
out << it.key() << "=" << toExportValue(it.value(), mapDir).toString() << "\n";
```
The same goes for the writing of layer properties.

The version of `toExportValue` taking an additional map directory is new and will handle the file path case.

dorkster: You're right. That syntax is even used on line 324. I fixed the two spots that needed it.

bjorn:
Thanks!

Issue #1310, opened 2016-06-18
Change duration of multiple frames in animation editor


IMMZ:
Sometimes in animation editor you want to change all frames' duration to some other value than default one. Doing it every frame when add them (new frames) or after adding could be annoying enough. It would be good to select some of or all the frames and change this value to what you need one. Also it could be useful to consider the next features:
- Add/Subtract[Multiply/Divide] selected durations by some number X.
- Set random duration values in range [X-Y] for selected frames.

Used ver: Version 2016.05.17

Alturos:
Gonna have a crack at this. @bjorn: Thoughts on the bullet point items?

bjorn:
@Alturos Before starting to implement it, please consider how the UI should work for this and make a proposal. As for the bullet points, I don't personally see these actions as particular common and hence not very useful. However, what I do think is frequently needed is to change the speed / duration of the entire animation, adjusting each frame's duration accordingly. Another thing to improve here is that Tiled currently always uses 100ms as the default frame duration, but I think this should be configurable.

Also note that on the `wip/tilesetdocument` branch, tileset editing has been moved away from map editing. This affects also the Tile Animation Editor. The code has already changed there, so modifying it on master will create conflicts, and actually I'd like to turn this animation editor dialog into a view of the new Tileset editing mode.

So hmm, after all this I'm not sure if this is the best moment to tackle this particular issue. :/

Alturos:
Funny how that timing works. The proposal I'd come to (by getting it most of the way to done), was the simplest approach:

- Add a toolbar button to the row above the listview
- Enable/disable based on having a selection IN the listview
- Hitting the button pops up a small dialog with a spinnerbox
- - Entering a value and hitting OK updates all the selected entries.

More or less end of list. At this rate, it looks like I could have that done by the end of the weekend. It's pretty light as features go, if you want to slip it in ahead of the tileset refactor. If not, also cool, I'll table it and look for something else to poke at.

bjorn:
Sounds fair. The other suggestions would of course be easy to add by means of a combo box in the dialog.

Actually it would be good to look at some other animation editors, to see how they work. I have personally never used an animation editor.

Alturos:
I may have jumped the gun on this one, but i was already 3/4's done, so figured I should just finish it. I think it's in line with the general UX theme of Tiled in general (or so my weary 4AM mind tells me). But it was light enough work that I didn't mind doing it, even if it's just throw away. :smile: :sleepy:

[...]

PR #1784, opened 2017-10-15
Animation Editor- Multiple Frame duration edit feature

satu0king:
Added a spinBox where the user can select a frame duration. 
The Set Frame Time button has two purposes 

1. If no Frame(s) are selected, then it sets the duration of all frames to the duration of the spinBox. It also resets the default Frame duration to that duration.
2. If frame(s) are selected, then it sets the duration of all those frames to the duration of the spinBox. Default Frame duration is not changed. 

Issues: The undo in situation 2 works but not all at a time. Each frame update is registered separately instead of all at a once. I am not sure how to fix this. Apart from this minor bug, everything seems to be all right. 

This is the first time I am contributing to open source. I am new to Tiled and QT,However I am not new to programming. Please guide me.


src/tiled/tileanimationeditor.cpp
```diff
@@ -67,17 +67,19 @@ class FrameListModel : public QAbstractListModel
...
-    static const int DEFAULT_DURATION = 100;
+    static int DEFAULT_DURATION;
 
     void addFrame(const Frame &frame);
 
     const Tileset *mTileset;
     QVector<Frame> mFrames;
 };
 
+int FrameListModel::DEFAULT_DURATION=100;
```
ketanhwr: I guess you can initialise this in the constructor.

src/tiled/tileanimationeditor.cpp
```diff
@@ -257,6 +259,13 @@ const QVector<Frame> &FrameListModel::frames() const
 {
     return mFrames;
 }
+void FrameListModel::setDefaultFrameTime(int duration)
```
ketanhwr: There should be a gap between the above two lines.

src/tiled/tileanimationeditor.cpp
```diff
@@ -257,6 +259,13 @@ const QVector<Frame> &FrameListModel::frames() const
...
+{
+    DEFAULT_DURATION=duration;
```
ketanhwr: Coding style: `DEFAULT_DURATION = duration;`

src/tiled/tileanimationeditor.cpp
```diff
@@ -257,6 +259,13 @@ const QVector<Frame> &FrameListModel::frames() const
...
+    for(Frame &F:mFrames){
```
ketanhwr: Coding style: `for (Frame &F:mFrames) {`

And since this is a single line statement, you can get rid of the braces.

src/tiled/tileanimationeditor.cpp
```diff
@@ -257,6 +259,13 @@ const QVector<Frame> &FrameListModel::frames() const
...
+        F.duration=duration;
```
ketanhwr: Spaces around the `=`

src/tiled/tileanimationeditor.cpp
```diff
@@ -391,6 +405,21 @@ void TileAnimationEditor::framesEdited()
                                             mFrameListModel->frames()));
     mApplyingChanges = false;
 }
+void TileAnimationEditor::setFrameTime()
```
ketanhwr: A gap between the two lines.

src/tiled/tileanimationeditor.cpp
```diff
@@ -391,6 +405,21 @@ void TileAnimationEditor::framesEdited()
...
+{
+    QItemSelectionModel *selectionModel = mUi->frameList->selectionModel();
+    QModelIndexList indexes = selectionModel->selectedIndexes();
+    if (indexes.isEmpty()){
```
ketanhwr: Coding stye: `if (indexes.isEmpty()) {`

src/tiled/tileanimationeditor.cpp
satu0king: I dont think this can be done in a constructor as its a static variable ..

src/tiled/tileanimationeditor.cpp
satu0king: Thanks, Change has been made.

src/tiled/tileanimationeditor.cpp
satu0king: Thanks, Change has been made.

src/tiled/tileanimationeditor.cpp
satu0king: Thanks, Change has been made.

src/tiled/tileanimationeditor.cpp
satu0king: Thanks, Change has been made.

src/tiled/tileanimationeditor.cpp
satu0king: Thanks, Change has been made.

satu0king:
#1784

src/tiled/tileanimationeditor.ui
```diff
@@ -20,6 +20,35 @@
...
+             <property name="maximum">
+              <number>10000</number>
+             </property>
+             <property name="singleStep">
+              <number>20</number>
+             </property>
+             <property name="value">
+              <number>100</number>
+             </property>
+         </widget>
```
bjorn: You would generally use Qt Designer or the Design mode in Qt Creator for editing `.ui` files, but if you do it by hand, please do take care that they are formatted consistently (easiest should be to open the file in Qt Designer and save it again).

src/tiled/tileanimationeditor.ui
```diff
@@ -20,6 +20,35 @@
    <item>
     <layout class="QHBoxLayout" name="horizontalLayout">
      <item>
+         <widget class="QSpinBox" name="frameTime">
+             <property name="suffix">
+              <string>ms</string>
+             </property>
+             <property name="minimum">
+              <number>20</number>
```
bjorn: Why is 20 ms the minimum frame time? I would just put it on 1 ms.

src/tiled/tileanimationeditor.ui
```diff
@@ -20,6 +20,35 @@
...
+             </property>
+             <property name="maximum">
+              <number>10000</number>
+             </property>
+             <property name="singleStep">
+              <number>20</number>
```
bjorn: Let's use 10 ms here.

src/tiled/tileanimationeditor.ui
```diff
@@ -20,6 +20,35 @@
    <item>
     <layout class="QHBoxLayout" name="horizontalLayout">
      <item>
+         <widget class="QSpinBox" name="frameTime">
+             <property name="suffix">
+              <string>ms</string>
```
bjorn: This should be " ms" (starting with a space).

src/tiled/tileanimationeditor.ui
```diff
@@ -20,6 +20,35 @@
...
+              <number>100</number>
+             </property>
+         </widget>
+
+
+     </item>
+     <item>
+         <widget class="QPushButton" name="setFrameTimeButton">
+             <property name="text">
+              <string>Set Frame Time</string>
```
bjorn: I think it would be nicer to use a label "Frame duration:", then the input box, and then the button with just "Apply" on it.

satu0king:
Really sorry that I didn't make the changes earlier. 
I fixed the undo issue the way @bjorn told to do and did all the changes requested by @bjorn and @ketanhwr.

satu0king:
Can you please check if my PR is alright, if not let me know the mistakes

src/tiled/tileanimationeditor.ui
```diff
@@ -19,6 +19,39 @@
   <layout class="QVBoxLayout" name="verticalLayout">
    <item>
     <layout class="QHBoxLayout" name="horizontalLayout">
+     <item>
+      <widget class="QLabel" name="label">
+       <property name="text">
+        <string>Frame-Duration: </string>
```
bjorn: Please change to "Frame duration:"

src/tiled/tileanimationeditor.cpp
```diff
@@ -392,6 +410,26 @@ void TileAnimationEditor::framesEdited()
     mApplyingChanges = false;
 }
 
+void TileAnimationEditor::setFrameTime()
+{
+    QItemSelectionModel *selectionModel = mUi->frameList->selectionModel();
+    const QModelIndexList indexes = selectionModel->selectedIndexes();
+
+    if (indexes.isEmpty()) {
+        mFrameListModel->setDefaultFrameTime(mUi->frameTime->value());
```
bjorn: I think the default animation time should change immediately when you change the value of the `frameTime` widget. And of course, without affecting the existing frames.

src/tiled/tileanimationeditor.cpp
```diff
@@ -258,6 +261,13 @@ const QVector<Frame> &FrameListModel::frames() const
     return mFrames;
 }
 
+void FrameListModel::setDefaultFrameTime(int duration)
+{
+    DEFAULT_DURATION = duration;
+    for(Frame &frame : mFrames)
+        frame.duration = duration;
```
bjorn: In my previous review I commented that I don't like this side-effect of `setDefaultFrameTime`. Please don't change the animation and just set the new default frame time (also this setter can be static then).

src/tiled/tileanimationeditor.cpp
```diff
@@ -67,17 +67,20 @@ class FrameListModel : public QAbstractListModel
 
     void setFrames(const Tileset *tileset, const QVector<Frame> &frames);
     void addTileIdAsFrame(int id);
+    void setDefaultFrameTime(int duration);
     const QVector<Frame> &frames() const;
 
 private:
-    static const int DEFAULT_DURATION = 100;
+    static int DEFAULT_DURATION;
```
bjorn: This variable used by the in `ALL_CAPS` because it was a constant. Now that it can be changed, please rename to `mDefaultDuration`.

src/tiled/tileanimationeditor.cpp
```diff
@@ -392,6 +410,26 @@ void TileAnimationEditor::framesEdited()
...
 }
 
+void TileAnimationEditor::setFrameTime()
+{
+    QItemSelectionModel *selectionModel = mUi->frameList->selectionModel();
+    const QModelIndexList indexes = selectionModel->selectedIndexes();
+
+    if (indexes.isEmpty()) {
+        mFrameListModel->setDefaultFrameTime(mUi->frameTime->value());
+        mUi->frameList->setFocus(); // This is necessary for the UI, It does not get updates without this
```
bjorn: This is a hack, which works because it triggers a repaint. The reason the UI isn't updating is that `setDefaultFrameTime` is modifying the frame durations without emitting the `dataChanged` signal.

Please write a separate function like `FrameListModel::applyFrameTime` that sets the duration like currently `setDefaultFrameTime` is doing, but in addition it should emit `dataChanged` with appropriate indexes. Something like:
```c++
emit dataChanged(index(0), index(mFrames.length() - 1), { Qt::EditRole, Qt::DisplayRole });
```
(since this will only be valid when there is at least one frame, you need to early-out if `mFrames.isEmpty()`)

satu0king: I now understood why it wasn't working and why the hack worked. I am going to use the function that was written previously. It internally emits a data change signal mFrameListModel->setData()

satu0king:
I think I got it right this time. I did the following things 

- Persistence of Frame Duration across Tiled sessions. (As implemented in NewMapDialog) 
- Removed side effect of `setDefaultFrameTime()` .
- Change default time immediately on a change of frame-time widget.
- Removed the hack and used data emit on changing frame time. 
- Name errors

bjorn:
All looking good, thanks for your contribution!

I didn't close #1310 since it had an idea which I think could still be interesting to add, which would be to add the ability to add/subtract/multiply with the duration of the selected animation frames. I could see this being added by way of adding a menu to the "Apply" button with the additional actions, but in that case I think it's a little strange to combine it with the automatic change of default frame duration (maybe that should then become another item in this menu). In any case I'm a little undecided about this, but if you like you could try to fit in that functionality as well.

Issue #1766, opened 2017-09-29
[USABILITY] Add an "inside tutorial" to Tiled


RhenaudTheLukark:
Hello there!

I am using Tiled for now some months, and I never really thought about participating in it.
However, I came back to my decision and I will expose what may not be very easy for new users to quickly know how to use Tiled.

Back in time, when I saw the main interface, I had some trouble handling it because there aren't a lot of explanation on where to start with. 
There aren't a lot of indications of what to do when you delve in the application itself.

Plus, I recommended the community I'm managing to use Tiled if they want to make maps they will use on the project I'm managing, but some people told me it's not very intuitive.

I thought it would be a good addition to add an in-project little tutorial.
Nothing too serious, I think an Audacity-level tutorial would be enough, to show what people can do with Tiled. 

Just the basics, that is to say some slides with screenshots and little texts to describe how to use the basic features of Tiled, and all of that on a pop-up which could be disabled on demand, maybe using a "Don't show this again" toggle option on the tutorial itself, which could be changed in the "Help" option.
Also, this tutorial could be closed easily using a cross, like in normal windows.

I would be glad to try myself on such thing, but I think it would be a bit big for my first contribution on Tiled, so I decided to expose my idea here.

I hope an answer will come as soon as possible, and best of luck for the future of Tiled.

Best Regards, RTL.

keshav2010:
Yup, being new to tiled myself, I think it would be good to include a short in-product tutorial, something like what earlier version of game maker engine adopted ( version 7 and 8), a side panel that contains instruction that user can follow to create his first tileset/map, this will make sure we need not change much of the existing code and only focus on the new panel/frame we are going to add and all the content that goes inside that panel/frame.

RhenaudTheLukark:
I never really tried Game Maker, but I will get more info about that.

And yes, it's just an additional panel and help for beginners. It doen't have to change absolutely everything, but it helps for usability.

I have to say I'm quite interested in that idea. I'll try to take a look at it soon.

keshav2010:
Yup same here, currently trying to understand codebase as well as learning Qt framework, once im little bit familiar with it, i will try to work on it :D , in case i make any further progress, i will keep you updated over here and in case you make any further progress, u can also update status here :) 
(btw here's facebook email id : [Click Here ](https://www.facebook.com/keshav15157 ))

bjorn:
[...]

PR #1789, opened 2017-10-18
Add 'New Tileset...' button when no tileset is opened

RhenaudTheLukark:
Feature announced in the issue #1766.

### DIFF
--- a/src/tiled/tilesetdock.cpp
+++ b/src/tiled/tilesetdock.cpp
@@ -53,6 +53,7 @@
 #include <QComboBox>
 #include <QDropEvent>
 #include <QFileDialog>
+#include <QPushButton>
 #include <QHBoxLayout>
 #include <QMenu>
 #include <QMessageBox>
@@ -575,9 +576,10 @@ void TilesetDock::deleteTilesetView(int index)
     delete view;                    // view needs to go before the tab
     mTabBar->removeTab(index);
 
-    // Make the "New Tileset..." special tab reappear if there is no opened tileset
-    if (mTilesets.count() == 0)
+    // Make the "New Tileset..." special tab reappear if there is no tileset open
+    if (mTilesets.count() == 0) {
         mSuperViewStack->setCurrentIndex(0);
+    }
 
     // Make sure we don't reference this tileset anymore
     if (mCurrentTiles && mCurrentTiles->referencesTileset(tileset)) {
--- a/src/tiled/tilesetdock.cpp
+++ b/src/tiled/tilesetdock.cpp
@@ -73,6 +73,21 @@ using namespace Tiled::Internal;
 
 namespace {
 
+class NewTilesetView : public QWidget
+{
+public:
+    explicit NewTilesetView(QWidget *tmb = nullptr)
+        : QWidget(tmb)
+    {
+        QWidget *w = new QWidget(this);
+
+        QPushButton *pbNewTileset = new QPushButton(w);
+        pbNewTileset->setText(QStringLiteral("New Tileset..."));
+
+        connect(pbNewTileset, SIGNAL(released()), tmb, SLOT(newTileset()));
+    }
+};
+
 class TilesetMenuButton : public QToolButton
 {
 public:
@@ -154,6 +169,7 @@ TilesetDock::TilesetDock(QWidget *parent)
     , mMapDocument(nullptr)
     , mTilesetDocumentsFilterModel(new TilesetDocumentsFilterModel(this))
     , mTabBar(new WheelEnabledTabBar)
+    , mSuperViewStack(new QStackedWidget)
     , mViewStack(new QStackedWidget)
     , mToolBar(new QToolBar)
     , mCurrentTile(nullptr)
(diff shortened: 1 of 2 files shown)


src/tiled/tilesetdock.cpp
```diff
@@ -553,6 +576,11 @@ void TilesetDock::deleteTilesetView(int index)
     delete view;                    // view needs to go before the tab
     mTabBar->removeTab(index);
 
+    // Make the "New Tileset..." special tab reappear if there is no tileset open
+    if (mTilesets.count() == 0) {
```
ketanhwr: Can remove the braces since it's only 1 line block.

src/tiled/tilesetdock.h
```diff
@@ -165,7 +165,7 @@ private slots:
     TilesetDocumentsFilterModel *mTilesetDocumentsFilterModel;
 
     QTabBar *mTabBar;
-    QStackedWidget *mViewStack;
+    QStackedWidget *mSuperViewStack, *mViewStack;
```
ketanhwr: Please define each variable in a new line.

src/tiled/tilesetdock.cpp
RhenaudTheLukark: I've already changed that in the second commit.

src/tiled/tilesetdock.h
RhenaudTheLukark: Will do on the third commit.

src/tiled/tilesetdock.cpp
```diff
@@ -553,6 +576,10 @@ void TilesetDock::deleteTilesetView(int index)
     delete view;                    // view needs to go before the tab
     mTabBar->removeTab(index);
 
+    // Make the "New Tileset..." special tab reappear if there is no tileset open
+    if (mTilesets.count() == 0)
```
bjorn: Please use `mTilesets.isEmpty()`

src/tiled/tilesetdock.cpp
```diff
@@ -73,6 +74,21 @@ using namespace Tiled::Internal;
...
+public:
+    explicit NewTilesetView(QWidget *tmb = nullptr)
+        : QWidget(tmb)
+    {
+        QWidget *w = new QWidget(this);
+
+        QPushButton *pbNewTileset = new QPushButton(w);
+        pbNewTileset->setText(QStringLiteral("New Tileset..."));
+
+        connect(pbNewTileset, SIGNAL(released()), tmb, SLOT(newTileset()));
```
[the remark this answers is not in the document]
RhenaudTheLukark: I changed it, instead of giving a QWidget, I'll specify you need to give the TilesetDock itself, which is what this QWidget is.

src/tiled/tilesetdock.cpp
```diff
@@ -73,6 +74,21 @@ using namespace Tiled::Internal;
 
 namespace {
 
+class NewTilesetView : public QWidget
+{
+public:
+    explicit NewTilesetView(TilesetDock *parent = nullptr)
+        : QWidget(parent)
+    {
+        QWidget *w = new QWidget(this);
```
bjorn: This widget serves no purpose apart from adding just another level to the widget hierarchy.

RhenaudTheLukark: Alright, I'll take a look at all of that.

I tried to center the button, but everything I tried didn't work.

bjorn: You basically just need spacer items on all 4 sides for centering the button, and for that it needs to be in a grid layout.

RhenaudTheLukark:
I tried to center the button, but no matter which way I took the problem from, I couldn't do better than that.

This is a simple task, yet somehow the QGridLayout widget is not taking the whole space the QDockWidget provides, so the button is centered horizontally but not vertically.

I surrounded it with QSpacerItem objects, but I couldn't center it vertically anyway.

Tell me if this is good enough (I really hope it is)

src/tiled/tilesetdock.cpp
```diff
@@ -73,6 +76,32 @@ using namespace Tiled::Internal;
...
+class NoTilesetWidget : public QWidget
+{
+public:
+    explicit NoTilesetWidget(TilesetDock *parent = nullptr)
+        : QWidget(parent)
+    {
+        QGridLayout *gridLayout = new QGridLayout(this);
+
+        QSpacerItem *spacerVertTop = new QSpacerItem(1, 1, QSizePolicy::Ignored, QSizePolicy::Expanding);
+        gridLayout->addItem(spacerVertTop, 1, 0, Qt::AlignCenter);
```
bjorn: This isn't doing what you think it does. You're actually passing `Qt::AlignCenter` as the row-span argument to `addItem`, which explains why the centering isn't working correctly.

I found out that the use of spacers isn't needed at all though, since we can just pass `Qt::AlignCenter` when adding the New Tileset button. I'll push this change in a moment.

bjorn:
> Tell me if this is good enough (I really hope it is)

Of course it isn't good enough to have a bunch of code sitting there not achieving what it is meant to do. :-)

It wasn't easy for me either to find out what was going on though. A compiler warning would have certainly helped, but in C++ passing an enum value to an `int` parameter is perfectly valid.

bjorn:
Thanks! :-)

Issue #1783, opened 2017-10-14
Tileset default filename


Ragath:
Would be nice if the default filename for a tileset was the name of the tileset instead of untitled.tsx.

PR #1790, opened 2017-10-20
Defaults filename to tileset name or image name

killerasus:
Solves issue #1783


src/tiled/documentmanager.cpp
```diff
@@ -510,7 +510,16 @@ bool DocumentManager::saveDocumentAs(Document *document)
         FormatHelper<TilesetFormat> helper(FileFormat::ReadWrite);
         filter = helper.filter();
 
-        fileName = getSaveFileName(QCoreApplication::translate("Tiled::Internal::MainWindow", "untitled.tsx"));
+        QString qName(QLatin1String("untitled.tsx"));
+        auto tileSetName = tilesetDocument->tileset()->name();
```
bjorn: There are some problems with this patch as-is, mostly notably it removes "untitled.tsx" as a translatable string and always tries to translate the suggested file name (which will never work). You also only trim when checking for empty but not when actually using the tileset name, and the naming style of `qName`, `tileSetName` and `cStrName` is off with the rest of the code.

I would implement this as follows:

```c++
auto suggestedFileName = tilesetDocument->tileset()->name().trimmed();
if (suggestedFileName.isEmpty())
    suggestedFileName = QCoreApplication::translate("Tiled::Internal::MainWindow", "untitled"));
suggestedFileName.append(QLatin1String(".tsx"));

fileName = getSaveFileName(suggestedFileName);
```

The above unifies the adding of the extension and reduces the translatable string. The same thing could be done for "untitled.tmx", so there is one less string to translate.

killerasus: Hi @bjorn thanks for reviewing my pull request. Would this issue also consider the changes to "untitled.tmx" or should I implement it as another issue?

killerasus: This would also consider updating translation files to refer to untitled string instead of untitled.tsx and untitled.tmx, right?

bjorn: > Would this issue also consider the changes to "untitled.tmx" or should I implement it as another issue?

Up to you, I'd be fine either way.

> This would also consider updating translation files to refer to untitled string instead of untitled.tsx and untitled.tmx, right?

No, I'm only updating the translation files before a new feature release (usually about 2 weeks in advance). This should be sometime next month.

bjorn:
Thanks!

Issue #1782, opened 2017-10-14
Tileset extension confirmation issue


Ragath:
When creating a new tileset and saving as an unsupported extension, answering No to the "are you sure?"-confirmation prompt doesn't take you back to the tileset creation dialog as expected, instead it ends the creation process.

bjorn:
Ah, you're right, and this will be a bit tricky to fix because the New Tileset dialog is already closed when the file name dialog is shown. So maybe that code should be moved into the dialog.

PR #1791, opened 2017-10-20
Reopen save dialog when user heeds file extension warning

agersant:
Proposed fix for https://github.com/bjorn/tiled/issues/1782.

With this change, the save dialog will re-open afters users answer no to the "Tiled may not automatically recognize your file when loading. Are you sure you want to save with this extension?" dialog.

The diff whitespace looks funny on Github for some reason, I'm not too sure why considering it's all spaces (and indented correctly).


bjorn:
Thanks! Your solution was simpler than I had expected and it also improves the same scenario when saving maps with different extensions.

Issue #1788, opened 2017-10-18
QXcbConnection: Could not connect to display 


cacciatc:
I recently tried to run Tiled with the --export-map flag as part of an automated build in a docker image and get the following error:
```
QXcbConnection: Could not connect to display 
```

Is it possible to run Tiled --export-map without a display? I am using the tiled package on Ubuntu.

bjorn:
Due to its use of `QGuiApplication` and `QPixmap`, Tiled does need an X server to function. However, you could run a headless X server, like `Xvfb`. That should allow Tiled to run and enable the use of `--export-map`.

cacciatc:
Thanks @bjorn. Installing xvfb and running Tiled ala:
```
xvfb-run tiled --export-map ...
```
did the trick. Is there a good place to document this in the Tiled docs? If so, happy to submit a PR.

bjorn:
> Is there a good place to document this in the Tiled docs? If so, happy to submit a PR.

That's probably something worth adding to [the page about the Export Formats](http://doc.mapeditor.org/en/latest/manual/export/), as part of a section about exporting on the command-line.

PR #1792, opened 2017-10-21
Added docs for running Tiled --export-map headless

cacciatc:
In response to issue #1788.


bjorn:
I've changed it a little since I considered it a bit out-of-place to put "Command-line" on the same level as all the export formats. Eventually there should probably be a separate page about all command-line options, but for now I've turned this into a note at the top where the command-line export option is mentioned.

Thanks for helping to make the docs more useful!

Issue #1722, opened 2017-09-06
Allow replacing existing objects with template instances


bjorn:
There should be a way to turn an already placed object into a template instance. Essentially the reverse of the "Detach" action that turns a template instance into a regular object.

Of course, the question is which template to use. I think there are two options:

* Add "Replace with Template" context menu action that uses the template selected in the Templates view.
* Add a menu similar to the one shown at issue #818, which allows selecting the template in the menu.

I'm somewhat leaning towards the first option.

bjorn:
@thabetx Is this something you may still get around to?

thabetx:
This looks like a good idea. I prefer the first option, mostly because it will have the same workflow as replacing the tile of an object. I will look into it.

bjorn:
Implemented in 6404b3ddd921684e3182cfb98867970ff5cfb6ff.

Issue #1732, opened 2017-09-06
Allow fixing wrong paths to tilesets in templates


thabetx:
Currently Tiled can fix a wrong path of a template referenced by a map. But doesn't offer fixing tileset references inside of template files.

Since template files aren't opened as top-level documents and any changes to them are saved immediately, they will need some special handling.

bjorn:
I've updated the title and description of this task to account for change ce60ac9100ef33931c54a8a648e2936ddd8fefc7, which removed the template groups.

thabetx:
The scenario for this is opening a map which has an instance of a template that has a broken link to a tileset. The main thing here is that there is no thing wrong with the MapDocument itself, and the map doesn't own the template and has no means of changing it.

This is exactly like fixing the image of the tileset, we open the tileset and fix it there.

So, what about offering to fix the template by loading it in the template viewer, there would be an option there to fix the tileset, I'm currently thinking of a context menu entry called "Replace Broken Tileset", and we should also be able to right click and replace the tile like #1525 (Replace object tile) .

I'm not sure if this is the best solution, and I'm open for suggestions.

bjorn:
@thabetx I like your suggestion, but I think it would need to be rather visible rather than hidden in a context menu, so for example a bar saying "Tileset not found" with a "Locate Tileset" button at the top of the template view.

Hmm, do we handle already images failing to load in a tileset referenced by a template? I think not...

In general I'm wondering if there isn't an easier way of dealing with broken references...

thabetx:
I'm thinking of adding a "Fix Tileset" button beside the tools in the templates dock, it only appears when there is a problem with the tileset, and it should work like the fix button in the broken links widget (promoting to change the path if it's the problem or open the tileset if the image is the problem).

bjorn:
@thabetx Yeah, that could work fine.

bjorn:
Closed by #1831.

PR #1793, opened 2017-10-21
Enable replacing an object with a template

thabetx:
Solves #1722

Works exactly like the replace tile feature.

replace_template

I had to move the selected template from the template creation tool into the tool manager so it can be used across all the tools, just like the tile.


src/tiled/changemapobject.cpp
```diff
@@ -245,19 +259,55 @@ void ResetInstances::undo()
...
+    emit mMapDocument->objectsChanged(mMapObjects);
+    emit mMapDocument->selectedObjectsChanged();
+}
+
+void ReplaceObjectsWithTemplate::undo()
+{
+    for (int i = 0; i < mMapObjects.size(); ++i) {
+        MapObject *current = mMapObjects.at(i);
+        MapObject *old = mOldMapObjects.at(i);
+        copyProperties(old, current);;
```
bjorn: Please remove second `;`

src/tiled/changemapobject.cpp
```diff
@@ -245,19 +259,55 @@ void ResetInstances::undo()
...
+        copyProperties(old, current);
+    }
+
+    emit mMapDocument->objectsChanged(mMapObjects);
+    emit mMapDocument->selectedObjectsChanged();
+}
+
+
+ReplaceObjectsWithTemplate::ReplaceObjectsWithTemplate(MapDocument *mapDocument,
+                                                      const QList<MapObject *> &mapObjects,
```
bjorn: Indentation missing a space.

src/tiled/changemapobject.cpp
```diff
@@ -245,19 +259,55 @@ void ResetInstances::undo()
...
+        MapObject *current = mMapObjects.at(i);
+        MapObject *old = mOldMapObjects.at(i);
+        copyProperties(old, current);;
     }
 
     emit mMapDocument->objectsChanged(mMapObjects);
     emit mMapDocument->selectedObjectsChanged();
 }
+
+ReplaceObjectsWithTemplate::~ReplaceObjectsWithTemplate()
```
bjorn: Please define the destructor right after the constructor (matching the order in the header file).

src/tiled/mapdocument.cpp
```diff
@@ -139,7 +139,7 @@ void MapDocument::saveSelectedObject(const QString &name, int groupIndex)
 
     if (ObjectTemplate *objectTemplate = model->saveObjectToDocument(object, name, groupIndex)) {
         // Convert the saved object into an instance and clear the changed properties flags
-        object->setTemplateRef({objectTemplate->templateGroup(), objectTemplate->id()});
+        object->setTemplateRef(objectTemplate->templateRef());
         object->setChangedProperties(0);
```
bjorn: Actually these lines should be covered by an undo command, right? Now that you've written `ReplaceObjectsWithTemplate`, you should probably use it here.

src/tiled/abstractobjecttool.cpp
```diff
@@ -249,6 +249,13 @@ void AbstractObjectTool::detachSelectedObjects()
     currentMapDocument->undoStack()->push(changeMapObjectCommand);
 }
 
+void AbstractObjectTool::replaceObjectsWithTemplate()
+{
+    mapDocument()->undoStack()->push(new ReplaceObjectsWithTemplate(mapDocument(),
+                                                                    mapDocument()->selectedObjects(),
+                                                                    {objectTemplate()->templateGroup(), objectTemplate()->id()}));
```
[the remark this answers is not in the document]
thabetx: For some reason I didn't commit that 😅.

src/tiled/mapdocument.cpp
thabetx: Hmm, currently this would be a partial undo, as removing from the template group is not implemented. But at least it would reset the object to being a normal object.

src/tiled/changemapobject.cpp
```diff
@@ -204,6 +204,20 @@ void DetachObjects::undo()
     emit mMapDocument->mapObjectModel()->objectsChanged(mMapObjects);
 }
 
+void copyProperties(const MapObject *source, MapObject *target) {
```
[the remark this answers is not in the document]
thabetx: Using the = operator looks like a good idea, it copies everything except id and position so I think i's okay but 'm not sure.

thabetx:
I like the option to add the name of the template in the context menu. It makes things clearer.

src/tiled/changemapobject.cpp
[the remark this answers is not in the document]
bjorn: Hmm, I guess `operator=` could be expected to copy those as well, so probably better to leave this as a special function.

src/tiled/changemapobject.cpp
```diff
@@ -204,6 +204,20 @@ void DetachObjects::undo()
...
+    target->setName(source->name());
+    target->setSize(source->size());
+    target->setType(source->type());
+    target->setTextData(source->textData());
+    target->setPolygon(source->polygon());
+    target->setShape(source->shape());
+    target->setCell(source->cell());
+    target->setRotation(source->rotation());
+    target->setVisible(source->VisibleProperty);
```
bjorn: I just noticed this little issue. I won't change it now on master to avoid conflicts, so please fix this to `source->isVisible()` as part of this PR.

src/tiled/mapdocument.cpp
bjorn: Yeah, I think partial undo would be fine here.

thabetx:
I cleaned the code and made some fixes and changes.

Replacing was enabled for tile objects only, so I fixed that as well, but now when replacing a non-tile object e.g. a polygon with a template, Tiled will crash. I think the reason is that the two objects have different sets of properties and when the properties dock is asked to update it can't set the properties of the new object.

So I will be looking into it and will do some more testing.

bjorn:
> Replacing was enabled for tile objects only, so I fixed that as well, but now when replacing a non-tile object e.g. a polygon with a template, Tiled will crash. I think the reason is that the two objects have different sets of properties and when the properties dock is asked to update it can't set the properties of the new object.

Ah, you're right. This new Replace by Template action is the first action that allows changing the actual type of an object, and some properties are only created for certain objects (like flipping for tile objects, and width/height only for rectangles, tiles and ellipses). So the PropertyBrowser will need some changes for this.

src/tiled/abstractobjecttool.cpp
```diff
@@ -358,13 +358,20 @@ void AbstractObjectTool::showContextMenu(MapObjectItem *clickedObjectItem,
...
-        auto replaceTemplateAction = menu.addAction(tr("Replace With Template"), this, SLOT(replaceObjectsWithTemplate()));
-        replaceTemplateAction->setEnabled(objectTemplate());
     }
 
+    // Create action for replacing an object with a template
+    auto selectedTemplate = objectTemplate();
+    auto replaceTemplateAction = menu.addAction(tr("Replace With Template"), this, SLOT(replaceObjectsWithTemplate()));
+
+    if (selectedTemplate)
+        replaceTemplateAction->setText(tr("Replace With \"") + selectedTemplate->name() + tr("\" Template"));
```
bjorn: This is not nice for the translator. You should instead use a parameter, like so:
```c++
tr("Replace With \"%1\" Template").arg(selectedTemplate->name())
```

thabetx: Alright.

bjorn:
Alright, I've merged `master` and in the process fixed that translation string and corrected `ReplaceObjectsWithTemplate::redo`.

The remaining issue is the crashing when changing the kind of object through replacing it with a template.

thabetx:
Hello bjron, thank you for the updates, I was stuck at updating the properties and the custom properties. I tested it and it's working great.

I will try working on #1823 .

Again thank you :).

bjorn:
> Hello bjorn, thank you for the updates, I was stuck at updating the properties and the custom properties. I tested it and it's working great.

No problem, I just needed this to be finished since I'm trying to move forward with a 1.1 release, which means finishing up the loose ends.

> I will try working on #1823.

Ah, I was about to look into it myself too, especially since I introduced this one recently, but feel free to have a look into it. It's probably not a hard one to fix.

thabetx:
Okay you can continue looking into it, I will try improving broken links.

Undoing after fixing doesn't reset and I will also look into fixing broken tileset link inside the template #1732 .

PR #1796, opened 2017-10-23
Fixed a bug where selection rectangle's shadow moved when zooming

agersant:
This change fixes an issue where the shadow of the selection rectangle would drift away from the selection based on the current zoom level (more zoom = more drift). With this change, the shadow is always on pixel away from the selection.

### DIFF
--- a/src/tiled/selectionrectangle.cpp
+++ b/src/tiled/selectionrectangle.cpp
@@ -20,6 +20,9 @@
 
 #include "selectionrectangle.h"
 
+#include "mapview.h"
+#include "zoomable.h"
+
 #include <QApplication>
 #include <QPainter>
 #include <QPalette>
@@ -45,7 +48,8 @@ QRectF SelectionRectangle::boundingRect() const
 }
 
 void SelectionRectangle::paint(QPainter *painter,
-                               const QStyleOptionGraphicsItem *, QWidget *)
+                               const QStyleOptionGraphicsItem *,
+                               QWidget *widget)
 {
     if (mRectangle.isNull())
         return;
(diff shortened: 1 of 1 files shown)


src/tiled/selectionrectangle.cpp
```diff
@@ -56,7 +60,8 @@ void SelectionRectangle::paint(QPainter *painter,
     QPen pen(black, 2, Qt::DotLine);
     pen.setCosmetic(true);
     painter->setPen(pen);
-    painter->drawRect(mRectangle.translated(1, 1));
+    const qreal scale = static_cast<MapView*>(widget->parent())->zoomable()->scale();
```
bjorn: I wish there was a better way to do this since this is a little hacky, but I think the alternative of somehow routing the scale to the `SelectionRectangle` would become too convoluted. To make the code a little saver and future-proof, I would prefer if it was written like this:
```c++
qreal scale = 1.0;
if (widget)
    if (MapView *mapView = dynamic_cast<MapView*>(widget->parent())
        scale = mapView->zoomable()->scale();
```

agersant: Good point. Not sure if that's preferable or not, but `widget->transform().m11()` (or .m22 for the y axis) also contains the desired scale factor.

bjorn: I guess you mean `painter->transform()`. Indeed it does, as long as the view is not rotated, which currently doesn't happen. It doesn't really matter, but I think I'd still prefer to access the zoomable.

agersant: Yeah, I agree none of the options are very elegant. Pushed changes making this safer.

bjorn:
Thanks a lot!

PR #1799, opened 2017-10-24
Added a new type of map object representing points

agersant:
Opening the PR while working on this so implementation can be discussed.

One thing I am not happy with at the moment is the various boundingBox functions. I don't think I am implementing this right. In my latest commits, I implemented OrthogonalRenderer::drawMapObject to draw pins. I then proceeded to update the bounding box functions like this:
1. `objectBounds` in `objectselectionitem.cpp`. As far as I can tell, this function is the one used to draw the animated selection rectangle when the item is highlighted. I put some magic numbers which match the pin drawing code I wrote in OrthogonalRenderer. I haven't looked at the other renderers yet but I'm worried they will need identical pin drawing functions too. **Where should the pin drawing code and the constants live**?
2. `OrthogonalRenderer::boundingRect`. This seems to be used to invalidate regions and repaint , so it needs a bit of padding (`extraSpace`) around the actual bounding box. Also uses the constants from the drawing code. Is this also used to position the resize handles? The handles end up a little bit removed from where they should be because of the padding. This doesn't happen for other object types so I must be doing something bad.
3. `OrthogonalRenderer::shape`. As far as I could tell, this one is used for mouse interactions and should match the visuals closely. (ie. identical to objectBounds from objectselectionitem). Also needs the drawing constants.

WIP screenshot:

point items

Progress:
- [x] Adding/removing points
- [x] Editing points
- [x] Points draw as pins
- [x] Save formats
- [x] Export formats
- [x] ~~Tool name localization~~
- [x] Tool icon
- [x] Documentation


bjorn:
One thing I forgot to mention, in order to fix the autobuilds (and compiling for me), please add new files also to the Qbs project (I'd also recommend using the Qbs project file, I'd like to get rid of the qmake one (#1484)).

agersant:
Last two commits clean up the duplicate functionality between renderers and bounding box functions. I ended up using MapRenderer as the home for the draw/shape utility functions, this makes them available to the renderer subclasses and also to the tmxrasterizer.

agersant:
New progress screenshot, now working on isometric maps and with tool icon:

point items 2

agersant:
Questions about the remaining items:

- Is there anything to do for localization support? I am under the impression that it's up to translators to re-generate .ts files and fill in missing texts?
- Is it safe to update the manual or should that only be done after a new release?

bjorn:
> Is there anything to do for localization support? I am under the impression that it's up to translators to re-generate .ts files and fill in missing texts?

For localization I do a string-freeze about two weeks before the release, which is when the .ts files are re-generated and all translators get a message asking them to start updating the translation. You don't have to do anything.

> Is it safe to update the manual or should that only be done after a new release?

You could definitely help updating the manual and you could do so as part of this PR. When I released Tiled 1.0 the manual unfortunately didn't really exist, which is why there is no 1.0-version of the manual hosted. Now the manual covers a lot of things, and many of them are already 1.1-specific. What I'll do upon release is to host both the 1.1 manual (which I could make the default) as well as master, so that manual updates can always go along with the changes without confusing readers using the Tiled releases. This is easy to do now that we are hosted by [readthedocs.org](https://readthedocs.org).

bjorn:
Great work, @agersant! I only found some minor issues and decided to just fix them instead of going through another review round.

PR #1800, opened 2017-10-26
Reduce translatable strings

killerasus:
Removes the need to translate untitled.tmx to just translate untitled, that is already used in saving a tsx file.

### DIFF
--- a/src/tiled/documentmanager.cpp
+++ b/src/tiled/documentmanager.cpp
@@ -496,7 +496,10 @@ bool DocumentManager::saveDocumentAs(Document *document)
         FormatHelper<MapFormat> helper(FileFormat::ReadWrite);
         filter = helper.filter();
 
-        fileName = getSaveFileName(QCoreApplication::translate("Tiled::Internal::MainWindow", "untitled.tmx"));
+        auto suggestedFileName = QCoreApplication::translate("Tiled::Internal::MainWindow", "untitled");
+        suggestedFileName.append(QLatin1String(".tmx"));
+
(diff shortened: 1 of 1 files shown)


src/tiled/documentmanager.cpp
```diff
@@ -495,7 +495,11 @@ bool DocumentManager::saveDocumentAs(Document *document)
         FormatHelper<MapFormat> helper(FileFormat::ReadWrite);
         filter = helper.filter();
 
-        fileName = getSaveFileName(QCoreApplication::translate("Tiled::Internal::MainWindow", "untitled.tmx"));
+        auto suggestedFileName = QCoreApplication::translate("Tiled::Internal::MainWindow", "untitled");
+        if(!suggestedFileName.isEmpty())
```
bjorn: Coding style: please put a space after `if`

However, in this case you can just remove the condition, since it would be an error for the translated string to be empty (it can never be, because in case of an empty translation, the source string is returned).

bjorn:
Ouch, it seems like you're not having much luck with git. :-/

bjorn:
I've done the interactive rebase for you, skipping commits that shouldn't be there and squashing the rest together.

killerasus:
Hey @bjorn, thanks. I messed up a bit using rebase. I was about to start over a new fork from the master.

bjorn:
> Hey @bjorn, thanks. I messed up a bit using rebase. I was about to start over a new fork from the master.

That's a bit radical! Please try to learn how to rebase and force-push. Doing a new fork should never be necessary. Play around with it. :-)

Issue #1603, opened 2017-06-08
Add ability to reorder terrains


bjorn:
Currently, it is not possible to change the order of the terrains defined in a tileset.

There should be only one difficulty with this, which is that reordering terrains affects the terrain IDs stored on the tiles, which will have to be adjusted accordingly. But this is similar to when a terrain is removed, which is already implemented.

Reported [on the forum](http://discourse.mapeditor.org/t/reordering-terrain-and-tilesets/2437).

eishiya:
It seems that the ability to reorder terrains was removed at some point, perhaps when they were merged with Wang tiles? It would be nice to have it back.

bjorn:
You're right, the ability to reorder Wang colors (and Wang sets) was never implemented, so now this feature is missing again.

UliAbo:
That would be quite helpful to have! Is this something complicated to be implemented? I would love to see that in, as I have lots of Terrains now in our Project.

bjorn:
> That would be quite helpful to have! Is this something complicated to be implemented? I would love to see that in, as I have lots of Terrains now in our Project.

Unfortunately it's not entirely trivial, since a `WangId` stores terrain _indexes_, which will all need updating when we allow terrains to be reordered. Alternatively we could switch to assigning terrain IDs. Either way it's not too complicated either, so I could probably still fit this into 1.9.

spiralhug:
Wanted to bump this to say that this is still an issue and one we'd love to have implemented for our project too!

PR #1802, opened 2017-10-27
Add ability to reorder terrains (#1603)

dorkster:


src/tiled/addremoveterrain.cpp
```diff
@@ -58,6 +58,22 @@ void AddRemoveTerrain::addTerrain()
     mTerrain = nullptr;
 }
 
+void AddRemoveTerrain::moveTerrainUp()
+{
+    Q_ASSERT(mTerrain);
```
bjorn: You're initializing the terrain to `nullptr` from the constructor of `MoveTerrainUp` and `MoveTerrainDown`, so this assertion failed for me. Should probably be `Q_ASSERT(!mTerrain);`? And then you can also get rid of the `mTerrain = nullptr;` below.

However, I wonder why these classes are derived from `AddRemoveTerrain`. What is the benefit of doing that? I'd suggest to make a separate `MoveTerrain` base class for these two instead.

src/tiled/terraindock.cpp
```diff
@@ -302,6 +314,8 @@ void TerrainDock::setCurrentTerrain(Terrain *terrain)
     mEraseTerrainButton->setChecked(terrain == nullptr);
 
     mRemoveTerrainType->setEnabled(terrain != nullptr);
+    mMoveTerrainTypeUp->setEnabled(terrain != nullptr);
+    mMoveTerrainTypeDown->setEnabled(terrain != nullptr);
```
bjorn: Could you look into also enabling/disabling these actions as appropriate based on the currently selected terrain? So that when the first terrain is selected, the "move terrain up" action is disabled?

src/tiled/tileseteditor.cpp
```diff
@@ -237,6 +237,8 @@ TilesetEditor::TilesetEditor(QObject *parent)
     connect(mTerrainDock, &TerrainDock::currentTerrainChanged, this, &TilesetEditor::currentTerrainChanged);
     connect(mTerrainDock, &TerrainDock::addTerrainTypeRequested, this, &TilesetEditor::addTerrainType);
     connect(mTerrainDock, &TerrainDock::removeTerrainTypeRequested, this, &TilesetEditor::removeTerrainType);
+    connect(mTerrainDock, &TerrainDock::moveTerrainTypeUpRequested, this, &TilesetEditor::moveTerrainTypeUp);
+    connect(mTerrainDock, &TerrainDock::moveTerrainTypeDownRequested, this, &TilesetEditor::moveTerrainTypeDown);
```
bjorn: I believe you could just implement these actions in the `TerrainDock` instead of routing them to the `TilesetEditor`. The terrain dock also has the document to which to push the undo command (`qobject_cast<TilesetDocument*>(mDocument)`).

src/tiled/addremoveterrain.cpp
dorkster: I've fixed this by creating a MoveTerrain class that only has the necessary members for moving terrains.

src/tiled/terraindock.cpp
dorkster: This has been implemented in e637e0c

src/tiled/tileseteditor.cpp
dorkster: Good call. When I was initially writing this, I definitely felt like I was going through one too many layers of indirection.

dorkster:
Thanks for being so thorough. I've made the changes you suggested.

src/tiled/terraindock.cpp
```diff
@@ -214,6 +227,13 @@ void TerrainDock::setDocument(Document *document)
         connect(terrainModel, &TilesetTerrainModel::terrainRemoved,
                 this, &TerrainDock::refreshCurrentTerrain);
 
+        /*
+         * The current terrain does not change when moving terrains.
+         * We need to refresh this in order to disable the up/down buttons when
+         * appropriate.
+         */
+        connect(terrainModel, &TilesetTerrainModel::updateTerrainDock,
```
bjorn: I think you could just use `QAbstractItemModel::rowsMoved` and drop the special `updateTerrainDock` signal.

dorkster: I tried that, but it gets triggered when `beginMoveRows()` executes, despite what the QT docs imply. We need a signal that happens after `endMoveRows()`, so I made `updateTerrainDock()`.

dorkster: Okay, I may have been wrong here. I think there is a bug in my code, as `QAbstractItemModel::rowsMoved` should really work.

dorkster: Ah, I was passing a TerrainModel pointer instead of TilesetTerrainModel to `connect()`. It works perfectly with the `QAbstractItemModel::rowsMoved` signal. I'll update my PR.

bjorn:
Works great and looks great, thanks!

PR #1803, opened 2017-10-27
Update .gitignore

dorkster:
- Add ignore for gmon.out, which gets created by gdb on Linux
- Add ignore for moc_predefs.h files, which appear to be auto-generated

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ moc_*.cpp
 ui_*.h
 qrc_*
 default/
+moc_predefs.h
 
 # Build artifacts on Windows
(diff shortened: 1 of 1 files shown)


bjorn:
`gmon.out` does not technically have anything to do with Tiled, and could instead be in your global git ignore list, but let's have it.

dorkster:
I never thought to use a global .gitignore. Thanks for the tip!

bjorn:
> I never thought to use a global .gitignore. Thanks for the tip!

See `core.excludesFile` in `git help config`. :-)

PR #1804, opened 2017-10-29
Ability to reset widgets layout arrangement to default setting

keshav2010:
This adds a missing feature in tiled that will allow a user to reset their editor layout to the default setting that tiled comes with.

### DIFF
--- a/src/tiled/mainwindow.cpp
+++ b/src/tiled/mainwindow.cpp
@@ -380,6 +380,9 @@ MainWindow::MainWindow(QWidget *parent, Qt::WindowFlags flags)
     mViewsAndToolbarsMenu = new QMenu(this);
     mViewsAndToolbarsAction = new QAction(tr("Views and Toolbars"), this);
     mViewsAndToolbarsAction->setMenu(mViewsAndToolbarsMenu);
+
+    mResetDefaultLayout = new QAction(tr("Reset To Default Layout"),this);
+
     mShowObjectTypesEditor = new QAction(tr("Object Types Editor"), this);
     mShowObjectTypesEditor->setCheckable(true);
     mUi->menuView->insertAction(mUi->actionShowGrid, mViewsAndToolbarsAction);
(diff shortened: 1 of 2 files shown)


src/tiled/mainwindow.cpp
```diff
@@ -380,6 +380,9 @@ MainWindow::MainWindow(QWidget *parent, Qt::WindowFlags flags)
     mViewsAndToolbarsMenu = new QMenu(this);
     mViewsAndToolbarsAction = new QAction(tr("Views and Toolbars"), this);
     mViewsAndToolbarsAction->setMenu(mViewsAndToolbarsMenu);
+
+    mResetDefaultLayout = new QAction(tr("Reset To Default Layout"),this);
```
bjorn: "to" should be in lowercase (English title casing rules).

Coding style: please put a space after all commas.

src/tiled/mainwindow.cpp
```diff
@@ -416,6 +419,8 @@ MainWindow::MainWindow(QWidget *parent, Qt::WindowFlags flags)
     connect(mDocumentManager, SIGNAL(reloadError(QString)),
             this, SLOT(reloadError(QString)));
 
+    connect(mResetDefaultLayout,SIGNAL(triggered(bool)),this,SLOT(resetDefaultLayout()));
```
bjorn: Please prefer modern connect syntax for new connections (eventually I'm going to do a commit changing all of the old ones I guess):
```c++
connect(mResetDefaultLayout, &QAction:triggered, this, &MainWindow::resetDefaultLayout);
```
The new syntax works more reliably when things are renamed / refactored, since the compiler can check the existence of the signals/slots and the compatibility of their arguments.

src/tiled/mainwindow.h
```diff
@@ -109,6 +109,7 @@ public slots:
     void dropEvent(QDropEvent *) override;
 
 private slots:
+    //File tab
```
bjorn: This would need to be "File menu" and there should be a space after `//`. But I don't really see a reason to have these comments.

src/tiled/mainwindow.h
```diff
@@ -211,6 +215,7 @@ private slots:
     QAction *mViewsAndToolbarsAction;
     QAction *mShowObjectTypesEditor;
 
+    QAction *mResetDefaultLayout;
```
bjorn: Should be named `mResetToDefaultLayout` or simply `mResetLayout`, since it's not resetting the default layout.

keshav2010:
Thanks for the feedback, I will be working on the new approach and would try to strictly follow the guidelines in next pull request :+1:

PR #1805, opened 2017-10-29
Update tiled_ru.ts

pavel-tkachenko:


bjorn:
@oktonion and @HellGhost, would you like to help with the review?

oktonion:
@bjorn yep

translations/tiled_ru.ts
```diff
@@ -4845,7 +4845,7 @@ Please select specific format.</source>
     <message>
         <location filename="../src/plugins/tmw/tmwplugin.cpp" line="+47"/>
         <source>Multiple collision layers found!</source>
-        <translation>Обнаружены несколько слоев столкновений!</translation>
+        <translation>Обнаружено несколько слоев столкновений!</translation>
```
oktonion: In my view the old version is closer to original.

translations/tiled_ru.ts
```diff
@@ -1433,7 +1433,7 @@ Line %1, column %2</source>
         <location filename="../src/tiled/preferencesdialog.cpp" line="+60"/>
         <location line="+131"/>
         <source>Native</source>
-        <translation>Нативный</translation>
+        <translation>Системный</translation>
```
oktonion: Need to know the context of this string. Changing 'Native' to 'System' may be not a good choice if it is not meant.

bjorn: This is about the theme that tries to look like the system theme, which can also be called the "native" theme. Up to you guys what is the best way to convey this in Russian.

pavel-tkachenko: There is another one - "По умолчанию" ("Default").

translations/tiled_ru.ts
pavel-tkachenko: Needs discussion.

translations/tiled_ru.ts
oktonion: New 'Системный' is better then.

translations/tiled_ru.ts
oktonion: ¯\_(ツ)_/¯

translations/tiled_ru.ts
bjorn: > There is another one - "По умолчанию" ("Default").

Yeah, I didn't go with "Default" because it's ambiguous which default is meant. In Tiled, the default theme on Windows and Linux is "Tiled Fusion".

translations/tiled_ru.ts
bjorn: Let's not discuss this string, because it is part of a plugin I just removed (2fd383858e6142ec7a4a2e03dd2c2e3e2e8e0ee0). :-)

bjorn:
Thanks! :-)

oktonion:
@bjorn you are welcome =)

PR #1808, opened 2017-11-01
Make some improvements to the Qbs build

jakepetroules:
Here's some improvements to the Qbs build as I suggested earlier on IRC. It should also fix the issue with the tmxviewer and automapping converter and other executables running from the wrong directory.


jakepetroules:
By the way, from the [Qbs 1.8 release blog](http://blog.qt.io/blog/2017/05/29/qbs-1-8-released/):

> Second, all files that are part of a bundle on Apple platforms like macOS and iOS are now tagged with “bundle.content”. If you want to install the files that constitute a bundle, this is the one and only tag that your installation Group’s fileTagsFilter should reference. Previous methods of installing bundles based on the file tags of the individual files that constitute them are not guaranteed to work.

bjorn:
Alright, finally got a recent Qbs to work on Travis! Next I should probably look into caching the build, but this has already taken me half a day... :-/

In any case, thanks a lot for the improvements! I think actually my next step should be looking into pushing the Qbs macOS build to itch.io as macOS development snapshot.

jakepetroules:
By the way, your setup-qbs.sh should probably run `make install`. I don't remember if we guarantee that the non-installed qmake build will actually work.

bjorn:
> By the way, your setup-qbs.sh should probably run make install. I don't remember if we guarantee that the non-installed qmake build will actually work.

I considered this, but didn't do it since it actually compile Tiled fine without installing. When I make the Abs build cached, I will definitely want to install it somewhere and cache only the installation directory.

PR #1809, opened 2017-11-02
Ability to reset MapEditor's layout to default layout without terminating/restarting Tiled

keshav2010:
This allows the user to reset the MapEditor's widget layout to default layout setting that tiled comes with, without requiring them to restart the Tiled.

### DIFF
--- a/src/tiled/editor.h
+++ b/src/tiled/editor.h
@@ -66,6 +66,8 @@ public:
     virtual StandardActions enabledStandardActions() const = 0;
     virtual void performStandardAction(StandardAction action) = 0;
 
+    virtual void resetLayout() = 0;
+
 signals:
     void enabledStandardActionsChanged();
 };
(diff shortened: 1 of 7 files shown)


bjorn:
@keshav2010 Sorry I'm simply busy.

The most important bit of feedback I can give you, is that the code setting up those views should not be duplicated. In your current patch you've leaving the initial setup in the constructor alone, but this code should be replaced by your new `resetLayout` function.

keshav2010:
Ohh alright, I'm sorry for being a little impatient for referencing this pull request a few days back, would avoid referencing it unnecessarily in future :+1: 
I will be committing the new changes within few days :+1:  Thank you for the feedback :+1:

bjorn:
I've rebased this on the latest master, because I couldn't test it due to a crash on startup (which affected my master for a short period). Then I've made some adjustments myself rather than doing another review, since I wanted to merge this change (Tiled 1.1 will enter String Freeze now, so otherwise it'd have to wait until the next release).

Thanks for working on adding this action!

keshav2010:
Thanks :) that was indeed motivating considering it was my first contribution to Tiled, meanwhile i studied the Tiled source code as well as Qt framework, so hopefully i will eventually be able to contribute again :) Thanks once again.

bjorn:
@keshav2010 I'm looking forward to your next PR! :-)

PR #1813, opened 2017-11-08
Refer to templates directly (by file name)

bjorn:
This change removes the "template group" file and instead makes it so that each template is stored in its own file. Saved template instances refer to the template by the file name.

It is an attempt at simplifying the system, which may be easier than filling in the remaining loose ends with the template group files and could also have a better usability.


bjorn:
@thabetx For various reasons I have decided to work on this change instead of tasks like "add ability to remove a template from a group" and "add action to unload a template group". It would be great if we can simplify the template system by dropping the concept of template groups entirely, so this is my work-in-progress attempt at doing this.

So far I've touched a lot of code and only made sure it still compiles and runs, but it does leave the templates unusable for now. The main component that still needs to be replaced is the `ObjectTemplateModel`. I think we can replace it with a `QFileSystemModel` that filters for template files with the user in control of its root directory. The other part that still needs some work is the modifying and saving of templates.

There will be a few conflicts with #1793, due to the removal of `TemplateRef` for example, but they should be pretty trivial.

Any feedback is welcome.

bjorn:
@thabetx Everything should now be fully functional. Will you have some time for looking into this change and giving feedback?

thabetx:
Alright, I will  try it and check the code on my next weekend (Friday, Saturday).

bjorn:
> Alright, I will try it and check the code on my next weekend (Friday, Saturday).

Thanks!

@Seanba Since you've already implemented support for templates in Tiled2Unity, your feedback on this change would also be welcome.

thabetx:
This is a big change :), I think template groups had some pros, mainly having separate bundles of templates, but I don't mind the new format, the system folders can represent the template groups and I like the option of setting the templates directory. Reading maps should be much easier now.

Having per map templates folder would be nice, but I think it better be part of the Tiled project feature in the future :).

I suggest doing a project wide search for templateGroup and removing any remaining obsolete code.

So far things have the same behavior. I ran into two problems:
- When closing Tiled, this assertion is hit: `ASSERT: "mTilesets.isEmpty()" in file tiled/src/libtiled/tilesetmanager.cpp, line 68`
- If an external tileset is used but not opened, Tiled crashes on startup. If the tileset is opened in Tiled but its tab is on the right of the map's tab, Tiled also crashes on startup.

bjorn:
> This is a big change :), I think template groups had some pros, mainly having separate bundles of templates, but I don't mind the new format, the system folders can represent the template groups and I like the option of setting the templates directory. Reading maps should be much easier now.

Yeah, I think regular folders are working quite well for grouping templates.

> Having per map templates folder would be nice, but I think it better be part of the Tiled project feature in the future :).

This will indeed be part of the future project support. I'm reluctant to introduce such a concept only for the templates now, since the same makes sense for image and tileset references. So they'll all either use the same "assets root" folder or we can have separate folders for each data type (I'm not entirely decided).

> I suggest doing a project wide search for templateGroup and removing any remaining obsolete code.

Thanks for the suggestion! As you can see I found some remaining stuff indeed. :-)

> So far things have the same behavior. I ran into two problems:

I'll definitely have a closer look at those issues soon. Thanks for testing!

bjorn:
@thabetx Both issues you found should be fixed now. Thanks again for catching these!

bjorn:
Dang, now I forgot to take care of this strange linker error that plagues the MinGW build since this change:

```
ERROR: "C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\BIN\x86_amd64\link.exe" /nologo /DLL /IMPLIB:C:\projects\tiled\release\json.05d97e6e\json.lib /INCREMENTAL:NO /MACHINE:X64 /OSVERSION:5.02 "/SUBSYSTEM:CONSOLE,5.02" C:\projects\tiled\release\json.05d97e6e\.obj\09d868704433c02e\json.cpp.obj C:\projects\tiled\release\json.05d97e6e\.obj\3a52ce780950d4d9\jsonplugin.cpp.obj C:\projects\tiled\release\json.05d97e6e\.obj\73489b871e127538\moc_jsonplugin.cpp.obj C:\projects\tiled\release\libtiled.080f31f8\tiled.lib Shell32.lib C:\Qt\5.9\msvc2013_64\lib\qtmain.lib C:\Qt\5.9\msvc2013_64\lib\Qt5Core.lib C:\Qt\5.9\msvc2013_64\lib\Qt5Gui.lib /OUT:C:\projects\tiled\release\json.05d97e6e\json.dll /LIBPATH:C:\Qt\5.9\msvc2013_64\lib
tiled.lib(tiled.dll) : error LNK2005: "public: __cdecl QMap<class QString,class QVariant>::~QMap<class QString,class QVariant>(void)" (??1?$QMap@VQString@@VQVariant@@@@QEAA@XZ) already defined in json.cpp.obj
   Creating library C:\projects\tiled\release\json.05d97e6e\json.lib and object C:\projects\tiled\release\json.05d97e6e\json.exp
C:\projects\tiled\release\json.05d97e6e\json.dll : fatal error LNK1169: one or more multiply defined symbols found
ERROR: Process failed with exit code 1169.
```
This seems to be saying that the destructor of the `QMap<class QString,class QVariant>` template instantiation is being defined in both `json.dll` and `tiled.dll`. I don't know how to fix that properly, but it can probably be avoided by not returning `QVariantMap` from `MapToVariantConverter::toVariant`.

bjorn:
Alright, needed a few more changes to get rid of that annoying linker error. Fixed now with change 73e8b5c2cd23d44baa49694f2f86f7ca31bccafd.

mitchcurtis:
I'm also getting that error with my project, and it mentions tiled:

```
tiled.lib(tiled.dll) : error LNK2005: "public: __cdecl QMap<class QString,class QVariant>::~QMap<class QString,class QVariant>(void)" (??1?$QMap@VQString@@VQVariant@@@@QEAA@XZ) already defined in SpriteSequence.cpp.obj
   Creating library C:\dev\tshnm-iso-qt5_14_d-Debug\Debug\libisle.7c81f238\isle.lib and object C:\dev\tshnm-iso-qt5_14_d-Debug\Debug\libisle.7c81f238\isle.exp
C:\dev\tshnm-iso-qt5_14_d-Debug\Debug\libisle.7c81f238\isle.dll : fatal error LNK1169: one or more multiply defined symbols found
```

I have no idea why though. Good to know that the fix is as easy as changing to QVariant though.

mitchcurtis:
Nope, still get it, even after changing to QVariant:

```
#ifndef SPRITESEQUENCE_H
#define SPRITESEQUENCE_H

#include <QQuickPaintedItem>
#include <QPixmap>
#include <QTimer>
#include <QVariant>

#include "IsleGlobal.h"

class SceneItemComponent;

class ISLE_EXPORT Sprite : public QObject
{
    Q_OBJECT
    Q_PROPERTY(QString name READ name WRITE setName NOTIFY nameChanged)
    Q_PROPERTY(int frameX READ frameX WRITE setFrameX NOTIFY frameXChanged)
    Q_PROPERTY(int frameY READ frameY WRITE setFrameY NOTIFY frameYChanged)
    Q_PROPERTY(int frameCount READ frameCount WRITE setFrameCount NOTIFY frameCountChanged)
    Q_PROPERTY(qreal frameRate READ frameRate WRITE setFrameRate NOTIFY frameRateChanged)
    Q_PROPERTY(qreal wanderCooldown READ wanderCooldown WRITE setWanderCooldown NOTIFY wanderCooldownChanged)
    Q_PROPERTY(bool reverse READ isReverse WRITE setReverse NOTIFY reverseChanged)
    Q_PROPERTY(int loops READ loops WRITE setLoops NOTIFY loopsChanged)
    Q_PROPERTY(QVariant wanderWeights READ wanderWeights WRITE setWanderWeights NOTIFY wanderWeightsChanged)

public:
    Sprite() = default;
    ~Sprite() override = default;

    QString name() const;
    void setName(const QString &name);

    int frameX() const;
    void setFrameX(int frameX);

    int frameY() const;
    void setFrameY(int frameY);

    int frameCount() const;
    void setFrameCount(int frameCount);

    qreal frameRate() const;
    void setFrameRate(const qreal &frameRate);

    qreal wanderCooldown() const;
    void setWanderCooldown(qreal wanderCooldown);

    bool isReverse() const;
    void setReverse(bool reverse);

    int loops() const;
    void setLoops(int loops);

    QVariant wanderWeights() const;
    void setWanderWeights(const QVariant &wanderWeights);

signals:
    void nameChanged();
    void frameXChanged();
    void frameYChanged();
    void frameCountChanged();
    void frameRateChanged();
    void wanderCooldownChanged();
    void reverseChanged();
    void wanderWeightsChanged();
    void started();
    void finished();
    void loopFinished();
    void loopsChanged();

private:
    friend class SpriteSequence;

    QString mName;
    int mFrameX = 0;
    int mFrameY = 0;
    int mFrameCount = 0;
    qreal mFrameRate = 0.0;
    bool mReverse = false;
    qreal mCooldown = 0.0;
    int mLoops = 1;
    QVariant mTo;

    qint64 mLastUsageSEpoch = 0;
};

ISLE_EXPORT QDebug operator<<(QDebug debug, Sprite *sprite);

class ISLE_EXPORT SpriteSequence : public QQuickPaintedItem
{
    Q_OBJECT
    Q_PROPERTY(bool running READ isRunning WRITE setRunning NOTIFY runningChanged)
    Q_PROPERTY(SceneItemComponent *_sceneItemComponent READ sceneItemComponent
        WRITE setSceneItemComponent NOTIFY sceneItemComponentChanged)
    Q_PROPERTY(QUrl source READ source WRITE setSource NOTIFY sourceChanged)
    Q_PROPERTY(int frameHeight READ frameHeight WRITE setFrameHeight NOTIFY frameHeightChanged)
    Q_PROPERTY(int frameWidth READ frameWidth WRITE setFrameWidth NOTIFY frameWidthChanged)
    Q_PROPERTY(QString currentSprite READ currentSprite NOTIFY currentSpriteChanged)
    Q_PROPERTY(QQmlListProperty<Sprite> sprites READ sprites)
    Q_CLASSINFO("DefaultProperty", "sprites")

public:
    SpriteSequence();
    ~SpriteSequence() override;

    bool isRunning() const;
    void setRunning(bool isRunning);

    SceneItemComponent *sceneItemComponent() const;
    void setSceneItemComponent(SceneItemComponent *sceneItemComponent);

    QUrl source() const;
    void setSource(const QUrl &source);

    int frameWidth() const;
    void setFrameWidth(int frameWidth);

    int frameHeight() const;
    void setFrameHeight(int frameHeight);

    QString currentSprite() const;

    QQmlListProperty<Sprite> sprites();
    // Just public for the QQmlListProperty.
    // If we didn't need to also do some other stuff (add the sprite to the hash),
    // we wouldn't need it.
    void appendSprite(Sprite *sprite);

    // Immediately starts playing \a spriteName, cancelling any currently playing animation.
    Q_INVOKABLE void jumpTo(const QString &spriteName);
    // Waits for the current sprite to finish, then plays \a path.
    // It is perfectly fine to pass a list with only one path: that tells SpriteSequence
    // to wait and then play, as usual.
    Q_INVOKABLE void followPath(const QStringList &path);

    void paint(QPainter *painter) override;

protected:
    void componentComplete() override;

signals:
    void runningChanged();
    void sceneItemComponentChanged();
    void sourceChanged();
    void frameWidthChanged();
    void frameHeightChanged();
    void currentSpriteChanged();
    void scheduledUpdate();

public slots:
    void onTimerElapsed();

private:
    friend class SpriteSequenceLayer;

    void setCurrentSprite(Sprite *currentSprite);
    Sprite *determineNextSprite();
    QPoint sourcePosForFrameIndex(int index);

    bool findPathToGoalSprite(Sprite *sprite, Sprite *goalSprite, QVector<QString> &path);
    bool isValidPathToGoalSprite(const QString &goalSpriteName, const QStringList &path) const;

    void doSetFrameWidth(int frameWidth, bool explicitlySet);
    void doSetFrameHeight(int frameHeight, bool explicitlySet);

    bool mComplete = false;
    bool mValid = false;

    bool mRunning = true;
    SceneItemComponent *mSceneItemComponent = nullptr;
    QUrl mSource;
    QPixmap mPixmap;
    int mFrameWidth = 0;
    int mFrameHeight = 0;
    bool mExplicitFrameWidth = false;
    bool mExplicitFrameHeight = false;
    QList<Sprite*> mSprites;
    // For quick lookup since we need a list for the QQmlListProperty.
    QHash<QString, Sprite*> mSpritesHash;

    Sprite *mCurrentSprite = nullptr;
    int mCurrentSpriteRemainingLoops = 0;
    QVector<QString> mRemainingPathToFollow;
    int mCurrentFrameIndex = -1;
    int mSourceX = 0;
    int mSourceY = 0;
    QTimer mTimer;
};

#endif // SPRITESEQUENCE_H

```

```
tiled.lib(tiled.dll) : error LNK2005: "public: __cdecl QMap<class QString,class QVariant>::~QMap<class QString,class QVariant>(void)" (??1?$QMap@VQString@@VQVariant@@@@QEAA@XZ) already defined in SpriteSequence.cpp.obj
   Creating library C:\dev\tshnm-iso-qt5_14_d-Debug\Debug\libisle.7c81f238\isle.lib and object C:\dev\tshnm-iso-qt5_14_d-Debug\Debug\libisle.7c81f238\isle.exp
C:\dev\tshnm-iso-qt5_14_d-Debug\Debug\libisle.7c81f238\isle.dll : fatal error LNK1169: one or more multiply defined symbols found
```

bjorn:
> Nope, still get it, even after changing to QVariant:

While you've posted the code in `SpriteSequence.h`, the linker error is about `SpriteSequence.cpp`. It must be calling the destructor of a QVariantMap somewhere.

I guess the correct place to fix this properly would be in libtiled, but I'm not sure how.

mitchcurtis:
The only instance of ~ in the .cpp is the SpriteSequence destructor:

```
SpriteSequence::~SpriteSequence()
{
}
```

There are also no deletes there, so I'll have to keep on looking.

mitchcurtis:
https://github.com/treefrogframework/treefrog-framework/issues/211 says:

> This is because TSession inherits from template class QVariantMap, which shouldn't be done - see section Things to avoid in https://wiki.qt.io/Coding_Conventions

https://wiki.qt.io/Coding_Conventions says:

>Things to avoid
>
>     Do not inherit from template/tool classes
>
>     The destructors are not virtual, leading to potential memory leaks
>     The symbols are not exported (and mostly inline), leading to interesting symbol clashes.
>     Example: Library A has
> 
>         class Q_EXPORT X: public QList<QVariant> {};
> 
>     and library B has
> 
>         class Q_EXPORT Y: public QList<QVariant> {};
> 
>     Suddenly, QList<QVariant>'s symbols are exported from two libraries - /clash/.
> 

The Properties class does that. Would it be possible to use composition instead of inheritance there?

mitchcurtis:
https://github.com/bjorn/tiled/pull/2679

bjorn:
> There are also no deletes there, so I'll have to keep on looking. Unless you mean that it's implicitly being called, which is definitely the case.

Yes, an implicit call would be enough to trigger this issue.

Thanks for finding out that it's related to inheriting from `QVariantMap`! In our case, there should not be problem with a non-virtual destructor since Properties are never heap-allocated on their own and especially not destructed through a base class pointer. But the symbol clashes are definitely a problem.

PR #1818, opened 2017-11-15
 Open file button added when no file opened 

RhenaudTheLukark:
I created the missing "Open File..." button in the view when no file is opened.

Result in QtCreator:

Result in Tiled:

### DIFF
--- a/src/tiled/noeditorwidget.cpp
+++ b/src/tiled/noeditorwidget.cpp
@@ -22,6 +22,7 @@
 #include "ui_noeditorwidget.h"
 
 #include "actionmanager.h"
+#include "documentmanager.h"
 
 namespace Tiled {
 namespace Internal {
@@ -31,9 +32,9 @@ NoEditorWidget::NoEditorWidget(QWidget *parent) :
     ui(new Ui::NoEditorWidget)
 {
     ui->setupUi(this);
-
     connect(ui->newMapButton, &QPushButton::clicked, this, &NoEditorWidget::newMap);
     connect(ui->newTilesetButton, &QPushButton::clicked, this, &NoEditorWidget::newTileset);
+    connect(ui->openFileButton, &QPushButton::clicked, this, &NoEditorWidget::openFile);
 }
 
 NoEditorWidget::~NoEditorWidget()
@@ -61,6 +62,12 @@ void NoEditorWidget::newMap()
 void NoEditorWidget::newTileset()
 {
     ActionManager::action("file.new_tileset")->trigger();
+}
+
+void NoEditorWidget::openFile()
+{
+    DocumentManager *documentManager = DocumentManager::instance();
+    documentManager->openFile();
 }
 
 } // namespace Internal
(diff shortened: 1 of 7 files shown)


src/tiled/noeditorwidget.cpp
```diff
@@ -31,9 +32,9 @@ NoEditorWidget::NoEditorWidget(QWidget *parent) :
     ui(new Ui::NoEditorWidget)
 {
     ui->setupUi(this);
-
```
bjorn: Please leave the empty line there. :-)

src/tiled/noeditorwidget.cpp
```diff
@@ -31,9 +32,9 @@ NoEditorWidget::NoEditorWidget(QWidget *parent) :
...
     connect(ui->newMapButton, &QPushButton::clicked, this, &NoEditorWidget::newMap);
     connect(ui->newTilesetButton, &QPushButton::clicked, this, &NoEditorWidget::newTileset);
+    connect(ui->openFileButton, &QPushButton::clicked, this, &NoEditorWidget::openFile);
```
bjorn: You can connect to this slot directly without needing `NoEditorWidget::openFile`:
```c++
connect(ui->openFileButton, &QPushButton::clicked, DocumentManager::instance(), &DocumentManager::openFile);
```

RhenaudTheLukark: I knew about that one, but I wanted to keep the style of it. I'll follow that way then.

bjorn:
Whoops! I didn't notice the failing autobuilds and the change was looking good to me, but actually it didn't compile. Fixed now with change 6fe30ecc55e802dcf2cfdad3ca5912831c95df6a. :-/

RhenaudTheLukark:
Okay, will update again

RhenaudTheLukark:
Actually one of the lines I added doesn't work for some reason, the one you told me to add.

It's weird, I swear it used to work before going to Qt 5.9...

bjorn:
> It's weird, I swear it used to work before moving to Qt 5.9...

No, it can't have worked. I wrote that line on GitHub so I hadn't tried to compile it either, but it couldn't work because there are two overloads of `DocumentManager::openFile`. Hence the cast is required for the connect call to know which one is meant. See the fix I referenced.

RhenaudTheLukark:
I see now, thank you for the enlightenment. It's the first time I see such construction, it seems I still have a lot to learn on C++.

bjorn:
> I see now, thank you for the enlightenment. It's the first time I see such construction, it seems I still have a lot to learn on C++.

It's casting to a certain member function signature (a member function of `DocumentManager` that returns `void` and takes no arguments). Anyway, even after coding C++ for almost 20 years I too need to look up the syntax for this... It's a lot easier with [qOverload](https://doc.qt.io/qt-5/qtglobal.html#qOverload) (if you ignore its template magic), but that's only available in Qt 5.7 and Tiled's minimum Qt version currently sits at Qt 5.6 (the last Qt version that supports Windows XP).

PR #1831, opened 2017-12-07
Fix broken tileset inside template

thabetx:
This pull request aims to enable fixing bad links inside templates (#1732).

Tileset bad reference and tileset bad image reference can be fixed from the templates view independent from the map. I provided a button and a label that shows the bad file link to allow fixing things and update the template and consequently the instances.

fixing

This works without using BrokenLinks at all, but there is still broken links integration which checks if there is any problem with any template (bad tileset link or image) and allows opening the template in the dock where we can carry on with the fixes.

If a tileset document is already opened then and we open it through the templates dock to fix the image, a new tab is opened for the same tileset document, I think the shared tileset pointers from the opened document and the tileset inside the template are not the same, so the document manager will open two instances of the same tileset document which won't be connected internally.

Testing and feedback on the UI and the flow are welcome.

### DIFF
--- a/src/tiled/templatesdock.cpp
+++ b/src/tiled/templatesdock.cpp
@@ -115,10 +115,11 @@ TemplatesDock::TemplatesDock(QWidget *parent):
 
     editingToolBar->addAction(mUndoAction);
     editingToolBar->addAction(mRedoAction);
+    editingToolBar->addSeparator();
     editingToolBar->addAction(mToolManager->registerTool(objectSelectionTool));
     editingToolBar->addAction(mToolManager->registerTool(editPolygonTool));
 
-    mFixTilesetButton = new QPushButton(tr("Fix Tileset"), this);
+    mFixTilesetButton = new QPushButton(QString(), this);
     connect(mFixTilesetButton, &QPushButton::clicked, this, &TemplatesDock::fixTileset);
     mFixTilesetButton->setVisible(false);
 
@@ -187,9 +188,18 @@ TemplatesDock::~TemplatesDock()
     delete mDummyMapDocument;
 }
 
-void TemplatesDock::openTemplate(const ObjectTemplate *objectTemplate)
+void TemplatesDock::openTemplate(const QString &path)
 {
-    setTemplate(TemplateManager::instance()->loadObjectTemplate(objectTemplate->fileName()));
+    bringToFront();
+    setTemplate(TemplateManager::instance()->loadObjectTemplate(path));
+    mTemplatesView->setSelectedTemplate(path);
+}
+
+void TemplatesDock::bringToFront()
+{
+    show();
+    raise();
+    setFocus();
 }
 
 void TemplatesDock::setSelectedTool(AbstractTool *tool)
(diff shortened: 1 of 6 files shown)


src/tiled/templatesdock.cpp
```diff
@@ -239,6 +267,46 @@ void TemplatesDock::setTemplate(ObjectTemplate *objectTemplate)
...
+
+    QString templateName = QFileInfo(mObjectTemplate->fileName()).fileName();
+
+    if (tileset->imageStatus() == LoadingError) {
+        mFixTilesetButton->setVisible(true);
+        mFixTilesetButton->setText(tr("Open Tileset"));
+        mFixTilesetButton->setToolTip(tileset->imageSource().fileName());
+
+        mDescriptionLabel->setVisible(true);
+        mDescriptionLabel->setText(tr("%1 :Couldn't find \"%2\"").arg(templateName,
```
bjorn: Please put a space after the colon instead of before (same below).

src/tiled/templatesdock.cpp
```diff
@@ -300,6 +371,54 @@ void TemplatesDock::retranslateUi()
...
+
+    Cell cell = mObject->cell();
+
+    if (cell.isEmpty())
+        return;
+
+    auto tileset = cell.tileset()->sharedPointer();
+
+    if (tileset->imageStatus() == LoadingError) {
+        // This code opens a new document even if there is a tileset document
```
bjorn: Outdated comment?

src/tiled/templatesdock.cpp
```diff
@@ -300,6 +371,54 @@ void TemplatesDock::retranslateUi()
...
+        auto tilesetDocument = DocumentManager::instance()->findTilesetDocument(tileset);
+        if (!tilesetDocument) {
+            tilesetDocument = new TilesetDocument(tileset, tileset->fileName());
+            DocumentManager::instance()->addDocument(tilesetDocument);
+        } else {
+            DocumentManager::instance()->openTileset(tileset);
+        }
+        connect(tilesetDocument, &TilesetDocument::tilesetChanged,
+                this, &TemplatesDock::applyChanges);
```
bjorn: This connection is a little strange, because the template didn't change when the tileset was changed. Would it work to only connect this to `checkTileset`?

Also, clicking the button multiple times would make multiple connections. I think we can solve that by passing `Qt::UniqueConnection`.

bjorn:
I've pushed a small patch tackling the "broken references to images in tilesets used by templates" case. I've decided to create a `TilesetDocument` on-demand in `DocumentManager::openTileset` (couldn't think of a reason to keep that assert in there). Please rebase this patch.

thabetx:
Hello bjorn, thanks for the great review, I've rebased on master and updated the code.

Now the templates dock is shown if it wasn't opened and the opened template is highlighted in the view. The index is valid even if the template doesn't exist in the view, in that case, setting the selection just removes any highlighting. I'm not sure if checking the index validity is useful, but i chose to check it anyway.

The "duplicate tileset document when the tileset is used by map and template" problem is still there despite using process tilesets, which should prevent duplicates. The problem is that shared pointers of the tilesets are different for some reason.

bjorn:
> The "duplicate tileset document when the tileset is used by map and template" problem is still there despite using process tilesets, which should prevent duplicates. The problem is that shared pointers of the tilesets are different for some reason.

Hmm, I've seen this as well and it is a little worrying since it probably indicates some issue elsewhere. These are external tilesets and they are expected to be loaded only once by the TilesetManager. Will need further investigation.

Since I think that must be a bug somewhere else, I'm accepting this patch now.

Issue #1213, opened 2016-02-27
Export tilesets to Lua


agersant:
At the moment, maps can be exported to Lua files but tilesets can not.

As a consequence, the entirety of the tileset data used by a map is embedded in its Lua export. When reusing a tileset across multiple maps, this wastes a lot of space by having each map contain the same tileset data.

A better solution would be to have the exported map include relative paths to exported Lua tilesets. eg:
`tilesets = {
    {
      firstgid = 1,
      filename = "../tilesets/town.lua",
    },
},`

This suggested behavior would also make the map export to Lua more consistent with the .json and .js equivalents.

bjorn:
Right, at the moment no export-only formats for tilesets are supported, but this would definitely make sense to add. The question is at which point should this file be saved and how should the map realize which file to reference upon export.

agersant:
Thanks for the quick reply! I gave this a lot more thinking and I think I came up with a decent solution which doesn't require drastic code changes.

There are fundamental difference between read-write tileset formats and read-only formats, and also between embedded vs external tilesets. In the current version, these difference are somewhat hidden from the user. This obfuscation creates a lot of confusion (eg. https://github.com/bjorn/tiled/issues/923, https://github.com/bjorn/tiled/issues/846) in addition to the limitation which made me open this ticket. Consequently, I think any good solution should involve not only code but also UI changes to make these differences a bit more apparent and let users make informed decisions.

Here are my proposed changes. I tried to preserve the existing workflow as much as possible while limiting the potential for user confusion. I also tried to be forward-compatible with the following hypothetical future features:
- Editing of tilesets that are not imported in the map (https://github.com/bjorn/tiled/issues/242)
- Project-level management of tilesets

## Adding tilesets

Currently, the "Map" menu has options for "New tileset" and "Add external tileset" but the Tileset panel only has "New tileset". This is very inconsistent and there's two ways to solve this:
- Have the two buttons in both locations ("Map" menu and "Tilesets" panel)
- Have the two buttons in a single location

My recommendation would be to remove the "New tileset" button from the "Tilesets" panel. The reasoning is that all other operations in that list of buttons operate on the current tileset, while adding a tileset really is a map operation.

A great compromise could be to have these options in the "Map" menu and also in the tileset dropdown in the top right of the "Tilesets" panel like this:

![Adding tilesets](http://i.imgur.com/Hf2umZR.png)

## Tileset operations

[...]

PR #1835, opened 2017-12-14
Export of tilesets to Lua (#1213)

radman0x:
These changes were prompted following [my forum post regarding this](http://discourse.mapeditor.org/t/export-tileset-to-lua/2884/2). The closest issue to what I raised is #1213, but the scope of that appears to be far wider that what I would like / am attempting. Simply I just want to be able to work on a tileset in isolation and then export that to a .lua file. The discussion in #1213 deals with a lot of issues which may impact this but are outside the driving reason for these particular changes.

With that in mind the changes I have here go a very small way toward addressing #1213. The changes simply enable the 'Export As' option for tilesets and I've implemented / modified  output to .lua files (the others will just error currently).

These changes are intentionally rough and I'd like to get feedback from @bjorn regarding their suitability from an architecture standpoint etc.

### DIFF
--- a/src/libtiled/mapformat.h
+++ b/src/libtiled/mapformat.h
@@ -97,7 +97,9 @@ class TILEDSHARED_EXPORT ReadableMapFormat : public MapFormat
     Q_INTERFACES(Tiled::MapFormat)
 
 public:
-    using MapFormat::MapFormat;
+    explicit ReadableMapFormat(QObject *parent = nullptr)
+        : MapFormat(parent)
+    {}
     Capabilities capabilities() const override { return Read; }
     bool write(const Map *, const QString &) override { return false; }
 };
@@ -112,7 +114,9 @@ class TILEDSHARED_EXPORT WritableMapFormat : public MapFormat
     Q_INTERFACES(Tiled::MapFormat)
 
 public:
-    using MapFormat::MapFormat;
+    explicit WritableMapFormat(QObject *parent = nullptr)
+        : MapFormat(parent)
+    {}
 
     Capabilities capabilities() const override { return Write; }
     Map *read(const QString &) override { return nullptr; }
--- a/src/libtiled/mapformat.h
+++ b/src/libtiled/mapformat.h
@@ -67,7 +67,7 @@ public:
     /**
      * Reads the map and returns a new Map instance, or 0 if reading failed.
      */
-    virtual Map *readMap(const QString &fileName) = 0;
+    virtual Map *read(const QString &fileName) = 0;
 
     /**
      * Writes the given \a map based on the suggested \a fileName.
@@ -79,7 +79,7 @@ public:
      * @return <code>true</code> on success, <code>false</code> when an error
      *         occurred. The error can be retrieved by errorString().
      */
-    virtual bool writeMap(const Map *map, const QString &fileName) = 0;
+    virtual bool write(const Map *map, const QString &fileName) = 0;
 };
 
 } // namespace Tiled
@@ -99,7 +99,7 @@ class TILEDSHARED_EXPORT ReadableMapFormat : public MapFormat
 public:
     using MapFormat::MapFormat;
     Capabilities capabilities() const override { return Read; }
-    bool writeMap(const Map *, const QString &) override { return false; }
+    bool write(const Map *, const QString &) override { return false; }
 };
 
 
@@ -115,7 +115,7 @@ public:
     using MapFormat::MapFormat;
 
     Capabilities capabilities() const override { return Write; }
-    Map *readMap(const QString &) override { return nullptr; }
+    Map *read(const QString &) override { return nullptr; }
     bool supportsFile(const QString &) const override { return false; }
 };
 
--- a/src/libtiled/mapformat.h
+++ b/src/libtiled/mapformat.h
@@ -28,8 +28,8 @@
 
 #pragma once
 
-#include "pluginmanager.h"
 #include "fileformat.h"
+#include "pluginmanager.h"
 
 #include <QObject>
 #include <QStringList>
--- a/src/libtiled/mapformat.h
+++ b/src/libtiled/mapformat.h
@@ -29,6 +29,7 @@
 #pragma once
 
 #include "pluginmanager.h"
+#include "fileformat.h"
 
 #include <QObject>
 #include <QStringList>
@@ -37,35 +38,21 @@
 namespace Tiled {
 
 class Map;
-class Tileset;
 
-class TILEDSHARED_EXPORT FileFormat : public QObject
+/**
+ * An interface to be implemented for adding support for a map format to Tiled.
+ * It can implement support for either loading or saving to a certain map
+ * format, or both.
+ */
+class TILEDSHARED_EXPORT MapFormat : public FileFormat
 {
     Q_OBJECT
+    Q_INTERFACES(Tiled::FileFormat)
 
 public:
-    enum Capability {
-        NoCapability    = 0x0,
-        Read            = 0x1,
-        Write           = 0x2,
-        ReadWrite       = Read | Write
-    };
-    Q_DECLARE_FLAGS(Capabilities, Capability)
-
-    explicit FileFormat(QObject *parent = nullptr)
-        : QObject(parent)
+    explicit MapFormat(QObject *parent = nullptr)
+        : FileFormat(parent)
     {}
-
-    /**
-     * Returns whether this format has Read and/or Write capabilities.
-     */
-    virtual Capabilities capabilities() const { return ReadWrite; }
-
-    /**
-     * Returns whether this format has all given capabilities.
-     */
-    bool hasCapabilities(Capabilities caps) const
-    { return (capabilities() & caps) == caps; }
 
     /**
      * Returns the absolute paths for the files that will be written by
@@ -76,49 +63,11 @@ public:
     virtual QStringList outputFiles(const Map *, const QString &fileName) const
     { return QStringList(fileName); }
 
-    /**
-     * Returns name filter for files in this map format.
-     */
-    virtual QString nameFilter() const = 0;
-
-    /**
-     * Returns short name for this map format
-     */
-    virtual QString shortName() const = 0;
-
-    /**
-     * Returns whether this map format supports reading the given file.
-     *
-     * Generally would do a file extension check.
-     */
-    virtual bool supportsFile(const QString &fileName) const = 0;
-
-    /**
-     * Returns the error to be shown to the user if an error occured while
-     * trying to read a map.
-     */
-    virtual QString errorString() const = 0;
-};
-
-
-/**
- * An interface to be implemented for adding support for a map format to Tiled.
- * It can implement support for either loading or saving to a certain map
- * format, or both.
- */
-class TILEDSHARED_EXPORT MapFormat : public FileFormat
-{
-    Q_OBJECT
-
-public:
-    explicit MapFormat(QObject *parent = nullptr)
-        : FileFormat(parent)
-    {}
 
     /**
      * Reads the map and returns a new Map instance, or 0 if reading failed.
      */
-    virtual Map *read(const QString &fileName) = 0;
+    virtual Map *readMap(const QString &fileName) = 0;
 
     /**
      * Writes the given \a map based on the suggested \a fileName.
@@ -130,14 +79,12 @@ public:
      * @return <code>true</code> on success, <code>false</code> when an error
      *         occurred. The error can be retrieved by errorString().
      */
-    virtual bool write(const Map *map, const QString &fileName) = 0;
-    virtual bool writeAsTileset(const Tiled::Tileset *tileset, const QString &fileName) { return false; }
+    virtual bool writeMap(const Map *map, const QString &fileName) = 0;
 };
 
 } // namespace Tiled
 
 Q_DECLARE_INTERFACE(Tiled::MapFormat, "org.mapeditor.MapFormat")
-Q_DECLARE_OPERATORS_FOR_FLAGS(Tiled::FileFormat::Capabilities)
 
 namespace Tiled {
 
@@ -150,8 +97,9 @@ class TILEDSHARED_EXPORT ReadableMapFormat : public MapFormat
     Q_INTERFACES(Tiled::MapFormat)
 
 public:
+    using MapFormat::MapFormat;
     Capabilities capabilities() const override { return Read; }
-    bool write(const Map *, const QString &) override { return false; }
+    bool writeMap(const Map *, const QString &) override { return false; }
 };
 
 
@@ -164,51 +112,13 @@ class TILEDSHARED_EXPORT WritableMapFormat : public MapFormat
     Q_INTERFACES(Tiled::MapFormat)
 
 public:
+    using MapFormat::MapFormat;
+
     Capabilities capabilities() const override { return Write; }
-    Map *read(const QString &) override { return nullptr; }
+    Map *readMap(const QString &) override { return nullptr; }
     bool supportsFile(const QString &) const override { return false; }
 };
 
-
-/**
- * Convenience class that can be used when implementing file dialogs.
- */
-template<typename Format>
-class FormatHelper
-{
-public:
-    FormatHelper(FileFormat::Capabilities capabilities,
-                 QString initialFilter = QString())
-        : mFilter(std::move(initialFilter))
-    {
-        PluginManager::each<Format>([this,capabilities](Format *format) {
-            if (format->hasCapabilities(capabilities)) {
-                const QString nameFilter = format->nameFilter();
-
-                if (!mFilter.isEmpty())
-                    mFilter += QLatin1String(";;");
-                mFilter += nameFilter;
-
-                mFormats.append(format);
-                mFormatByNameFilter.insert(nameFilter, format);
-            }
-        });
-    }
-
-    const QString &filter() const
-    { return mFilter; }
-
-    const QList<Format*> &formats() const
-    { return mFormats; }
-
-    Format *formatByNameFilter(const QString &nameFilter) const
-    { return mFormatByNameFilter.value(nameFilter); }
-
-private:
-    QString mFilter;
-    QList<Format*> mFormats;
-    QMap<QString, Format*> mFormatByNameFilter;
-};
 
 /**
  * Attempt to read the given map using any of the map formats added
--- a/src/libtiled/mapformat.h
+++ b/src/libtiled/mapformat.h
@@ -37,6 +37,7 @@
 namespace Tiled {
 
 class Map;
+class Tileset;
 
 class TILEDSHARED_EXPORT FileFormat : public QObject
 {
@@ -130,6 +131,7 @@ public:
      *         occurred. The error can be retrieved by errorString().
      */
     virtual bool write(const Map *map, const QString &fileName) = 0;
+    virtual bool writeAsTileset(const Tiled::Tileset *tileset, const QString &fileName) { return false; }
 };
 
 } // namespace Tiled
--- a/src/libtiled/mapwriter.cpp
+++ b/src/libtiled/mapwriter.cpp
@@ -74,7 +74,7 @@ public:
     void writeMap(const Map *map, QIODevice *device,
                   const QString &path);
 
-    void writeTileset(const Tileset &tileset, QIODevice *device,
+    void writeTileset(const Tileset *tileset, QIODevice *device,
                       const QString &path);
 
     void writeObjectTemplate(const ObjectTemplate *objectTemplate, QIODevice *device,
@@ -88,7 +88,7 @@ public:
 
 private:
     void writeMap(QXmlStreamWriter &w, const Map &map);
-    void writeTileset(QXmlStreamWriter &w, const Tileset &tileset,
+    void writeTileset(QXmlStreamWriter &w, const Tileset *tileset,
                       unsigned firstGid);
     void writeLayers(QXmlStreamWriter &w, const QList<Layer *> &layers);
     void writeTileLayer(QXmlStreamWriter &w, const TileLayer &tileLayer);
@@ -163,7 +163,7 @@ void MapWriterPrivate::writeMap(const Map *map, QIODevice *device,
     writer.writeEndDocument();
 }
 
-void MapWriterPrivate::writeTileset(const Tileset &tileset, QIODevice *device,
+void MapWriterPrivate::writeTileset(const Tileset *tileset, QIODevice *device,
                                     const QString &path)
 {
     mMapDir = QDir(path);
@@ -196,7 +196,7 @@ void MapWriterPrivate::writeObjectTemplate(const ObjectTemplate *objectTemplate,
     if (Tileset *tileset = objectTemplate->object()->cell().tileset()) {
         unsigned firstGid = 1;
         mGidMapper.insert(firstGid, tileset->sharedPointer());
-        writeTileset(writer, *tileset, firstGid);
+        writeTileset(writer, tileset, firstGid);
     }
 
     writeObject(writer, *objectTemplate->object());
@@ -250,7 +250,7 @@ void MapWriterPrivate::writeMap(QXmlStreamWriter &w, const Map &map)
     mGidMapper.clear();
     unsigned firstGid = 1;
     for (const SharedTileset &tileset : map.tilesets()) {
-        writeTileset(w, *tileset, firstGid);
+        writeTileset(w, tileset.data(), firstGid);
         mGidMapper.insert(firstGid, tileset);
         firstGid += tileset->nextTileId();
     }
@@ -293,14 +293,14 @@ static bool includeTile(const Tile *tile)
     return false;
 }
 
-void MapWriterPrivate::writeTileset(QXmlStreamWriter &w, const Tileset &tileset,
+void MapWriterPrivate::writeTileset(QXmlStreamWriter &w, const Tileset *tileset,
                                     unsigned firstGid)
 {
     w.writeStartElement(QLatin1String("tileset"));
     if (firstGid > 0)
         w.writeAttribute(QLatin1String("firstgid"), QString::number(firstGid));
 
-    const QString &fileName = tileset.fileName();
+    const QString &fileName = tileset->fileName();
     if (!fileName.isEmpty()) {
         QString source = fileName;
         if (!mUseAbsolutePaths)
@@ -312,13 +312,13 @@ void MapWriterPrivate::writeTileset(QXmlStreamWriter &w, const Tileset &tileset,
         return;
     }
 
-    w.writeAttribute(QLatin1String("name"), tileset.name());
+    w.writeAttribute(QLatin1String("name"), tileset->name());
     w.writeAttribute(QLatin1String("tilewidth"),
-                     QString::number(tileset.tileWidth()));
+                     QString::number(tileset->tileWidth()));
     w.writeAttribute(QLatin1String("tileheight"),
-                     QString::number(tileset.tileHeight()));
-    const int tileSpacing = tileset.tileSpacing();
-    const int margin = tileset.margin();
+                     QString::number(tileset->tileHeight()));
+    const int tileSpacing = tileset->tileSpacing();
+    const int margin = tileset->margin();
     if (tileSpacing != 0)
         w.writeAttribute(QLatin1String("spacing"),
                          QString::number(tileSpacing));
@@ -326,16 +326,16 @@ void MapWriterPrivate::writeTileset(QXmlStreamWriter &w, const Tileset &tileset,
         w.writeAttribute(QLatin1String("margin"), QString::number(margin));
 
     w.writeAttribute(QLatin1String("tilecount"),
-                     QString::number(tileset.tileCount()));
+                     QString::number(tileset->tileCount()));
     w.writeAttribute(QLatin1String("columns"),
-                     QString::number(tileset.columnCount()));
+                     QString::number(tileset->columnCount()));
 
-    if (tileset.backgroundColor().isValid()) {
+    if (tileset->backgroundColor().isValid()) {
         w.writeAttribute(QLatin1String("backgroundcolor"),
-                         colorToString(tileset.backgroundColor()));
+                         colorToString(tileset->backgroundColor()));
     }
 
-    const QPoint offset = tileset.tileOffset();
+    const QPoint offset = tileset->tileOffset();
     if (!offset.isNull()) {
         w.writeStartElement(QLatin1String("tileoffset"));
         w.writeAttribute(QLatin1String("x"), QString::number(offset.x()));
@@ -343,19 +343,19 @@ void MapWriterPrivate::writeTileset(QXmlStreamWriter &w, const Tileset &tileset,
         w.writeEndElement();
     }
 
-    if (tileset.orientation() != Tileset::Orthogonal || tileset.gridSize() != tileset.tileSize()) {
+    if (tileset->orientation() != Tileset::Orthogonal || tileset->gridSize() != tileset->tileSize()) {
         w.writeStartElement(QLatin1String("grid"));
-        w.writeAttribute(QLatin1String("orientation"), Tileset::orientationToString(tileset.orientation()));
-        w.writeAttribute(QLatin1String("width"), QString::number(tileset.gridSize().width()));
-        w.writeAttribute(QLatin1String("height"), QString::number(tileset.gridSize().height()));
+        w.writeAttribute(QLatin1String("orientation"), Tileset::orientationToString(tileset->orientation()));
+        w.writeAttribute(QLatin1String("width"), QString::number(tileset->gridSize().width()));
+        w.writeAttribute(QLatin1String("height"), QString::number(tileset->gridSize().height()));
         w.writeEndElement();
     }
 
     // Write the tileset properties
-    writeProperties(w, tileset.properties());
+    writeProperties(w, tileset->properties());
 
     // Write the image element
-    const QUrl &imageSource = tileset.imageSource();
+    const QUrl &imageSource = tileset->imageSource();
     if (!imageSource.isEmpty()) {
         w.writeStartElement(QLatin1String("image"));
         QString source;
@@ -365,25 +365,25 @@ void MapWriterPrivate::writeTileset(QXmlStreamWriter &w, const Tileset &tileset,
             source = toFileReference(imageSource, mMapDir);
         w.writeAttribute(QLatin1String("source"), source);
 
-        const QColor transColor = tileset.transparentColor();
+        const QColor transColor = tileset->transparentColor();
         if (transColor.isValid())
             w.writeAttribute(QLatin1String("trans"), transColor.name().mid(1));
 
-        if (tileset.imageWidth() > 0)
+        if (tileset->imageWidth() > 0)
             w.writeAttribute(QLatin1String("width"),
-                             QString::number(tileset.imageWidth()));
-        if (tileset.imageHeight() > 0)
+                             QString::number(tileset->imageWidth()));
+        if (tileset->imageHeight() > 0)
             w.writeAttribute(QLatin1String("height"),
-                             QString::number(tileset.imageHeight()));
+                             QString::number(tileset->imageHeight()));
 
         w.writeEndElement();
     }
 
     // Write the terrain types
-    if (tileset.terrainCount() > 0) {
+    if (tileset->terrainCount() > 0) {
         w.writeStartElement(QLatin1String("terraintypes"));
-        for (int i = 0; i < tileset.terrainCount(); ++i) {
-            const Terrain *t = tileset.terrain(i);
+        for (int i = 0; i < tileset->terrainCount(); ++i) {
+            const Terrain *t = tileset->terrain(i);
             w.writeStartElement(QLatin1String("terrain"));
 
             w.writeAttribute(QLatin1String("name"), t->name());
@@ -397,7 +397,7 @@ void MapWriterPrivate::writeTileset(QXmlStreamWriter &w, const Tileset &tileset,
     }
 
     // Write the properties for those tiles that have them
-    for (const Tile *tile : tileset.tiles()) {
+    for (const Tile *tile : tileset->tiles()) {
         if (imageSource.isEmpty() || includeTile(tile)) {
             w.writeStartElement(QLatin1String("tile"));
             w.writeAttribute(QLatin1String("id"), QString::number(tile->id()));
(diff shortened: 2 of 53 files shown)


src/plugins/lua/luaplugin.cpp
```diff
@@ -199,14 +228,20 @@ static bool includeTile(const Tile *tile)
 }
 
 void LuaPlugin::writeTileset(LuaTableWriter &writer, const Tileset *tileset,
-                             unsigned firstGid)
+                             unsigned firstGid, bool standalone)
```
radman0x: I've just realised that I've got the semantics of this bool backwards. Please disregard as this will certainly be rectified if these changes move forward.

src/tiled/document.h
```diff
@@ -130,6 +138,9 @@ class Document : public QObject
     Object *mCurrentObject;             /**< Current properties object. */
 
     bool mIgnoreBrokenLinks;
+
+    QString mLastExportFileName;
+    MapFormat* mExportFormat;
```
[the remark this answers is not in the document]
radman0x: This was one of the hacks that I made to just get it working. I copy pasted directly from ``MapDocument`` to ``Document`` which is why it's a ``MapFormat`` type. I'll look at making it a ``FileFormat`` as suggested to make it fit with the major design change.

I did have an issue here that you might be able to assist with. It looks like ``QPointer`` doesn't work with forward declared types, which is why I made it a raw pointer as a hack. Pretty sure there is a circular dependency somewhere between ``mapformat.h`` and here so including it wasn't an option. Any ideas on this front?

radman0x:
@bjorn, as far as the style changes go I'm happy to fix all of them up. Mainly they're the product of my IDE being set up contrary, 2 space indent, *& attached to type rather than var name etc. My next round of changes will be less hack and I'll try to adhere everywhere, if I miss any please point it out as my eyes are ingrained to looking for different formatting.

Putting this here so as not to make the same response on each issue raised.

src/tiled/document.h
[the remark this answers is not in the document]
bjorn: Regarding the moving from `MapDocument` to `Document`, that makes total sense so it's not much of a hack, as long as you do change the type to the more generic one. :-)

I think including `mapformat.h` in `document.cpp` should be enough to fix your compile issues.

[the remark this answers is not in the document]
radman0x: I'll give that a go, that might just do it :)

src/plugins/lua/luaplugin.h
```diff
@@ -79,9 +70,66 @@ class LUASHARED_EXPORT LuaPlugin : public Tiled::WritableMapFormat
...
+    Q_OBJECT
+    Q_INTERFACES(Tiled::Plugin)
+    Q_PLUGIN_METADATA(IID "org.mapeditor.Plugin" FILE "plugin.json")
+
+public:
+    void initialize() override;
+};
+
+
+class LUASHARED_EXPORT LuaMapFormat : public Tiled::WritableMapFormat, public LuaUtilWriter
```
bjorn: Please instantiate the `LuaUtilWriter` on the stack instead of inheriting from it. At least currently I don't see the reason for inheritance. Then you also avoid naming conflicts to you can drop the "Lua" from "writeMapLua" and "writeTilesetLua".

src/plugins/lua/luaplugin.h
```diff
@@ -45,23 +47,12 @@ class LuaTableWriter;
 /**
  * This plugin allows exporting maps as Lua files.
  */
-class LUASHARED_EXPORT LuaPlugin : public Tiled::WritableMapFormat
+class LuaUtilWriter
```
bjorn: When allocated on the stack (see other comment), this class isn't used anymore outside of `luaplugin.cpp`, so it can be moved from the header to the source file.

src/tiled/mapdocument.h
```diff
@@ -319,35 +315,22 @@ public slots:
...
     QPointer<MapFormat> mWriterFormat;
-    QPointer<MapFormat> mExportFormat;
     Map *mMap;
     LayerModel *mLayerModel;
     QRegion mSelectedArea;
     QList<MapObject*> mSelectedObjects;
     MapRenderer *mRenderer;
     Layer* mCurrentLayer;
     MapObjectModel *mMapObjectModel;
+    QPointer<MapFormat> mExportFormat;
```
bjorn: Why move this pointer to the end?

src/tiled/document.cpp
```diff
@@ -112,5 +114,11 @@ void Document::setIgnoreBrokenLinks(bool ignoreBrokenLinks)
     emit ignoreBrokenLinksChanged(ignoreBrokenLinks);
 }
 
+//FileFormat* Document::exportFormat() const
+//{
+//  return mExportFormat;
+//}
```
bjorn: Please remove commented code.

src/tiled/mainwindow.cpp
```diff
@@ -791,13 +883,13 @@ void MainWindow::export_()
     QString exportFileName = mapDocument->lastExportFileName();
 
     if (!exportFileName.isEmpty()) {
-        MapFormat *exportFormat = mapDocument->exportFormat();
+        MapFormat *exportFormat = qobject_cast<MapFormat*>(mapDocument->exportFormat());
```
bjorn: Cast not needed because `MapDocument::exportFormat` returns `MapFormat*`.

src/tiled/mapdocument.cpp
```diff
@@ -218,6 +208,17 @@ QString MapDocument::displayName() const
...
+MapFormat *MapDocument::exportFormat() const
+{
+    return mExportFormat;
+}
+
+void MapDocument::setExportFormat(FileFormat *format)
+{
+    mExportFormat = qobject_cast<MapFormat*>(format);
+    Q_ASSERT(mExportFormat);
+}
```
bjorn: Please don't needlessly move these functions around (move them back up after `setWriterFormat`.

src/tiled/propertybrowser.cpp
```diff
@@ -753,9 +754,9 @@ void PropertyBrowser::addTilesetProperties()
         auto property = addProperty(FileNameProperty, filePathTypeId(), tr("Filename"), groupProperty);
 
         QString filter = QCoreApplication::translate("MainWindow", "All Files (*)");
-        FormatHelper<TilesetFormat> helper(FileFormat::Read, filter);
+//        FormatHelper<TilesetFormat> helper(FileFormat::Read, filter);
 
-        property->setAttribute(QStringLiteral("filter"), helper.filter());
+//        property->setAttribute(QStringLiteral("filter"), helper.filter());
```
bjorn: What's happening here?

src/tiled/tilesetdock.cpp
```diff
@@ -28,6 +28,7 @@
 #include "containerhelpers.h"
 #include "documentmanager.h"
 #include "erasetiles.h"
+#include "fileformat.h"
```
bjorn: Already includes `tmxmapformat.h` and `tilesetformat.h`, so this is superfluous.

src/tiled/tilesetdocument.cpp
```diff
@@ -20,6 +20,7 @@
 
 #include "tilesetdocument.h"
 
+#include "fileformat.h"
```
bjorn: Already includes `tilesetformat.h`, so this is superfluous.

src/tiled/tilesetdocument.h
```diff
@@ -64,6 +63,8 @@ class TilesetDocument : public Document
     void setWriterFormat(TilesetFormat *format);
 
     QString displayName() const override;
+    TilesetFormat* exportFormat() const override;
+    void setExportFormat(FileFormat *format) override;
```
bjorn: Please move declarations up to avoid having unrelated function `displayName` in between.

src/plugins/lua/luaplugin.cpp
```diff
@@ -81,38 +107,72 @@ bool LuaPlugin::write(const Map *map, const QString &fileName)
...
-    if (color.alpha() != 255)
-        writer.writeValue(color.alpha());
-    writer.writeEndTable();
-    writer.setSuppressNewlines(false);
+    return true;
+}
+
+bool LuaTilesetFormat::writeTileset(const Tileset &tileset, const QString &fileName)
+{
+  SaveFile file(fileName);
```
bjorn: Coding style: please correct the indentation.

src/plugins/lua/luaplugin.h
```diff
@@ -79,9 +70,66 @@ class LUASHARED_EXPORT LuaPlugin : public Tiled::WritableMapFormat
...
+protected:
+    QString mError;
+
+};
+
+
+class LUASHARED_EXPORT LuaTilesetFormat : public Tiled::WritableTilesetFormat, public LuaUtilWriter
+{
+    Q_OBJECT
+//    Q_INTERFACES(Tiled::WritableTilesetFormat)
```
bjorn: You can remove this. There is no such interface declared, and it isn't useful on top of the `Tiled::TilesetFormat` interface.

src/plugins/lua/luaplugin.h
```diff
@@ -79,9 +70,66 @@ class LUASHARED_EXPORT LuaPlugin : public Tiled::WritableMapFormat
...
+
+public:
+    void initialize() override;
+};
+
+
+class LUASHARED_EXPORT LuaMapFormat : public Tiled::WritableMapFormat, public LuaUtilWriter
+{
+    Q_OBJECT
+//    Q_INTERFACES(Tiled::MapFormat)
```
bjorn: You can remove this because `WritableMapFormat` already specified this interface.

src/tiled/mainwindow.cpp
```diff
@@ -99,6 +102,95 @@ using namespace Tiled::Internal;
 using namespace Tiled::Utils;
 
 
+namespace {
+
+template <typename Format>
+struct ExportDetails
+{
+   Format* mFormat;
```
bjorn: Coding style: `*` should hug the variable name.

It's easy to remember if you realize this is how C++ associates the pointer: `int a = 3, *b = &a;`

src/tiled/mainwindow.cpp
```diff
@@ -99,6 +102,95 @@ using namespace Tiled::Internal;
...
+   bool valid() { return static_cast<bool>(mFormat); }
+};
+
+template <typename Format>
+ExportDetails<Format> chooseExportDetails(const QString &fileName,
+                                          const QString &lastExportName,
+                                          const QString &lastExportFilter,
+                                          QWidget* window)
+{
+    FormatHelper<Format> helper(FileFormat::Write, window->tr("All Files (*)"));
```
bjorn: You can't use `window->tr`, since that way `lupdate` is not able to determine the context for the translatable string. Fortunately the function is static, so you can write `MainWindow::tr` instead.

src/tiled/mainwindow.cpp
```diff
@@ -90,6 +91,8 @@
 #include <QUndoStack>
 #include <QUndoView>
 
+#include <tuple>
```
bjorn: Not actually used, right?

src/tiled/mainwindow.cpp
```diff
@@ -813,115 +905,11 @@ void MainWindow::export_()
...
-    mSettings.setValue(QLatin1String("lastUsedExportFilter"), selectedFilter);
-
-    if (!chosenFormat->write(mapDocument->map(), fileName)) {
-        QMessageBox::critical(this, tr("Error Exporting Map"),
-                              chosenFormat->errorString());
-    } else {
-        // Remember export parameters, so subsequent exports can be done faster
-        mapDocument->setLastExportFileName(fileName);
-        mapDocument->setExportFormat(chosenFormat);
+    else if (auto tilesetDocument = qobject_cast<TilesetDocument*>(mDocument)) {
```
bjorn: Coding style: `else` on the same line as the `}`.

src/tiled/mainwindow.cpp
```diff
@@ -1509,6 +1497,99 @@ void MainWindow::retranslateUi()
...
+                                                        selectedFilter,
+                                                        this);
+    if (!exportDetails.valid()) {
+        return;
+    }
+
+    // Check if writer will overwrite existing files here because some writers
+    // could save to multiple files at the same time. For example CSV saves
+    // each layer into a separate file.
+    QStringList outputFiles = exportDetails.mFormat->outputFiles(mapDocument->map(), exportDetails.mFileName);
```
bjorn: This `outputFiles` call and related overwrite check should technically also by applied when exporting tilesets, even though right now we know that there are no tileset formats that export multiple files. The code could just be moved into `chooseExportDetails`, right?

src/tiled/mainwindow.cpp
```diff
@@ -99,6 +102,95 @@ using namespace Tiled::Internal;
...
+   Format* mFormat;
+   QString mFileName;
+
+   ExportDetails() {}
+   ExportDetails(Format* format, const QString& fileName)
+     : mFormat(format)
+     , mFileName(fileName)
+   {}
+
+   bool valid() { return static_cast<bool>(mFormat); }
```
bjorn: Naming style: please call this `isValid`. For implementation I prefer `return mFormat != nullptr;`

src/tiled/mainwindow.cpp
```diff
@@ -99,6 +102,95 @@ using namespace Tiled::Internal;
...
 
+namespace {
+
+template <typename Format>
+struct ExportDetails
+{
+   Format* mFormat;
+   QString mFileName;
+
+   ExportDetails() {}
```
bjorn: Should initialize `mFormat` to `nullptr`.

src/tiled/mainwindow.cpp
```diff
@@ -1509,6 +1497,99 @@ void MainWindow::retranslateUi()
...
+    QString fileName = tilesetDocument->fileName();
+    QString selectedFilter =
+            mSettings.value(QLatin1String("lastUsedExportFilter")).toString();
+    auto exportDetails = chooseExportDetails<TilesetFormat>(fileName,
+                                                    tilesetDocument->lastExportFileName(),
+                                                    selectedFilter,
+                                                    this);
+    if (!exportDetails.valid()) {
+        return;
+    }
```
bjorn: Coding style: for single-line bodies, leave out the brackets.

radman0x:
Thanks for the detailed feedback @bjorn, I'll hopefully be able to integrate all of that and get a final changeset done today.

src/libtiled/mapformat.h
```diff
@@ -75,49 +63,11 @@ class TILEDSHARED_EXPORT FileFormat : public QObject
...
-public:
-    explicit MapFormat(QObject *parent = nullptr)
-        : FileFormat(parent)
-    {}
 
     /**
      * Reads the map and returns a new Map instance, or 0 if reading failed.
      */
-    virtual Map *read(const QString &fileName) = 0;
+    virtual Map *readMap(const QString &fileName) = 0;
```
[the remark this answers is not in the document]
radman0x: Done

src/libtiled/mapwriter.cpp
```diff
@@ -74,7 +74,7 @@ class MapWriterPrivate
     void writeMap(const Map *map, QIODevice *device,
                   const QString &path);
 
-    void writeTileset(const Tileset &tileset, QIODevice *device,
+    void writeTileset(const Tileset *tileset, QIODevice *device,
```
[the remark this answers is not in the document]
radman0x: Done.

The reason I changed this was to make it consistent with ``writeMap`` just above. It was an unnecessary change, just something I did while I was thinking about something else.

src/libtiled/mapwriter.h
```diff
@@ -83,7 +83,7 @@ class TILEDSHARED_EXPORT MapWriter
      * Error checking will need to be done on the \a device after calling this
      * function.
      */
-    void writeTileset(const Tileset &tileset, QIODevice *device,
+    void writeTileset(const Tileset *tileset, QIODevice *device,
```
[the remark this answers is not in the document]
radman0x: Done

src/libtiled/tilesetformat.h
```diff
@@ -79,5 +102,3 @@ TILEDSHARED_EXPORT SharedTileset readTileset(const QString &fileName,
 TILEDSHARED_EXPORT TilesetFormat *findSupportingTilesetFormat(const QString &fileName);
 
 } // namespace Tiled
-
-Q_DECLARE_INTERFACE(Tiled::TilesetFormat, "org.mapeditor.TilesetFormat")
```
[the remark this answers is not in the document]
radman0x: Looks like I put it on line 71 above as ``WritableTilesetFormat`` is a new addition and I was keeping it in the same spot relative to the ``TilesetFormat`` class. I can move this down if desired?

src/plugins/lua/luaplugin.cpp
radman0x: Done

src/plugins/lua/luaplugin.h
radman0x: Done

src/plugins/lua/luaplugin.h
radman0x: Done

src/plugins/lua/luaplugin.h
radman0x: Done

src/plugins/lua/luaplugin.h
radman0x: Done

Inheritance was just a quick way of sharing the implementation across the two classes. It felt natural given that the methods were all pulled directly from the pevious objects private implementation. Composition works just as well though so I've done as requested.

src/tiled/document.cpp
radman0x: Done

src/tiled/mainwindow.cpp
radman0x: Done

Good spotting, I was going to use a tuple before I created the ``ExportDetails`` struct, it caused issues and I forgot to remove the include.

src/tiled/mainwindow.cpp
radman0x: Done

I've been writing C++ with the opposite left adjusted style e.g. ``int* number`` for many years so it's just a matter of habit. Happy to conform to your code style so thanks for picking it up :)

src/tiled/mainwindow.cpp
radman0x: Done

src/tiled/mainwindow.cpp
radman0x: Done

src/tiled/mainwindow.cpp
radman0x: Done

Thanks for the tip on that, I was just trying to keep the same calls.

src/tiled/mainwindow.cpp
radman0x: Done

src/tiled/mainwindow.cpp
radman0x: Done

src/tiled/mainwindow.cpp
radman0x: This section relies on ``MapFormat::outputFiles`` and this isn't currently a part of ``TilesetFormat`` which is why I separated it from ``chooseExportDetails``, if it was then it could easily be included.

What would you like done here?

src/tiled/mainwindow.cpp
radman0x: Done

src/tiled/mapdocument.cpp
radman0x: Done

src/tiled/mapdocument.h
radman0x: Fixed

Not sure how I moved it down there... I suspect I must have removed it and then added it as a new variable as I always add new variables at the end.

src/tiled/propertybrowser.cpp
radman0x: Fixed

I was having compile issues so I suspect I commented this and hadn't put it back yet.

src/tiled/tilesetdock.cpp
radman0x: Done

I was having forward declare compile issue so I had tried to include directly to mitigate. Wasn't necessary and I hadn't reverted, good catch.

src/tiled/tilesetdocument.cpp
radman0x: Done

src/tiled/tilesetdocument.h
radman0x: Done

src/plugins/lua/luaplugin.cpp
```diff
@@ -44,15 +44,41 @@
...
-LuaPlugin::LuaPlugin()
+void LuaPlugin::initialize()
+{
+    addObject(new LuaMapFormat(this));
+    addObject(new LuaTilesetFormat(this));
+}
+
+bool LuaMapFormat::supportsFile(const QString &fileName) const
 {
+    return true;
```
radman0x: @bjorn, can you please provide some direction on how implement this. I could copy as best I can from other places but it'd be good to not waste my time doing the wrong thing if possible ;)

radman0x:
Phew! Finally got those cross compile build errors under control :O Changes should now be tidied and I've integrated all comments, barring those I've flagged with a question. It looks like exporting to json also works now for free and I think that's true for any of the types that tilesets are supported for.

Somewhat related question; Why is one of the Windows compilers MSVC2013 and not something more recent? Also why the need for multiple different compilers: Mingw, MSVC, clang, gcc etc? (I'm sure there's a good reason)

src/tiled/mainwindow.cpp
bjorn: Sorry, I didn't realize `outputFiles` was only part of the `MapFormat`. I see in the current code it was part of `FileFormat`, but that didn't make much sense since it receives a map as parameter. So it's good that you moved it to `MapFormat` and then you can disregard this comment. :-)

src/tiled/tilesetdocument.h
```diff
@@ -63,6 +62,9 @@ class TilesetDocument : public Document
     FileFormat *writerFormat() const override;
     void setWriterFormat(TilesetFormat *format);
 
+    TilesetFormat* exportFormat() const override;
```
bjorn: Coding style: please put `*` against the function name.

src/tiled/tilesetdocument.h
```diff
@@ -22,13 +22,12 @@
 
 #include "document.h"
 #include "tileset.h"
+#include "tilesetformat.h"
```
bjorn: I believe you shouldn't need this include and can keep relying on the forward declaration.

src/tiled/mapdocument.h
```diff
@@ -93,17 +93,14 @@ class MapDocument : public Document
...
-
     MapFormat *readerFormat() const;
     void setReaderFormat(MapFormat *format);
 
     FileFormat *writerFormat() const override;
     void setWriterFormat(MapFormat *format);
 
-    MapFormat *exportFormat() const;
-    void setExportFormat(MapFormat *format);
+    MapFormat* exportFormat() const override;
```
bjorn: Coding style: please put `*` against the function name.

src/tiled/mapdocument.h
```diff
@@ -24,6 +24,7 @@
 
 #include "document.h"
 #include "layer.h"
+#include "mapformat.h"
```
bjorn: I believe you shouldn't need this include and can keep relying on the forward declaration.

src/tiled/mainwindow.cpp
```diff
@@ -1509,6 +1493,98 @@ void MainWindow::retranslateUi()
...
+
+    Preferences *pref = Preferences::instance();
+
+    pref->setLastPath(Preferences::ExportedFile, QFileInfo(exportDetails.mFileName).path());
+    mSettings.setValue(QLatin1String("lastUsedExportFilter"), selectedFilter);
+
+    auto exportResult = exportDetails.mFormat->write(mapDocument->map(), exportDetails.mFileName);
+    if (!exportResult)
+    {
```
bjorn: Coding style: please put `{` on the same line as the `if`

src/tiled/mainwindow.cpp
```diff
@@ -1509,6 +1493,98 @@ void MainWindow::retranslateUi()
...
+        return;
+
+    Preferences *pref = Preferences::instance();
+
+    pref->setLastPath(Preferences::ExportedFile, QFileInfo(exportDetails.mFileName).path());
+    mSettings.setValue(QLatin1String("lastUsedExportFilter"), selectedFilter);
+
+    auto exportResult = exportDetails.mFormat->write(*tilesetDocument->tileset(), exportDetails.mFileName);
+    if ( ! exportResult )
+    {
```
bjorn: Coding style: please put `{` on the same line as the `if`, don't put spaces on the inside of parentheses and no space after the `!`.

src/tiled/mainwindow.cpp
```diff
@@ -99,6 +99,95 @@ using namespace Tiled::Internal;
...
+    Format *mFormat = nullptr;
+    QString mFileName;
+
+    ExportDetails() {}
+    ExportDetails(Format *format, const QString& fileName)
+        : mFormat(format)
+        , mFileName(fileName)
+    {}
+
+    bool isValid() { return mFormat != nullptr; }
```
bjorn: Method can be marked `const`

src/tiled/document.cpp
```diff
@@ -112,5 +113,6 @@ void Document::setIgnoreBrokenLinks(bool ignoreBrokenLinks)
     emit ignoreBrokenLinksChanged(ignoreBrokenLinks);
 }
 
+
```
bjorn: No need for adding a newline here.

src/tiled/document.cpp
```diff
@@ -20,6 +20,7 @@
 
 #include "document.h"
 
+#include "fileformat.h"
```
bjorn: Include not needed.

src/tiled/brokenlinks.cpp
```diff
@@ -22,6 +22,7 @@
 
 #include "changetileimagesource.h"
 #include "documentmanager.h"
+#include "fileformat.h"
```
bjorn: Include not needed.

src/tiled/abstractobjecttool.cpp
```diff
@@ -23,6 +23,7 @@
 #include "addremovetileset.h"
 #include "changemapobject.h"
 #include "documentmanager.h"
+#include "fileformat.h"
```
bjorn: Include not needed.

src/plugins/lua/luaplugin.h
```diff
@@ -45,23 +47,12 @@ class LuaTableWriter;
...
-    bool write(const Tiled::Map *map, const QString &fileName) override;
-    QString nameFilter() const override;
-    QString shortName() const override;
-    QString errorString() const override;
-
-private:
     void writeMap(LuaTableWriter &, const Tiled::Map *);
     void writeProperties(LuaTableWriter &, const Tiled::Properties &);
-    void writeTileset(LuaTableWriter &, const Tiled::Tileset *, unsigned firstGid);
+    void writeTileset(LuaTableWriter &, const Tiled::Tileset &, unsigned firstGid, bool standalone=true);
```
bjorn: Coding style: please put spaces around the `=`.

src/plugins/lua/luaplugin.cpp
```diff
@@ -697,4 +763,11 @@ void LuaPlugin::writeTextProperties(LuaTableWriter &writer, const MapObject *map
     }
 }
 
+void LuaUtilWriter::mapDir(const QString &dir)
```
bjorn: Naming style: should be called `setMapDir`.

src/plugins/lua/luaplugin.cpp
```diff
@@ -198,61 +258,67 @@ static bool includeTile(const Tile *tile)
...
 }
 
-void LuaPlugin::writeTileset(LuaTableWriter &writer, const Tileset *tileset,
-                             unsigned firstGid)
+void LuaUtilWriter::writeTileset(LuaTableWriter &writer, const Tileset &tileset,
+                                    unsigned firstGid, bool standalone)
 {
-    writer.writeStartTable();
+    if (standalone) {
+      writer.writeStartTable();
```
bjorn: Coding style: please correct the indentation (also below).

src/plugins/lua/luaplugin.cpp
```diff
@@ -44,15 +44,41 @@
...
+        writer.writeStartTable(name);
+        writer.setSuppressNewlines(true);
+        writer.writeValue(color.red());
+        writer.writeValue(color.green());
+        writer.writeValue(color.blue());
+        if (color.alpha() != 255)
+            writer.writeValue(color.alpha());
+        writer.writeEndTable();
+        writer.setSuppressNewlines(false);
+    }
```
bjorn: I see no reason for moving this function. Please just keep it where it was as `static`.

src/libtiled/tilesetformat.h
```diff
@@ -56,14 +57,38 @@ class TILEDSHARED_EXPORT TilesetFormat : public FileFormat
...
     virtual bool write(const Tileset &tileset, const QString &fileName) = 0;
+
+private:
+    QPointer<TilesetFormat> mExportFormat;
+};
+
+} // namespace Tiled
+
+Q_DECLARE_INTERFACE(Tiled::TilesetFormat, "org.mapeditor.TilesetFormat")
+Q_DECLARE_OPERATORS_FOR_FLAGS(Tiled::FileFormat::Capabilities)
```
bjorn: Please move this operators macro to `fileformat.h`.

src/plugins/lua/luaplugin.h
```diff
@@ -79,9 +70,72 @@ class LUASHARED_EXPORT LuaPlugin : public Tiled::WritableMapFormat
...
+    bool supportsFile(const QString &fileName) const override;
+
+    bool write(const Tiled::Map *map, const QString &fileName) override;
+
+    QString nameFilter() const override;
+    QString shortName() const override;
+    QString errorString() const override;
+
+protected:
+    LuaUtilWriter mLuaWriter;
```
bjorn: I suggested to allocate the `LuaUtilWriter` on the stack. Was there a reason it should be a member instead?

src/plugins/lua/luaplugin.h
bjorn: I would rename this class to just `LuaWriter`.

src/plugins/lua/luaplugin.cpp
```diff
@@ -198,61 +258,67 @@ static bool includeTile(const Tile *tile)
...
+    } else {
+      writer.writeStartReturnTable();
+    }
 
-    writer.writeKeyAndValue("name", tileset->name());
-    writer.writeKeyAndValue("firstgid", firstGid);
+    writer.writeKeyAndValue("name", tileset.name());
+    if (standalone) {
+      writer.writeKeyAndValue("firstgid", firstGid);
+    }
```
bjorn: Coding style: in case of single-line body, please leave out the brackets (also above).

radman0x:
Hmmm I took a look at the [build log](https://ci.appveyor.com/project/bjorn/tiled/build/1.0.2898/job/i5bqnu6df332vset) you linked and it looks like your container / build node just doesn't have the newer version of MSVC installed:

> C:\projects\tiled\dist\win\installer.wxs(96) : error LGHT0103 : The system cannot find the file 'C:/Program Files (x86)/Microsoft Visual Studio 14.0/VC/bin/amd64/../..\redist\x64\Microsoft.VC120.CRT\msvcp120.dll'.

Not sure how you would get it installed but it looks like everything compiled fine, just it couldn't link...

Actually I think it's trying to link an older version of the dll ``msvcp120.dll`` where it should be getting msvcp140.dll or something similar. Chances are this is hardcoded in the Wix file. There's also a mixture of / and \ in the path, but that's "probably" not the problem.

I do have some experience with Wix so if I find myself on Windows I might try to build it and see if I can work it out quickly.

src/libtiled/tilesetformat.h
radman0x: Done

src/plugins/lua/luaplugin.cpp
radman0x: Done

bjorn:
> Actually I think it's trying to link an older version of the dll msvcp120.dll where it should be getting msvcp140.dll or something similar. Chances are this is hardcoded in the Wix file. There's also a mixture of / and \ in the path, but that's "probably" not the problem.

Indeed it's just trying to include the wrong version of the DLL in the installer. I do not know which DLL is the right one to ship for MSVC 2015. If it's just a change of file name, then it would need to be updated in `dist/win/installer.wxs` and `dist/distribute.qbs`.

src/plugins/lua/luaplugin.cpp
radman0x: Done

Gah! Thought I found all of those before.

src/plugins/lua/luaplugin.cpp
radman0x: Done

src/plugins/lua/luaplugin.cpp
radman0x: Done

src/plugins/lua/luaplugin.h
radman0x: Done

src/plugins/lua/luaplugin.h
radman0x: Done

src/plugins/lua/luaplugin.h
radman0x: It has ``mMapDir`` and is used across calls e.g. ``writeProperties`` which seems to use the member assuming it has been set previously. I could be wrong, please feel free to correct me.

src/tiled/abstractobjecttool.cpp
radman0x: This uses ``FormatHelper`` which is now from ``fileformat.h``. It will compile without it explicitly included though because doubtless something in the chain pulls it in. In this situation I favour an explicit include but I'm happy to defer to you. Please confirm that you definitely want it changed.

src/tiled/brokenlinks.cpp
radman0x: Same as above, please confirm and I'll remove.

src/tiled/document.cpp
radman0x: Done

src/tiled/document.cpp
radman0x: Done

Apologies, slip of the keyboard.

src/tiled/mainwindow.cpp
radman0x: Done

Certainly can, good catch.

src/tiled/mainwindow.cpp
radman0x: Done

src/tiled/mainwindow.cpp
radman0x: Done

src/tiled/mapdocument.h
radman0x: Ahh this one is actually legit. I've modified ``exportFormat`` to have a covariant return type, so that derived classes can return a pointer that is descended from ``FileFormat`` e.g. ``MapFormat`` etc which is a little nicer. Because of this the [full definition of the class needs to exist](https://stackoverflow.com/questions/29817748/invalid-covariant-return-type-error) and [a forward declaration is inadequate](https://stackoverflow.com/questions/2159390/in-c-is-it-possible-to-forward-declare-a-class-as-inheriting-from-another-cla), so that the compiler knows that ``MapFormat`` is derived from ``FileFormat`` and the covariant return is valid.

src/tiled/abstractobjecttool.cpp
bjorn: In that case let's keep it. Generally I consider includes superfluous when a subclass is already included (like `tmxmapformat.h` in this case), but since `FormatHelper` is another hierarchy we can keep it.

src/tiled/mapdocument.h
radman0x: Done

src/tiled/tilesetdocument.h
radman0x: Same reasoning as described above for ``mapdocument.h``

src/tiled/tilesetdocument.h
radman0x: Done

src/tiled/mapdocument.h
bjorn: Good point (and what an annoying top-voted answers in your second SO link...). :-)

radman0x:
The name will be with a ``14`` replacing ``12`` where it is currently. Confusingly [Visual Studio versioning is separate from MSVC++ versioning](https://en.wikipedia.org/wiki/Microsoft_Visual_C%2B%2B). Visual Studio 2017 ships with MSVC++ 14.1 :P 

I'd also say that you can probably comfortably jump to an even newer compiler, e.g. Visual Studio 2017. There are unlikely to be any compilation incompatibilities.

radman0x:
WRT inheriting constructors, It's an addition from C++11 that allows a [using declaration](http://en.cppreference.com/w/cpp/language/using_declaration)(see Inheriting Constructors section) of a constructor that essentially allows the derived class to say, I expose the same constructors as my base. This is convenient in not having to repeat code. MSVC2013 is the only one that had an issue, it really is pretty stone age now as far as the new C++ standards go (C++11, C++14, C++17).

As an aside there are also [delegating constructors](https://www.google.com.au/url?sa=t&rct=j&q=&esrc=s&source=web&cd=2&cad=rja&uact=8&ved=0ahUKEwiWgcO0nJjYAhWHU7wKHWdwBIwQygQINDAB&url=http%3A%2F%2Fen.cppreference.com%2Fw%2Fcpp%2Flanguage%2Finitializer_list%23Delegating_constructor&usg=AOvVaw2jDzlUTNmgVNKx5fNkoJO6) which let you call a constructor of the same class (just like a base class) to help eliminate code duplication if constructors share a lot of the same initialisation logic.

src/tiled/abstractobjecttool.cpp
radman0x: Ok sounds good :+1:

src/tiled/mapdocument.h
radman0x: Haha yeah, at least the top voted comment for that answer pulls it back into line a bit.

bjorn:
> The name will be with a 14 replacing 12 where it is currently. Confusingly Visual Studio versioning is separate from MSVC++ versioning. Visual Studio 2017 ships with MSVC++ 14.1 :P

Ok, let's see what 33637fa54539cbbd354efe8a51935f670f776b91 yields.

> As an aside there are also delegating constructors which let you call a constructor of the same class (just like a base class) to help eliminate code duplication if constructors share a lot of the same initialisation logic.

Yeah, I've been using those a lot already so VS 2013 seems to support those fine.

src/plugins/lua/luaplugin.h
bjorn: I've pushed a change which puts it on the stack, since I wanted to make sure anyway that it would work. Now the `mMapDir` thing works even better I think, since by requiring it in the constructor we make sure that it is set appropriately. Only the static methods could be called without initializing it (though they aren't).

radman0x:
Huzzah! I'm AFK for the rest of the night so it's good to hear!

src/plugins/lua/luaplugin.h
radman0x: Looks great :)

PR #1838, opened 2017-12-21
Update libs.yml

pamtdoh:
Add Phaser

### DIFF
--- a/_data/libs.yml
+++ b/_data/libs.yml
@@ -24,3 +24,11 @@
   status: Inactive since 2012
   features: Reads TMX
   notes: Depends on [RapidXml](http://rapidxml.sourceforge.net/)
+
+- language: JavaScript
+  name: Phaser
+  url: https://github.com/photonstorm/phaser
+  updated: 21-12-2017
+  status: Active
+  Features: Supports Tilemaps and Tile Sprites
+  notes: Can't read TMX, must be exported to csv or json.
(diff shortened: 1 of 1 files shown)


_data/libs.yml
```diff
@@ -24,3 +24,11 @@
...
   features: Reads TMX
   notes: Depends on [RapidXml](http://rapidxml.sourceforge.net/)
+
+- language: JavaScript
+  name: Phaser
+  url: https://github.com/photonstorm/phaser
+  updated: 21-12-2017
+  status: Active
+  Features: Supports Tilemaps and Tile Sprites
+  notes: Can't read TMX, must be exported to csv or json.
```
bjorn: `features:` should be written without capital.

Their tile sprites seem to have nothing to do with tile maps or Tiled, so that isn't worth mentioning. I think the noteworthy parts are that it "Reads tilemaps from CSV or JSON and supports rendering"

In the `notes:` it could be worth mentioning that the library supports runtime modifications (though I expect this to be true for most libraries, so it might make more sense to only make note about when this isn't the case). The idea is to stay brief on this page while more details are found on the wiki. If there's nothing to note the `notes:` can be left out.

pamtdoh:
I've made the changes needed. As for the notes, I really can't think of anything else worth noting.

bjorn:
Thanks!

Will you continue with the next task, to actually use Phaser to make a small demo of your own, or, since it appears to already have demos, maybe see if you can improve something about the Tiled support in Phaser?

PR #1840, opened 2017-12-22
Fix a few bugs in libtiled-java port.

theKidOfArcrania:
I tried to solve a few bugs I was having with this code. There might still be other bugs out there.


bjorn:
Thanks for these fixes! They all look good to me. Would you like to have a look as well, @mikepthomas?

bjorn:
Forgot about this before 1.1.0 / 1.1.1 releases, but at least it will be included in 1.1.2. Thanks!

PR #1846, opened 2018-01-04
Fix return in Tileset::orientationToString

jubalh:
This part should not be reached in theory.
Let's return an empty QString anyways to be on the save side and to make
the compiler happy.

Fixes 'Program returns random data in a function'.


bjorn:
Which compiler is giving you that warning?

I used to have a `default:` label in there, but then I got "warning: default label in switch which covers all enumeration values".

Anyway, maybe returning an empty string is anyway preferable to choosing one of the others as default.

jubalh:
gcc7, but it might have been that another rpm check tool reported the error.

PR #1861, opened 2018-01-18
Update Python Plugin API to support new features

ilanosortap:


bjorn:
Just including the additional headers is enough? I would expect that you'll also need to actually define the classes to expose and which methods to expose.

Also, please include the updated generated file in your commit. `pybindgen` should not be required for building Tiled, which is why the generated file is in the repository.

bjorn:
I've rebased your change so that PyBindGen 0.18.0 can be used to generate the `pythinbind.cpp` file. But note that adding the include files does not suffice. You will need to make additional changes as seen in PR #1867.

Do you still plan to work on this?

ilanosortap:
I ll make the additional changes and create a pull request soon.

Regards,

Sonali Patro

Contact: 00917381809123/0097455579707

bjorn:
@ilanosortap Shall we close this or do you think you'll still continue the work sometime?

bjorn:
Closing this due to lack of updates.

PR #1863, opened 2018-01-23
Tileset offset and Layer offset now considered when drawing with libtiled-java

digitalhoax:
Fix for orthogonal and isometric view.


bjorn:
Would there be any way around all these `null` checks? They make the whole thing very verbose.

Otherwise this change looks fine me. Do you have anything to add, @mikepthomas?

mikepthomas:
I'm sure you can set the default value to zero using the jaxb commons [defaut-value-plugin](https://github.com/javaee/jaxb2-commons/tree/master/default-value-plugin) rather than having to do all the null checking... would need to take a closer look though to be sure :)

bjorn:
Ah, that stuff makes me dizzy. :-(

I'll just accept this patch as-is for now, anyway.

bjorn:
Thanks, @digitalhoax!

Issue #1585, opened 2017-06-01
JSON Tile-Set Format tiles array is needlessly difficult to parse.


JeremyWildsmith:
Basically, the way it seems to be structured, the tiles value is an object with a variable number of fields where the field name represents the id of the tile in an array of tiles. Since the ID starts from 0 and grows up, it would make a lot more sense if this was just an array.

This is really awkward to process, you can't really de-serialize this structure into a class without some awkward hacks with a deserialization framework. It's a minor issue, but its very odd.

I understand tilewidth and type is also in the tiles, but this just makes it more difficult to work with. Why not separate them?

Consider the current format:
```
"tiles":
    {
     "0":
        {
         "image":"necro3_0\/5.png",
         "imageheight":32,
         "imagewidth":32
        },
     "1":
        {
         "image":"necro3_0\/4.png",
         "imageheight":32,
         "imagewidth":32
        }
    }
```

Why not:

```
"tiles":
    [
        {
         "image":"necro3_0\/5.png",
         "imageheight":32,
         "imagewidth":32
        },
        {
         "image":"necro3_0\/4.png",
         "imageheight":32,
         "imagewidth":32
        }
    ]
```

bjorn:
Unfortunately when the format was set up this way, I was not aware of the difficulties this would cause when working with some JSON serializers. Of course, changing it now will break all existing readers, but it may be worth it anyway for Tiled 1.1.

However, tile IDs are not necessarily consecutive. When a tile is removed (when using an image collection tileset like in your quoted examples), the IDs of other tiles do not change since that would break tile references. This means the alternative format using an array would need to include the tile IDs in the tile objects.

JeremyWildsmith:
Hey,

Yeah, I just noticed this with the IDs as well. I think it would definitely be better to have the ID as apart of the rest of the object.

I think if Tiled were to move in this direction for 1.1, it may be a good idea to do a quick pass over the rest of the formats as well to make sure they make sense. It's not just about ease of parsing, they'd also just generally become simpler.

I noticed an oddity in the world JSON format as well where not all objects in an array were of the same type. Consider the layers array, where we can have objectgroup and layergroup and tilegroup where type is specified by type field. It may actually be easier to keep this as it is (and it isn't a big issue to work around), but it is definitely worth considering since any serialization framework I've come across treats every element in an array as the same type. (The work-around is to just make one unionized type and processes the deserialized data per the type field)

[...]

PR #1868, opened 2018-01-29
Jsonformat

saeedakhter:
Addresses issue https://github.com/bjorn/tiled/issues/1585

This introduces a new json revision (version: 1.2) that is much easier to serialize and de-serialize.  For an example of de-serializing the new format in Unity3d, see https://github.com/saeedakhter/TiledJsonUnity3d

**Problem:**
Tiles were serialized as a map with the tile ID as the key:
```
"tiles":
    {
     "0": { "image":"necro3_0\/5.png", "imageheight":32, "imagewidth":32 },
     "1": { "image":"necro3_0\/4.png", "imageheight":32, "imagewidth":32 }
    }
```
In Unity3d, this makes it difficult (impossible?) to use Unity3d's JSON deserialization:
https://docs.unity3d.com/Manual/JSONSerialization.html 

**Solution (this pull request):**
Serialize tiles and properties as a list.  If an ID is required, put it inside the list object.  Example:
```
 "tiles":[
        {
         "terrain":[0, 2, 2, 2],
         "tile":0
        }, 
        {
         "properties":[
                {
                 "name":"Prefrab",
                 "type":"string",
                 "value":"Floor*"
                }, 
                {
                 "name":"SortingLayer",
                 "type":"string",
                 "value":"Floor"
                }],
         "terrain":[0, 0, 0, 0],
         "tile":1
        }, 
```

### DIFF
--- a/docs/reference/json-map-format.rst
+++ b/docs/reference/json-map-format.rst
@@ -29,7 +29,7 @@ Map
 | orientation       | string   | ``orthogonal``, ``isometric``, ``staggered`` or          |
 |                   |          | ``hexagonal``                                            |
 +-------------------+----------+----------------------------------------------------------+
-| properties        | object   | String key-value pairs                                   |
+| properties        | array    | A list of properties (name, value, type).                |
 +-------------------+----------+----------------------------------------------------------+
 | renderorder       | string   | Rendering direction (orthogonal maps only)               |
 +-------------------+----------+----------------------------------------------------------+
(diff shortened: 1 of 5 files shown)


bjorn:
Thanks a lot for your work on this!

bitinn:
Thx for your good work @saeedakhter!

Just wondering @bjorn, does the snapshot build contain this change? I was writing a parser on C# for my specific use-case and realize it would have been much easier with this updated as I just happen to heavily rely on tileset properties.

bjorn:
@bitinn Yes, the changes are included in the [latest snapshot build](https://thorbjorn.itch.io/tiled/devlog/24167/multi-layer-editing-and-other-stuff). The [JSON format documentation](http://docs.mapeditor.org/en/latest/reference/json-map-format/) is also already updated, which means it's now actually time to host explicit "Tiled 1.1" documentation (probably as default) as well to avoid confusion.

bitinn:
@bjorn fantastic! and yeah I was reading the doc and realize `tileproperties` was missing and find my way here :)

PR #1883, opened 2018-02-16
Added option to lock/unlock all other layers

kralle333:
Added option to lock/unlock all other layers except for the currently selected one. Works similar to show/hide all layers function.


src/tiled/layermodel.cpp
```diff
@@ -534,3 +534,36 @@ void LayerModel::toggleOtherLayers(Layer *layer)
...
+    else
+        undoStack->beginMacro(tr("Unlock Other Layers"));
+
+    for (Layer *l : otherLayers) {
+        if (locked != l->isLocked())
+            undoStack->push(new SetLayerLocked(mMapDocument, l, locked));
+    }
+
+    undoStack->endMacro();
+}
```
ketanhwr: Insert a new line at the end.

src/tiled/mapdocumentactionhandler.cpp
```diff
@@ -125,6 +125,11 @@ MapDocumentActionHandler::MapDocumentActionHandler(QObject *parent)
     mActionToggleOtherLayers->setIcon(
             QIcon(QLatin1String(":/images/16x16/show_hide_others.png")));
 
+    mActionToggleLockOtherLayers = new QAction(this);
+    mActionToggleLockOtherLayers->setShortcut(tr("Ctrl+Shift+U"));
```
bjorn: I guess `Ctrl+Shift+U` was chosen because `Ctrl+Shift+L` is already taken by "Save All"? I wonder if it wouldn't be more intuitive to use `Ctrl+Shift+L` here and maybe drop the "Save All" shortcut again (or switch that one to `Ctrl+Alt+S`, which is what Notepad++ did).

kralle333: Yes I had put the shortcut to Ctrl+Shift+L first. I agree with your suggestion.

bjorn: Sorry, we can have only one `Ctrl+Shift+S` shortcut, and now it means both "Save As" and "Save All" (in which case neither will trigger). Please change "Save As" to `Ctrl+Alt+S`.

kralle333: Hi Bjorn, I can not see that Save As has the "Ctrl+Shift+S" shortcut. Have I overlooked it somewhere?

bjorn: Yes, it happens in this line, because `Ctrl+Shift+S` is a common shortcut for "Save As":
```c++
mUi->actionSaveAs->setShortcuts(QKeySequence::SaveAs);
```

bjorn:
Thanks for this useful addition!

PR #1894, opened 2018-02-26
Fix #91 object tile position, to be same as other objects

### LINKED ISSUE #91: TileObject: y-pos wrong
http://imageshack.us/photo/my-images/543/objectbugypos.png/

when i place a tileObject ( object with an Image), the y-pos is to hight.
f.e. when i place a tileobject at (2,13), the pos is (2,14).

justgook:
Maybe You can improve - but that fixes issues when You create tile in objecLayer, so now it will be offset by height in tiled Prieview  same as other objects - so as result parsing tiles-objecsts can be done same as other objects.. #91 

* Btw im total newbie in C++/QT, but that bug just make me mad..

### DIFF
--- a/src/libtiled/orthogonalrenderer.cpp
+++ b/src/libtiled/orthogonalrenderer.cpp
@@ -96,7 +96,7 @@ QRectF OrthogonalRenderer::boundingRect(const MapObject *object) const
 
         const QPointF bottomLeft = bounds.topLeft();
         boundingRect = QRectF(bottomLeft.x() + (tileOffset.x() * scale.width()),
-                              bottomLeft.y() + (tileOffset.y() * scale.height()) - objectSize.height(),
+                              bottomLeft.y() + (tileOffset.y() * scale.height()),
                               objectSize.width(),
                               objectSize.height()).adjusted(-1, -1, 1, 1);
     } else {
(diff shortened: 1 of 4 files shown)


justgook:
looks like I forgot to change scaling origin.. hope some one can help.. or Ill try dot tomorrow..

bjorn:
While this may be helpful for those looking to patch up their version of Tiled to behave differently, it's not something I can merge because it would break everybody's existing maps and frameworks.

Hence, this change will need to be introduced as an option. Since changing the alignment of objects is a rare thing to do, I would be inclined to allow specifying it at the map level. There would need to be at least two settings for this "object origin" setting:

* The current behavior, maybe we call it "classic" or "legacy". It's not any particular alignment, because it differs between object types as well as between map types (orthogonal tile objects use bottom-left, whereas tile objects on isometric maps use bottom-center).
* The generally desired new behavior, probably "top-left".

Eventually there is also reason to support this as a per-instance option, though that should more likely take the form of a pixel offset, as done in #1323.

justgook:
maybe there is point to close it - as i have seen some other ideas abut it, and my solution without configs is kind of pointless

bjorn:
@justgook Yeah, I hope to get around to this soon, but it will indeed be a much larger change.

Issue #1762, opened 2017-09-26
Enable compiling of the Python plugin when using Visual Studio


bjorn:
The Tiled project files (Qbs) currently disable the Python plugin when using Visual Studio. On Windows it is only compiled when using MinGW. Somebody should figure out how we can enable this plugin on Visual Studio, especially since this is currently preventing the plugin from shipping with Windows 64-bit builds.

This task may be easier when first the Python dependency is updated from Python 2 to 3 (#1759), but it could be done independently of that task in general.

Ideally the only thing a developer would need to do to get this to work is to install Python in its default location, as is currently the case when using Python 2.7 and MinGW.

PR #1900, opened 2018-03-03
python3 support

### LINKED ISSUE #1759: Update Python plugin to Python 3
Currently the Python plugin depends on Python 2.7. This version is increasingly old, as one can read [on the Python website](https://wiki.python.org/moin/Python2orPython3):

> Short version: Python 2.x is legacy, Python 3.x is the present and future of the language

The Python plugin shipping with Tiled was originally developed by @stt. I've since made minor changes, mostly to keep it compiling and to adjust it to changes to the interfaces it implements. But it has never seen proper maintenance, and I'm personally too occupied with other issues to spend time making sure the Plugin is really usable (there are no autotests), to document its use (it's currently not mentioned in [the manual](http://doc.mapeditor.org/) at all). The plugin is also not shipping with the 64-bit version of Tiled for Windows, which actually happens to be the most popular download (this is because I haven't figured out how to set up the project files to enable the plugin for Visual Studio builds).

So, there is actually a lot to do on this plugin, but I think updating it to the latest version of Python would be a good first step. Here's a list of issues covering the other things:

* Document how to use the Python plugin (#1760)
* Set up autotests for the Python plugin (#1761)
* Enable compiling of the Python plugin when using Visual Studio (#1762)

Note that if possible this should be done without relying on any compatibility flags (not sure if they have any).

stt:
### linux (mint 18) + qt5.6, and osx (10.12.6) with qt5.10

Built simply by `qmake . && make` and after `cp -a src/plugin/python/scripts ~/.tiled` at least fotf.py works out of the box (don't have data files to test all the other ones atm).

Noticed the new qbs build files later on. Since osx ships with py2 preinstalled but not with py3 I've changed the build to use pkgconfig for osx as well, which seems to work (after finally figuring out PkgConfigProbe defaulted to qbs.sysroot i.e. /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk for .pc files)

### win10, msvc2017 qt5.9 x64 runtime

Steps:

-  install [py3.6.4 x64](https://www.python.org/ftp/python/3.6.4/python-3.6.4-amd64.exe), and set the checkbox under advanced options to add it to environment variables (PATH) (there's [some reasons](https://bugs.python.org/issue3561#msg164635) why it's not on by default)
- in Qt Designer Kit-config set it to use msvc2017 as C/C++ x64 compiler
- add PYTHONHOME in Qt Designer project config, this is to locate header files
- building Release config now includes the plugin linked to py3
- copy src/plugin/python/scripts to %USERPROFILE%/.tiled .. this is something that should perhaps be changed, maybe the included scripts could be in a zip file in plugin directory and homedir .tiled is only for user's own scripts?

Some dlls changed names in msvc2017, msvcp120 became 140, and apparently due to ["Universal CRT"](https://stackoverflow.com/a/35806815/7571258) msvcr is no more. Build should work now with either set of dlls but if someone has a some other Visual Studio versions it'd be good to test?

**Troubleshooting**
If during building it errors saying something about 32/64bit, make sure it's not trying to use some other python installation.

If, when running Tiled it says "The specified module could not be found" this might mean it can't find the dynamically linked python3.dlls (one reason could be qt designer wasn't restarted after the PATH envvar was changed).

**Misc**
Should py3 be added in appveyor somehow?

Btw I'd first installed qt for windows UWP and spent an hour pondering error relating to QFileSystemWatcher being an "undeclared type" (due to qtcore-config.h defining QT_NO_FILESYSTEMWATCHER) before it dawned on me it was the wrong platform. Everything worked after reinstalling with msvc2017 x64 runtime. Perhaps it'd help others to put a check for that #define in a build script somewhere and alert people to check their qt version?

Closes #1759 and #1762

### DIFF
--- a/src/plugins/python/python.qbs
+++ b/src/plugins/python/python.qbs
@@ -7,33 +7,53 @@ TiledPlugin {
     Depends { name: "Qt"; submodules: ["widgets"] }
 
     condition: {
-        if (qbs.targetOS.contains("windows")) {
-            return File.exists( Environment.getEnv("PYTHONHOME") );
-        }
+        if (qbs.targetOS.contains("windows"))
+            return File.exists(Environment.getEnv("PYTHONHOME"));
 
-        return pkgConfig.found;
+        return pkgConfigPython3.found ||
+               pkgConfigPython3_6.found ||
+               pkgConfigPython3_4.found;
     }
 
     Probes.PkgConfigProbe {
-        id: pkgConfig
-        // the default qbs.sysroot basically just messes up system defaults
-        sysroot: ''
-        packageNames: ["python3"]
+        id: pkgConfigPython3
+        name: "python3"
+    }
+    Probes.PkgConfigProbe {
+        id: pkgConfigPython3_6
+        name: "python-3.6"
+    }
+    Probes.PkgConfigProbe {
+        id: pkgConfigPython3_4
+        name: "python-3.4"
     }
 
     Properties {
-        condition: pkgConfig.found
-
-        cpp.cxxFlags: pkgConfig.cflags
-        cpp.linkerFlags: pkgConfig.libs
-        // was ldflags broken for some py2 pkgconfigs? seems to work with py3
-        //cpp.dynamicLibraries: ["python3"]
+        condition: pkgConfigPython3.found
+        cpp.cxxFlags: pkgConfigPython3.cflags
+        cpp.dynamicLibraries: pkgConfigPython3.libraries
+        cpp.libraryPaths: pkgConfigPython3.libraryPaths
+        cpp.linkerFlags: pkgConfigPython3.linkerFlags
+    }
+    Properties {
+        condition: !pkgConfigPython3.found && pkgConfigPython3_6.found
+        cpp.cxxFlags: pkgConfigPython3_6.cflags
+        cpp.dynamicLibraries: pkgConfigPython3_6.libraries
+        cpp.libraryPaths: pkgConfigPython3_6.libraryPaths
+        cpp.linkerFlags: pkgConfigPython3_6.linkerFlags
+    }
+    Properties {
+        condition: !pkgConfigPython3.found && !pkgConfigPython3_6.found && pkgConfigPython3_4.found
+        cpp.cxxFlags: pkgConfigPython3_4.cflags
+        cpp.dynamicLibraries: pkgConfigPython3_4.libraries
+        cpp.libraryPaths: pkgConfigPython3_4.libraryPaths
+        cpp.linkerFlags: pkgConfigPython3_4.linkerFlags
     }
 
     Properties {
         condition: qbs.targetOS.contains("windows")
-        cpp.includePaths: [Environment.getEnv("PYTHONHOME")+"/include"]
-        cpp.libraryPaths: [Environment.getEnv("PYTHONHOME")+"/libs"]
+        cpp.includePaths: [Environment.getEnv("PYTHONHOME") + "/include"]
+        cpp.libraryPaths: [Environment.getEnv("PYTHONHOME") + "/libs"]
         cpp.dynamicLibraries: ["python3"]
     }
 
--- a/src/plugins/python/python.qbs
+++ b/src/plugins/python/python.qbs
@@ -2,6 +2,7 @@ import qbs 1.0
 import qbs.Probes as Probes
 import qbs.File
 import qbs.Environment
+import qbs.FileInfo
 
 TiledPlugin {
     Depends { name: "Qt"; submodules: ["widgets"] }
@@ -10,22 +11,17 @@ TiledPlugin {
         if (qbs.targetOS.contains("windows"))
             return File.exists(Environment.getEnv("PYTHONHOME"));
 
-        return pkgConfigPython3.found ||
-               pkgConfigPython3_6.found ||
-               pkgConfigPython3_4.found;
+        return pkgConfigPython3.found;
     }
 
     Probes.PkgConfigProbe {
         id: pkgConfigPython3
         name: "python3"
     }
-    Probes.PkgConfigProbe {
-        id: pkgConfigPython3_6
-        name: "python-3.6"
-    }
-    Probes.PkgConfigProbe {
-        id: pkgConfigPython3_4
-        name: "python-3.4"
+
+    PythonProbe {
+        id: pythonDllProbe
+        pythonDir: Environment.getEnv("PYTHONHOME")
     }
 
     Properties {
@@ -35,26 +31,19 @@ TiledPlugin {
         cpp.libraryPaths: pkgConfigPython3.libraryPaths
         cpp.linkerFlags: pkgConfigPython3.linkerFlags
     }
-    Properties {
-        condition: !pkgConfigPython3.found && pkgConfigPython3_6.found
-        cpp.cxxFlags: pkgConfigPython3_6.cflags
-        cpp.dynamicLibraries: pkgConfigPython3_6.libraries
-        cpp.libraryPaths: pkgConfigPython3_6.libraryPaths
-        cpp.linkerFlags: pkgConfigPython3_6.linkerFlags
-    }
-    Properties {
-        condition: !pkgConfigPython3.found && !pkgConfigPython3_6.found && pkgConfigPython3_4.found
-        cpp.cxxFlags: pkgConfigPython3_4.cflags
-        cpp.dynamicLibraries: pkgConfigPython3_4.libraries
-        cpp.libraryPaths: pkgConfigPython3_4.libraryPaths
-        cpp.linkerFlags: pkgConfigPython3_4.linkerFlags
-    }
 
     Properties {
-        condition: qbs.targetOS.contains("windows")
+        condition: qbs.targetOS.contains("windows") && !qbs.toolchain.contains("mingw")
         cpp.includePaths: [Environment.getEnv("PYTHONHOME") + "/include"]
         cpp.libraryPaths: [Environment.getEnv("PYTHONHOME") + "/libs"]
         cpp.dynamicLibraries: ["python3"]
+    }
+
+    Properties {
+        condition: qbs.targetOS.contains("windows") && qbs.toolchain.contains("mingw")
+        cpp.includePaths: [Environment.getEnv("PYTHONHOME") + "/include"]
+        cpp.libraryPaths: [Environment.getEnv("PYTHONHOME") + "/libs"]
+        cpp.dynamicLibraries: [FileInfo.joinPaths(Environment.getEnv("PYTHONHOME"), pythonDllProbe.fileNamePrefix + ".dll")]
     }
 
     files: [
--- a/src/plugins/python/python.qbs
+++ b/src/plugins/python/python.qbs
@@ -1,56 +1,40 @@
 import qbs 1.0
 import qbs.Probes as Probes
 import qbs.File
+import qbs.Environment
 
 TiledPlugin {
     Depends { name: "Qt"; submodules: ["widgets"] }
 
     condition: {
-        if (qbs.targetOS.contains("linux")) {
-            return pkgConfig.found;
-        } else if (qbs.targetOS.contains("windows")) {
-            // On Windows, currently only the default install location of
-            // Python 2.7 is supported, and only when compiling with MinGW in
-            // release mode. Also, avoid Python 2.7.10, since it results in a
-            // linker error.
-            //
-            return File.exists("C:/Python27") &&
-                    qbs.toolchain.contains("mingw") &&
-                    !qbs.debugInformation;
-        } else if (qbs.targetOS.contains("darwin")) {
-            return true;
+        if (qbs.targetOS.contains("windows")) {
+            return File.exists( Environment.getEnv("PYTHONHOME") );
         }
 
-        return false;
+        return pkgConfig.found;
     }
 
     Probes.PkgConfigProbe {
         id: pkgConfig
-        name: "python-2.7"
+        // the default qbs.sysroot basically just messes up system defaults
+        sysroot: ''
+        packageNames: ["python3"]
     }
 
     Properties {
-        condition: qbs.targetOS.contains("linux")
+        condition: pkgConfig.found
 
         cpp.cxxFlags: pkgConfig.cflags
-        // This should be it, but it doesn't work because the -lpython2.7 ends
-        // up too early on the command line.
-        //cpp.linkerFlags: pkgConfig.libs
-        cpp.dynamicLibraries: ["python2.7"]
+        cpp.linkerFlags: pkgConfig.libs
+        // was ldflags broken for some py2 pkgconfigs? seems to work with py3
+        //cpp.dynamicLibraries: ["python3"]
     }
 
     Properties {
         condition: qbs.targetOS.contains("windows")
-        cpp.includePaths: ["C:/Python27/include"]
-        cpp.libraryPaths: ["C:/Python27/libs"]
-        cpp.dynamicLibraries: ["python27"]
-    }
-
-    Properties {
-        condition: qbs.targetOS.contains("darwin")
-        cpp.includePaths: ["/System/Library/Frameworks/Python.framework/Versions/2.7/include/python2.7"]
-        cpp.libraryPaths: ["/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/config"]
-        cpp.dynamicLibraries: ["python2.7"]
+        cpp.includePaths: [Environment.getEnv("PYTHONHOME")+"/include"]
+        cpp.libraryPaths: [Environment.getEnv("PYTHONHOME")+"/libs"]
+        cpp.dynamicLibraries: ["python3"]
     }
 
     files: [
--- a/src/plugins/python/pythonplugin.cpp
+++ b/src/plugins/python/pythonplugin.cpp
@@ -71,11 +71,18 @@ void PythonPlugin::initialize()
         Py_NoSiteFlag = 1;
         Py_NoUserSiteDirectory = 1;
 
+        #if PY_VERSION_HEX >= 0x03000000
+        PyImport_AppendInittab("tiled", PyInit_tiled);
+        PyImport_AppendInittab("tiled.qt", PyInit_tiled);
+        PyImport_AppendInittab("tiled.Tiled", PyInit_tiled);
+        Py_Initialize();
+        #else
         Py_Initialize();
         inittiled();
+        #endif
 
-        // Get reference to base class to find its extensions later on
         PyObject *pmod = PyImport_ImportModule("tiled");
+
         if (pmod) {
             PyObject *tiledPlugin = PyObject_GetAttrString(pmod, "Plugin");
             Py_DECREF(pmod);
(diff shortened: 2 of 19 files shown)


src/plugins/python/pythonplugin.cpp
```diff
@@ -71,11 +71,18 @@ void PythonPlugin::initialize()
         Py_NoSiteFlag = 1;
         Py_NoUserSiteDirectory = 1;
 
+        #if PY_VERSION_HEX >= 0x03000000
```
bjorn: What is the point in still supporting older versions of Python here? It seems that the build files won't ever build against Python 2.7 and nor does `pythonbind.cpp` support older versions.

src/plugins/python/pythonplugin.cpp
```diff
@@ -133,6 +141,7 @@ void PythonPlugin::initialize()
 void PythonPlugin::log(Tiled::LoggingInterface::OutputType type,
                        const QString &msg)
 {
+    printf("%s\n", msg.toStdString().c_str());
```
bjorn: Was this intended to stay?

src/plugins/python/scripts/lib/lbm.py
```diff
@@ -57,28 +57,29 @@ class CMAP(CpyStruct("BYTE color[3];", True)):
...
-      yield QColor(c.color[0],c.color[1],c.color[2]).rgb()
+      yield T.qt.QColor(c.color[0],c.color[1],c.color[2]).rgb()
 
 def uncomp(dat):
   i = 0
   ret = bytearray()
   while i < len(dat):
-    v = struct.unpack('b', dat[i])[0]
+    #v = struct.unpack('b', dat[i:i+1])[0]
+    v = (256-dat[i])*-1 if dat[i] > 127 else dat[i]
```
bjorn: This seems like a great improvement in readability in Python 3. o_O

src/plugins/python/scripts/lib/utils.py
```diff
@@ -0,0 +1,17 @@
+import os.path
+
+
+def find_sensitive_path(dir, insensitive_path):
```
bjorn: What is a sensitive or insensitive path?

bjorn:
On my Arch Linux with Python 3.6.4 installed, I get the following when it tries to load the Python plugin:
```
Error: Cannot load library .../install-root/lib/tiled/plugins/libpython.so:
(.../install-root/lib/tiled/plugins/libpython.so: undefined symbol: _Py_NoneStruct)
```

src/plugins/python/pythonplugin.cpp
stt: I was actually wondering if we should bindings for both, but no, you're right, time to let 2 go

src/plugins/python/python.qbs
```diff
@@ -1,56 +1,40 @@
...
 
-        return false;
+        return pkgConfig.found;
     }
 
     Probes.PkgConfigProbe {
         id: pkgConfig
-        name: "python-2.7"
+        // the default qbs.sysroot basically just messes up system defaults
+        sysroot: ''
```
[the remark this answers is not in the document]
stt: Perhaps the reason it defaults to a toolchain path instead of systemwide pkgconfigs is to support cross-compiling? Could perhaps just remove the slightly passive aggressive comment :)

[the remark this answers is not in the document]
stt: qbs manual says of sysroot "This property is typically set in a profile for cross-compiling. Default: Undefined",  my profile was created just by `qbs setup-qt`, not sure what to report yet.

src/plugins/python/python.qbs
```diff
@@ -1,56 +1,40 @@
...
+        condition: pkgConfig.found
 
         cpp.cxxFlags: pkgConfig.cflags
-        // This should be it, but it doesn't work because the -lpython2.7 ends
-        // up too early on the command line.
-        //cpp.linkerFlags: pkgConfig.libs
-        cpp.dynamicLibraries: ["python2.7"]
+        cpp.linkerFlags: pkgConfig.libs
+        // was ldflags broken for some py2 pkgconfigs? seems to work with py3
+        //cpp.dynamicLibraries: ["python3"]
```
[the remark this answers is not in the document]
stt: yep was intended as a review note, it works for me but should some different linux distributions be tested?

src/plugins/python/scripts/lib/lbm.py
stt: I take it you weren't serious?-) (the old way works too with the slice notation, I'll switch it back)

src/plugins/python/scripts/lib/utils.py
stt: case-in/sensitive, since DOS-based systems don't care about cases but *nix does. script should find case-sensitive path by using incorrect casing.

stt:
hm, what does `ldd lib/tiled/plugins/libpython.so |grep py` and `pkg-config python3 --cflags` tell you?
If it's not there already could you try adding `-rdynamic` to cpp.cxxFlags in python.qbs?

src/plugins/python/python.qbs
```diff
@@ -1,56 +1,40 @@
...
-        return false;
+        return pkgConfig.found;
     }
 
     Probes.PkgConfigProbe {
         id: pkgConfig
-        name: "python-2.7"
+        // the default qbs.sysroot basically just messes up system defaults
+        sysroot: ''
+        packageNames: ["python3"]
```
bjorn: On Ubuntu I had to change this to "python-3.6".

bjorn:
Alright, the Python plugin builds fine on Arch Linux now and should also work on Ubuntu. Next step would be making sure it works on Travis CI.

> Should py3 be added in appveyor somehow?

I've pushed a change that may enable Python3 on AppVeyor, let's see.

bjorn:
Ok, I removed the `qbs.sysroot` thing because for me it was `undefined` by default. Is the default value problematic on Windows or macOS?

bjorn:
@stt Any idea what could be causing those linker errors for the MinGW builds?

bjorn:
Rebased on latest master to fix up the merge conflict.

Also, I'm not sure if the Python 3.4 fallback is meaningful. Apparently on Travis CI that "python3" pkg-config already worked, even though normally on Ubuntu this doesn't appear to work.

stt:
On Mint18 `/usr/lib/x86_64-linux-gnu/pkgconfig/python3.pc` comes from package python3-dev and `apt depends python3-dev` says it `Depends: python3.5-dev` which itself provides .pc file for 3.5, but there's no reverse dependency. So, people might have 3.x pkgconfigs without python3.pc. This is a bit odd, I wonder if build scripts for other sw have a fallback hierarchy for every minor release.

bjorn:
> On Mint18 `/usr/lib/x86_64-linux-gnu/pkgconfig/python3.pc` comes from package `python3-dev`

Arg... I had installed `libpython3-dev` and not `python3-dev`, and so hence was missing the `python3.pc` file. I find this packaging method odd indeed...

> I wonder if build scripts for other sw have a fallback hierarchy for every minor release.

Well, that does mean all tested systems so far have a `python3.pc` once you install the Python dependency in the right way, so I will remove the fallback chain again. That also explains why Travis CI was happily building the Python plugin before I put that in place.

bjorn:
@stt Do you have any thoughts on the failing MinGW builds on AppVeyor?

stt:
Hey @bjorn, sry for late reply, was traveling a bit last week, had some time to look into it now. Got thrown off at first by there not being mingw kit or tools options in Qt installer for linux or osx, and no up-to-date distribution packages either. Almost started cross-compiling Qt for mingw on linux before realizing that the Windows Qt installer had all the mingw stuff, that would've been a waste of a day. :)

Anyway, those symbols are in a seperate dll that msvc seems to find ok when just python3.dll is linked, not sure how all of the dll magic works with windows but my local mingw started compiling by changing `cpp.dynamicLibraries: ["python3"]` to "python36".

If there's distribution packaging scripts, e.g. `debian/rules` somewhere then python3-dev could be added to Build-Depends field, and to BuildRequires field for rpm spec etc, to avoid that missing .pc file issue, and the 3.4 fallback can be taken out, but I'm not sure how to avoid specifying the minor version for mingw.

stt:
AppVeyor seems to work now but Travis complains about yarn command?

bjorn:
> AppVeyor seems to work now but Travis complains about yarn command?

I think that part is non-critical. The actual problem with the build seems to be:
```
$ qbs install --install-root install release projects.Tiled.version:$TILED_VERSION
ERROR: Unexpected command line parameter 'release'.
Expected an assignment of the form <property>:<value>, profile:<profile-name> or config:<configuration-name>.
```

This used to work fine, but apparently something changed in Qbs 1.11 (which I guess was just released and is now installed by `brew`) that makes this command no longer work.

bjorn:
Pushed a test fix for the macOS build as f280afc8bc793e9e594ad957d292372d8ed31e91, will merge to master when it works and then this PR can be rebased on top.

bjorn:
`set |grep PYT` does not output anything for me.

Maybe I could fix this by shipping everything in `libpython3.4-minimal` (and maybe it's a good idea to also include everything in `libpython3.4-stdlib`). However, I wonder if this is the way to go? Wouldn't users expect to be able to use any Python module they have installed on their system in their Tiled scripts?

To that end, probably we want to load the system Python library. But then, is there anything we can do to make this work with both Python 3.4 and 3.6?

bjorn:
I think we'll want to link against `libpython3` and make sure to only use the [stable ABI](https://www.python.org/dev/peps/pep-0384/). But when I tried to compile with `Py_LIMITED_API=0x03040000` defined, I had to make a few changes:

```diff
--- src/plugins/python/pythonplugin.cpp
+++ src/plugins/python/pythonplugin.cpp
@@ -72,10 +72,6 @@ void PythonPlugin::initialize()
     addObject(&mLogger);
 
     if (!Py_IsInitialized()) {
-        // PEP370
-        Py_NoSiteFlag = 1;
-        Py_NoUserSiteDirectory = 1;
-
         PyImport_AppendInittab("tiled", PyInit_tiled);
         PyImport_AppendInittab("tiled.qt", PyInit_tiled);
         PyImport_AppendInittab("tiled.Tiled", PyInit_tiled);
@@ -368,7 +364,7 @@ QString PythonMapFormat::nameFilter() const
         PySys_WriteStderr("** Uncaught exception in script **\n");
     } else {
         PyObject* pyStr = PyUnicode_AsEncodedString(pinst, "utf-8", "Error ~");
-        ret = PyBytes_AS_STRING(pyStr);
+        ret = PyBytes_AsString(pyStr);
         Py_XDECREF(pyStr);
         Py_DECREF(pinst);
     }
@@ -396,7 +392,7 @@ QString PythonMapFormat::shortName() const
         PySys_WriteStderr("** Uncaught exception in script **\n");
     } else {
         PyObject* pyStr = PyUnicode_AsEncodedString(pinst, "utf-8", "Error ~");
-        ret = PyBytes_AS_STRING(pyStr);
+        ret = PyBytes_AsString(pyStr);
         Py_XDECREF(pyStr);
         Py_DECREF(pinst);
     }
```

The remaining issue is that there is no `PyRun_SimpleString`. Seems like we need to use some more low-level functions to replicate that functionality.

bjorn:
Hmm, even if we manage to fix up `pythonplugin.cpp` to only use stable ABI, I think we won't get very far because it does not look like PyBindGen likes the limited API. :-(

stt:
Yep PEP384 makes sense to me. Did the changes to get rid of PyRun, also added include path for application scripts in addition to user's own scripts that (meant to do that long time ago).
Seems to work ok with Py_LIMITED_API 0x03040000, on linux and mac at least, did you find something that had a problem with it?

One thing I'm still working on is getting stderr to use LoggingInterface's error output without having to change LoggingInterface itself, I'll make a PR once it's figured out.

bjorn:
> Seems to work ok with Py_LIMITED_API 0x03040000, on linux and mac at least, did you find something that had a problem with it?

I tried it only Linux but was having compile issues on the generated code, so I was ready to give up. The compile errors I ran into were:

```
/usr/lib/ccache/g++ -g -O0 -Wall -Wextra -m64 -pipe -fexceptions -fvisibility=hidden -fvisibility-inlines-hidden -I/usr/include/python3.6m -I/usr/include/x86_64-linux-gnu/python3.6m -fPIC -DQT_DEPRECATED_WARNINGS -DQT_DISABLE_DEPRECATED_BEFORE=0x050900 -DQT_NO_FOREACH -DQT_NO_URL_CAST_FROM_STRING -DPy_LIMITED_API=0x03040000 -DQT_CORE_LIB -DQT_GUI_LIB -DQT_WIDGETS_LIB -I/home/bjorn/Qt/5.10.1/gcc_64/include -I/home/bjorn/Qt/5.10.1/gcc_64/include/QtCore -I/home/bjorn/Qt/5.10.1/gcc_64/mkspecs/linux-g++ -I/home/bjorn/build-tiled-Desktop_Qt_5_10_1_GCC_64bit2-Debug/qtc_Desktop_Qt_5_10_1_GCC_64bit2_Debug/python.4235227b/qt.headers -I/home/bjorn/Qt/5.10.1/gcc_64/include/QtGui -I/home/bjorn/Qt/5.10.1/gcc_64/include/QtWidgets -I/home/bjorn/tiled/src/libtiled -std=c++11 -o /home/bjorn/build-tiled-Desktop_Qt_5_10_1_GCC_64bit2-Debug/qtc_Desktop_Qt_5_10_1_GCC_64bit2_Debug/python.4235227b/3a52ce780950d4d9/pythonbind.cpp.o -c /home/bjorn/tiled/src/plugins/python/pythonbind.cpp
/home/bjorn/tiled/src/plugins/python/pythonbind.cpp:14:9: error: ‘PyAsyncMethods’ does not name a type; did you mean ‘Py_tp_methods’?
 typedef PyAsyncMethods* cmpfunc;
         ^~~~~~~~~~~~~~
         Py_tp_methods
/home/bjorn/tiled/src/plugins/python/pythonbind.cpp: In function ‘void _wrap_PyQPointF__tp_dealloc(PyQPointF*)’:
/home/bjorn/tiled/src/plugins/python/pythonbind.cpp:423:18: error: invalid use of incomplete type ‘struct _typeobject’
     Py_TYPE(self)->tp_free((PyObject*)self);
                  ^~
In file included from /usr/include/python3.6m/Python.h:71:0,
                 from /home/bjorn/tiled/src/plugins/python/pythonbind.cpp:8:
/usr/include/python3.6m/object.h:109:12: note: forward declaration of ‘struct _typeobject’
     struct _typeobject *ob_type;
            ^~~~~~~~~~~
/home/bjorn/tiled/src/plugins/python/pythonbind.cpp: At global scope:
/home/bjorn/tiled/src/plugins/python/pythonbind.cpp:459:14: error: variable ‘PyTypeObject PyQPointF_Type’ has initializer but incomplete type
 PyTypeObject PyQPointF_Type = {
              ^~~~~~~~~~~~~~
/home/bjorn/tiled/src/plugins/python/pythonbind.cpp:466:6: error: ‘printfunc’ was not declared in this scope
     (printfunc)0,                      /* tp_print */
      ^~~~~~~~~
/home/bjorn/tiled/src/plugins/python/pythonbind.cpp:466:6: note: suggested alternative: ‘printf’
     (printfunc)0,                      /* tp_print */
      ^~~~~~~~~
      printf
/home/bjorn/tiled/src/plugins/python/pythonbind.cpp:466:16: error: expected ‘}’ before numeric constant
     (printfunc)0,                      /* tp_print */
                ^
/home/bjorn/tiled/src/plugins/python/pythonbind.cpp:466:16: error: expected ‘,’ or ‘;’ before numeric constant
/home/bjorn/tiled/src/plugins/python/pythonbind.cpp:511:1: error: expected declaration before ‘}’ token
 };
 ^
/home/bjorn/tiled/src/plugins/python/pythonbind.cpp:427:1: warning: ‘PyObject* _wrap_PyQPointF__tp_richcompare(PyQPointF*, PyQPointF*, int)’ defined but not used [-Wunused-function]
 _wrap_PyQPointF__tp_richcompare (PyQPointF *PYBINDGEN_UNUSED(self), PyQPointF *other, int opid)
 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/bjorn/tiled/src/plugins/python/pythonbind.cpp:407:20: warning: ‘PyQPointF_methods’ defined but not used [-Wunused-variable]
 static PyMethodDef PyQPointF_methods[] = {
                    ^~~~~~~~~~~~~~~~~
/home/bjorn/tiled/src/plugins/python/pythonbind.cpp:335:1: warning: ‘int _wrap_PyQPointF__tp_init(PyQPointF*, PyObject*, PyObject*)’ defined but not used [-Wunused-function]
 _wrap_PyQPointF__tp_init(PyQPointF *self, PyObject *args, PyObject *kwargs)
 ^~~~~~~~~~~~~~~~~~~~~~~~
/home/bjorn/tiled/src/plugins/python/pythonbind.cpp:327:20: warning: ‘tiled_qt_functions’ defined but not used [-Wunused-variable]
 static PyMethodDef tiled_qt_functions[] = {
                    ^~~~~~~~~~~~~~~~~~
Process failed with exit code 1.
```
If you are able to get it working, I'm really looking forward to a pull request! :-)

stt:
Odd, I'm also compiling against py3.6 without errors, although with clang, make shows this command:

        /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang++ -c -pipe -stdlib=libc++ -O2 -std=gnu++11  -arch x86_64 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk -mmacosx-version-min=10.10 -fvisibility=hidden -fvisibility-inlines-hidden -Wall -W -fPIC -DPYTHON_LIBRARY -DQT_NO_DEBUG -DQT_PLUGIN -DQT_WIDGETS_LIB -DQT_GUI_LIB -DQT_CORE_LIB -DPy_LIMITED_API=0x03040000 -I. -I../../libtiled -I/usr/local/opt/python/Frameworks/Python.framework/Versions/3.6/include/python3.6m -I/usr/local/Cellar/qt/5.10.1/lib/QtWidgets.framework/Headers -I/usr/local/Cellar/qt/5.10.1/lib/QtGui.framework/Headers -I/usr/local/Cellar/qt/5.10.1/lib/QtCore.framework/Headers -I. -I/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk/System/Library/Frameworks/OpenGL.framework/Headers -I/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk/System/Library/Frameworks/AGL.framework/Headers -I/usr/local/Cellar/qt/5.10.1/mkspecs/macx-clang -F/usr/local/Cellar/qt/5.10.1/lib -o pythonplugin.o pythonplugin.cpp

Re: LoggingInterface, spent couple hours considering all sorts of wrappers and c++11 lambdas for the logs before realising that since it iterates through all the scripts on startup why not just move the previous PyRun stdout/stderr redirect into a script, guess it was just too obvious. :)

I'll open the PR so we can continue there.

bjorn:
@stt The bounties on bountysource.com for issues #1759 and #1762 are still open. You totally deserve them for your efforts, but if you don't intend to claim them please let me know, then I'll look for other issues to put them on.

stt:
Oh man totally forgot about it after a while. I'm ok with redistributing it to new bounties. Should I claim it and add the sum back as a contributor or how is it easiest to do?
Merry Christmas @bjorn 🎅

bjorn:
@stt I approved the claims. Up to you what to do with it, it's yours! I'm not so excited about BountySource specifically given their pushing of useless CAN tokens, so maybe you should just have it paid out. In any case, Merry Christmas to you as well! :christmas_tree:

Issue #1665, opened 2017-07-27
Tiled Projects


griest024:
Apologies if this is a duplicate issue. It is at least partially duplicated by a number of issues (#945, #233, #70). However many of those issues are quite old and don't pertain to Tiled 1.x. In addition, those issues bring up a number of good ideas but I think it would be good to bring them together in one place and codify them into a concrete proposal.

# Proposal

Tiled should have projects.

## Data Structure

A project is a data structure that can be dumped to a `.json` file or an XML-based file similar to `.tmx`.

A project contains a number of fields including:

### "Orphan" Data

Data that doesn't belong to any particular data structure such as

- Commands
- Object types

*Note on orphan data merge strategy: I think it still makes sense to have globally defined commands and object types available in projects. In the case where there is a conflict, the project data should take precedence.*

### Session Data

*Edit: After looking at this again, it doesn't make sense for session data to be kept in a project file. It should instead be kept in some kind of workspace file, similar to the behavior of sublime text and other IDEs.*

Data that describes the state that the project was last in

- Open files: which files are open as tabs
- Current file: which file is currently being edited
- Current file history: the edit history of the current file (possible performance concerns)
- Currently selected object, tile, map, or tileset whose properties are displayed

### Directory Data

Paths relative to the project file that define a number of important directories including

- Image directory: if this is defined, all paths pointing to images (like in tilesets) are relative to this directory.
- Tileset directory: if this is defined, all paths pointing to tilesets (like in maps) are relative to this directory.

### File Data

Paths relative to the project file that point to files included in this project.

- Tilesets
- Maps

### Miscellaneous Data

- Custom properties: similar to other data structures, a project should allow for custom properties

## UI

Along with the project data structure, this proposal includes UI changes. These changes make it easier to interact with the project data structure and its members directly.

### Project Properties

A project should enumerate its members in the property browser, similar to other data structures (map, tileset, etc.)

### Project Browser

The UI should include a new panel (widget?) called the *project browser*. This panel is an accordion, and lists the 4 major components of a project: maps, tilesets, commands, and object types. Each accordion section includes a list with multi-select capability.

![image](https://user-images.githubusercontent.com/5853222/28741405-687c09c2-73e3-11e7-8b15-433fc653c236.png)

An asterisk denotes that a file has unsaved changes, like the tab.

Map and tileset list items should have a context menu with the following items whose functions are identical to those in the file submenu:

- Open
- Export (This would effectively close #1117)
- Export As...
- Save
[...]

PR #1901, opened 2018-03-04
Add support to Lua exporter for object templates

Spice-King:
Properties are copied from the template, then overwritten by the instances's properties before being written to the export.

Pros:
* Backwards compatible with older tools/importers
* No need to check templates for properties
* No extra runtime overhead, compared to not using templates

Cons:
* Can't directly tell if the object had a template
* Can't tell if any properties were from the template or not

### DIFF
--- a/src/plugins/lua/luaplugin.cpp
+++ b/src/plugins/lua/luaplugin.cpp
@@ -656,13 +656,14 @@ void LuaWriter::writeMapObject(LuaTableWriter &writer,
         break;
     }
 
-    Properties props;
-    if (const ObjectTemplate *objectTemplate = mapObject->objectTemplate()) {
-        props.merge(objectTemplate->object()->properties());
+    if (const MapObject *base = mapObject->templateObject()) {
+        // Include template properties
+        Properties props = base->properties();
+        props.merge(mapObject->properties());
+        writeProperties(writer, props);
+    } else {
+        writeProperties(writer, mapObject->properties());
     }
-    // Override template properties if they exist
-    props.merge(mapObject->properties());
-    writeProperties(writer, props);
 
     writer.writeEndTable();
 }
--- a/src/plugins/lua/luaplugin.cpp
+++ b/src/plugins/lua/luaplugin.cpp
@@ -28,6 +28,7 @@
 #include "map.h"
 #include "mapobject.h"
 #include "objectgroup.h"
+#include "objecttemplate.h"
 #include "properties.h"
 #include "savefile.h"
 #include "terrain.h"
@@ -655,7 +656,13 @@ void LuaWriter::writeMapObject(LuaTableWriter &writer,
         break;
     }
 
-    writeProperties(writer, mapObject->properties());
+    Properties props;
+    if (const ObjectTemplate *objectTemplate = mapObject->objectTemplate()) {
+        props.merge(objectTemplate->object()->properties());
+    }
+    // Override template properties if they exist
+    props.merge(mapObject->properties());
+    writeProperties(writer, props);
 
     writer.writeEndTable();
 }


src/plugins/lua/luaplugin.cpp
```diff
@@ -655,7 +656,13 @@ void LuaWriter::writeMapObject(LuaTableWriter &writer,
         break;
     }
 
-    writeProperties(writer, mapObject->properties());
+    Properties props;
+    if (const ObjectTemplate *objectTemplate = mapObject->objectTemplate()) {
+        props.merge(objectTemplate->object()->properties());
```
bjorn: Here, we could just assign instead of calling `merge`, given that `props` is still empty anyway.

Please also remove the brackets, since there is only a single statement.

Spice-King:
Funny thing, I was a good chunk of the way into doing something close to export options. One of the things I did want to include was CLI overrides for the export command. Spent a chunk of time tossing around what would do the least damage and where to integrate export options.

I had not figured out where I wanted to store it, stuffing it into the map file seemed too granular, but sticking it in the global preferences felt kinda buried. A middle ground of a project or a workspace kind of thing feels more like where is should be, but we don't have that yet. Might tackle that as a project later.

Also, we get into a bit of a odd spot when an exporter does not react to an export option. Someone will get confused by that. No nice idea comes to mind to solve that. Can't assume that the plugin list is fixed, so I can't just label which options work with what formats. Might just wing it for now and figure it out later.

I'll give it some TLC over the next day and work on the export options.

bjorn:
> Funny thing, I was a good chunk of the way into doing something close to export options. One of the things I did want to include was CLI overrides for the export command. Spent a chunk of time tossing around what would do the least damage and where to integrate export options.

I find it usually quite annoying that such a large amount of time in programming is spent weighing in questions about what approach to take (or even just how to name something), rather than actually getting something done. But such is the life of a programmer, or maybe mostly the life of a perfectionist programmer... Anyway, if you have other suggestions I'd love to hear them. :-)

> I had not figured out where I wanted to store it, stuffing it into the map file seemed too granular, but sticking it in the global preferences felt kinda buried. A middle ground of a project or a workspace kind of thing feels more like where is should be, but we don't have that yet. Might tackle that as a project later.

Yes, there are already many things that would work better when stored in a project or workspace file. This is covered by issue #1665 and it's on the [Roadmap](https://github.com/bjorn/tiled/wiki/Roadmap) for Tiled 1.3.

> Also, we get into a bit of a odd spot when an exporter does not react to an export option. Someone will get confused by that. No nice idea comes to mind to solve that. Can't assume that the plugin list is fixed, so I can't just label which options work with what formats. Might just wing it for now and figure it out later.

Actually all plugins are part of Tiled so we can make sure the setting is recognized in all places where the user would expect it.

Of course, there will probably also be a use-case for format-specific options. I have currently no design for this though, especially regarding how the UI would present such options to the user. I imagine a plugin would need to register or expose its options somewhere, such that a UI can be built for it dynamically. But this exporting option is pretty universal, as would be the option to embed tilesets, which can be added in the same way.

> I'll give it some TLC over the next day and work on the export options.

Looking forward to it!

bjorn:
So, I ended up implementing the "Detach Templates" export option independently of the export format (fe13589e7eaa8d5564df5397ffa5755d106cc47c), so that that it is immediately supported by all formats without needing to modify each of them.

I think that obsoletes this change, though I'll keep the PR open for now as a reminder that the Lua format should still be fixed to at least refer to the template files if templates are not detached on export.

bjorn:
In hindsight I decided to merge this change in addition to the recently added export options, because I realized that in case of the Lua plugin, templates were already written out in a detached form (with all attributes being simply included, regardless of whether the object overrides them), with the exception of the custom properties. I believe this exception was just an oversight.

PR #1917, opened 2018-03-27
Tile scalefactor property

kralle333:
Made it possible to set a scalefactor on tiles on tilesets. When placing an object on a map, the object will be scaled up/down using this scalefactor. This is useful for sprites that are scaled down to reduce file sizes and scaled up again when the map is loaded in a game.

### DIFF
--- a/src/libtiled/mapreader.cpp
+++ b/src/libtiled/mapreader.cpp
@@ -427,6 +427,9 @@ void MapReaderPrivate::readTilesetTile(Tileset &tileset)
     Tile *tile = tileset.findOrCreateTile(id);
 
     tile->setType(atts.value(QLatin1String("type")).toString());
+    QStringRef scaleFactor = atts.value(QLatin1String("scalefactor"));
+    if(!scaleFactor.isEmpty())
+        tile->setScaleFactor(scaleFactor.toDouble());
 
     // Read tile quadrant terrain ids
     QString terrain = atts.value(QLatin1String("terrain")).toString();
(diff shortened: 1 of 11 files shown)


bjorn:
This seems like a fast way of causing placed tile objects to be automatically given a larger size than the tile image. Did you consider to create scaled templates of your tile objects instead and place instances of those?

kralle333:
Ah I see, I haven't used the template feature before. Having the object given a larger size than the image size is indeed what i want. I see in the documentation that groups of templates can be made, however when I get the newest windows build from itch.io I can not see this feature. I it quite tedious if I have to create a file for each map object, e.g. I might have say 30 tiles in which I would then need to make 30 template files.

bjorn:
> I see in the documentation that groups of templates can be made, however when I get the newest windows build from itch.io I can not see this feature.

The screenshots in the documentation are a little out of date. The templates are grouped by regular file system folders now, which avoids the management of the template group files.

> I it quite tedious if I have to create a file for each map object, e.g. I might have say 30 tiles in which I would then need to make 30 template files.

That is indeed tedious, so I can see your change is more usable for your needs. I still feel we're duplicating some functionality here and wonder if there would be a different way to solve the problem of tediousness.

For this particular case, I could imagine to support a scale property for the "Insert Tile" tool, which would mainly be useful if you often use the same scale to place multiple tiles. Shortcuts could be added to scale up/down the current tile and a scale entry widget could be added to the tool bar. However, this does not help if you need a specific fixed scale factor that is different for each tile. Is that the case for you?

kralle333:
Sounds like it could be useful, but it sounds like it would require a lot of code (There are no properties UI for tools right now). For now the solution is simply to have scaled objects outside the level area that is just copy/pasted.

bjorn:
> Sounds like it could be useful, but it sounds like it would require a lot of code (There are no properties UI for tools right now).

Actually there is a tool-specific tool bar since Tiled 1.1, which is used by some tools to show options or trigger certain actions. So adding a "scale" text entry when using the Insert Tile tool should not be too much trouble.

> For now the solution is simply to have scaled objects outside the level area that is just copy/pasted.

Right, objects that are already placed can be copy/pasted or duplicated easily.

PR #1931, opened 2018-04-09
Bugfix - Layer not selected if clicked in objects dock view

kralle333:


src/tiled/objectsdock.cpp
```diff
@@ -342,7 +342,7 @@ void ObjectsView::mousePressEvent(QMouseEvent *event)
             DocumentManager::instance()->centerMapViewOn(center + offset);
         }
     } else if (Layer *layer = mapObjectModel()->toLayer(index)) {
-        mMapDocument->setCurrentObject(layer);
+        mMapDocument->setCurrentLayer(layer);
```
bjorn: Please keep the call to `setCurrentObject`, since changing the current layer will only automatically change the current object when either there's isn't a current object or it is a layer. But when the user clicks a layer he may expect to see its properties, for which we need to change the current object.

In any case I agree it would be expected that the current layer changes as well. I do not remember why this code was not doing that in the first place. Maybe there was no good reason, or maybe we'll find out later. :-)

bjorn:
Thanks!

Did you check how this works together with multiple selected layers?

kralle333:
Haven't seen issues when selecting multiple layers.

bjorn:
> Haven't seen issues when selecting multiple layers.

Alright, indeed there isn't really a problem, it just changes the selection to that one clicked object layer.

It is however possible to select multiple object layers also in the Objects view, but this currently does not lead to multiple selected layers. I'm not sure if we want to go that far though, since layer selection in general is currently not reflected in the Objects view, and when doing so we would mix object and layer selection in a single view. It actually makes little sense to deselect all objects when clicking on a layer, and nor would it make sense to deselect the current layers when clicking on an object. But that is exactly what the Qt selection behavior would do if we stick it into a single view. It may of course be feasible to customize this behavior, but I think that's only worth looking into when we go as far as to unify the Objects and Layers views into a single view (though that raises additional questions).

PR #1942, opened 2018-05-06
tiledquickplugin: render horizontally and vertically flipped tiles

### LINKED ISSUE #1751: Add support for horizontally and vertically flipped tiles in tiledquickplugin

mitchcurtis:
Fixes #1751

### DIFF
--- a/src/tiledquickplugin/tilelayeritem.cpp
+++ b/src/tiledquickplugin/tilelayeritem.cpp
@@ -165,6 +165,8 @@ static void drawOrthogonalTileLayer(QSGNode *parent,
             data.y = (y + 1) * tileHeight - tileset->tileHeight() + offset.y();
             data.width = size.width();
             data.height = size.height();
+            data.flippedHorizontally = cell.flippedHorizontally();
+            data.flippedVertically = cell.flippedVertically();
             helper.setTextureCoordinates(data, cell);
             tileData.append(data);
         }
--- a/src/tiledquickplugin/tilelayeritem.cpp
+++ b/src/tiledquickplugin/tilelayeritem.cpp
@@ -244,6 +244,8 @@ void IsometricRenderHelper::appendTileData(int x, int y)
     const QSize size = cell.tile()->size();
     data.width = size.width();
     data.height = size.height();
+    data.flippedHorizontally = cell.flippedHorizontally();
+    data.flippedVertically = cell.flippedVertically();
     mTilesetHelper.setTextureCoordinates(data, cell);
     mTileData.append(data);
 }
--- a/src/tiledquickplugin/tilesnode.cpp
+++ b/src/tiledquickplugin/tilesnode.cpp
@@ -94,6 +94,17 @@ void TilesNode::processTileData(const QVector<TileData> &tileData)
         v[3].tx = s_tx;                 v[4].tx = s_tx + s_width;
         v[3].ty = s_ty + s_height;      v[4].ty = s_ty + s_height;
 
+        if (data.flippedHorizontally) {
+            std::swap(v[0].tx, v[4].tx);
+            std::swap(v[1].tx, v[5].tx);
+            std::swap(v[2].tx, v[3].tx);
+        }
+        if (data.flippedVertically) {
+            std::swap(v[0].ty, v[4].ty);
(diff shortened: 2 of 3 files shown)


mitchcurtis:
Oops, missed some changes...

bjorn:
That's great and would indeed to the trick!

What about diagonal flipping? I guess it's rather more complicated, except for square tiles where simply the top-right and bottom-left corners could be flipped. For non-square tiles the width/height would need to be flipped and possibly the location would be affected as well.

As for the compile error, I will try to get around to merging `master` into the `wip/qtquick` branch again. Hopefully it will solve both issues. However, if you did not have these issues last time you worked on this branch, then I don't really understand why you have them now. Also I see neither of these problems on Linux with GCC 8.1 and Qt 5.10.1.

mitchcurtis:
> That's great and would indeed do the trick!
>
> What about diagonal flipping? I guess it's rather more complicated, except for square tiles where simply the top-right and bottom-left corners could be flipped. For non-square tiles the width/height would need to be flipped and possibly the location would be affected as well.

How does diagonal flipping work? This patch supports a combination of horizontal and vertical flipping, if that's what you mean?

In any case, it seems like something that could be added later if the need arises.

> As for the compile error, I will try to get around to merging master into the wip/qtquick branch again. Hopefully it will solve both issues. However, if you did not have these issues last time you worked on this branch, then I don't really understand why you have them now. Also I see neither of these problems on Linux with GCC 8.1 and Qt 5.10.1 (compiling with Qbs).

I probably just screwed something up with Git at some point, then... I'm not sure. If I run into the same problem later I'll double check that I'm doing it properly and if I still run into it, then I'll come here and complain. :)

bjorn:
> How does diagonal flipping work? This patch supports a combination of horizontal and vertical flipping, if that's what you mean?

It's flipping of the tile over its diagonal, effectively exchanging the top-right and bottom-left corners. This operation enables rotation by 90-degree increments.

But indeed, support for that flag can be added in a later patch.

bjorn:
> Does the wip/qtquick need to be updated against the latest master?

It was over a year out of date, but I've just merged `master` to `wip/qtquick` and it seems to work fine. I expect there will be issues with support for infinite maps, though.

In general, I still think this is an exciting branch and I do hope to have time to come back to it at some point...

mitchcurtis:
> It was over a year out of date, but I've just merged master to wip/qtquick and it seems to work fine. I expect there will be issues with support for infinite maps, though.

Oh sweet, thanks!

> In general, I still think this is an exciting branch and I do hope to have time to come back to it at some point...

That's great to hear!

bjorn:
> Oh sweet, thanks! I haven't been using the infinite maps feature, which is probably a good thing while all tiles are considered visible. :)

Well, each layer keeps a bounds that is the rectangle around the actually used part. Alternatively the exact region of non-empty tiles can be obtained. So actually I think performance could generally be better if we use that instead of the static size. :-)

mitchcurtis:
Would that also work with tiledquickplugin, where the visible area isn't currently respected?

mitchcurtis:
>Actually the visible area is respected by the tiledquickplugin, through MapItem.visibleArea, which is used to determine TileLayerItem::mVisibleTiles (though only for orthogonal maps at the moment).
>
> But for isometric maps, using mLayer->bounds() is now more efficient than it was before the merge, since it will now return a rect that covers the area filled with tiles, or an empty area when a layer is empty.

Oh, so visibleArea does actually work for isometric tiles now, but through bounds() instead, or? I'm just making sure because I had it in the back of my mind as something that I'd need to make work when I start making bigger maps, but if it's already done, that's awesome!

bjorn:
> Oh, so visibleArea does actually work for isometric tiles now, but through bounds() instead, or? I'm just making sure because I had it in the back of my mind as something that I'd need to make work when I start making bigger maps, but if it's already done, that's awesome!

No, currently `visibleArea` only does something useful for orthogonal maps, but `TileLayer::bounds()` now returns a bounding rect of only the actually used area of the layer, rather than the entire layer. So that still benefits also isometric maps, but only speeds up processing of empty layers or layers of which only a small part is used.

That said, it should not be too hard to optimize isometric maps to use the visible area as well.

PR #1947, opened 2018-05-11
Allow qrc-based tileset images

mitchcurtis:
While irrelevant for the end users of Tiled as an application, this
allows those of us using Tiled as a library to have maps whose tilesets
are stored in resources.


bjorn:
Looks good, thanks!

PR #1954, opened 2018-05-20
Support for PEP384 + separate user script dir

stt:
Todo: assumes `scripts` directory is under QCoreApplication::applicationDirPath(), if this is ok packaging scripts would need to be updated.

Actually for me doing `make install INSTALL_ROOT=/tmp/tiled-pkg` doesn't copy any of the examples etc directories to INSTALL_ROOT that dist/distribute.qbs seems to indicate should be copied in the package?

### DIFF
--- a/src/plugins/python/pythonplugin.cpp
+++ b/src/plugins/python/pythonplugin.cpp
@@ -1,6 +1,6 @@
 /*
  * Python Tiled Plugin
- * Copyright 2012-2013, Samuli Tuomola <samuli@tuomola.net>
+ * Copyright 2012-2018, Samuli Tuomola <samuli@tuomola.net>
  *
  * This file is part of Tiled.
  *
@@ -18,6 +18,7 @@
  * this program. If not, see <http://www.gnu.org/licenses/>.
  */
 
+#define Py_LIMITED_API 0x03040000
 #include "pythonplugin.h"
 
 #include "map.h"
@@ -25,6 +26,7 @@
 #include <QDateTime>
 #include <QDir>
 #include <QDirIterator>
+#include <QCoreApplication>
 
 namespace Python {
 
@@ -40,7 +42,8 @@ static void handleError()
 }
 
 PythonPlugin::PythonPlugin()
-    : mScriptDir(QDir::homePath() + "/.tiled")
+    : mSysScriptDir(QCoreApplication::applicationDirPath() + "/scripts")
+    , mUserScriptDir(QDir::homePath() + "/.tiled")
     , mPluginClass(nullptr)
 {
     mReloadTimer.setSingleShot(true);
@@ -52,7 +55,7 @@ PythonPlugin::PythonPlugin()
             this, [this] { mReloadTimer.start(); });
 
     connect(&mReloadTimer, &QTimer::timeout,
-            this, &PythonPlugin::reloadModules);
+            this, &PythonPlugin::reloadUserModules);
 }
 
 PythonPlugin::~PythonPlugin()
@@ -72,9 +75,9 @@ void PythonPlugin::initialize()
     addObject(&mLogger);
 
     if (!Py_IsInitialized()) {
-        // PEP370
-        Py_NoSiteFlag = 1;
-        Py_NoUserSiteDirectory = 1;
+        QString scriptPath = mSysScriptDir + ":" + mUserScriptDir + ":" + QString::fromWCharArray(Py_GetPath());
+        log(tr("Python paths: %1").arg(scriptPath));
+        Py_SetPath(scriptPath.toStdWString().c_str());
 
         PyImport_AppendInittab("tiled", PyInit_tiled);
         PyImport_AppendInittab("tiled.qt", PyInit_tiled);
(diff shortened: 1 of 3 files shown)


src/plugins/python/pythonplugin.cpp
```diff
@@ -18,13 +18,15 @@
  * this program. If not, see <http://www.gnu.org/licenses/>.
  */
 
+#define Py_LIMITED_API 0x03040000
```
bjorn: I think this explains why you didn't run into the errors I was running into. The whole plugin needs to compile with this define, not only this file! In particular, we need `pythonbind.cpp` to also take this define.

Rather than putting this here, I have added the following line to `python.qbs` (see fa939ef4b4e92a74771ce8426b68ab22fd79856a):

```qbs
cpp.defines: base.concat(["Py_LIMITED_API=0x03040000"])
```

bjorn:
> Actually for me doing make install INSTALL_ROOT=/tmp/tiled-pkg doesn't copy any of the examples etc directories to INSTALL_ROOT that dist/distribute.qbs seems to indicate should be copied in the package?

Indeed, the example files were never installed, though they are listed in `distribute.qbs` to make them part of the released archive (though this may only be used by the hidden 7-zip packages for Windows at the moment). In any case, I'd rather just get rid of the qmake files.

I think I preferred the contents of `_preload.py` to be in the C++ code though, and I guess we can still keep things that was if we switch to using `Py_CompileString` and then somehow execute / evaluate the compiled code.

src/plugins/python/pythonplugin.cpp
stt: Ah, I added the define in the Makefile but guess I never did `make clean` to force it to recompile the bind sources as well.
Now it does give those same errors, will check if there's something that can be done about them.

stt: yep it does indeed look like pybindgen would require many changes to generate stable ABI compatible code .. should’ve perhaps read past the PEP384 Abstract and/or realised that just removing one PyRun was way too easy. :)
The biggest hurdle seems to be the new dynamic type objects. There is however a tool that can automate the conversion to some extent http://svn.python.org/projects/python/branches/pep-0384/Tools/scripts/abitype.py ..I’ll give that a go some day next week when there’s a free moment.

stt:
Ok, having modified pythonbind.cpp to use dynamic type declarations (PyType_FromSpec) + few other changes, the plugin now compiles with -DPy_LIMITED_API, and fotf.py and mappy.py successfully load maps on mac + linux as they used to.
Obviously the file can't be regenerated anymore without losing the changes. I opened an issue with pybindgen project for adding stable API support, if no one picks it up in a day or two I'll look into developing that feature.

Re: _preload, I was thinking maybe there could be a feature for batch converting bunch of levels from custom binary formats to tmx and with that users might want to get any errors to stderr, having the logging redirection in a script gives them an option to customize it as they want to.

src/plugins/python/pythonplugin.cpp
```diff
@@ -102,35 +104,20 @@ void PythonPlugin::initialize()
...
+    reloadModules(mSysScriptDir);
+    reloadModules(mUserScriptDir);
+
+    if (QFile::exists(mUserScriptDir)) {
+        mFileSystemWatcher.addPath(mUserScriptDir);
 
-    if (QFile::exists(mScriptDir))
-        mFileSystemWatcher.addPath(mScriptDir);
+        connect(&mFileSystemWatcher, SIGNAL(directoryChanged(QString)),
+                &mReloadTimer, SLOT(start()));
```
bjorn: This connection is now already made in the constructor, so it can be removed here.

src/plugins/python/pythonplugin.cpp
```diff
@@ -144,20 +131,25 @@ void PythonPlugin::log(const QString &msg)
...
+void PythonPlugin::reloadUserModules()
+{
+    reloadModules(mUserScriptDir);
+}
+
 /**
  * (Re)load modules in the script directory
  */
-void PythonPlugin::reloadModules()
+void PythonPlugin::reloadModules(QString path)
```
bjorn: Please use `const QString &path` to avoid unnecessary reference counting.

src/plugins/python/pythonplugin.cpp
```diff
@@ -40,7 +41,8 @@ static void handleError()
 }
 
 PythonPlugin::PythonPlugin()
-    : mScriptDir(QDir::homePath() + "/.tiled")
+    : mSysScriptDir(QCoreApplication::applicationDirPath() + "/scripts")
```
bjorn: Probably would prefer to call this `mAppScriptDir`.

Unfortunately, initializing it like this will only work on Windows. Check out the `LanguageManager` constructor for how we could make it work on macOS and Linux (and of course, install rules will need to be written for this - see `translations.qbs`).

src/plugins/python/pythonbind.cpp
```diff
@@ -10,7 +10,7 @@
 
 
 #if PY_VERSION_HEX >= 0x03000000
-#if PY_VERSION_HEX >= 0x03050000
+#if PY_VERSION_HEX >= 0x03050000 && 0   // <- TODO: this is not Py_LIMITED_API compatible?
```
bjorn: `PyAsyncMethods` is not available in the limited API, so that check should probably be:
```c++
#if PY_VERSION_HEX >= 0x03050000 && !defined(Py_LIMITED_API)
```

Issue #1322, opened 2016-07-25
Allow setting collisions for multiple tiles at once


adsilcott:
Often a large number of tiles on a sheet will have exactly the same collision boundaries and properties (basic walls, platforms, etc). It would save a lot of time if collisions could be applied to multiple selected tiles in the Tile Collision Editor. 

This would likely include a popup to warn the user that individual tile collision assignments would be overwritten. 

Alternatively, maybe the default behavior could be to "add" the new collision to to the tiles rather than overwriting them, along with an option to clear the previous assignments from the selected tiles. This could work well if the dialog displayed all of the selected tile's collisions superimposed on each other (uneditable unless they all matched). Not sure about the usefulness versus complexity of this approach. 

As discussed in this issue:
https://github.com/bjorn/tiled/issues/5#issuecomment-234915789

bjorn:
Also related to issue #1281, which suggests to move the collision editing out of a dialog and instead allow editing collision shapes of all tiles side-by-side.

The two approaches seem not entirely compatible, unless maybe we would move to a mode where all collision info is displayed side-by-side, but there is still a separate view for editing the shapes, which would modify the info on all selected tiles.

adsilcott:
The nice thing about the separate view approach that you mentioned, is that you could have a zoomed in view for detailed editing of the collisions, while having a zoomed out view showing you the overall tileset, which ones you've missed and which ones you are editing, and how they all fit together. I think in that way they could work well together.

Another thought: If you went with only the whole tilesheet view, then if you could create big collider objects across multiple tiles and have it split it up and assign it to all tiles that it touches individually, then if you grouped together all similar tiles on the tilesheet then you could assign them all at once, giving a similar functionality. For example put all solid tiles in one section and draw a big square across them, and group corner tiles together so you can just draw a diamond shape across all four of them, etc.

bjorn:
> Another thought: If you went with only the whole tilesheet view, then if you could create big collider objects across multiple tiles and have it split it up and assign it to all tiles that it touches individually, then if you grouped together all similar tiles on the tilesheet then you could assign them all at once, giving a similar functionality. For example put all solid tiles in one section and draw a big square across them, and group corner tiles together so you can just draw a diamond shape across all four of them, etc.

[...]

PR #1960, opened 2018-06-18
Apply collision to multiple tiles, autodetect tile extents

robinmacharg:
Adds two small collision mask editor features requested in issue #1322:

* A context menu entry that applies the selected mask(s) in the current tile (i.e. the last selected one) to all selected tiles.  Offers the user the choice of appending to existing masks or replacing existing ones.

* An additional tool on the collision mask editor toolbar that autodetects the (rectangular) extents of the tile and creates a collision mask automatically.  It uses the existing magic wand icon since the intent feels similar enough to the traditional image-related use.


src/tiled/tilecollisiondock.cpp
```diff
@@ -127,6 +136,51 @@ TileCollisionDock::~TileCollisionDock()
...
+                    left = x;
+                    x = right; // shortcut to only search for new right bound from here
+                }
+            }
+        }
+        if (rowFilled) {
+            top = std::min(top, y);
+            bottom = y;
+        }
+    }
```
bjorn: Though actually, after doing some digging in how the `QGraphicsPixmapItem` determines its shape, the following would do the same:
```c++
QRect content = QRegion(mTile->image().createHeuristicMask()).boundingRect();
```
It does involve the allocation of a temporary `QBitmap` and the construction of a `QRegion`, but I think it will not matter much for performance and it keeps this code a lot shorter.

src/tiled/tilecollisiondock.cpp
```diff
@@ -127,6 +136,51 @@ TileCollisionDock::~TileCollisionDock()
...
+
+    // Create a group for collision objects if none exists
+    if (!mTile->objectGroup()) {
+        mTile->setObjectGroup(new ObjectGroup);
```
bjorn: You can't just directly modify the tile, since that way the operation cannot be undone.

For the undo system to work in the tile collision editor, the undo stack of the `mDummyMapDocument` should be used. This dummy map is guaranteed to already have an object layer (see `TileCollisionDock::setTile`). You'd only need to use something like the following to add the object:
```c++
    auto objectGroup = static_cast<ObjectGroup*>(mDummyMapDocument->map()->layerAt(1));
    auto command = new AddMapObjects(mDummyMapDocument, objectGroup, newObject);
    mDummyMapDocument->undoStack()->push(command);
```
That way the UI will also update automatically, since that is triggered by the undo command.

(Note that you'll need to rebase your changes, since `AddMapObjects` used to be called `AddMapObject` and could only add one object, but I've recently changed that.)

src/tiled/abstractobjecttool.cpp
```diff
@@ -454,6 +523,15 @@ void AbstractObjectTool::showContextMenu(MapObject *clickedObject,
     QAction *removeAction = menu.addAction(tr("Remove %n Object(s)", "", selectedObjects.size()),
                                            this, SLOT(removeObjects()));
 
+    // Allow the currently selected collision mask to be applied to all selected tiles in the tileset editor
+    if (auto document = DocumentManager::instance()->currentDocument()) {
+        if (auto tilesetDocument = qobject_cast<TilesetDocument*>(document)) {
+            if (tilesetDocument->selectedTiles().count() > 1) {
+                menu.addAction(tr("Apply Collision(s) to Selection"), this, SLOT(applyCollisionsToSelection()));
```
bjorn: This is a very hidden location in the UI and somewhat of a strange place also code-wise. Of course I understand the latter naturally followed from the wish to put this action in the context menu.

I think I'd rather see this action as a button alongside this magic wand button you've added to the tile collision editor. There it is easier to discover and we don't need this somewhat hacky check to see if the user is editing a tileset. Of course, it will be a nice challenge to consider a good icon and probably we need to give it an elaborate tool tip to explain the functionality.

src/tiled/tilecollisiondock.cpp
```diff
@@ -80,6 +81,13 @@ TileCollisionDock::TileCollisionDock(QWidget *parent)
     CreateObjectTool *polygonObjectsTool = new CreatePolygonObjectTool(this);
     CreateObjectTool *templatesTool = new CreateTemplateTool(this);
 
+    // Autodetection of tile extents
+    QIcon autoDetectMaskIcon(QLatin1String(":images/22x22/stock-tool-fuzzy-select-22.png"));
+    QAction *autoDetectMask = new QAction(this);
+    autoDetectMask->setText(tr("Detect Shape"));
```
bjorn: This icon and text really suggests a more advanced functionality than actually implemented. I think the icon should include a rectangle and the text should say "Autodetect Collision Box".

src/tiled/abstractobjecttool.cpp
```diff
@@ -246,6 +247,74 @@ void AbstractObjectTool::removeObjects()
...
+                msgBox.setText(tr("Should Collision Masks be appended to, or replace, selected tiles' masks?"));
+                msgBox.addButton(tr("Append"), QMessageBox::ActionRole);
+                QPushButton *overwriteBtn = msgBox.addButton(tr("Replace"), QMessageBox::ActionRole);
+                QPushButton *abortButton = msgBox.addButton(QMessageBox::Cancel);
+                msgBox.setDefaultButton(overwriteBtn);
+                msgBox.exec();
+
+                if (msgBox.clickedButton() != abortButton) {
+
+                    QList<Tile *>selectedTiles = tilesetDocument->selectedTiles();
```
bjorn: Since we're not going to change the list of selected tiles, this can be:
```c++
const auto &selectedTiles = tilesetDocument->selectedTiles();
```
Or even just:
```c++
for (Tile *tile : tilesetDocument->selectedTiles()) {
```

src/tiled/abstractobjecttool.cpp
```diff
@@ -246,6 +247,74 @@ void AbstractObjectTool::removeObjects()
...
+
+                    // Add each collision object to each selected tile, as long as the tile is not the current one
+                    for (Tile* tile : selectedTiles) {
+
+                        // Ignore the currently selected tile
+                        if (tile != currentTile) {
```
bjorn: To avoid writing a lot of code with large indentation, I prefer to this this check as follows:
```c++
if (tile == currentTile)
    continue;
```

src/tiled/abstractobjecttool.cpp
```diff
@@ -246,6 +247,74 @@ void AbstractObjectTool::removeObjects()
...
+ * all selected tiles.
+ */
+void AbstractObjectTool::applyCollisionsToSelection()
+{
+    // The selected Collision Masks
+    QList<MapObject*> selectedObjects = mapDocument()->selectedObjects();
+
+    if (auto document = DocumentManager::instance()->currentDocument()) {
+        if (auto tilesetDocument = qobject_cast<TilesetDocument*>(document)) {
+            if (Tile *currentTile = dynamic_cast<Tile *>(tilesetDocument->currentObject())) {
```
bjorn: Avoid indenting the whole function by doing early-outs for these basic sanity checks:
```c++
Tile *currentTile = dynamic_cast<Tile *>(tilesetDocument->currentObject());
if (!currentTile)
    return;
```

src/tiled/abstractobjecttool.cpp
```diff
@@ -246,6 +247,74 @@ void AbstractObjectTool::removeObjects()
...
+/**
+ * @brief AbstractObjectTool::applyCollisionsToSelection
+ *
+ * Add the selected collision masks for the currently selected tile - the last selected tile - to
+ * all selected tiles.
+ */
+void AbstractObjectTool::applyCollisionsToSelection()
+{
+    // The selected Collision Masks
+    QList<MapObject*> selectedObjects = mapDocument()->selectedObjects();
```
bjorn: Since we're not going to modify the list of selected objects, we can avoid the extra reference count (and detaching, as will happen when iterating the non-const implicit copy later on):
```c++
const QList<MapObject*> &selectedObjects = mapDocument()->selectedObjects();
```

src/tiled/abstractobjecttool.cpp
```diff
@@ -246,6 +247,74 @@ void AbstractObjectTool::removeObjects()
...
+                        // Ignore the currently selected tile
+                        if (tile != currentTile) {
+
+                            // If we should overwrite, then remove all objects prior to adding the new ones
+                            if (msgBox.clickedButton() == overwriteBtn) {
+                                if (tile->objectGroup()) {
+                                    for (MapObject *object : tile->objectGroup()->objects()) {
+                                        tile->objectGroup()->removeObject(object);
+                                    }
+                                    tile->objectGroup()->resetObjectIds();
```
bjorn: This is pointless after removing all objects. There's nothing left to reset.

src/tiled/abstractobjecttool.cpp
```diff
@@ -246,6 +247,74 @@ void AbstractObjectTool::removeObjects()
...
+                                }
+                            }
+
+                            // Create a group for collision objects if none exists
+                            if (!tile->objectGroup()) {
+                                tile->setObjectGroup(new ObjectGroup);
```
bjorn: We can't modify the tiles directly since that breaks the undo system. What we need to do here is to clone the existing object group (or create a new one, if there is none or if we should overwrite) and then push a `ChangeTileObjectGroup` undo command onto the `tilesetDocument->undoStack()`. See `TileCollisionDock::applyChanges` for an example (except we don't need to change `mApplyingChanges`).

src/tiled/abstractobjecttool.cpp
```diff
@@ -246,6 +247,74 @@ void AbstractObjectTool::removeObjects()
...
+                            }
+
+                            // Copy across the selected collision mask shapes
+                            for (MapObject* object : selectedObjects) {
+                                MapObject* newObject = new MapObject;
+                                newObject->copyPropertiesFrom(object);
+                                QPointF position = QPointF(object->position().x(), object->position().y());
+                                newObject->setPosition(position);
```
bjorn: This should do the trick as well:
```c++
MapObject *newObject = object->clone();
newObject->resetId();
```

src/tiled/abstractobjecttool.cpp
```diff
@@ -246,6 +247,74 @@ void AbstractObjectTool::removeObjects()
...
+                // Ask the user if they want to append or overwrite the collision mask
+                QMessageBox msgBox;
+                msgBox.setIcon(QMessageBox::Information);
+                msgBox.setWindowTitle(tr("Overwrite Collision Masks?"));
+                msgBox.setText(tr("Should Collision Masks be appended to, or replace, selected tiles' masks?"));
+                msgBox.addButton(tr("Append"), QMessageBox::ActionRole);
+                QPushButton *overwriteBtn = msgBox.addButton(tr("Replace"), QMessageBox::ActionRole);
+                QPushButton *abortButton = msgBox.addButton(QMessageBox::Cancel);
+                msgBox.setDefaultButton(overwriteBtn);
+                msgBox.exec();
```
bjorn: This prompt is going to get annoying fast. At the least it should only happen when any of the selected tiles already has some collision shapes. But if this action is moved to a button on the tool bar in the collision editor, then alternatively the button could also be turned into a forced menu button using `QToolButton::InstantPopup` that presents both options before triggering the action.

robinmacharg:
Thanks for the detailed feedback; it's really appreciated.  This is my
first time contributing to a C++ project so your comments will not only
help make Tiled better, but me as well.  Work's hectic for the next week or
two (contracting, deadlines, etc.  but nothing compared to an Always a
Deadline new child, I know this) but I'll work through the comments and get
back to you when I can.

Git remains a devil to me - amazingly powerful but with the potential for
idiot mistakes like mine; I'll try and fix this with the next PR.

Thanks,
Robin

bjorn:
I'm glad the detailed review is appreciated, makes it worth the time! :-)

> Work's hectic for the next week or
> two (contracting, deadlines, etc.  but nothing compared to an Always a
> Deadline new child, I know this) but I'll work through the comments and get
> back to you when I can.

Alright, there's no rush! I do hope to be able to release Tiled 1.2 in about three weeks, but this might be too optimistic anyway.

> Git remains a devil to me - amazingly powerful but with the potential for
> idiot mistakes like mine; I'll try and fix this with the next PR.

Ah, I remember giving up on Git when I tried it for the first time instead of Subversion. Only switched back to it a year later after also having used Mercurial, but since then I had never looked back and love its speed and flexibility. :-)

Be sure to force-push into this PR (just keep using your `ApplyCollisionToMultipleTiles` branch) instead of opening a new one, btw.

rosshadden:
This would be so very helpful! It's quite tedious adding the same collision shape to hundreds or thousands of tiles.

bjorn:
> It's quite tedious adding the same collision shape to hundreds or thousands of tiles.

While that is certainly true, this was really not the intended use-case for the tile collision editor. If you have some trivial collision shape (possibly just filling the whole tile with a rectangle), it would be much faster and convenient to set a custom property like "collides: true" on those tiles. If you select all the tiles you can set this custom property at once on all of them.

rosshadden:
That's certainly true. Good point. Silly me, I even used to set a collision boolean on tiles back like six years ago before Tiled had collision polygons -_-.

robinmacharg:
@bjorn Apologies for leaving this hanging.  I've gained some Git(hub)-fu in the intervening couple of years and might even have some time to implement the changes you requested.  With the changes that have taken place in the last couple of years with Tiled is this proposed feature still valid and wanted/required?

bjorn:
@robinmacharg No problem about leaving it hanging and glad to hear you're still considering to finish this change! Indeed it is still relevant since no one else addressed #1322 so far.

To get you started I've just rebased your work onto the latest `master` branch (so be careful when updating your local checkout). I tried to make the changes minimal to make it compile, and will leave working on the feedback to you. :-)

robinmacharg:
@bjorn Thanks.  Glad to hear it's still potentially a useful addition, and thanks for the rebase.  New job, new laptop, new Qt install etc. so may take a couple of days to get back up to speed.

tliron:
@bjorn Any estimate as to when this feature will be merged and released? This remains the most painful aspect of using Tiled for me. :( I have lots of tilesets with many tile variants and creating collision for them takes many hours of work... This feature would be such a timesaver!

bjorn:
@tliron Since @robinmacharg appears to be busy elsewhere I can definitely consider picking this one up. I'll have a look on Tuesday how much work is remaining.

robinmacharg:
> @tliron Since @robinmacharg appears to be busy elsewhere...

You're very polite!  Yet another new job started today, following a fairly intense 6 months in the last one.  I can only apologise; I had the best of intentions but simply didn't have (and likely won't have), the time to follow this up.

tliron:
@robinmacharg You've created a very popular tool here! Thank you! Do you think it might be best to look for a new maintainer if you don't have the time?

bjorn:
I've rebased the changes to the latest `master` and addressed all the minor comments (and added [this answer](https://stackoverflow.com/a/63010657/355419) to Stack Overflow regarding finding the bounding box of an image). However, there are a few bigger issues I had pointed out that prevent this from being merged:

* Neither of the actions use the undo system.
* The icon and text for "Detect Shape" should be changed to reflect it's only detecting the bounding box (I've been considering whether we should add detection of convex polygon, since it seems not too hard and would be nice to have as well).
* The popup that asks whether the user would like to add to or replace the existing shapes on the selected tiles should be replaced by explicit actions to do one or the other. As I suggested before, it could be a good idea to move these actions from the context menu to the tool bar (though, in that case it will be less clear / intuitive that only the selected objects are used).

Apart from that, it would still be nice if you could just select multiple tiles and change all their collision shapes at once, but this is outside the scope of this change and adds the additional problem of handling the case where the selected tiles have different shapes set on them.

I'll look into the above issues next.

bjorn:
All done and will be available in today's development snapshot. It took me about 6 hours of work. Please considering [sponsoring me](https://github.com/sponsors/bjorn) if this feature is useful to you, to ensure I have the time to make similar improvements. Thanks!

tliron:
I wanted to test this and built from master, but am getting "corrupt data layer" errors for my project. I guess some formats have changed since the last release?

bjorn:
@tliron Hmm, what version were you using before and which file format are you using? I think there actually isn't any change that would be expected to cause such errors. Can you still open the files fine in the previous version you were using?

Note that you don't need to build it yourself. If you install the latest "snapshot" build on itch.io (see [this news post](https://thorbjorn.itch.io/tiled/devlog/163796/improvements-to-tmxrasterizer-collision-editor-and-more)) then you should see these new actions.

src/tiled/abstractobjecttool.cpp
bjorn: I decided to keep the actions in the context menu, because of the challenges of making them part of the tool bar, including finding a good icon as well as making sure the user understands that only selected objects will be copied.

tliron:
@bjorn Thanks! The snapshot on itch.io works fine with my existing files and also has the new features. I'm not sure why my self-built version was different, possibly different libraries on my OS (Fedora 32).

PR #1969, opened 2018-07-02
Include columns in lua export

MDrollette:
`columns` field was not being included.


bjorn:
This property is not generally needed since it can be derived from the tile size, margin, spacing and image width, but there is no harm in including it. Thanks for your patch!

Tiled mostly has this property since it's an editable property for image-collection tilesets and to be able to adjust tile indices in case the image width is changed (for example to add more tiles to it).

PR #1992, opened 2018-08-19
Display full path in recent files menu

gogoprog:
This is useful when different files have the same name

### DIFF
--- a/src/tiled/mainwindow.cpp
+++ b/src/tiled/mainwindow.cpp
@@ -474,6 +474,7 @@ MainWindow::MainWindow(QWidget *parent, Qt::WindowFlags flags)
                  this, &MainWindow::openRecentFile);
     }
     mUi->menuRecentFiles->insertSeparator(mUi->actionClearRecentFiles);
+    mUi->menuRecentFiles->setToolTipsVisible(true);
 
     setThemeIcon(mUi->menuNew, "document-new");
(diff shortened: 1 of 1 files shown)


bjorn:
Thanks for the patch! But let's use http://doc.qt.io/qt-5/qmenu.html#toolTipsVisible-prop, and set the full path on the tool tip, to avoid the menu becoming very wide in some cases. Usually it's not very common to have many times the same file name.

gogoprog:
Thanks for reviewing,
here are my new changes

src/tiled/mainwindow.cpp
```diff
@@ -1328,6 +1329,7 @@ void MainWindow::updateRecentFilesMenu()
         mRecentFiles[i]->setText(QFileInfo(files[i]).fileName());
         mRecentFiles[i]->setData(files[i]);
         mRecentFiles[i]->setVisible(true);
+        mRecentFiles[i]->setToolTip(QFileInfo(files[i]).filePath());
```
bjorn: Just a small suggestion: please re-use the QFileInfo instance. Not that the code is performance-critical, but in general it's good to avoid duplicating work.

gogoprog: Sure, I just updated

bjorn:
Nice addition, thanks!

PR #2019, opened 2018-10-10
GH1976 Labels change color when layer color is changed

### LINKED ISSUE #1976: Label color not updated when changing the color of an object layer
Changing the color of an object layer can affect the color of object name labels, but the labels are not updated as necessary in this case.

JustinZhengBC:
Closes #1976.

Adds new slot to `ObjectSelectionItem` that listens for `objectGroupChanged` from `MapDocument` (a signal that is triggered whenever the layer color is changed) and updates color labels when a signal is received.

### DIFF
--- a/src/tiled/objectselectionitem.cpp
+++ b/src/tiled/objectselectionitem.cpp
@@ -272,6 +272,9 @@ ObjectSelectionItem::ObjectSelectionItem(MapDocument *mapDocument,
     connect(mapDocument, &MapDocument::objectsChanged,
             this, &ObjectSelectionItem::syncOverlayItems);
 
+    connect(mapDocument, &MapDocument::objectGroupChanged,
+            this, &ObjectSelectionItem::objectGroupChanged);
+
     connect(mapDocument, &MapDocument::hoveredMapObjectChanged,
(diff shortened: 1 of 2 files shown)


src/tiled/objectselectionitem.cpp
```diff
@@ -420,6 +423,10 @@ void ObjectSelectionItem::layerChanged(Layer *layer)
     }
 }
 
+void ObjectSelectionItem::objectGroupChanged(ObjectGroup * objectGroup) {
```
bjorn: Coding style: please put the `{` on a new line for function bodies.

src/tiled/objectselectionitem.cpp
```diff
@@ -420,6 +423,10 @@ void ObjectSelectionItem::layerChanged(Layer *layer)
...
+    updateObjectLabelColors();
```
bjorn: In fact, since this function does not do anything else than call `updateObjectLabelColors`, you can connect the `objectGroupChanged` signal directly to that function.

JustinZhengBC:
@bjorn thanks for the suggestions, the code is a lot simpler now

bjorn:
Thanks! :-)

PR #2021, opened 2018-10-11
Add missing include.

kensington:
Under certain conditions, this seems to be causing a [build failure](https://667000.bugs.gentoo.org/attachment.cgi?id=547810) with qt-5.11.

### DIFF
--- a/src/tiled/newsfeed.cpp
+++ b/src/tiled/newsfeed.cpp
@@ -22,6 +22,7 @@
 
 #include "preferences.h"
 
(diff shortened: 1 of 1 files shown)


bjorn:
I wonder what these conditions are, since I didn't ran into it myself even when compiling against Qt 5.11 (on Arch Linux, with either system Qt 5.11.2 or downloaded Qt 5.11.1).

But it is appropriate to include this header given the use of `qWarning`, so thanks for the fix!

Issue #2014, opened 2018-10-02
Deleting command from context menu crashes the application


cdefgah:
Bug report. When you try to delete a command from context menu (right mouse click on the command), the application crashes.

To reproduce the issue:
1. Add a command via menu: File -> Commands -> Edit Commands
2. Close "Edit Commands" dialog
3. Open commands dialog again: File -> Commands -> Edit Commands
4. Select previously added command and make right click by mouse, to invoke the context menu
5. Select "Delete" for the selected command

Environment:

OS: Windows 10 64-bit
Application version: 1.2.0

Check the attached gif to view the issue in action:

![1](https://user-images.githubusercontent.com/10169696/46372618-5ba80e80-c694-11e8-985f-0db4fc620b41.gif)

rlewicki:
After quick investigation  it seems that it occurs when list of commands is empty after performing delete action. Reopening dialog windows is not necessary to reproduce effect. I'll make further investigation and try to fix it.

rlewicki:
It seems that application is recieving `SIGSEGV` when calling remove rows implementation at `CommandDataModel::removeRows(...)`, at the `beginRemoveRows(...)` function call to be exact. Not sure what is origin of this, since all values seems to be correct. Does any one know what might be going on there?

PR #2022, opened 2018-10-11
Fixed crash on remove command attempt (#2014)

rlewicki:
Fixed crash caused by badly calculated argument in beginRemoveRows function call that led to SIGSEGV when attepting to remove command and only one left in list


bjorn:
What a great find, thanks for tracking this down!

PR #2025, opened 2018-10-17
GH1996 tool probabilities update with tileset

### LINKED ISSUE #1996: Changed tile probability does not affect brush until the stamp is changed
There are a few occasions upon which the `RandomPicker<Cell>` used by the stamp brush is updated, but changing the probability of a tile is not one of them.

Of course, since it would be silly to update it each time a probability changes, the change should only invalidate this "cache" such that it is updated when the stamp brush is used again.

Problem reported [on the forum](https://discourse.mapeditor.org/t/automapping-works-just-once/418/5).

JustinZhengBC:
Closes #1996.

A signal for tile probability changes is added to `MapDocument` and triggered by a probabilty update in the tileset editor. `MapEditor` connects the signal from `MapDocument` to the update function in the brush. `StampBrush::updateRandomList` is made a public method to make the connection path possible.

### DIFF
--- a/src/tiled/abstracttiletool.cpp
+++ b/src/tiled/abstracttiletool.cpp
@@ -40,7 +40,6 @@ AbstractTileTool::AbstractTileTool(const QString &name,
                                    BrushItem *brushItem,
                                    QObject *parent)
     : AbstractTool(name, icon, shortcut, parent)
-    , mRandomCacheValid(false)
     , mTilePositionMethod(OnTiles)
     , mBrushItem(brushItem)
     , mBrushVisible(false)
@@ -106,16 +105,8 @@ void AbstractTileTool::mouseMoved(const QPointF &pos, Qt::KeyboardModifiers)
 void AbstractTileTool::mapDocumentChanged(MapDocument *oldDocument,
                                           MapDocument *newDocument)
 {
+    Q_UNUSED(oldDocument)
     mBrushItem->setMapDocument(newDocument);
-
-    if (oldDocument) {
-        disconnect(oldDocument, &MapDocument::tileProbabilityChanged,
-                   this, &AbstractTileTool::invalidateRandomCache);
-    }
-    if (newDocument) {
-        connect(newDocument, &MapDocument::tileProbabilityChanged,
-                this, &AbstractTileTool::invalidateRandomCache);
-    }
 }
 
 void AbstractTileTool::updateEnabledState()
@@ -236,9 +227,4 @@ void AbstractTileTool::setBrushVisible(bool visible)
     mBrushVisible = visible;
     updateStatusInfo();
     updateBrushVisibility();
-}
-
-void AbstractTileTool::invalidateRandomCache()
-{
-    mRandomCacheValid = false;
 }
--- a/src/tiled/abstracttiletool.cpp
+++ b/src/tiled/abstracttiletool.cpp
@@ -40,6 +40,7 @@ AbstractTileTool::AbstractTileTool(const QString &name,
                                    BrushItem *brushItem,
                                    QObject *parent)
     : AbstractTool(name, icon, shortcut, parent)
+    , mRandomCacheValid(false)
     , mTilePositionMethod(OnTiles)
     , mBrushItem(brushItem)
     , mBrushVisible(false)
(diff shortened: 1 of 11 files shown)


src/tiled/mapdocument.h
```diff
@@ -299,6 +299,8 @@ class MapDocument : public Document
 
     void tileLayerChanged(TileLayer *layer, TileLayerChangeFlags flags);
 
+    void tileProbabilityChanged(Tile * tile);
```
bjorn:
```suggestion
    void tileProbabilityChanged(Tile *tile);
```

src/tiled/tilesetdocument.cpp
```diff
@@ -373,6 +375,11 @@ void TilesetDocument::onTerrainRemoved(Terrain *terrain)
         setCurrentObject(nullptr);
 }
 
+void TilesetDocument::onTileProbabilityChanged(Tile *tile) {
```
bjorn:
```suggestion
void TilesetDocument::onTileProbabilityChanged(Tile *tile)
{
```

src/tiled/mapeditor.cpp
```diff
@@ -404,6 +404,8 @@ void MapEditor::setCurrentDocument(Document *document)
     if (mapDocument) {
         connect(mapDocument, &MapDocument::currentLayerChanged,
                 this, &MapEditor::updateLayerComboIndex);
+        connect(mapDocument, &MapDocument::tileProbabilityChanged,
+                mStampBrush, &StampBrush::updateRandomList);
```
bjorn: It is really the responsibility of the stamp brush itself to update its random list when needed. The current map document is already available to the stamp brush, so it could make this connection itself (and keep `updateRandomList` private).

Also, please note what I wrote at #1996 about only "invalidating" the random picker when a tile probability changes rather than immediately updating it.

JustinZhengBC:
@bjorn I've made the requested code style changes. As for avoiding `MapEditor` in the connection path, I couldn't access the map document directly from `TileSetDocument` because the getter returns null during the constructor, so I put the relevant connection in `AbstractTileTool`. The paint tool suffers from the same bug, so a tool-wide change could be justified. I've also used a bool to track whether the cache is invalid, and `StampBrush` checks it only when painting. I had to change a few other lines in `StampBrush` to force a new choice in the preview on painting if the cache is invalid, otherwise the preview might be left with a tile that had its probability recently set to zero, and drawn to the map.

src/tiled/stampbrush.cpp
```diff
@@ -337,6 +337,12 @@ void StampBrush::doPaint(int flags, QHash<TileLayer*, QRegion> *paintedRegions)
     if (!preview)
         return;
 
+    if(!mRandomCacheValid) {
+        updateRandomList();
+        mRandomCacheValid = true;
+        updatePreview();
```
bjorn: I think it is wrong for `doPaint` to call `updatePreview`. Maybe it would be better to check `mRandomCacheValid` in `updatePreview` instead?

src/tiled/abstracttiletool.cpp
```diff
@@ -40,6 +40,7 @@ AbstractTileTool::AbstractTileTool(const QString &name,
                                    BrushItem *brushItem,
                                    QObject *parent)
     : AbstractTool(name, icon, shortcut, parent)
+    , mRandomCacheValid(false)
```
bjorn: I'm not sure about putting this bool here, since not all tile tools have a random cache (actually only the `AbstractTileFillTool` and `StampBrush` keep around their `RandomPicker` instance). I think I'd currently prefer only those classes to include the bool and live with a bit of code duplication for now. Note that `mapDocumentChanged` is virtual so they can override it for connecting to the `tileProbabilityChanged` signal.

src/tiled/stampbrush.cpp
```diff
@@ -582,6 +588,7 @@ void StampBrush::updatePreview(QPoint tilePos)
             break;
         case Line:
         case Free:
+            drawPreviewLayer(QVector<QPoint>() << tilePos);
```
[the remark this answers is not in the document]
JustinZhengBC: My bad, that line turned out to be unnecessary

src/tiled/abstracttiletool.cpp
JustinZhengBC: I'm moved it to `StampBrush`. After my implementation there is approved I can do something similar to the `AbstractTileFillTool`.

src/tiled/stampbrush.cpp
JustinZhengBC: Ok. I put it in `drawPreviewLayer` so it's only checked when the stamp is being used randomly

src/tiled/stampbrush.cpp
```diff
@@ -56,6 +56,7 @@ StampBrush::StampBrush(QObject *parent)
     , mIsRandom(false)
     , mIsWangFill(false)
     , mWangSet(nullptr)
+    , mRandomCacheValid(false)
```
bjorn: I believe the value should initially be `true`. Since `mStamp` will be an empty stamp, the initially empty random list is valid.

bjorn:
Works great, thanks!

PR #2029, opened 2018-10-31
GH1997 Add QFileSystemWatcher for rules.txt

### LINKED ISSUE #1997: Automapping rules not reloaded when changed
Changing the `rules.txt` file does not cause it to be reloaded. It would probably be a good idea to use `QFileSystemWatcher` to watch for changes to this file as well as to any loaded maps, and set `mLoaded` to `false` when a change is detected.

Currently the rules are reloaded when switching maps. This could be changed to only reload when the new map uses a different `rules.txt` file.

Problem reported [on the forum](https://discourse.mapeditor.org/t/automapping-works-just-once/418).

JustinZhengBC:
closes #1997 

Automatic update of mapping rules whenever rules.txt is changed, and rules are only reloaded when switching maps if the maps are in different directories

### DIFF
--- a/src/tiled/automappingmanager.cpp
+++ b/src/tiled/automappingmanager.cpp
@@ -42,11 +42,13 @@ AutomappingManager::AutomappingManager(QObject *parent)
     , mMapDocument(nullptr)
     , mLoaded(false)
 {
+    mWatcher = new QFileSystemWatcher;
 }
 
 AutomappingManager::~AutomappingManager()
 {
     cleanUp();
+    delete mWatcher;
 }
 
 void AutomappingManager::autoMap()
@@ -94,8 +96,7 @@ void AutomappingManager::autoMapInternal(const QRegion &where,
     const bool automatic = touchedLayer != nullptr;
 
     if (!mLoaded) {
-        const QString mapPath = QFileInfo(mMapDocument->fileName()).path();
-        const QString rulesFileName = mapPath + QLatin1String("/rules.txt");
+        const QString rulesFileName = getRulesFileName();
         if (loadFile(rulesFileName)) {
             mLoaded = true;
         } else {
@@ -205,21 +206,53 @@ bool AutomappingManager::loadFile(const QString &filePath)
 void AutomappingManager::setMapDocument(MapDocument *mapDocument)
 {
     cleanUp();
-    if (mMapDocument)
+
+    QString oldRules;
+    if (mMapDocument) {
         mMapDocument->disconnect(this);
+        oldRules = getRulesFileName();
+        QFileInfo oldFile(oldRules);
+        if (oldFile.exists()) {
+            mWatcher->disconnect(SIGNAL(fileChanged(const QString &)));
+            mWatcher->removePath(oldRules);
+        }
+    }
 
     mMapDocument = mapDocument;
 
+    QString newRules;
     if (mMapDocument) {
         connect(mMapDocument, &MapDocument::regionEdited,
                 this, &AutomappingManager::onRegionEdited);
+        newRules = getRulesFileName();
+        QFileInfo newFile(newRules);
+        if (newFile.exists()) {
+            mWatcher->addPath(newRules);
+            connect(mWatcher, SIGNAL(fileChanged(const QString &)),
+                this, SLOT(onFileChanged()));
+        }
     }
 
-    mLoaded = false;
+    if (QString::compare(newRules, oldRules) != 0) {
+        mLoaded = false;
+    }
 }
 
 void AutomappingManager::cleanUp()
 {
     qDeleteAll(mAutoMappers);
     mAutoMappers.clear();
+}
+
+void AutomappingManager::onFileChanged()
+{
+    mLoaded = false;
+    cleanUp();
+}
+
+QString AutomappingManager::getRulesFileName()
+{
+    const QString mapPath = QFileInfo(mMapDocument->fileName()).path();
+    const QString rulesFileName = mapPath + QLatin1String("/rules.txt");
+    return rulesFileName;
 }
--- a/src/tiled/automappingmanager.cpp
+++ b/src/tiled/automappingmanager.cpp
@@ -29,6 +29,7 @@
 
 #include <QFileInfo>
 #include <QTextStream>
+#include <QFileSystemWatcher>
 
 #include "qtcompat_p.h"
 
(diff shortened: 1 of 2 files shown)


src/tiled/automappingmanager.h
```diff
@@ -134,6 +136,10 @@ private slots:
      * behavior.
      */
     QString mWarning;
+
+    QFileSystemWatcher *mWatcher;
+
+    QString getRulesFileName();
```
bjorn: Naming style: please call this function simply `rulesFileName`. Please also mark it `const`.

src/tiled/automappingmanager.cpp
```diff
@@ -205,21 +206,53 @@ bool AutomappingManager::loadFile(const QString &filePath)
 void AutomappingManager::setMapDocument(MapDocument *mapDocument)
 {
     cleanUp();
```
bjorn: If you only set `mLoaded` to `false` when the rules file changes, then I don't think you can call `cleanUp` unconditionally. Probably it would be good to set `mLoaded` to `false` in `cleanUp` and to call that when the rules files changes.

src/tiled/automappingmanager.cpp
```diff
@@ -205,21 +206,53 @@ bool AutomappingManager::loadFile(const QString &filePath)
...
+            mWatcher->addPath(newRules);
+            connect(mWatcher, SIGNAL(fileChanged(const QString &)),
+                this, SLOT(onFileChanged()));
+        }
     }
 
-    mLoaded = false;
+    if (QString::compare(newRules, oldRules) != 0) {
+        mLoaded = false;
+    }
```
bjorn: Coding style: generally leave out the braces when there is only one line.

src/tiled/automappingmanager.cpp
```diff
@@ -205,21 +206,53 @@ bool AutomappingManager::loadFile(const QString &filePath)
...
+        QFileInfo newFile(newRules);
+        if (newFile.exists()) {
+            mWatcher->addPath(newRules);
+            connect(mWatcher, SIGNAL(fileChanged(const QString &)),
+                this, SLOT(onFileChanged()));
+        }
     }
 
-    mLoaded = false;
+    if (QString::compare(newRules, oldRules) != 0) {
```
bjorn: Can just be `newRules != oldRules`.

src/tiled/automappingmanager.cpp
```diff
@@ -205,21 +206,53 @@ bool AutomappingManager::loadFile(const QString &filePath)
...
+    QString newRules;
     if (mMapDocument) {
         connect(mMapDocument, &MapDocument::regionEdited,
                 this, &AutomappingManager::onRegionEdited);
+        newRules = getRulesFileName();
+        QFileInfo newFile(newRules);
+        if (newFile.exists()) {
+            mWatcher->addPath(newRules);
+            connect(mWatcher, SIGNAL(fileChanged(const QString &)),
+                this, SLOT(onFileChanged()));
```
bjorn: We don't need to connect/disconnect to this signal all the time. It can be made once in the constructor instead, and the disconnect will happen automatically.

Also, please use the new connect style:
```c++
    connect(mWatcher, &QFileSystemWatcher::fileChanged,
            this, &AutomappingManager::onFileChanged);
```

src/tiled/automappingmanager.cpp
```diff
@@ -205,21 +206,53 @@ bool AutomappingManager::loadFile(const QString &filePath)
...
 
     mMapDocument = mapDocument;
 
+    QString newRules;
     if (mMapDocument) {
         connect(mMapDocument, &MapDocument::regionEdited,
                 this, &AutomappingManager::onRegionEdited);
+        newRules = getRulesFileName();
+        QFileInfo newFile(newRules);
+        if (newFile.exists()) {
```
bjorn: Please merge the above two lines using the static helper function:
```c++
        if (QFile::exists(newRules)) {
```

src/tiled/automappingmanager.cpp
```diff
@@ -205,21 +206,53 @@ bool AutomappingManager::loadFile(const QString &filePath)
...
+
+    QString oldRules;
+    if (mMapDocument) {
         mMapDocument->disconnect(this);
+        oldRules = getRulesFileName();
+        QFileInfo oldFile(oldRules);
+        if (oldFile.exists()) {
+            mWatcher->disconnect(SIGNAL(fileChanged(const QString &)));
+            mWatcher->removePath(oldRules);
+        }
```
bjorn: When we move the connect to the constructor, we can unconditionally call `removePath`:
```c++
mWatcher->removePath(rulesFileName());
```
Since it does not matter whether the file exists or not.

We can also do the `removePath`/`addPath` only when the rules file changes.

JustinZhengBC:
@bjorn I think I've made all your requested changes, except for the part about unconditionally calling `removePath`. I added a bit of logic and duplicated code for the sake of not having to reload `rules.txt` and all associated files in the case of switching documents in the same directory. Do you think it's worth the tradeoff? Or should the watcher definitely be cleared every time?

src/tiled/automappingmanager.cpp
```diff
@@ -204,22 +207,46 @@ bool AutomappingManager::loadFile(const QString &filePath)
...
-        connect(mMapDocument, &MapDocument::regionEdited,
-                this, &AutomappingManager::onRegionEdited);
+        newRules = rulesFileName();
+        if (newRules != oldRules) {
+            cleanUp();
+            if (!mWatcher.files().isEmpty())
+                mWatcher.removePaths(mWatcher.files());
+        }
+        if (QFile::exists(newRules))
+            mWatcher.addPath(newRules);
```
bjorn: It doesn't really make sense to watch the file if it hasn't been loaded yet, so I would move this to the `loadFile` function. 

In addition, that will fix the reload in case the user modifies any rules file that is referenced by the main rules file (it is possible to include another rules file in a rules file by just listing a `.txt` file).

Finally, I guess it would make sense to move the clearing of `mWatcher` to the `cleanUp` function. Once the loaded rules have been cleaned, it is no longer interesting to watch any files for changes after all.

JustinZhengBC:
@bjorn I've made the requested changes

bjorn:
Thanks, great change! I've decided to merge this into the `1.2` branch so that it can be included in an eventual Tiled 1.2.2 release.

PR #1925, opened 2018-04-01
Build updates to tiled-java

mikepthomas:
Update Tiled map schema, Java and dependency versions and file licences and ensure project can be opened and built in Eclipse

PR #2030, opened 2018-11-05
Fix @mikepthomas PR + remove unnecessary files

cdefgah:
+ Fixed Mike's PR, AppVeyor and Travis CI tests now green.
+ Removed unnecessary project files. The project can be imported to any modern IDE as a Maven project. All sufficient project information is encapsulated in pom.xml file(s).

Please note, that fixing the broken tests is a temporary solution. It would be better to redesign the whole functionality (in the all methods) that deals with file paths/urls. Now it is not platform independent, and on Windows may cause file not found exceptions, if path contains linux delimiters (accepted by Windows), but the code is trying to use platform specific File.separator char to split path and filename.


cdefgah:
I will redo this PR without removing unnecessary files. and will check getParent() method.

cdefgah:
getParent() did not worked for the current state of the code. 

I have reverted removal of unnecessary files, noted, that it is a temporary workaround. 

All changes were pushed to the master and should be available in the PR.

Now AppVeyor and Travis are working, and I think it will be ok, because on my computer all tests were green.

cdefgah:
@bjorn AppVeyor and Travis checks passed ok, please check the PR.

bjorn:
> getParent() did not work for the current state of the code.

Can you explain what problem you encountered?

bjorn:
Closing this since #1925 was fixed up as well and is now merged. Thanks @cdefgah for your help with this!

PR #2038, opened 2018-11-28
Add qtdeclarative5-dev to dependencies

iand:
Fixes the error `Unknown module(s) in QT: qml`


bjorn:
That's right, thanks!

Issue #249, opened 2012-07-29
Invert Y axis option


Beastifus:
Currently the tiling y-axis starts with the height at 0 at the top and incrementing as it goes down the map.   I think it would a be a nice feature, but relatively simple task, to add some kind of checkbox option to invert the Y axis so that (0,0) starts at the bottom and then (0, 1) is the tile above.   

Thanks!

jrmgx:
I'm much in favor of this feature request! :)

bjorn:
Also requested [on the forum](http://discourse.mapeditor.org/t/change-tiled-coordinate-system-to-left-bottom/1704).

ShipshapeDetour:
This feature would be very helpful.

bjorn:
> I think this is a good idea, rather than having to adjust the code in game it would make more sense just to have an option here. Then you wouldn't need to make boilerplate code just to invert the access in game for every object.

@davos555 I realize I'm a bit late with this reply, but I just thought I'd point out that I see this option as only a display option in Tiled. It would not affect the actual coordinates stored in a TMX file. So if you want to use the inverted coordinates in your game, you'll still have to invert them while or after loading the map. That is because if the TMX format would support storing inverted coordinates, then every loader out there would suddenly need to add such conversion code.

kibertoad:
@bjorn Why can't this conversion be applied transparently during the save process so that loaders do not have to do anything at all?.. Aren't coords actually saved?

ShipshapeDetour:
That's a sound idea, but the most important part for me is changing the y
coordinate in the status bar. It's much easier to implement.

bjorn:
> @bjorn Why can't this conversion be applied transparently during the save process so that loaders do not have to do anything at all?.. Aren't coords actually saved?

@kibertoad There are many loaders either integrated into various game development frameworks or provided by 3rd parties, that both read the map file and render the map. When an option to invert the Y axis in the TMX file is introduced, then in order to render the map correctly, they would all need to be aware of this option and invert the Y axis when appropriate while rendering.

I think it is better to keep a particular coordinate system on the file format level so that we are all on the same page. At the same time, I understand that some people are working with their coordinates inverted, and that it could help if Tiled could display them as such.

@ShipshapeDetour In addition to the status bar I think it'll need to affect the Properties view as well, but still you're right that this is less work than changing the interpretation of the data.

kibertoad:
[...]

PR #1897, opened 2018-03-01
Added Invert Y coordinate option to the Preferences dialog

divitto:
Added a helper class to allow users to choose whether or not to display the Y coordinates going top-down or bottom-up.

PR #2040, opened 2018-12-02
Added Y flip option to the coordinates

AdrianFL:
Added the possibility of flipping the Y coordinate, to set the origin at the bottom and not at the top, as discussed in the issue #249.

This doesn't change any internal behavior or logic, and only the visual representation of the numbers.


src/tiled/invertyaxishelper.h
```diff
@@ -0,0 +1,55 @@
...
+class InvertYAxisHelper{
+    public:
+        InvertYAxisHelper() {}
+        ~InvertYAxisHelper() {}
+        
+        // Inverts Y coordinate in grid
+        float getY(float y) const
+        {
+            // Check if Invert Y Axis is set
+            if(Tiled::Internal::Preferences::instance()->invertYAxis())
```
bjorn: Please put the helper class itself in the `Tiled::Internal` namespace, so we don't need these qualifications here.

src/tiled/invertyaxishelper.h
```diff
@@ -0,0 +1,55 @@
...
+            {
+                return Tiled::Internal::MapDocumentActionHandler::instance()->mapDocument()->map()->height() - 1 - y;
+            }
+            return y;
+        }
+
+        // Inverts Y coordinate in pixels
+        float getPixelY(float y) const
```
bjorn: Please use `qreal` (alias for `double`) instead of `float`.

src/tiled/invertyaxishelper.h
```diff
@@ -0,0 +1,55 @@
...
+#include "preferences.h"
+#include "mapdocumentactionhandler.h"
+
+class InvertYAxisHelper{
+    public:
+        InvertYAxisHelper() {}
+        ~InvertYAxisHelper() {}
+        
+        // Inverts Y coordinate in grid
+        float getY(float y) const
```
bjorn: Please call this `tileY`, to make it more clear that it's the function inverting tile coordinates (also in the Qt coding style we do not use a "get" prefix).

src/tiled/invertyaxishelper.h
```diff
@@ -0,0 +1,55 @@
...
+                return Tiled::Internal::MapDocumentActionHandler::instance()->mapDocument()->map()->height() - 1 - y;
+            }
+            return y;
+        }
+
+        // Inverts Y coordinate in pixels
+        float getPixelY(float y) const
+        {
+            // Obtain the map document
+            auto map = Tiled::Internal::MapDocumentActionHandler::instance()->mapDocument()->map();
```
bjorn: The helper can't simply assume that the currently selected map is the one it has to invert coordinates for. The `MapDocument*` will need to be provided, I think preferably to the constructor of `InvertYAxisHelper`.

src/tiled/propertybrowser.cpp
```diff
@@ -323,6 +327,12 @@ void PropertyBrowser::wangSetChanged(Tileset *tileset, int index)
         updateProperties();
 }
 
+void PropertyBrowser::invertYAxisChanged()
+{
+    if(mObject && mObject->typeId() == Object::MapObjectType)
```
bjorn: Coding style: please put a space after `if`.

src/tiled/invertyaxishelper.h
```diff
@@ -0,0 +1,55 @@
...
+ * You should have received a copy of the GNU General Public License along with
+ * this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "preferences.h"
+#include "mapdocumentactionhandler.h"
+
+class InvertYAxisHelper{
```
bjorn: Coding style: please put the `{` on its own line for class and function definitions.

src/tiled/invertyaxishelper.h
```diff
@@ -0,0 +1,55 @@
...
+    public:
```
bjorn: Coding style: please do not indent the access specifiers.

src/tiled/invertyaxishelper.h
```diff
@@ -0,0 +1,55 @@
...
+        InvertYAxisHelper() {}
+        ~InvertYAxisHelper() {}
```
bjorn: There's no point in declaring the trivial destructor here. The compiler will generate it automatically.

AdrianFL:
Thanks for the detailed feedback! I'll work through all of it and commit again with the changes made as soon as possible. I hope the feature gets merged this time too :)

Sorry for having so many coding style mistakes, I'll pay more attention to it next time.

src/tiled/invertyaxishelper.h
```diff
@@ -24,32 +24,41 @@
...
 
-class InvertYAxisHelper{
-    public:
-        InvertYAxisHelper() {}
-        ~InvertYAxisHelper() {}
-        
-        // Inverts Y coordinate in grid
-        float getY(float y) const
+namespace Tiled{
+    namespace Internal {
```
bjorn: Coding style: please don't indent for namespaces.

src/tiled/invertyaxishelper.h
```diff
@@ -24,32 +24,41 @@
...
+
+        class InvertYAxisHelper
         {
-            // Check if Invert Y Axis is set
-            if(Tiled::Internal::Preferences::instance()->invertYAxis())
+        public:
+            // Constructors
+            InvertYAxisHelper()  {}
+            InvertYAxisHelper(Map* m) : target(m) {}
```
bjorn: You'll need to take a `MapDocument` and not a bare `Map` (see your implementation of `tileY`, which still uses the `MapDocumentActionHandler`).

Coding style: please put the `*` against the variable name, so `MapDocument *target`.

src/tiled/invertyaxishelper.h
```diff
@@ -24,32 +24,41 @@
...
+                }
+                return y;
             }
-            return y;
-        }
 
-};
\ No newline at end of file
+        private:
+            Map* target;
```
bjorn: Coding style: please prefix member variables with `m`, so `mTarget`.

src/tiled/invertyaxishelper.h
```diff
@@ -24,32 +24,41 @@
...
+            InvertYAxisHelper()  {}
+            InvertYAxisHelper(Map* m) : target(m) {}
+                
+            // Inverts Y coordinate in grid
+            qreal tileY(qreal y) const
             {
-                return Tiled::Internal::MapDocumentActionHandler::instance()->mapDocument()->map()->height() - 1 - y;
+                // Check if Invert Y Axis is set
+                if(Preferences::instance()->invertYAxis())
+                {
```
bjorn: Coding style: please put the `{` on the same line as the `if` (and put a space after `if`):
```c++
if (Preferences::instance()->invertYAxis()) {
```

src/tiled/invertyaxishelper.h
```diff
@@ -24,32 +24,41 @@
...
-        }
 
-};
\ No newline at end of file
+        private:
+            Map* target;
+        };
+
+    } // Namespace Internal
+} // Namespace Tiled
```
bjorn: Please do make sure there is a newline at the end of every file.

src/tiled/propertybrowser.cpp
```diff
@@ -1618,7 +1618,7 @@ void PropertyBrowser::updateProperties()
         if (auto visibleProperty = mIdToProperty[VisibleProperty])
             visibleProperty->setValue(mapObject->isVisible());
         mIdToProperty[XProperty]->setValue(mapObject->x());
-        mIdToProperty[YProperty]->setValue(InvertYAxisHelper().getPixelY(mapObject->y()));
+        mIdToProperty[YProperty]->setValue(InvertYAxisHelper(MapDocumentActionHandler::instance()->mapDocument()->map()).pixelY(mapObject->y()));
```
bjorn: In `applyMapObjectValueTo`, you used `mMapDocument`. You should do the same here to make sure the right map is used.

src/tiled/invertyaxishelper.h
```diff
@@ -24,32 +24,41 @@
...
         {
-            // Check if Invert Y Axis is set
-            if(Tiled::Internal::Preferences::instance()->invertYAxis())
+        public:
+            // Constructors
+            InvertYAxisHelper()  {}
+            InvertYAxisHelper(Map* m) : target(m) {}
+                
+            // Inverts Y coordinate in grid
+            qreal tileY(qreal y) const
```
bjorn: I think using `int` would be better for the tile coordinates.

src/tiled/invertyaxishelper.h
```diff
@@ -24,32 +24,41 @@
...
+namespace Tiled{
+    namespace Internal {
+
+        class InvertYAxisHelper
         {
-            // Check if Invert Y Axis is set
-            if(Tiled::Internal::Preferences::instance()->invertYAxis())
+        public:
+            // Constructors
+            InvertYAxisHelper()  {}
```
bjorn: You should drop this constructor.

src/tiled/invertyaxishelper.h
AdrianFL: Thanks for the feedback, but i would like to understand the logic behind taking the `MapDocument `and not the `Map`. From my point of view, you are only going to use the Map, so what's the reason behind giving the whole `MapDocument` to the helper class?

Anyway I'll make the changes ASAP!

src/tiled/invertyaxishelper.h
AdrianFL: I maintained it to be able to use the class without providing a Map, like I'm doing at the tileY method. Should I give a map anyway?

bjorn: Yes, because actually `tileY` can't be implemented correctly without a map, and currently its gets the map from the `MapDocumentActionHandler`, which is not always referring to the right map.

src/tiled/invertyaxishelper.h
bjorn: Because just the map is not enough to know where the origin is for implementing `pixelY`. It is currently hardcoded to work for orthogonal maps, but it is returning the wrong thing for staggered or isometric maps. See https://github.com/bjorn/tiled/pull/1897/files#r174102431

AdrianFL:
Thanks for all the feedback provided, you are awesome :) I think I've finally understood how `MapDocument` and the `MapRenderer` work internally in the project thanks to it (I hope I can use that knowledge to be able to help in other issues in the future).

Regarding my last commit, I think I've implemented all the feedback, and after testing it I think the feature should be done, but, I have one question: You suggested to change the `tileY` function to use `int` instead of `qreal`. Should I change `pixelY` to use int too?

src/tiled/abstractobjecttool.cpp
```diff
@@ -160,7 +161,10 @@ void AbstractObjectTool::mouseMoved(const QPointF &pos,
     const QPointF tilePosF = mapDocument()->renderer()->screenToTileCoords(offsetPos);
     const int x = qFloor(tilePosF.x());
     const int y = qFloor(tilePosF.y());
-    setStatusInfo(QString(QLatin1String("%1, %2 (%3, %4)")).arg(x).arg(y).arg(pixelPos.x()).arg(pixelPos.y()));
+    setStatusInfo(QString(QLatin1String("%1, %2 (%3, %4)")).arg(x)
+                                                           .arg(InvertYAxisHelper(mapDocument()).tileY(y))
+                                                           .arg(pixelPos.x())
+                                                           .arg(InvertYAxisHelper(MapDocumentActionHandler::instance()->mapDocument()).pixelY(pixelPos.y())));
```
bjorn:
```suggestion
                                                           .arg(InvertYAxisHelper(mapDocument()).pixelY(pixelPos.y())));
```
You got it right for `tileY`, why not for `pixelY`? :-)

src/tiled/invertyaxishelper.h
```diff
@@ -0,0 +1,64 @@
...
+#pragma once
+
+#include "preferences.h"
+#include "mapdocumentactionhandler.h"
+#include "maprenderer.h"
+
+namespace Tiled {
+namespace Internal {
+
+    class InvertYAxisHelper
```
bjorn: Coding style: please do not indent for namespaces.

src/tiled/invertyaxishelper.h
```diff
@@ -0,0 +1,64 @@
...
+    {
+    public:
+        // Constructors
```
bjorn: Superfluous comment.

src/tiled/invertyaxishelper.h
```diff
@@ -0,0 +1,64 @@
...
+        InvertYAxisHelper(MapDocument *m) : mTarget(m) {}
+                
+        // Inverts Y coordinate in grid
+        int tileY(int y) const
+        {
+            // Check if Invert Y Axis is set
+            if (Preferences::instance()->invertYAxis()) {
+                const MapRenderer *renderer = mTarget->renderer();
+                return renderer->map()->height() - 1 - y;
```
bjorn: It's not necessary to involve the renderer here:
```suggestion
                return mTarget->map()->height() - 1 - y;
```

src/tiled/invertyaxishelper.h
```diff
@@ -0,0 +1,64 @@
...
+            return y;
+        }
+
+        // Inverts Y coordinate in pixels
+        qreal pixelY(qreal y) const
+        {
+            // Obtain the map document
+            if (Preferences::instance()->invertYAxis()) {
+                const MapRenderer *renderer = mTarget->renderer();
+                return (renderer->map()->height() * renderer->map()->tileHeight())- y;
```
bjorn: You're only using the renderer to get access to the map directly, which completely defeats the purpose of going through the renderer. The idea is to use the renderer, which abstracts the rendering of the map, to find out where the origin is instead of hardcoding `height * tileHeight` here, which only works for orthogonal maps.

It should look something like this:
```c++
QRect boundingRect = renderer->boundingRect(QRect(QPoint(), mTarget->map()->size()));
return boundingRect.height() - y;
```
But please test it.

src/tiled/preferences.cpp
```diff
@@ -683,6 +684,17 @@ void Preferences::setWheelZoomsByDefault(bool mode)
     mSettings->setValue(QLatin1String("Interface/WheelZoomsByDefault"), mode);
 }
 
+void Preferences::setInvertYAxis(bool enabled)
+{
+    if(mInvertYAxis == enabled)
```
bjorn: Coding style: please put a space after `if`.

AdrianFL:
> No. The reason to use int for the tile coordinates is because there don't exist sub-tile coordinates. But objects, which use pixel coordinates, do support sub-pixel positions as well.
Okey, didn't knew about sub-pixel positions, thanks for clarifying.

Added all the suggestion to the most recent commit :) I had to modify a bit the code suggested  for the boundingRect to make it work, I hope using the `Map` of the `MapDocument` to obtain the size it's fine, I think that's what you intended.

bjorn:
> I had to modify a bit the code suggested for the boundingRect to make it work, I hope using the `Map` of the `MapDocument` to obtain the size it's fine, I think that's what you intended.

I don't understand, because the code in the commit looks exactly the same as the code I suggested, or am I missing something?

Thanks for addressing all the comments! :-)

AdrianFL:
Well, when I copied that code fragment to my editor, for some reason instead of `mTarget->map()->size()` it appeared `mTarget->size()`, and I needed to add that `map()` call in order to compile. But looking at your code here, it seems it was my mistake copying, so sorry for that.

Well, I hope everything is fine now :)

AdrianFL:
Sorry to bother you, but I would like to ask if additional change is needed to this pull request, or if it is ready to be merged?

AdrianFL:
@bjorn Just to be sure I am understanding this correctly. What is happening is that the object's position is placed using the top-left corner as a reference, but besides that the inversion of the Y axis is working properly, right?

Well, in my opinion, that seems a problem that needs to be solved before merging the whole functionality of this PR, because it may lead to some confusion to the users if not. But depending on the approach, I think it could even be a completelly different functionality. 

Do you think users may be interested in choosing the corner which the object is aligned? Because in that case, it could even be a different customizable option, implemented separately from this one. I'm not sure about if this has any utility for users, but if we are going to do the work, we could assess if it's worth doing it.

Anyway, in my opinion, I think we should work on changing the objects corner-alignment before merging it. I can work on this, but I'll need some time to familiarize myself with the code better before doing any changes. Is it possible to see a class diagram or something similar to understand better of the code is structured?

elleyer:
Was this one added in some another PR?

AdrianFL:
> Was this one added in some another PR?

Hi @elleyer , unfortunately not, life got in the way and I never had the chance to properly work on it. I also don't think it has been merged in another PR, but I would be glad to help making the necessary core changes to get it done!

Jonatino:
What's left for this to be merged? This would be a huge QoL for me personally.

bjorn:
@Jonatino This PR got stuck on the issue I commented on at https://github.com/mapeditor/tiled/pull/2040#issuecomment-450242395. What's your take on this?

Jonatino:
@bjorn I would say that's definitely unexpected behaviour. I can imagine that changing the anchoring logic for all objects will be quite a bit of work if it was never designed to do so.

Albeit, personally I wouldn't say this is enough of a blocker to stop this PR from being merged. It's unexpected behaviour that humans would notice and be able to fix within seconds. Could always track it in another issue.

Having the option will be a nice quality of life improvement for people who use bottom-left aligned maps to be able to quickly see the x/y of a tile without needing to subtract the height manually. Very minor change which saves a lot of manual work.

vinnydiehl:
Could someone point me towards the object anchoring logic? I'd like to take a look at this soon.

bjorn:
Closing as superseded by #3679.

PR #2045, opened 2018-12-20
Throw an exception here

TheRamenChef:
Rather than printing a message to STDOUT and doing nothing, it would be much better if this method would instead throw an exception on illegal input.


bjorn:
I agree, thanks!

Issue #1856, opened 2018-01-13
Suggestion: Rearranging Image Collection by Image Name


hepari:
As a developer of an ongoing project, I have to add new images to an existing collection. 
The problem is that since the editor shows all images arranged by their id instead of the order of image names, my collection becomes messy very quickly once I start adding new objects.

**Example (current)**
id1: stone1
id2: window

**Now, if I add an image named 'stone2':**
id1: stone1
id2: window
id3: stone2

And the system gets lost right away.
That's why I would love to arrange the objects by name. Is it possible?

bjorn:
> That's why I would love to arrange the objects by name. Is it possible?

In this case, the order in the file is irrelevant and you're just looking for an option to order by name when displaying the tileset, right?

Unfortunately right now this is indeed not an option, but we anyway need a better tileset view that is more suitable for image collections. There should be a thumbnail view, where the tiles are displayed in a uniform size and there should also be an option to display the name. This new view should then definitely also be able to display the tiles sorted by name.

> This is a really pressing matter to me, and would be grateful for an answer

It is hard to say when I would get around to this. There are multiple other features on my priority list that I'm currently working on, so it will take at least some weeks. Of course, maybe somebody else would be interested in implementing this.

keshav2010:
I will try to devote some time on this issue  (although my first internal examination is going to start from 19nth of February) i will be digging up the source code and see how much of the requested feature i can implement.

hepari:
@keshav2010 : Thank you for showing interest in this issue. I actually meant the roster of objects under the 'tilesets' window! Right now, they are always sorted by id numbers which are assigned in the order of being added to the tileset file.

[ID-01] [ID-02] [ID-03]
[ID-04] [ID-05] ...

But I want them to be sorted automatically by their file name, like in the following example!

![capture-20180207001](https://user-images.githubusercontent.com/10383640/35944171-d276fbbe-0c5b-11e8-888c-1e23c092ee26.png)

I'd be really, really grateful if you could implement this one.

keshav2010:
Oh i see, so you basically want to re-arrange those individual tiles in a given Tileset.
From the build I've on my system, currently their isn't by default any Name property _**for each tile**_ in a Tileset and although you can sure add a Custom property for name but that wouldn't be really helpful since you might not be willing to assign a name to all tiles in a given tileset. 

[...]

PR #2046, opened 2018-12-20
Rearranging Image Collection by Image Name 

keshav2010:
This PR is related to issue #1856

### DIFF
--- a/src/libtiled/tile.cpp
+++ b/src/libtiled/tile.cpp
@@ -42,7 +42,9 @@ Tile::Tile(int id, Tileset *tileset):
     mProbability(1.0),
     mCurrentFrameIndex(0),
     mUnusedTime(0)
-{}
+{
+    mName = QString::number(mId);
+}
 
 Tile::Tile(const QPixmap &image, int id, Tileset *tileset):
     Object(TileType),
@@ -54,7 +56,9 @@ Tile::Tile(const QPixmap &image, int id, Tileset *tileset):
     mProbability(1.0),
     mCurrentFrameIndex(0),
     mUnusedTime(0)
-{}
+{
+    mName = QString::number(mId);
+}
 
 Tile::~Tile()
 {
--- a/src/libtiled/tile.h
+++ b/src/libtiled/tile.h
@@ -103,6 +103,7 @@ public:
     ~Tile();
 
     int id() const;
+    const QString &name() const;
 
     Tileset *tileset() const;
     QSharedPointer<Tileset> sharedTileset() const;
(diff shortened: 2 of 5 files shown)


hepari:
Can someone please take care of this? :) I don't understand that error message, but I need this functionality.

hepari:
@bjorn Could you have a look at this pull request please?

bjorn:
The main problem with this patch is that whenever some tiles in a collection tileset have the same name, only one of them will be displayed.

Personally I do not really understand the need for the name to be editable. In addition, the functionality could alternatively be implemented by using a different view for image collection tilesets (maybe QListView), which allows sorting the tiles by their image file name.

hepari:
@bjorn : I just need the image collection be sorted by image file names, and not in the order these files were added. I'm still using the old version because that's how important it is to me. :(

bjorn:
@hepari I'd love to help you but I am currently way too busy as I wrote about in my [latest post on Patreon](https://www.patreon.com/posts/final-home-24609471). Maybe I could have a look at this in March.

hepari:
@bjorn I will happily wait, though it would be cool if someone could take care of it in the meantime. Good luck with the move!

bjorn:
Now that the tiles can be manually reordered as of #2983, this change no longer really makes sense.

There could still be an option to sort the tiles by name, but it's a little unclear to me whether this should affect their stored order and how it would interact with manual reordering, so we'd need somebody with a use-case for it first.

Thanks @keshav2010 for your work in any case!

hepari:
> 
> 
> Now that the tiles can be manually reordered as of #2983, this change no longer really makes sense.
> 
> There could still be an option to sort the tiles by name, but it's a little unclear to me whether this should affect their stored order and how it would interact with manual reordering, so we'd need somebody with a use-case for it first.
> 
> Thanks @keshav2010 for your work in any case!

I'd say it should affect the stored order and overwrite it. But I think 'reorder by file name' should be only an option for image collections, not tilesets.
(I was actually surprised to find that I now have the option to manually re-order tiles in a tileset as I never thought about it.)

The manual reordering feature makes this new version at least light-years better than the old one, so thank you for that. :)

bjorn:
> I'd say it should affect the stored order and overwrite it. But I think 'reorder by file name' should be only an option for image collections, not tilesets.

Yeah, the only problem with this is that it's a little destructive to carefully organized tilesets, so it would have to be implemented as some manually triggered action with undo support.

hepari:
I agree. :)

PR #2050, opened 2019-01-01
Animation preview follows zoom factor for tiles

huksley:
On HiDPI displays preview box tile animation is very small.
This patch adds zooming to scale defined for animation editor popup.

image


bjorn:
Hey, sorry this improvement managed to get buried in my inbox until I was reminded about it by a [recent forum post](https://discourse.mapeditor.org/t/how-to-activate-this-window/3953/6?u=bjorn).

Thanks for the patch!

Two possible ways to improve this further:

* Use the zoom factor also in the frame list (did you decide not do to this on purpose, or because the patch would get rather more involved?).
* Update the size of the preview as soon as the zoom factor changes rather than waiting until the next frame.

PR #2079, opened 2019-02-27
Allow qrc-based tileset images in new tileset code

mitchcurtis:
89aa41b and 9764a01 already made these kind of changes, so this
just fixes it in another place after 9202e25.


mitchcurtis:
By the way, I've updated my fork to master and everything seems to be working besides this. :D

bjorn:
Thanks! :-)

PR #2084, opened 2019-03-12
Added export to Defold .collection files

CodeSpartan:
This pull request adds a plugin to Tiled, which allows exporting a map into a .collection (Defold's prefab).

Exporting into a collection allows the use of:
- Group layers (**only top-level group layers are supported, not nested ones!**)
- Multiple Tilesets per Tilemap

Upon export:
- The **Path** property of each Tileset may need to be set up manually in Defold after each export. However, Tiled will attempt to find the .tilesource file corresponding with the name of the Tileset in Tiled in your project's /tilesources/ directory. If one is found, manual adjustments won't be necessary. The way Tiled finds the root of your project is by looking for a file called "game.project" while going up the hierarchy, starting from the directory where you save the collection. Therefore, if you save your Collection outside your project, manually setting up the **Path** property will be necessary afterwards.
- If you create custom properties on your map called "x-offset" and "y-offset", these values will be used as coordinates for your top-level GameObject in the Collection.

All layers of a Tilemap will have **Z**-index property assigned with values ranging between **0** and **0.1**.
The plugin supports the use of 9999 Group Layers and 9999 Tile Layers per Group Layer.
Z-Index is assigned by combining Group Layer's order (from bottom up) with Tile Layer's order (from bottom up) in the following way: group's order is divided by 10000, layer's order is divided by 1'0000'0000, and their sum is the Z-index. So, for example, if the Group Layer's order was 15, and the Tile Layer's order was 200, the resulting Z-index would be 0.00150200. This guarantees correct superimposing of all elements in Defold.

### DIFF
--- a/src/plugins/defoldcollection/defoldcollectionplugin.cpp
+++ b/src/plugins/defoldcollection/defoldcollectionplugin.cpp
@@ -326,7 +326,7 @@ bool DefoldCollectionPlugin::write(const Tiled::Map *map, const QString &collect
                 QVariantHash map_h;
 
                 QString layers;
-                int cells_on_this_tilemap = 0;
+                int cellsOnThisTilemap = 0;
                 int layerOrder = 0;
                 for (auto &subLayer : layer->asGroupLayer()->layers()) {
                     auto tileLayer = subLayer->asTileLayer();
@@ -355,13 +355,13 @@ bool DefoldCollectionPlugin::write(const Tiled::Map *map, const QString &collect
                             cell_h["h_flip"] = cell.flippedHorizontally() ? 1 : 0;
                             cell_h["v_flip"] = cell.flippedVertically() ? 1 : 0;
                             cells.append(replaceTags(QLatin1String(cell_t), cell_h));
-                            cells_on_this_tilemap++;
+                            cellsOnThisTilemap++;
 
                             // Create a component for this embedded instance only when the first cell is found.
                             // If 0 cells are found, this component is not necessary.
                             // If more than 1 cells are found, recreating it would be redundant.
                             // Hence, only when encountering the first cell do we create it.
-                            if (cells_on_this_tilemap == 1) {
+                            if (cellsOnThisTilemap == 1) {
                                 QVariantHash component_h;
                                 component_h["tilemap_name"] = mapName + "-" + layer->name() + "-" + tileset->name();
                                 component_h["tilemap_rel_path"] = tilesetRelativePath(tilemapFilePath);
@@ -382,7 +382,7 @@ bool DefoldCollectionPlugin::write(const Tiled::Map *map, const QString &collect
                 map_h["tile_set"] = "/tilesources/" + tileset->name() + ".tilesource";
 
                 // avoid saving tilemaps with 0 cells
-                if (cells_on_this_tilemap == 0)
+                if (cellsOnThisTilemap == 0)
                     continue;
 
                 QString result = replaceTags(QLatin1String(map_t), map_h);
--- a/src/plugins/defoldcollection/defoldcollectionplugin.cpp
+++ b/src/plugins/defoldcollection/defoldcollectionplugin.cpp
@@ -38,83 +38,83 @@
 
 namespace DefoldCollection {
 
-typedef QList<const Tiled::Cell*> CellsList;
-
 static const char cellTemplate[] =
-"  cell {\n\
-    x: {{x}}\n\
-    y: {{y}}\n\
-    tile: {{tile}}\n\
-    h_flip: {{h_flip}}\n\
-    v_flip: {{v_flip}}\n\
-  }\n";
+R"(  cell {
+    x: {{x}}
+    y: {{y}}
+    tile: {{tile}}
+    h_flip: {{h_flip}}
+    v_flip: {{v_flip}}
+  }
+)";
 
 static const char layerTemplate[] =
-"layers {\n\
-  id: \"{{id}}\"\n\
-  z: {{z}}\n\
-  is_visible: {{is_visible}}\n\
-{{cells}}\
-}\n";
+R"(layers {
+  id: "{{id}}"
+  z: {{z}}
+  is_visible: {{is_visible}}
+{{cells}}}
+)";
 
 static const char tileMapTemplate[] =
-"tile_set: \"{{tile_set}}\"\n\
-{{layers}}\n\
-material: \"{{material}}\"\n\
-blend_mode: {{blend_mode}}\n";
+R"(tile_set: "{{tile_set}}"
+{{layers}}
+material: "{{material}}"
+blend_mode: {{blend_mode}}
+)";
 
 static const char collectionTemplate[] =
-"name: \"default\"\n\
-scale_along_z: 0\n\
-{{embedded-instances}}\n\
-\n";
+R"(name: "default"
+scale_along_z: 0
+{{embedded-instances}}
+
+)";
 
 static const char componentTemplate[] =
-"  \"components {\\n\"\n\
-  \"  id: \\\"{{tilemap_name}}\\\"\\n\"\n\
-  \"  component: \\\"{{tilemap_rel_path}}\\\"\\n\"\n\
-  \"  position {\\n\"\n\
-  \"    x: 0.0\\n\"\n\
-  \"    y: 0.0\\n\"\n\
-  \"    z: 0.0\\n\"\n\
-  \"  }\\n\"\n\
-  \"  rotation {\\n\"\n\
-  \"    x: 0.0\\n\"\n\
-  \"    y: 0.0\\n\"\n\
-  \"    z: 0.0\\n\"\n\
-  \"    w: 1.0\\n\"\n\
-  \"  }\\n\"\n\
-  \"}\\n\"\n\
-";
+R"(  "components {\n"
+  "  id: \"{{tilemap_name}}\"\n"
+  "  component: \"{{tilemap_rel_path}}\"\n"
+  "  position {\n"
+  "    x: 0.0\n"
+  "    y: 0.0\n"
+  "    z: 0.0\n"
+  "  }\n"
+  "  rotation {\n"
+  "    x: 0.0\n"
+  "    y: 0.0\n"
+  "    z: 0.0\n"
+  "    w: 1.0\n"
+  "  }\n"
+  "}\n"
+)";
 
 static const char childTemplate[] =
-"  children: \"{{child-name}}\"\n\
-";
+R"(  children: "{{child-name}}"
+)";
 
 static const char emdeddedInstanceTemplate[] =
-"embedded_instances {\n\
-  id: \"{{instance-name}}\"\n\
-{{children}}\
-  data: {{components}}\n\
-  \"\"\n\
-  position {\n\
-    x: {{pos-x}}\n\
-    y: {{pos-y}}\n\
-    z: 0.0\n\
-  }\n\
-  rotation {\n\
-    x: 0.0\n\
-    y: 0.0\n\
-    z: 0.0\n\
-    w: 1.0\n\
-  }\n\
-  scale3 {\n\
-    x: 1.0\n\
-    y: 1.0\n\
-    z: 1.0\n\
-  }\n\
-}\n\
-";
+R"(embedded_instances {
+  id: "{{instance-name}}"
+{{children}}  data: {{components}}
+  ""
+  position {
+    x: {{pos-x}}
+    y: {{pos-y}}
+    z: 0.0
+  }
+  rotation {
+    x: 0.0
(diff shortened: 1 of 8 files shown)


src/plugins/defoldcollection/defoldcollectionplugin.cpp
```diff
@@ -0,0 +1,473 @@
...
+
+QString DefoldCollectionPlugin::errorString() const
+{
+    return mError;
+}
+
+/*
+ * Returns a new filepath relative to the root of the Defold project if we're in one.
+ * Determines the root of the project by looking for a file called "game.project".
+ * If no such file is found by going up the heirarchy, return filename from the filepath.
```
bjorn:
```suggestion
 * If no such file is found by going up the hierarchy, return filename from the \a filePath.
```

src/plugins/defoldcollection/defoldcollectionplugin.cpp
```diff
@@ -0,0 +1,473 @@
...
+*/
+QString DefoldCollectionPlugin::TilesetRelativePath(QString filePath)
```
bjorn: Coding style: please rename to `tilesetRelativePath`.

Also please pass `QString` as `const QString &` when possible.

Finally I think this function could be a static local function instead of a member.

src/plugins/defoldcollection/defoldcollectionplugin.cpp
```diff
@@ -0,0 +1,473 @@
...
+{
+    QString gameproject = "/game.project";
+    QFileInfo fi(filePath);
+    QDir Qd = fi.dir();
```
bjorn: Coding style: please rename to `dir`.

src/plugins/defoldcollection/defoldcollectionplugin.cpp
```diff
@@ -0,0 +1,473 @@
...
+
+    while (Qd.exists() && !Qd.isRoot())
+    {
```
bjorn: Coding style: please put `{` on the same line (also for `if`, `for` and `else`)

src/plugins/defoldcollection/defoldcollectionplugin.cpp
```diff
@@ -0,0 +1,473 @@
...
+        if (QFileInfo::exists(Qd.path() + gameproject))
+        {
+            // return relative path
+            filePath.replace(Qd.path(), "");
+            return filePath;
+        }
+        else if (!Qd.cdUp())
```
bjorn: Coding style: please put `}` on the same line as the `else`.

src/plugins/defoldcollection/defoldcollectionplugin.cpp
```diff
@@ -0,0 +1,473 @@
...
+*/
+QString DefoldCollectionPlugin::TilesetRelativePath(QString filePath)
+{
+    QString gameproject = "/game.project";
+    QFileInfo fi(filePath);
+    QDir Qd = fi.dir();
+
+    while (Qd.exists() && !Qd.isRoot())
+    {
+        if (QFileInfo::exists(Qd.path() + gameproject))
```
bjorn: I think it's nicer to remove the `/` from `gameproject` and use [`QDir::filePath`](https://doc.qt.io/qt-5/qdir.html#filePath).

src/plugins/defoldcollection/defoldcollectionplugin.cpp
```diff
@@ -0,0 +1,473 @@
...
+    QString tilesetFileDir = outputFilePath;
+    tilesetFileDir.chop(outputFileName.length());
+
+    // if there are no group layers, all tilemaps are components of the top gameobject called "tilemaps"
+    if (map->groupLayerCount() == 0)
+    {
+        QString components;
+
+        // create a tilemap for each tileset this map uses
+        for (auto& tileset_current : map->tilesets())
```
bjorn: Coding style: please don't use underscores in variable names. Can be renamed to simply `tileset` I think. Also, please place the `&` against the variable name instead of the type.

src/plugins/defoldcollection/defoldcollectionplugin.cpp
```diff
@@ -0,0 +1,473 @@
...
+            map_h["layers"] = layers;
+            map_h["material"] = "/builtins/materials/tile_map.material";
+            map_h["blend_mode"] = "BLEND_MODE_ALPHA";
+            // Below, we input a value that's not necessarily correct in Defold, but it lets the user know what tilesource to link this tilemap with manually.
+            // However, if the user keeps all tilesources in /tilesources/ and the name of the tilesource corresponds with the name of the tileset in Defold,
+            // the value will be automatically correct.
+            map_h["tile_set"] = "/tilesources/" + tileset_current->name() + ".tilesource";
+
+            QString result = replaceTags(QLatin1String(map_t), map_h);
+            Tiled::SaveFile mapFile(tilesetFilePath);
```
bjorn: Please rename to `tilesetFile` to avoid confusion. :-)

src/plugins/defoldcollection/defoldcollectionplugin.cpp
```diff
@@ -0,0 +1,473 @@
...
+
+            QVariantHash emdedded_instance_h;
+            emdedded_instance_h["instance-name"] = layer->name();
+
+            QString components;
+
+            // write as many tilemaps as there are tilesets per group layer
+            for (auto& tileset_current : map->tilesets())
+            {
+                QString tilesetFilePath = tilesetFileDir;
```
bjorn: Even though you're writing one tilemap per tileset, this is still a tile map file, so please rename to `tilemapFilePath` to avoid confusion.

src/plugins/defoldcollection/defoldcollectionplugin.cpp
```diff
@@ -0,0 +1,473 @@
...
+        QVariantHash collection_h;
+        // optional custom properties that allow exporting a map with an offset
+        collection_h["pos-x"] = map->property(QLatin1String("x-offset")).toInt();
+        collection_h["pos-y"] = map->property(QLatin1String("y-offset")).toInt();
+        collection_h["components"] = QString("");
+        collection_h["children"] = children;
+        collection_h["embedded-instances"] = embeddedInstances;
+
+        QString result = replaceTags(QLatin1String(collection_t), collection_h);
+        Tiled::SaveFile mapFile(fileName);
```
bjorn: Maybe rename this to `collectionFile`?

CodeSpartan:
> I've left a bunch of inline comments, mostly regarding coding style.
Done, let me know if I missed anything. The only thing that I didn't change was underscores in names like "cell_t", "cell_h", because I'm assuming that this is fine? This is how it is in the original defold plugin.

> I am wondering if it still makes sense to keep the original Defold export plugin, or does it make sense to replace it with this one?
I would absolutely keep both, not everyone needs collections. Defold heavily emphasizes light-weightness and performance, and they didn't implement out-of-the-box multiple tilesets per map because each tileset adds a draw call. So I would leave both and add a warning that using multiple tilesources results in multiple drawcalls, which can introduce a source of poor performance if users mix tiles from **many** tilesets.

bjorn:
> Done, let me know if I missed anything. The only thing that I didn't change was underscores in names like "cell_t", "cell_h", because I'm assuming that this is fine? This is how it is in the original defold plugin.

Yeah, I realize we have precedence there. It's not my style, but I had decided to leave that be. Personally I don't like these names, because I do not know what the "t" or "h" mean (though my guesses would be "template" and "hash"). For string constants I usually use UPPERCASED_UNDERSCORED_STYLE.

> I would absolutely keep both, because not everyone needs collections. Besides, Defold heavily emphasizes light-weightness and performance, and they didn't implement out-of-the-box multiple tilesets per map because each tileset adds a draw call. So I would leave both and add a warning that using multiple tilesets results in multiple drawcalls, which can introduce a source of poor performance if users mix tiles from **many** tilesets.

Alright. You're definitely right that using multiple tilesets will result in multiple drawcalls, unless of course some technical solution is implemented to avoid that, but either will have a runtime impact that can be avoided by creating the map using only a single tileset in the first place.

But when this plugin creates the same tilemap file as the other plugin, in addition to the collection file, then somebody using Defold can always choose to ignore the collection file and use the tilemap file directly if that suits his needs better. Then we avoid maintenance and a bit of confusion on the user side about which option to choose. Alternatively, we could merge the plugins so that a single plugin exposes the two different export formats, to share some of the code.

> Regarding Travis-CI failing, it looks like it's an internal error on their side, correct?

Yes, it appears the Qt download is failing. I may need to make it more verbose somehow, but not too verbose. Right now it fails due to 10 minutes of no output while it is downloading Qt (which is quite large). But when I make the download verbose, it fails because it reaches the maximum log length...

src/plugins/defoldcollection/defoldcollectionplugin.h
```diff
@@ -0,0 +1,52 @@
...
+    QString shortName() const override;
+
+protected:
+    QString nameFilter() const override;
+
+private:
+    QString mError;
+};
+
+static QString tilesetRelativePath(const QString & filePath);
```
bjorn: It makes no sense to declare this function as static in the header file. The static will make its symbol local to the translation unit, so it can't be used somewhere else anyway. Just remove the declaration and move the `static` keyword to the definition.

src/plugins/defoldcollection/defoldcollectionplugin.cpp
```diff
@@ -0,0 +1,444 @@
...
+            mError = mapFile.errorString();
+            return false;
+        }
+
+        if (!mapFile.commit()) {
+            mError = mapFile.errorString();
+            return false;
+        }
+    } else {
+        // if there are group layers, each group layer is a gameobject, parented to the top gameobject, while being a parent to multiple tilemap components
```
bjorn: Do you think it would be possible to remove this big distinction between "if there are group layers" and "if there are no group layers"? It seems weird to me, that if I group some of my tile layers, that it will go into the group layers branch and ignore any top-level tile layers.

CodeSpartan: Good point, I actually hadn't thought that someone might want to mix the two types of structures. I'll see if I can remove the distinction.

CodeSpartan:
Rewrote a lot of stuff today:
- Added the possibility of having both top-level tile layers and top-level group layers
- Now tilemaps only contain layers that have cells on them
- Fixed a bug where the max amount of layers wasn't 9999, but 1000
- Removed underscores from all variable names, even ones that were "_t" and "_h"

Independently of whether you intend to keep the old export format or not, I would suggest leaving both in at least for a while, until this one is proven to be stable. I tested in all ways I could think of, but I don't think it should count as "thorough". And I don't intend on working on the world for my game until much later, so if there's any bugs, I won't catch them myself. So consider keeping both export options at least for now.

Let me know if there's anything else that needs to be corrected.

britzl:
Wow, this is great! @CodeSpartan could you please take a look at the requested changes so that this can get merged?

bjorn:
@britzl I believe @CodeSpartan has already responded to all my requested changes and I've made some tweaks myself already as well. I'm not sure why I haven't merged this yet, but I'll be sure to take another look at it this week.

stefankendall:
Any timeline for merging this?

bjorn:
@stefankendall I really need to learn how not to forget about stuff I plan to do. I tend to get driven by email notifications and my last comment generated an email I expected to get around to, but I ended up working on other things that popped up afterwards.

My next full Tiled development day is Wednesday and I have set a reminder about this in my calendar, but I'll try to get around to it earlier.

britzl:
@bjorn You could use something like [Pull Panda](https://github.com/marketplace/pull-panda) to get reminders on open pull requests. Its integrated into the GitHub Marketplace ecosystem and is free to use.

bjorn:
@britzl Hmm, it appears Pull Panda can only be installed on an organization account. Well, let's see whether using my calendar works first. :-)

britzl:
> @britzl Hmm, it appears Pull Panda can only be installed on an organization account. Well, let's see whether using my calendar works first. :-)

Ah, I see. Well let's hope the calendar works.

bjorn:
So in fact I'm working on Tiled today. I've looked over the code again and did a quick test and it seemed to work fine. Unfortunately, I didn't test the merged result so I had to push a quick follow-up fix (72ba25d354286e328565692f1d00c9b35fa7ec9c) due to a signature change.

Thanks a lot @CodeSpartan for your contribution! The plugin will be included in the next development snapshot (probably later this week). Do you think you'd have time to add a note about this new export option to the [Export Formats](https://doc.mapeditor.org/en/latest/manual/export/#defold) section of the manual?

CodeSpartan:
@bjorn done: https://github.com/bjorn/tiled/pull/2186

PR #2085, opened 2019-03-16
Allow loading a TSX file from Python plugins

Ruin0x11:
This pull request adds a method to the Python API, `loadSharedTilesetFromTsx`, which allows loading an existing TSX file as a tileset.

The use case I'm thinking of is embedding the path to a specific version of a TSX in a map format, then attempting to resolve it when the map is loaded in Tiled, and sharing the same TSX across multiple map files. The data system we have may support several groups of tiles loaded from outside sources like mods which can be enabled and disabled, so I was thinking about pregenerating a TSX for a single supported configuration of enabled/disabled mods, caching the tile images somewhere, and allowing them to be used in the editor.

### DIFF
--- a/src/plugins/python/tiledbinding.py
+++ b/src/plugins/python/tiledbinding.py
@@ -379,18 +379,12 @@ static bool loadTilesetFromFile(Tiled::Tileset *ts, const QString &file)
 }
 """)
 
-mod.add_function('loadTileset', 'bool', [('Tiled::SharedTileset&','sts'),('QString','file')])
+mod.add_function('loadTileset', 'Tiled::SharedTileset', [('QString','file')])
 
 mod.body.writeln("""
-static bool loadTileset(Tiled::SharedTileset& sts, const QString &file)
+static Tiled::SharedTileset loadTileset(const QString &file)
 {
-    auto result = Tiled::TilesetManager::instance()->loadTileset(file);
-    if (!result || result->status() == Tiled::LoadingStatus::LoadingError)
-    {
-        return false;
-    }
-    sts = result;
-    return true;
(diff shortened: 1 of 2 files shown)


bjorn:
Thanks for your patch! It definitely makes sense for the Python plugins to be able to load external tilesets while loading a map.

Since generally we want to avoid loading the same tileset multiple times, there is the `TilesetManager` through which tilesets are loaded. Also, I think it would make sense for this function to work with any supported tileset format rather than just TSX. So, what about renaming this to simply `loadTileset` and implementing it based on `TilesetManager::loadTileset`?

Ruin0x11:
Yes, that makes sense. It also prevents the behavior where the same tileset shows up in the editor multiple times.

src/plugins/python/tiledbinding.py
```diff
@@ -374,6 +379,21 @@ def _decorate(obj, *args, **kwargs):
 }
 """)
 
+mod.add_function('loadTileset', 'bool', [('Tiled::SharedTileset&','sts'),('QString','file')])
+
+mod.body.writeln("""
+static bool loadTileset(Tiled::SharedTileset& sts, const QString &file)
+{
+    auto result = Tiled::TilesetManager::instance()->loadTileset(file);
+    if (!result || result->status() == Tiled::LoadingStatus::LoadingError)
```
bjorn: Could this function be implemented to simply return the `Tiled::SharedTileset`? I think for `Tiled::LoadingStatus::LoadingError` we'll want to anyway set it, since it can be useful to load the map with a tileset that failed to load entirely (allowing the user to fix up references in that tileset, for example).

bjorn:
Thanks!

Issue #2065, opened 2019-02-02
Changing maps in World shouldn't reset tileset and layer


RainingChain:
I have a large world with many maps. They all share the same tilesets and layers. Let say I'm adding trees from tileset 2 on layer TreeLayer on the first map. Now, I want to do the same on another map. 

When I click the other map, the tileset is reset to tileset 1 and the layer is reset to something else. This forces me to change back the tileset and layer everytime. This is especially annoying because I sometime forget to change the layer and I start drawing on the wrong layer by mistake. 

Wanted behaviour:
When clicking on a new map in the world:
- If one of the layers on the new map as the same name as the current layer of the old map, change the layer to this one.
- If the new map has the same tileset as the current tileset of the old map, change the tileset to this one.

Thanks

bjorn:
That makes total sense, thanks for opening such a detailed issue about it!

bjorn:
@RainingChain I plan to push a new development snapshot next week that will include this improvement.

PR #2087, opened 2019-03-18
Try to keep layer by name in worlds (#2065)

JustinZhengBC:
I was working on #2065 and I added a method `setCurrentLayerByName` to `MapDocument` so that after `DocumentEditor` switches documents within a world, it can try to also change the layer to one with an identical name, as the original issue requested.

I was not able to reproduce the second request in the original issue. Changing the active tileset already persists between maps for me.

### DIFF
--- a/src/tiled/mapdocument.cpp
+++ b/src/tiled/mapdocument.cpp
@@ -248,6 +248,24 @@ void MapDocument::setSelectedLayers(const QList<Layer *> &layers)
     emit selectedLayersChanged();
 }
 
+void MapDocument::switchCurrentLayer(Layer *layer)
+{
+    setCurrentLayer(layer);
+
+    // Automatically select the layer if it isn't already
+    if (layer && !mSelectedLayers.contains(layer))
+        setSelectedLayers({ layer });
+}
+
+void MapDocument::switchSelectedLayers(const QList<Layer *> &layers)
+{
+    setSelectedLayers(layers);
+
+    // Automatically make sure the current layer is one of the selected ones
+    if (!layers.contains(mCurrentLayer))
+        setCurrentLayer(layers.isEmpty() ? nullptr : layers.first());
+}
+
 void MapDocument::resizeMap(QSize size, QPoint offset, bool removeObjects)
 {
     static_cast<EditableMap*>(editable())->resize(size, offset, removeObjects);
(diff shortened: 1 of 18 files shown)


src/tiled/documentmanager.cpp
```diff
@@ -290,11 +290,26 @@ bool DocumentManager::switchToDocument(Document *document)
...
-void DocumentManager::switchToDocument(MapDocument *mapDocument, QPointF viewCenter, qreal scale)
+void DocumentManager::switchToDocument(MapDocument *mapDocument,
+                                       QPointF viewCenter,
+                                       qreal scale,
+                                       bool tryPreserveLayer)
 {
+    const QString *currentLayerName = nullptr;
+    if (tryPreserveLayer) {
+        Document *currentDocument = mMapEditor->currentDocument();
+        if (auto currentMapDocument = qobject_cast<MapDocument*>(currentDocument)) {
```
bjorn: Since this keeping of the current layer is a very world-specific feature, I would suggest to put the logic entirely in `MapItem::mouseReleaseEvent` for now. That way, we avoid the boolean parameter as well.

src/tiled/mapdocument.cpp
```diff
@@ -246,6 +246,19 @@ void MapDocument::setCurrentLayer(Layer *layer)
             setCurrentObject(layer);
 }
 
+void MapDocument::setCurrentLayerByName(const QString *name)
+{
+    for (int i = 0; i < mLayerModel->rowCount(); ++i) {
```
bjorn: Please don't implement this based on the layer model. Also I think we don't need an extra method for this. We can use `Map::findLayer` to find the layer by name and then simply call `MapDocument::setCurrentLayer`.

bjorn:
> I was not able to reproduce the second request in the original issue. Changing the active tileset already persists between maps for me.

I think it will only persist if both maps are referencing the same tileset (so it must be an external tileset). If they both embed a tileset which is similar (potentially based on the same tileset image), then it will not work. To make that work, we could remember the previously selected tileset and then use `Tileset::findSimilarTileset` to find a similar one among the tilesets shown for the newly selected map.

JustinZhengBC:
@bjorn I've made the requested changes and I was also able to replicate the tileset issue after embedding the tilesets

bjorn:
@JustinZhengBC Do you think you'll have time to review and test the functionality after my changes above? Maybe we can merge this change this week.

JustinZhengBC:
@bjorn yes, I can definitely find time to test the changes in the next few days.

JustinZhengBC:
@bjorn I got a chance to try out your changes, and I can confirm your changes fix the additional test cases.

bjorn:
Hmm, there's still a bug, or potentially an issue introduced in my latest change, which breaks using Ctrl to select multiple layers in the Layers view.

bjorn:
Alright, I think I got it all working now... but this could use some more testing. I'll merge it now, probably will do a new development snapshot next week.

bjorn:
Thanks again for your work on this, @JustinZhengBC!

PR #2088, opened 2019-03-18
Mention clojure library

oakes:


bjorn:
Thanks for your patch, @oakes! I've merged it as 524130a6 on the 1.2 branch (just so that if I tag another release, the link will be included in the "stable" version of the docs). It'll go online in the "latest" version of the manual when I merge 1.2 into master.

PR #2089, opened 2019-03-19
single JAXBContext instance for Tiles and AnimatedTiles

richardfoltin:
Greatly increases loading time in case of a lot of tile properties.

### DIFF
--- a/util/java/libtiled-java/src/main/java/org/mapeditor/io/TMXMapReader.java
+++ b/util/java/libtiled-java/src/main/java/org/mapeditor/io/TMXMapReader.java
@@ -98,7 +98,10 @@ public class TMXMapReader {
     private TreeMap<Integer, TileSet> tilesetPerFirstGid;
     public final TMXMapReaderSettings settings = new TMXMapReaderSettings();
     private final HashMap<String, TileSet> cachedTilesets = new HashMap<>();
-
+  
+    private static JAXBContext tileContext;
+    private static JAXBContext animatedTileContext;
+    
     public static final class TMXMapReaderSettings {
 
         public boolean reuseCachedTilesets = false;
@@ -154,12 +157,9 @@ public class TMXMapReader {
         }
     }
 
-    
-    private JAXBContext tileContext;
-    private JAXBContext animatedTileContext;
-
     private <T> T unmarshalClass(Node node, Class<T> type) throws JAXBException {
         JAXBContext context;
+        
         if (type == Tile.class) {
             if (tileContext == null) tileContext = JAXBContext.newInstance(Tile.class);
             context = tileContext;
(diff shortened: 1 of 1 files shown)


bjorn:
@mikepthomas Can you review this, since you introduced the JAXBContext?

A few things I'm wondering about, also for @richardfoltin:

* If `JAXBContext.newInstance` is slow, it would make sense to set up a cache or to pre-create all instances as static members `TMXMapReader`, for all serialized classes, right? Of course some things are only going to be present a few times, but even then we probably only need one JAXBContext instance (like for tileset elements). And other things may also occur frequently, like object elements.

* The context seems to be only used to create the unmarshaller. Should we cache / pre-create the unmarshallers instead? I'm not how fast `context.createUnmarshaller` is, but if there is no need to create a new unmarshaller each time then let's avoid it.

mikepthomas:
I like the way they are being Lazy loaded the first time they are used 👍 may be worth looking at setting them up as static members but dont think that would speed anything up any more than you have

richardfoltin:
Fixed it to static property.

The speed increase could be substantial; My .tmx contains about 1500 rows and loading time decreased from about 30 sec to 1 sec.

bjorn:
> The speed increase could be substantial; My .tmx contains about 1500 rows and loading time decreased from about 30 sec to 1 sec.

While that is great of course, I'd seriously consider moving away from the "XML" storage method for tile layer data. One second is still rather long, unless you're also including the loading time for the images. The base64+zlib option is generally the fastest to load.

@richardfoltin @mikepthomas Neither of you know whether it could be preferable to keep around the Unmarshaller instead of the JAXBContext?

mikepthomas:
> Neither of you know whether it could be preferable to keep around the Unmarshaller instead of the JAXBContext?

The context is used to create the unmarshaller:
https://github.com/bjorn/tiled/blob/af13741ccdcb31d38089cc74e6e018b80f205011/util/java/libtiled-java/src/main/java/org/mapeditor/io/TMXMapReader.java#L173

bjorn:
Hmm, actually, according to the documentation, we can create a single JAXBContext instance, passing in all the classes that we want it to be able to handle (see [JAXBContext.createInstance(Class...)](https://docs.oracle.com/javase/7/docs/api/javax/xml/bind/JAXBContext.html#newInstance(java.lang.Class...))). This will automatically include aggregated classes, so we may only need to list a few top-level classes like the map and the tileset. In addition, it seems we just need a single Unmarshaller instance as well.

So, I would suggest to change this patch to use a static Unmarshaller member, which is initialized on-demand in the constructor based on a single JAXBContext using the required classes.

bjorn:
Closing this since the optimization was also part of #2006, which was just merged.

PR #2092, opened 2019-03-24
Fix an error in the docs.

TheRamenChef:
There is a backtick block that isn't properly closed, and every time I read it it manages to bug me. I'm making this pull request to fix it.

Edit 4/18: A different pull request fixed the backtick block, but it broke the entire table, which this pull request fixes.


bjorn:
Whoops, seems I forgot about this one. Thanks!

PR #2096, opened 2019-03-31
Create tiled_sv.ts

Regnander:
Uploading the translation from the Weblate which I made about a month ago. I thought it would be uploaded here automatically, but I guess not. 😕

### DIFF
--- a/translations/tiled_sv.ts
+++ b/translations/tiled_sv.ts
@@ -1,0 +1,6284 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!DOCTYPE TS>
+<TS version="2.1" language="sv">
+<context>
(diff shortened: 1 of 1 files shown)


bjorn:
It's a manual process since the automated process leads to a lot of "commit spam", and unfortunately I do not get notified by Weblate when there is a new translation. I forgot to check for updates when I pushed out Tiled 1.2.3, sorry about that.

I'll do a merge from Weblate now to the `1.2` branch, so that the translation may be included in an eventual 1.2.4 release.

bjorn:
Pushed now as change af87b4dc19c3eae9968dec1fc88a22e99fdcf2bc. Thanks a lot for translating Tiled to Swedish!

PR #2099, opened 2019-04-03
Documented the rect() type

Ablu:


bjorn:
Please add `Qt.point` as well, while you're at it. :-)

bjorn:
Ah, still missing `Qt.size`. :-)

docs/reference/scripting.rst
```diff
@@ -838,28 +839,34 @@ SelectedArea
 Functions
 ~~~~~~~~~
 
-SelectedArea.set(rect : rect) : void
+SelectedArea.boundingRect() : :ref:`script-rect`
+    Returns a ectangle which surrounds the selected area.
```
bjorn: You've exposed `boundingRect` as a property, so it shouldn't be documented as a function.

src/tiled/editableselectedarea.cpp
```diff
@@ -33,6 +33,11 @@ EditableSelectedArea::EditableSelectedArea(MapDocument *mapDocument, QObject *pa
 {
 }
 
+RegionValueType EditableSelectedArea::get()
```
bjorn: As `const` it didn't work?

src/tiled/editableselectedarea.h
```diff
@@ -38,6 +38,8 @@ class EditableSelectedArea : public QObject
 public:
     explicit EditableSelectedArea(MapDocument *mapDocument,
                                   QObject *parent = nullptr);
+    Q_PROPERTY(QRect boundingRect READ boundingRect)
```
bjorn: Please move the `Q_PROPERTY` definition up between `Q_OBJECT` and the `public:`, where we usually put it.

src/tiled/editableselectedarea.h
```diff
@@ -55,6 +57,8 @@ public slots:
 
 private:
     void set(const QRegion &region);
+    QRect boundingRect();
```
bjorn: Please mark as `const`. Also I think this getter shouldn't be private.

PR #2108, opened 2019-04-14
Add Tile.probability field in JSON spec

DanTup:
While parsing some of the example files, I noticed this field was in the exported file but not in my parsing. I don't know if it's all that relevant for a non-editor (since it says this is used in by the editor) but since it can be in the file it seems like it should be listed for completeness.


DanTup:
(Sorry about the bad diff - I increased the width of the table a little to avoid wrapping onto three lines, but GH didn't do a good job of representing that! There were no significant changes to any lines except for the probability field)

bjorn:
Thanks!

PR #2109, opened 2019-04-14
Add possible renderorder values to spec

DanTup:
This makes it easier to convert to an enum if generated classes from this spec.


bjorn:
Thanks! :-)

PR #2110, opened 2019-04-22
Remove an unmatched closing brace in README

Freeyorp:
Nitpicky, but just in case nobody else got to it yet. ^^


bjorn:
Thanks! :-)

These qmake instructions should really be replaced with qbs instructions...

PR #2112, opened 2019-04-26
Update support-for-tmx-maps.rst

qwebltd:
Added Qiso reference under the Corona SDK framework. Also added note against Dusk reference, as it hasn't been updated for over 4 years now.


bjorn:
Looks fine, thanks!

PR #2114, opened 2019-04-28
Add "source" field for Tileset object

DanTup:
If you reference an external tileset and then save as JSON, there's a `source` field but it's not mentioned in the docs (which means if you code-gen from the spec, you don't have it :-)).

I didn't copy the whole description of this from the XML spec as it'd get a bit messy in this table, but it miht be easier to do that if this file changes to using the CSV format.


bjorn:
Thanks!

PR #2115, opened 2019-05-03
Add missing "only" to JSON spec

DanTup:
When re-generating after changing to CSVs, I noticed this property moved from ImageLayer down to Layer. I think this was unintended that this should be marked as imagelayer *only*.


bjorn:
That's correct. :-)

PR #2118, opened 2019-05-07
Clarify the meaning of the "staggered" orientation

Phlosioneer:
There are three uses of staggered tiles: Staggered Isometric, Staggered Hexagonal, and Staggered Orthogonal (which is normal squares, but each tile has 6 neighbors instead of 9). So the name `staggered` could refer to either the Isometric or the Orthogonal meanings (hexagonal is clearly documented as using staggered axes in addition to the `staggered` orientation).


bjorn:
I think there could be a lot more to clarify about it, but maybe this helps. Thanks! :-)

PR #2124, opened 2019-05-14
Add Tile.type() and MapObject.effectiveType() to Python API

### LINKED ISSUE #2123: MapObject.type() in Python API returns blank string if tile with predefined type is used
If a `MapObject` uses a tile with a predefined type that is set in the tileset, then calling `MapObject.type()` from the Python API will return a blank string. If the object uses a tile with no type and instead it is set in the `Properties` pane manually, it is returned correctly.

Also, there is no way to get the `type` field of a `Tile` in the current Python API, so there is no way to access this field currently.

Ruin0x11:
Close #2123.

### DIFF
--- a/src/plugins/python/pythonbind.cpp
+++ b/src/plugins/python/pythonbind.cpp
@@ -3925,6 +3925,18 @@ _wrap_PyTiledTile_tileset(PyTiledTile *self)
 
 
 PyObject *
+_wrap_PyTiledTile_type(PyTiledTile *self)
+{
+    PyObject *py_retval;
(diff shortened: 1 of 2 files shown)


bjorn:
Hey, that looks fine, thanks for adding these methods!

I think this can go into Tiled 1.2.4, so I'll rebase this change.

bjorn:
Alright, changed my mind about `1.2` since some additions were already made on the `master` branch, leading to conflicts. I think it's not really worth it, since those who'd need these functions can rely on the development snapshots.

PR #2125, opened 2019-05-14
Allow retrieving the type of a property from Python

Ruin0x11:
Adds `Object.propertyType()` to the Python API. This way plugin code can properly encode the types of properties without having to guess what they are using `Object.propertyAsString()`.


bjorn:
I've no idea what you did to duplicate all those commits, but I've removed them now by rebasing the two actual changes onto the latest master. :-)

Thanks for the useful addition!

Issue #2121, opened 2019-05-13
Customizable Chunk Size for Infinite Maps


RPGHacker:
Hello,

I have a minor feature request for Tiled (but one that I hope should not take long to implement).

I've noticed that when working with infinitely sized maps, Tiled automatically organizes tiles in the .tmx file into chunks. These chunks are sized 16x16 tiles by default. This can be changed by modifying/writing .tmx files by hand. Tiled seems to read the chunks in correctly from the .tmx file, no matter their size. However, once saving back to the .tmx file, Tiled always seems to rearrange the chunks into 16x16 blocks. Would it be possible to add support for overriding this default output chunk size (via a property in the tile layer or something like that)?

Background: I'm currently writing a script that converts one of my .tmx files into a format usable by some game engine, and while doing that, I want to perform certain optimizations on the output file. Specifially, the engine in question is very bad at working with a high number of tiles and becomes really slow, so I want to optimize performance by batching and prerendering tiles into screen-sized textures. The game uses NES-styled graphics with a screen resolution of 256x224 and tiles with a size of 16x16, amounting to 16x14 tiles per screen. So naturally, I wondered if it was possible to directly organize my tiles in the .tmx file into chunks of exactly this size, since it would make it easier to write my conversion script (I could use the data from the chunks directly and wouldn't have to add a step to reorganize the chunks into 16x14 blocks).

I hope my explanation was understandable, if not, let me know and I'll try to clarify.
Thanks!

bjorn:
Hmm, right now, the 16x16 chunks match the in-memory representation of tile layers so it is the most straight-forward way to output a tile map. Anything else will require writing some custom code that iterates a tile layer to figure out which of those custom chunks contain any tiles, though that shouldn't be too involved.

And indeed Tiled supports reading of arbitrary chunks, even chunks of different sizes, which is mostly for future compatibility reasons. I could certainly imagine adding a custom chunk size option.

Is it maybe something you'd like to try contributing? I'm generally available for questions regarding the code here or on Discord.

RPGHacker:
Ah, yeah, if this matches the internal memory representation, that makes a lot of sense.

Yeah, I think I'll give it a try. Since I'll have to write some code for rearrangeing chunks, anyways, I could try to do it directly in Tiled and see how far I'll get with it. Will check this out at home.

bjorn:
The list of chunks to write is currently determined by the following function:

[...]

PR #2130, opened 2019-05-17
Custom output chunk size can now be specified in the map properties

RPGHacker:


RPGHacker:
This pull request would close #2121.

RPGHacker:
Whoopsie about the tabs. Something Visual Studio does by default. Thought I had converted them all to spaces, but seems like I missed some. Will add all of the suggested fixes later today.

As for the problem of ambiguity: the simplest "solution" I can think of would be to simply rename the new attributes I added to "outputchunkwidth" and "outputchunkheight" respectively. That would make it clear that those attributes only refer to the output format specifically, whereas the sizes on each chunk refer to the input sizes.

The alternative would be to drop the attributes entirely and just set this size based on whatever chunk sizes are found in the input file, but yeah, that would get problematic as soon as differently-sized chunks were present in any given input file. I think differentiating between "input" and "output" attributes, as described above, would be the more robust solution here.

bjorn:
@RPGHacker I think that rename is not a bad solution.

Ah, the other thing I don't really like is that we're introducing another property that in the UI is presented on the map, but in the saved format is put on each layer. I know this is the case for the compression as well. What do you think of either moving these settings up in the TMX hierarchy or having these values on each TileLayer instance?

bjorn:
I'm personally thinking of having a configuration section right after the map element, like:
```xml
<map ...>
  <editorsettings>
    <chunksize width="16" height="16"/>
    <tilelayerformat encoding="base64" compression="zlib"/>
  </editorsettings>
  ...
</map>
```
This would make it clear that some things are set in the editor on a map-basis, and at the same time  that these things can be ignored by readers in general.

RPGHacker:
Haha, ironically, having the property in the layer properties was actually how I solved it originally, but then moved it to the map properties since you suggested taking a look at the encoding format for reference. Since that one is edited via the map properties, I also moved this property there. I can change that, though, no problem.

Yeah, the editorsettings element sounds like a reasonable solution. Do you want me to try implementing it, or would it be okay to keep it as something to add down-the-line (and just have the property in the layer settings for now)? Adding this element seems like it would be a slightly bigger undertaking, since things like backwards-compatibility and supporting all the different formats would have to be considered.

bjorn:
> Adding this element seems like it would be a slightly bigger undertaking, since things like backwards-compatibility and supporting all the different formats would have to be considered.

Well, it's a bigger undertaking, but it avoids any backwards-compatibility issues.

> Do you want me to try implementing it, or would it be okay to keep it as something to add down-the-line (and just have the property in the layer settings for now)?

Well, at least it would need to happen before releasing Tiled 1.3. Doesn't need to be part of this pull request though. But there is no need to move those settings to the layer properties if we're going to have them on the map in the end anyway.

RPGHacker:
Alright, I'll leave them in the map properties for now and implement all the other suggestions, then I can try implementing the new element in a another pull request.

>> Well, it's a bigger undertaking, but it avoids any backwards-compatibility issues.

But if the settings move from the <data> element of each layer to this new <editorsettings> element, won't that introduce an incompatibility? Or would the idea be to still check for the old attributes in the <data> element, "just in case"?

bjorn:
> But if the settings move from the element of each layer to this new element, won't that introduce an incompatibility? Or would the idea be to still check for the old attributes in the element, "just in case"?

The idea was to explicitly mark this section as "editor settings" and not to change the rest of the format in any way. So the editor settings do not influence the way a map is loaded, apart from that those settings are read and may affect other things, in this case how the map is saved. I could imagine grid rendering options to become part of this section as well.

In addition, I think in Tiled 1.4, we'll see these editor settings show up in project files, and then in map files they would only serve to override the project settings.

RPGHacker:
(Whoops, just realized that my previous comment was a bit messed up because GitHub interpreted some parts as HTML)

Okay, just to make sure I'm understanding correctly. After the proposed changes, when opening, let's say, a .tmx file with the following contents:

```
...
  <editorsettings>
    <chunksize width="16" height="16"/>
    <tilelayerformat encoding="base64" compression="zlib"/>
  </editorsettings>
...
```

Tiled would set its compression to "zlib", and when opening a .tmx file with the following contents:

```
...
      <data compression="zlib">
         ...
      </data>
...
```

Tiled would also set its compression to "zlib". So effectively, it would just keep whatever was read last, right?

bjorn:
> Tiled would also set its compression to "zlib". So effectively, it would just keep whatever was read last, right?

Right, for backwards-compatibility reasons, we need to keep setting that value also based on those attributes, but I'd say when an "editorsettings/tilelayerformat" element is present, only the value specified there applies to the map.

RPGHacker:
Alright, thanks for the clarification. Will give it a try. :)

bjorn:
Alright, after some tweaks I think this is fine to merge. I'll look into introducing the "editorsettings" section as a follow-up change.

bjorn:
Last week I introduced the `<editorsettings>` section and now I've finally moved this information there rather than repeating it on the `<data>` elements. :)

PR #2141, opened 2019-06-15
Add options to customize "MatchOutsideMap" behavior

JoaoBaptMG:
I decided to edit Tiled's source code after viewing [this discussion about the behavior of automapping near the map's edges](https://discourse.mapeditor.org/t/automapper-ignore-the-empty-void-beyond-the-perimeter-of-map/) and I was not satified with the state of the art of automapping near the borders.

Normally, when you try to automap near the edges, the edges are seen as empty tiles, failing to apply, for example, Remex's auto-generated rules (the ones originated from RPG Maker):
image

This pull request add two options (as map properties) to customize the behavior of automapping near the edges.
- **OverflowBorder**: this option makes the automapper treats the out-of-bounds areas of the map as if they were prolongations of the border tiles, thus applying the rules along the borders as if they used the same tiles as the nearest inbound tile. This is actually the default behavior of every RPG Maker since XP.
image
- **WrapBorder**: this option makes the automapper wrap the map around the edges, effectively applying the rules also to the other side of the border; this can be useful to create looped maps, to make it seamless around the borders:
image

I coded those two features so that if both **OverflowBorder** and **WrapBorder** are true, **WrapBorder** takes precedence. And, of course, it will not work with infinite maps, since the notion of "border" in them has no sense.

### DIFF
--- a/src/tiled/automapper.cpp
+++ b/src/tiled/automapper.cpp
@@ -42,6 +42,11 @@
 
 using namespace Tiled;
 
+// Should i leave this function here?
+inline int wrap(int value, int bound) {
+    return (value % bound + bound) % bound;
+}
+
 /*
  * About the order of the methods in this file.
  * The Automapper class has 3 bigger public functions, that is
@@ -134,6 +139,17 @@ bool AutoMapper::setupRuleMapProperties()
                        "Ignoring this property.")
                 .arg(mRulePath, name, value.toString()) + QLatin1Char('\n');
     }
+
+    // OverflowBorder and WrapBorder make no sense for infinite maps
+    if (mMapWork->infinite()) {
+        mOptions.overflowBorder = false;
+        mOptions.wrapBorder = false;
+    }
+
+    // Each of the border options imply MatchOutsideMap
+    if (mOptions.overflowBorder || mOptions.wrapBorder)
+        mOptions.matchOutsideMap = true;
+
     return true;
 }
 
@@ -618,8 +634,7 @@ static bool layerMatchesConditions(const TileLayer &setLayer,
                                    const InputConditions &conditions,
                                    const QRegion &ruleRegion,
                                    const QPoint offset,
-                                   const AutoMapper::Options &options,
-                                   bool isMapInfinite)
+                                   const AutoMapper::Options &options)
 {
     const auto &listYes = conditions.listYes;
     const auto &listNo = conditions.listNo;
@@ -638,27 +653,21 @@ static bool layerMatchesConditions(const TileLayer &setLayer,
 #endif
         for (int x = rect.left(); x <= rect.right(); ++x) {
             for (int y = rect.top(); y <= rect.bottom(); ++y) {
-                if (!options.matchOutsideMap &&
-                        !setLayer.contains(x + offset.x(), y + offset.y()))
-                    return false;
-
                 int xd = x + offset.x();
                 int yd = y + offset.y();
 
-                if (!isMapInfinite) {
-                    if (options.wrapBorder) {
-                        int width = setLayer.size().width();
-                        int height = setLayer.size().height();
-                        xd = (xd % width + width) % width;
-                        yd = (yd % height + height) % height;
-                    } else if (options.overflowBorder) {
-                        if (xd < 0) xd = 0;
-                        else if (xd >= setLayer.size().width())
-                            xd = setLayer.size().width() - 1;
-                        if (yd < 0) yd = 0;
-                        else if (yd >= setLayer.size().height())
-                            yd = setLayer.size().height() - 1;
-                    }
+                if (!options.matchOutsideMap &&
+                        !setLayer.contains(xd, yd))
+                    return false;
+
+                // Those two options are guaranteed to be false if the map is infinite,
+                // so no "invalid" width/height accessing here.
+                if (options.wrapBorder) {
+                    xd = wrap(xd, setLayer.width());
+                    yd = wrap(yd, setLayer.height());
+                } else if (options.overflowBorder) {
+                    xd = qBound(0, xd, setLayer.width() - 1);
+                    yd = qBound(0, yd, setLayer.height() - 1);
                 }
 
                 const Cell &setCell = setLayer.cellAt(xd, yd);
@@ -751,7 +760,7 @@ QRect AutoMapper::applyRule(int ruleIndex, const QRect &where)
                 const int i = mMapWork->indexOfLayer(name, Layer::TileLayerType);
                 const TileLayer &setLayer = (i >= 0) ? *mMapWork->layerAt(i)->asTileLayer() : dummy;
 
-                if (!layerMatchesConditions(setLayer, conditions, ruleInputRegion, QPoint(x, y), mOptions, mMapWork->infinite())) {
+                if (!layerMatchesConditions(setLayer, conditions, ruleInputRegion, QPoint(x, y), mOptions)) {
                     allLayerNamesMatch = false;
                     break;
                 }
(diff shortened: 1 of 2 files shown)


src/tiled/automapper.cpp
```diff
@@ -631,8 +642,26 @@ static bool layerMatchesConditions(const TileLayer &setLayer,
...
+                        int height = setLayer.size().height();
+                        xd = (xd % width + width) % width;
+                        yd = (yd % height + height) % height;
+                    } else if (options.overflowBorder) {
+                        if (xd < 0) xd = 0;
+                        else if (xd >= setLayer.size().width())
+                            xd = setLayer.size().width() - 1;
+                        if (yd < 0) yd = 0;
+                        else if (yd >= setLayer.size().height())
+                            yd = setLayer.size().height() - 1;
```
bjorn: Please shorten this to:
```c++
xd = qBound(0, xd, setLayer.width() - 1);
yd = qBound(0, yd, setLayer.height() - 1);
```

src/tiled/automapper.cpp
```diff
@@ -631,8 +642,26 @@ static bool layerMatchesConditions(const TileLayer &setLayer,
...
 
-                const Cell &setCell = setLayer.cellAt(x + offset.x(),
-                                                      y + offset.y());
+                int xd = x + offset.x();
+                int yd = y + offset.y();
+
+                if (!isMapInfinite) {
+                    if (options.wrapBorder) {
+                        int width = setLayer.size().width();
+                        int height = setLayer.size().height();
```
bjorn: Please use `setLayer.width()` and `setLayer.height()`. :-)

src/tiled/automapper.cpp
```diff
@@ -631,8 +642,26 @@ static bool layerMatchesConditions(const TileLayer &setLayer,
                         !setLayer.contains(x + offset.x(), y + offset.y()))
                     return false;
 
-                const Cell &setCell = setLayer.cellAt(x + offset.x(),
-                                                      y + offset.y());
+                int xd = x + offset.x();
+                int yd = y + offset.y();
```
bjorn: Now that we have these variables, maybe move them up so that you can also use them in the `contains` check.

src/tiled/automapper.cpp
JoaoBaptMG: Thanks! I have never written a line of Qt code before, so I didn't know of that function. That could be very useful, indeed!

JoaoBaptMG:
>     * Would it make sense to have the OverflowBorder and WrapBorder options imply MatchOutsideMap = true? It feels superfluous to require the user to specify both.

I will incorporate this into my code as such, but I think it might be useful if the user disables MatchOutsideMap but forgets to do so with OverflowBorder/WrapBorder (well, that doesn't make any sense, but I guess it's a possible scenario?)

>     * Maybe instead of passing in whether the map is infinite, we can make sure these options are never enabled for infinite maps.

That could also be possible, and will remove excessive parameter passing.

src/tiled/automapper.cpp
```diff
@@ -124,6 +139,17 @@ bool AutoMapper::setupRuleMapProperties()
...
+    // OverflowBorder and WrapBorder make no sense for infinite maps
+    if (mMapWork->infinite()) {
+        mOptions.overflowBorder = false;
+        mOptions.wrapBorder = false;
+    }
+
+    // Each of the border options imply MatchOutsideMap
+    if (mOptions.overflowBorder || mOptions.wrapBorder)
+        mOptions.matchOutsideMap = true;
+
```
JoaoBaptMG: As requested, I put the options to make the border options imply **MatchOutsideMap** and prevent infinite maps from turnin on the border options.

src/tiled/automapper.cpp
JoaoBaptMG: Moved up.

src/tiled/automapper.cpp
```diff
@@ -627,12 +653,24 @@ static bool layerMatchesConditions(const TileLayer &setLayer,
...
+                // so no "invalid" width/height accessing here.
+                if (options.wrapBorder) {
+                    xd = wrap(xd, setLayer.width());
+                    yd = wrap(yd, setLayer.height());
+                } else if (options.overflowBorder) {
+                    xd = qBound(0, xd, setLayer.width() - 1);
+                    yd = qBound(0, yd, setLayer.height() - 1);
+                }
+
+                const Cell &setCell = setLayer.cellAt(xd, yd);
```
JoaoBaptMG: Incorporated `qBound` as requested and add a helper function `wrap` to eliminate excessive code duplication.

src/tiled/automapper.cpp
```diff
@@ -42,6 +42,11 @@
 
 using namespace Tiled;
 
+// Should i leave this function here?
+inline int wrap(int value, int bound) {
+    return (value % bound + bound) % bound;
+}
+
```
JoaoBaptMG: I haven't found a better place to this function, so I will leave it here.

src/tiled/automapper.cpp
```diff
@@ -627,12 +653,24 @@ static bool layerMatchesConditions(const TileLayer &setLayer,
...
-                const Cell &setCell = setLayer.cellAt(x + offset.x(),
-                                                      y + offset.y());
+                // Those two options are guaranteed to be false if the map is infinite,
+                // so no "invalid" width/height accessing here.
+                if (options.wrapBorder) {
+                    xd = wrap(xd, setLayer.width());
+                    yd = wrap(yd, setLayer.height());
+                } else if (options.overflowBorder) {
+                    xd = qBound(0, xd, setLayer.width() - 1);
+                    yd = qBound(0, yd, setLayer.height() - 1);
```
JoaoBaptMG: Incorporated `qBound` as requested and add a helper function wrap to eliminate excessive code duplication.

src/tiled/automapper.cpp
```diff
@@ -42,6 +42,11 @@
 
 using namespace Tiled;
 
+// Should i leave this function here?
+inline int wrap(int value, int bound) {
```
bjorn: The location is fine for now, but please define the function `static` and put the `{` on the next line.

JoaoBaptMG: `inline static` or `static inline`? I know there's no difference, but maybe there is a convention in Tiled's code?

src/tiled/automapper.cpp
```diff
@@ -42,6 +42,12 @@
 
 using namespace Tiled;
 
+// Should i leave this function here?
+inline static int wrap(int value, int bound)
+{
```
JoaoBaptMG: Change done!

src/tiled/automapper.cpp
bjorn: Actually I doubt there is any precedence here since I don't really see the point of defining a function both `inline` and `static`.

I actually only use `inline` when I need to use it. That is, to avoid multiple-definition problems for function definitions in header files that are outside class definitions. In any other case it's at best a hint that the compiler is free to ignore (and probably will, it'll inline only when it thinks it's a good idea anyway).

The `static` however is appropriate here, since we don't intend the function to be used from another translation unit. Defining it as static makes it inaccessible outside of its translation unit.

bjorn: > Actually I doubt there is any precedence here since I don't really see the point of defining a function both `inline` and `static`.

Maybe I should search the code before raising doubts. So, there's precedence for both `inline static` and `static inline`, with the latter a clear winner. However, of 9 total occurrences, 8 were inherited from 3rd party code. The one remaining one, I'm probably going to change to just `static` soon. :P

JoaoBaptMG: Not a problem, I will leave it just as `static`. Commiting the code again.

src/tiled/automapper.cpp
JoaoBaptMG: Now it's just `static`

bjorn:
@JoaoBaptMG Thanks again for this great improvement!

Would you like to have a go at updating [the documentation](https://github.com/bjorn/tiled/blob/master/docs/manual/automapping.rst#map-properties) for these attributes as well?

JoaoBaptMG:
Yeah, I can do that. However, since the PR is already merged, should I open another PR for it or can I commit it to my fork?

bjorn:
@JoaoBaptMG Yes, just open a new PR! :-)

the-simian:
I'm late to the party here, but let me say this is awesome as can be, and I am so happy to see this work @JoaoBaptMG , @bjorn 

So very cool! WhenI work on my project create-phaser-app I will include this change and make that how the levels work

PR #2152, opened 2019-07-09
Allow tmxrasterizer to support additional map formats 

gauauu:
A simple change that loads the plugin manager, and uses `readMap` from mapformat.h to allow reading other map types in addition to tmx.

I don't have a great grasp of how the plugin system works, so if I should do something different other than 
```
 PluginManager::instance()->loadPlugins();
```
to load the plugins, let me know.  (Or feel free to reject the PR if I'm way off the mark)


src/tmxrasterizer/tmxrasterizer.cpp
```diff
@@ -32,9 +32,11 @@
 #include "imagelayer.h"
 #include "isometricrenderer.h"
 #include "map.h"
+#include "mapformat.h"
```
bjorn: You don't seem to need this include.

src/tmxrasterizer/tmxrasterizer.cpp
```diff
@@ -116,6 +118,8 @@ bool TmxRasterizer::shouldDrawLayer(const Layer *layer) const
 int TmxRasterizer::render(const QString &fileName,
                           const QString &imageFileName)
 {
+    PluginManager::instance()->loadPlugins();
```
bjorn: I'd prefer if this was put in the `main` function rather than in the `TmxRasterizer`.

src/tmxrasterizer/tmxrasterizer.cpp
gauauu: Hmm, the build fails for me without it (MapReader.readMap doesn't seem to look for all readers, so I'm using the readMap from mapformat.cpp).  That said, now that you mention this, I should remove the now-unnecessary MapReader instances that get created.

src/tmxrasterizer/tmxrasterizer.cpp
gauauu: Will do. I'll drop it in right after the app initialization calls in main unless you have another convention of where it should go.

src/tmxrasterizer/tmxrasterizer.cpp
bjorn: Whoops, I didn't realize I put `readMap` in `mapformat.h`!

bjorn:
Great improvement, thanks!

Issue #44, opened 2011-07-27
Auto-refresh of tileset files only works for 1 refresh


Romulas:
It happens only in photoshop (version doesnt seems to make difference), MSPaint is working fine.
Tested with png files. Restarting photoshop has no effect.

fo-fo:
The cause of the issue is in FileSystemWatcher::onFileChanged() (in filesystemwatcher.cpp). When saving from Photoshop, mWatcher->files().contains(path) returns false, but QFile::exists(path) returns false also, so the file is not readded to the watcher.

Not sure what would be a proper way to fix it. One way might be to also watch the directory where the file is/was, and add the file back to the watcher if/when it reappears in the directory.

bjorn:
Thanks for investigating this @fo-fo! At least this gives us a clue about why it is breaking with Photoshop and we can consider ways to fix it, but indeed it seems we have to start tracking the directories also.

PR #2158, opened 2019-07-16
Possibly fixes issue #44 -- Improves how file system watcher handles replaced files 

gauauu:
It appears that when this bug  ( https://github.com/bjorn/tiled/issues/44 ) happens, the watcher starts handling the filesystem change before the file actually gets recreated. So the new version of the file doesn't exist, and never gets added back to the watcher.

This PR moves the functionality of re-adding the file to the watcher into the timeout handler. In my tests, this gives it enough time for the file to exist again, and thus gets re-added.

I'm not well-versed in QT, so if I should be handling the iteration with a different type of Iterator, or use a different style here, please let me know.

### DIFF
--- a/src/libtiled/filesystemwatcher.cpp
+++ b/src/libtiled/filesystemwatcher.cpp
@@ -100,8 +100,6 @@ void FileSystemWatcher::clear()
 
 void FileSystemWatcher::onFileChanged(const QString &path)
 {
-
-
     mChangedFiles.insert(path);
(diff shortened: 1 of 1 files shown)


bjorn:
I've made some tweaks as you can see and I've removed the merge commits. If you need help getting rid of those yourself let me know on Discord. :-)

I'm pretty confident this change will fix issue #44, thanks for looking into it!

Issue #2156, opened 2019-07-12
loading tmx files is unnecessarily slow for big worlds


oncer:
In the project I'm working on right now we have a world file with ~200 TMX maps loaded. When creating a new MapScene, i.e. by clicking on a map that is not currently opened in a tab, we experienced loading times of several seconds.
Any time a map is opened in a new tab, MapScene::refreshScene() is called, which calls DocumentManager::loadDocument for each map in the world.

DocumentManager::loadDocument in turn loops through the instances in Document::sDocumentInstances to look up an already loaded map. As I understand it, sDocumentInstances is a lookup index for tmx files that have been loaded previously, in this case _at least_ all the maps in the current world. Since we are looping through all previously loaded maps inside a loop of all maps of the world, this turns into a classic O(n²) scenario.

The loading time was reduced greatly when we replaced QList with QHash for Document::sDocumentInstances. See [this commit](https://github.com/oncer/tiled-hydra/commit/36ca7c8e0d8be7032b713d426cb53fa8854ca4e5) for reference.

This change alone made a big difference for us, decreasing the loading time of several seconds to less then a second. Maybe it's also possible to re-use the existing world structure when a map file from the same world is loaded, eliminating the need for the loop inside refreshScene in many cases.

bjorn:
That's a great performance analysis and speed improvement!

I think normally the linear search would not have been much of a problem even with 200 maps, but what made it slow is that for each map the path was converted to the canonical file path, which takes time mostly because it requires file system access. Btw, there is still a linear search done at the start of `DocumentManager::loadDocument` over the open files, which I think can just be removed.

Since we should never have the same file loaded multiple times, turning the list into a map should be totally fine. The only worry I have is what happens if you overwrite a loaded file with another open file, since this is a sure way to produce two Document instances with the same canonical file path. Of course, that's a rather rare case and in the end it may not really matter because the map is only used to find existing loaded documents for a given file name, and in this case we probably want to use the last one that was given this file name, which is exactly the one that will be in the map. That said, when the destructor is called it should only remove the entry when it points to the same instance.

Would you like to open a pull request with that change?

> This change alone made a big difference for us, decreasing the loading time of several seconds to less then a second. Maybe it's also possible to re-use the existing world structure when a map file from the same world is loaded, eliminating the need for the loop inside refreshScene in many cases.

[...]

PR #2159, opened 2019-07-16
use QHash instead of QList for Document::sDocumentInstances

oncer:
See issue #2156

### DIFF
--- a/src/tiled/document.cpp
+++ b/src/tiled/document.cpp
@@ -36,21 +36,18 @@ Document::Document(DocumentType type, const QString &fileName,
     : QObject(parent)
     , mType(type)
     , mFileName(fileName)
-    , mCurrentObject(nullptr)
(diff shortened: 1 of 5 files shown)


bjorn:
Thanks for this optimization!

I've pushed some improvements based on https://github.com/bjorn/tiled/issues/2156#issuecomment-511937254, though I did not have time to test it. Can you help with testing the change?

bjorn:
At least I could have made sure it compiles... :-)

oncer:
It compiles for me, and works well for my use-case. I can confirm there is another performance boost when using the cached canonical path.

bjorn:
Yeah, it didn't compile when I first pushed my commit, but I force-pushed a working change after the CI system told me about the problem. :-)

You're welcome and thanks again for bringing this up!

PR #2170, opened 2019-07-23
TMXMapReader can load maps from resources using readMapFromResources

mschroeder-github:
Added that TMXMapReader can read maps from resources instead of only files.

### DIFF
--- a/util/java/libtiled-java/src/main/java/org/mapeditor/core/TileSet.java
+++ b/util/java/libtiled-java/src/main/java/org/mapeditor/core/TileSet.java
@@ -41,17 +41,15 @@ import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.NoSuchElementException;
 import java.util.TreeMap;
-
 import javax.imageio.ImageIO;
 import javax.xml.bind.Marshaller;
 import javax.xml.bind.Unmarshaller;
 import javax.xml.bind.annotation.XmlAccessType;
 import javax.xml.bind.annotation.XmlAccessorType;
 import javax.xml.bind.annotation.XmlRootElement;
-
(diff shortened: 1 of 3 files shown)


bjorn:
From the tests we can see that it is possible to create a `URL` that points to a resource:

https://github.com/bjorn/tiled/blob/233ff4e1e7324b8d7de3f59d2238d1d92a8cd1e1/util/java/libtiled-java/src/test/java/org/mapeditor/io/MapReaderTest.java#L290-L293

Do you think it would be possible to adjust the API to use `URL` instead of `String`, so that the `boolean fromResources` can be avoided?

mschroeder-github:
I also saw it and I tried to do it this way. Somehow my tests failed. But you are right, I tested it again and it could work (also for built fat jar). My test code:
```java
public class Main {
    public static void main(String[] args) throws IOException {
        ClassLoader classLoader = Main.class.getClassLoader(); 
        URL url = classLoader.getResource("path/to/test.txt"); //do not use starting '/' here
        System.out.println(url);
        System.out.println(url.openStream());
    }
}
```
It works when started in IDE:
```
file:/(...)/target/classes/path/to/test.txt
java.io.BufferedInputStream@6d06d69c
```
And it works when using `java -jar fat.jar`:
```
jar:file:/(...)/target/fat.jar!/path/to/test.txt
sun.net.www.protocol.jar.JarURLConnection$JarURLInputStream@6bc7c054
```

This means my implementation could be done easier. Directly pass the `URL` with `Map readMap(URL url)` as you suggest. 

I could make another pull request.

bjorn:
> I could make another pull request.

I think you can just continue the work in this pull request?

bjorn:
Superseded by #2829.

PR #2173, opened 2019-07-24
Link zstd static library instead of dynamic library on windows

Oipo:
For some reason, cpp.staticLibraries also looks for dynamic libraries. Made it so that the dynamic library does not get build on windows, so it cannot be linked. Turns out that there's no need to differentiate between platforms anymore.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -45,7 +45,6 @@ build_script:
   - cd zstd/lib
   - set CC=gcc
   - mingw32-make -j2 libzstd.a
-  - ls -lR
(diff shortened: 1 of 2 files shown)


Oipo:
Would recommend squashing the commits, as I couldn't be bothered using proper commit messages.

bjorn:
Right, I was just testing a similar commit (d4d9b8505fddc6c9f4c5586773f672900d20cfe5). Thanks for cleaning this up!

Oipo:
Happy to help!

PR #2177, opened 2019-08-05
Add note that <object> can have <point> as a child

Zulu-Inuoe:
This is noted in the section for <point>, but <object> was missing this as a possibility.


bjorn:
Thanks!

PR #2182, opened 2019-08-14
Clarify object rotation

Phlosioneer:
Clarify the center of rotation for objects

Also adds a note about rotation for tile-objects.


bjorn:
Sounds good, thanks for providing some more details!

In the future I do hope to get around to supporting different alignment options, especially for tile objects, but it would not change the fact that rotation is applied around their position.

PR #2183, opened 2019-08-14
Add technical details for font display

Phlosioneer:
Explain in detail how fonts are laid out in the object. Also note in the documentation that fonts are clipped outside of the bounds of their object.

These details were determined from how Tiled handles text objects.

For a quick explanation of font terminology, see this image: https://i.stack.imgur.com/ntg5f.png


bjorn:
Sounds good, thanks!

I think it would be nice to add a no-clip option or auto-size, which can be convenient if you only care about the width or if you wrap manually. But indeed for now clipping is always done.

PR #2185, opened 2019-08-18
Specify defaults for name and type

Phlosioneer:
It wasn't clear from the documentation whether the `name` and `type` fields of `object` tags could be safely omitted from TMX files. Giving them a default value makes clear that they can be left out.


bjorn:
Thanks! In general I did try to specify the default for optional properties, so it's a good idea to do it here as well.

PR #2187, opened 2019-08-22
add objectgroup rendering to tmxrasterizer

oncer:
also do not abort rendering if not all tmx files inside a world can be loaded

### DIFF
--- a/src/tmxrasterizer/tmxrasterizer.cpp
+++ b/src/tmxrasterizer/tmxrasterizer.cpp
@@ -89,7 +89,7 @@ void TmxRasterizer::drawMapLayers(MapRenderer &renderer,
 
         const TileLayer *tileLayer = dynamic_cast<const TileLayer*>(layer);
         const ImageLayer *imageLayer = dynamic_cast<const ImageLayer*>(layer);
-        const ObjectGroup *objectGroup = static_cast<const ObjectGroup*>(layer);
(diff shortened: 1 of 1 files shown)


src/tmxrasterizer/tmxrasterizer.cpp
```diff
@@ -89,11 +89,35 @@ void TmxRasterizer::drawMapLayers(MapRenderer &renderer,
 
         const TileLayer *tileLayer = dynamic_cast<const TileLayer*>(layer);
         const ImageLayer *imageLayer = dynamic_cast<const ImageLayer*>(layer);
+        const ObjectGroup *objectGroup = static_cast<const ObjectGroup*>(layer);
```
bjorn: I can see that due to `shouldDrawLayer` excluding the group layer and there only being a limited number of layers, this potentially wrongly cast `ObjectGroup` is guaranteed to be only used when the cast is valid. Yet, I'd still prefer a `dynamic_cast` was used.

Or, since we're now handling almost every type of layer, maybe it's time to just do a `switch` on the layer type as done in the `MiniMapRenderer`.

I guess after this it would be worth moving the `MiniMapRenderer` into `libtiled`, so that it can also be used by the `tmxrasterizer` to avoid the code duplication.

oncer:
Thank you for the hints, I've fixed the code as requested

bjorn:
Thanks!

Issue #1956, opened 2018-05-29
Documentation for generating Visual Studio Solution from qbs


shiyuge:
I am a windows user and I would like to generate a Visual Studio Solution file.
I think there are many developers like me who prefer to use some other 3rd party IDE other than Qt Creator.

Normally, one could generate a vs sln by migrating the `.pro` file from a vs plugin, Qt VS Tools.
This plugin uses qmake. 
But in tiled.pro, I see
```
win* {
    message("Building Tiled for Windows using qmake is no longer supported")
    error("Use the tiled.qbs project file instead")
}
```
so it's impossible to do it this way.

Getting the error message, I looked up qbs online and run these commands:
```
qbs setup-qt --detect
qbs setup-toolchains --detect
qbs generate -g visualstudio2017
```

No error message occurred but I can't find a vs file anywhere. I get one `default.bg.lock` which I believe is some intermediate product in the qbs generate process. I think the process could be like this: the qbs generator locks current profile, then something goes wrong that breaks the generating process leaving only the lock.

I think the root course for my problem could be some qbs bugs, but I can't help but wonder:

1. Why is it that building Tiled for Windows using qmake is no longer supported? Can we restore it?
2. Would it be helpful if we have some additional documentation on how to generate a vs solution file? Many other programmers could have the same demand.

Thank you.

bjorn:
> Why is it that building Tiled for Windows using qmake is no longer supported? Can we restore it?

It is time consuming to maintain multiple build systems, and qmake is more hacky and tricky to support. I don't remember what issue prompted me to put that message there, but in general my goal is to get rid of the qmake project files entirely (issue #1484). Of course, it could be restored somehow, but I think there are better ways to spend the time.

> Would it be helpful if we have some additional documentation on how to generate a vs solution file? Many other programmers could have the same demand.

As far as I know you're the first to try generating VS project files using Qbs for Tiled. I do not know how well this option is supported.

@jakepetroules, would you be able to help us getting that to work? Or is there another way in which the Qbs project files can be used in Visual Studio?

bjorn:
Jake has let me know that he has no time to help out with this issue since he's no longer working on Qbs.

[...]

PR #2188, opened 2019-08-24
Update windows build instructions

Phlosioneer:
Explain that QtCreator is required to build the project, link to the download, and mention the possibility of making a Visual Studio file.

Mentions #1956 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,12 @@ You will still need to install a development environment alongside and some
 libraries depending on your system, for example:
 
 * On Ubuntu/Debian: `sudo apt install build-essential zlib1g-dev libgl1-mesa-dev`
-* On Windows:       Choose "MinGW" Qt version, or install Visual Studio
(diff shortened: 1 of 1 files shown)


bjorn:
I've updated the readme, replacing the qmake instructions with qbs instructions and mentioning Qt Creator. Please feel free to let me know if something doesn't work or needs further instructions.

I don't feel it make sense to mention the option to generate Visual Studio project files, since this is not a supported method of compiling Tiled. Issue #1956 is a bit misleading in that it may be interpreted as if there's just documentation missing, while in fact getting it to work probably requires changes to the Qbs project files or may not work at all depending on which features are actually supported in this mode.

bjorn:
Since the generated Visual Studio 2017 project was reported to work fine in https://github.com/bjorn/tiled/issues/1956#issuecomment-538725686, I've decided to add a note about that after all.

PR #2196, opened 2019-09-25
Fix build warnings on macOS with Qt 5.13.1

mitchcurtis:
- src/libtiled/tileset.cpp:517:19: warning: 'swap' is deprecated: Use QList<T>::swapItemsAt() [-Wdeprecated-declarations]
- src/libtiled/wangset.cpp:596:5: warning: 'qStableSort<QList<Tiled::WangTile>::iterator>' is deprecated: Use std::stable_sort [-Wdeprecated-declarations]
- src/libtiled/logginginterface.h:51:1: warning: 'Issue' defined as a struct here but previously declared as a class [-Wmismatched-tags]
- src/tiled/commanddatamodel.cpp:489:19: warning: 'swap' is deprecated: Use QList<T>::swapItemsAt() [-Wdeprecated-declarations]
- src/tiled/changepolygon.h:100:9: warning: private field 'mObjectIndex' is not used [-Wunused-private-field]
- src/tiled/tiledproxystyle.cpp:126:24: warning: 'find' is deprecated: Use bool find(const QString &, QPixmap *) instead [-Wdeprecated-declarations]
- src/tiled/tiledproxystyle.cpp:373:64: warning: 'foreground' is deprecated: Use QPalette::windowText() instead [-Wdeprecated-declarations]
- src/tiled/tiledproxystyle.cpp:881:49: warning: 'foreground' is deprecated: Use QPalette::windowText() instead [-Wdeprecated-declarations]
- src/libtiled/logginginterface.h:51:1: warning: 'Issue' defined as a struct here but previously declared as a class [-Wmismatched-tags]
- src/tiled/issuesdock.h:31:1: warning: class 'Issue' was previously declared as a struct [-Wmismatched-tags]

### DIFF
--- a/src/libtiled/wangset.cpp
+++ b/src/libtiled/wangset.cpp
@@ -593,7 +593,7 @@ void WangSet::removeWangTile(const WangTile &wangTile)
 QList<WangTile> WangSet::sortedWangTiles() const
 {
     QList<WangTile> wangTiles = mWangIdToWangTile.values();
-    qStableSort(wangTiles.begin(), wangTiles.end());
+    std::stable_sort(wangTiles.begin(), wangTiles.end());
     return wangTiles;
 }
 
--- a/src/tiled/tiledproxystyle.cpp
+++ b/src/tiled/tiledproxystyle.cpp
@@ -123,7 +123,7 @@ static void qt_fusion_draw_arrow(Qt::ArrowType type, QPainter *painter, const QS
     QString cacheKey = uniqueName(QLatin1String("fusion-arrow"), option, rect.size())
             % HexString<uint>(type)
             % HexString<uint>(color.rgba());
-    if (!QPixmapCache::find(cacheKey, cachePixmap)) {
+    if (!QPixmapCache::find(cacheKey, &cachePixmap)) {
         cachePixmap = styleCachePixmap(rect.size());
         cachePixmap.fill(Qt::transparent);
         QPainter cachePainter(&cachePixmap);
@@ -370,7 +370,7 @@ void TiledProxyStyle::drawPrimitive(PrimitiveElement element,
             int state = option->state;
 
             QColor pressedColor = mergedColors(option->palette.base().color(),
-                                               option->palette.foreground().color(), 85);
+                                               option->palette.windowText().color(), 85);
             painter->setBrush(Qt::NoBrush);
 
             // Gradient fill
@@ -878,7 +878,7 @@ void TiledProxyStyle::drawComplexControl(ComplexControl control,
             QColor alphaOutline = outline;
             alphaOutline.setAlpha(180);
 
-            QColor arrowColor = option->palette.foreground().color();
+            QColor arrowColor = option->palette.windowText().color();
             arrowColor.setAlpha(220);
 
             const QColor bgColor = mPalette.color(QPalette::Base);
--- a/src/libtiled/logginginterface.h
+++ b/src/libtiled/logginginterface.h
@@ -48,8 +48,9 @@ class Object;
 class Tile;
 class Tileset;
 
-struct TILEDSHARED_EXPORT Issue
+class TILEDSHARED_EXPORT Issue
 {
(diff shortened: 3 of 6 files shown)


mitchcurtis:
Ah... I probably should have checked which Qt version you're using. :D

src/tiled/tiledproxystyle.cpp
```diff
@@ -370,7 +370,7 @@ void TiledProxyStyle::drawPrimitive(PrimitiveElement element,
             int state = option->state;
 
             QColor pressedColor = mergedColors(option->palette.base().color(),
-                                               option->palette.foreground().color(), 85);
+                                               option->palette.windowText().color(), 85);
```
bjorn: Strange, according to the documentation, `foreground()` was apparently [already obsolete in Qt 5.6](https://doc.qt.io/archives/qt-5.6/qpalette-obsolete.html#foreground) or even earlier, but I'm compiling with `QT_DISABLE_DEPRECATED_BEFORE=0x050900` and this still worked... apparently it's not covered.

src/tiled/tiledproxystyle.cpp
```diff
@@ -123,7 +123,7 @@ static void qt_fusion_draw_arrow(Qt::ArrowType type, QPainter *painter, const QS
     QString cacheKey = uniqueName(QLatin1String("fusion-arrow"), option, rect.size())
             % HexString<uint>(type)
             % HexString<uint>(color.rgba());
-    if (!QPixmapCache::find(cacheKey, cachePixmap)) {
+    if (!QPixmapCache::find(cacheKey, &cachePixmap)) {
```
bjorn: Same here... already [obsolete in Qt 5.6](https://doc.qt.io/archives/qt-5.6/qpixmapcache-obsolete.html#find-1), but still working even with `QT_DISABLE_DEPRECATED_BEFORE=0x050900`.

Btw, I think this change is a bit sad. If you pass in a variable by writable reference, Qt Creator will put it in italic so you can clearly see it may get modified. The `&` was originally meant to make this clearer when we were all coding in Vim or Emacs, but these days I think it makes it less clear. Anyway, enough shedding of tears.

mitchcurtis: Is it worth reporting any of these as issues? At least the `QT_DISABLE_DEPRECATED_BEFORE=0x050900` issue?

bjorn: I guess it's anyway not a good idea to remove functions "retroactively". Since I made my application compile with `QT_DISABLE_DEPRECATED_BEFORE=0x050900`, I would not expect a function that was already deprecated in 5.6 without being wrapped in `QT_DEPRECATED_SINCE` to suddenly disappear when compiling against Qt 5.13. I see they did the expected thing in Qt 5.13, and actually disabled those functions from 5.13 onwards:

```c++
#if QT_DEPRECATED_SINCE(5, 13)
    QT_DEPRECATED_X("Use bool find(const QString &, QPixmap *) instead")
    static QPixmap *find(const QString &key);
    QT_DEPRECATED_X("Use bool find(const QString &, QPixmap *) instead")
    static bool find(const QString &key, QPixmap &pixmap);
#endif
```

src/libtiled/wangset.cpp
```diff
@@ -593,7 +593,11 @@ void WangSet::removeWangTile(const WangTile &wangTile)
 QList<WangTile> WangSet::sortedWangTiles() const
 {
     QList<WangTile> wangTiles = mWangIdToWangTile.values();
+#if QT_VERSION < QT_VERSION_CHECK(5, 13, 0)
     qStableSort(wangTiles.begin(), wangTiles.end());
+#else
+    std::stable_sort(wangTiles.begin(), wangTiles.end());
+#endif
```
bjorn: Since the `std::stable_sort` also works with older Qt versions, there is no need for the Qt version check here.

bjorn:
Thanks!

PR #2198, opened 2019-09-25
Tiled Quick: don't render hidden layers

### LINKED ISSUE #2197: Hidden layers are still rendered in Tiled Quick
I'm using MapItem in my game and noticed this when I tried to hide a layer for the first time in the game.

mitchcurtis:
Fixes #2197.


mitchcurtis:
This doesn't handle changes in visibility after startup, but... it's still an improvement. :)

bjorn:
I'm thinking this may not be the way we want to do this, but I guess it does not hurt to have this for now.

Technically you have a bug in there, that if `layerVisibilityChanged` _would_ get called when visibility changes, then the layer will not have the right visible area. So when becoming visible, this function should call `updateVisibleTiles`.

Why did you go for changing the `ItemHasContents` flag rather than calling `setVisible`?

mitchcurtis:
> I'm thinking this may not be the way we want to do this, but I guess it does not hurt to have this for now.

Haha, I thought you may say something like that. :D

> Technically you have a bug in there, that if `layerVisibilityChanged` _would_ get called when visibility changes, then the layer will not have the right visible area. So when becoming visible, this function should call `updateVisibleTiles`.

Done.

> Why did you go for changing the `ItemHasContents` flag rather than calling `setVisible`?

Dunno. :) I'll change it.

bjorn:
> Haha, I thought you may say something like that. :D

So I've merged it for now, but I'm having a hard time suppressing the itch to refactor it all. :-)

mitchcurtis:
> > Haha, I thought you may say something like that. :D
> 
> So I've merged it for now, but I'm having a hard time suppressing the itch to refactor it all. :-)

I imagine that it's a lack of time that's stopping you and not hurting my feelings, otherwise I'd say don't let that stop you! :D

bjorn:
Right, it's just that getting Tiled 1.3 out is my current priority. :-)

PR #2202, opened 2019-10-02
added some note on how firstgid works

MatthewAlner:
I had to look this up else where when I was dynamically creating maps in the fly for phaser, so I thought id add it.


bjorn:
That's fine, thanks!

In general, the JSON format documentation is very sparse on additional information, since it would largely duplicate what is written on the [TMX Map Format](https://doc.mapeditor.org/en/stable/reference/tmx-map-format/) page. That page has some information on this "firstgid" in [the data section](https://doc.mapeditor.org/en/stable/reference/tmx-map-format/#data).

PR #2203, opened 2019-10-14
Update tiled_sv.ts

Regnander:
Fixed one small typo, which I also fixed on Weblate just in case you would import the translations from there in the future.


bjorn:
Eventually I'd have merged it from Weblate, but thanks for opening a pull request, that's fine too!

I'll merge this into `master` because I do not intend to make another 1.2.x release at this point. Note that soon a major update is going to be needed to all translations when we enter string freeze for 1.3.

Issue #2184, opened 2019-08-15
Add a way to reset the editor scrolling position to X=0, Y=0


thomasgoldstein:
Hello,

Sorry if this has already been suggested. My main grip with Tiled is that I often scroll too much and end up "out of bounds" (like plus or minus huge horizontal/vertical value), and then I struggle to go to the exact position where my map is. My map is tile based (32x32), 80x10 tiles. Nothing special.

Maybe I'm missing something, but I couldn't find a way to easily reset the scrolling position.
Suggestion:
- Make it so clicking the grey square between both scrollbars (at the bottom right) resets the scrolling position to 0,0.
- And/or add a keyboard shortcut to reset the scrolling position to 0,0
- And/or simply make it impossible to scroll far away from the map

bjorn:
Thanks for the suggestion! I think it's not a good idea to prevent scrolling away entirely, but there should definitely be an easy way to get back to the origin both as a shortcut and some button in the UI.

I using the little square in the bottom right is a nice idea, though I could also imagine using that as a clickable area for panning in both directions (but, I know this is already possible with middle-click and spacebar as well...).

cctechwiz:
I'd like to try taking this one if you don't mind.

My initial instinct is that `src/tiled/mapview.cpp(h)` or `src/tiled/mapscene.cpp(h)` would be where I need to look to implement this behavior? At least that seems like where the code for handling those events is at. I'm assuming the little square in question is defined in one of the `.ui` files, I can't seem to identify which one though.

This is really my first time contributing to someone else's project so any guidance or feedback is gladly accepted.

bjorn:
@cctechwiz Hmm, sorry for the late reply! I hope you'd still have time to look into this. Indeed the MapView class could implement the method. I think it would be good to add it as a menu item in the "View" menu, which can be edited in the `mainwindow.ui` file and handled in `mainwindow.cpp` (just check how some of the other actions are connected).

If you run into any other questions, also feel free to ask in the [Tiled Discord channel](https://discord.gg/39wbTv7).

We'll also need to decide what this action should do _exactly_. The request is to scroll to 0,0 but that's the top-left corner of the map. Even though this is the origin, I don't see that being very useful in general. Maybe it would be better to have an action that centers the map, zooming out so you will see the entire map? So a "Fit in View" action essentially. I think that'd also help @thomasgoldstein.

PR #2206, opened 2019-10-21
Added fit in view action

mateosss:
From #2184 discussion.

It implements the "Fit In View" action.

It uses `Ctrl-Shift-J` (like gimp) however it may have been better to use `5` (like inkscape) as `2`, `4`, `6` and `8` are already used for viewport navigation. Unfortunately when assigning `5` as the action shortcut `QAction::event: Ambiguous shortcut overload: 5` is printed and I wasn't able to find the cause.

Another thing to consider is that I reused `zoom-original.png` (same as `actionZoomNormal`), is there a place to get a proper icon for this action?


bjorn:
I've made the changes that I requested and some other tweaks and merged this now, because I'm working towards string freeze for Tiled 1.3.

mateosss:
Sorry I couldn't make time during the week. But that's great thanks!

PR #2211, opened 2019-10-26
Extension Mismatch dialog: Ask in the main text

encukou:
The MessageBox shows Yes/No buttons, but the question it is asking is hidden in [the fine print](https://doc.qt.io/qt-5/qmessagebox.html#informativeText-prop).
I believe it would make more sense to ask questions first, and explain later.

### DIFF
--- a/src/tiled/documentmanager.cpp
+++ b/src/tiled/documentmanager.cpp
@@ -666,13 +666,13 @@ bool DocumentManager::saveDocumentAs(Document *document)
             {
                 QMessageBox messageBox(QMessageBox::Warning,
                                        QCoreApplication::translate("Tiled::MainWindow", "Extension Mismatch"),
-                                       QCoreApplication::translate("Tiled::MainWindow", "The file extension does not match the chosen file type."),
+                                       QCoreApplication::translate("Tiled::MainWindow", "Are you sure you want to save with this extension?"),
                                        QMessageBox::Yes | QMessageBox::No,
                                        mWidget->window());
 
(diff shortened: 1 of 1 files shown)


encukou:
Looking further, I see this happens on other dialogs as well. Do you think it's an issue?

bjorn:
> Looking further, I see this happens on other dialogs as well. Do you think it's an issue?

I guess it could be an issue. I think your suggestion definitely make sense, but it depends a little on what other applications do to see what the user could really expect. I noticed the example in the QMessageBox documentation also asks the question in the informative text:

image

I do think this only really matters on macOS, I'm not sure if the message box has different formatting of the text on other platforms. I also think for Tiled 1.3 it's more important to keep the string freeze now, so I'd merge this after branching.

Feel free to extend this pull request with fixes to the other dialogs you've noticed this on.

encukou:
Ah. If this is usual on Macs, it's probably not worth fixing, though it still seems weird to me. But, I certainly won't be able to prioritize working on this :(
And changing just one dialog is probably worse than not chaging anything
Please re-open if you think it's worth fixing.

bjorn:
I don't know what's usual on macOS. The QMessageBox example may not be ideal either. In the case of having unsaved changes, the usual popup seems rather more helpful:

Screen Shot 2019-10-29 at 11 56 00

Though it is true that the question is in the title, like you are suggesting.

> And changing just one dialog is probably worse than not chaging anything
> Please re-open if you think it's worth fixing.

Yeah, I'll reopen since I think it's worth improving this, but I'll check the other cases of "informative text" before merging.

PR #2212, opened 2019-10-26
issuescounter.cpp: Use %n for plural translations

encukou:
With `tr("%1...", "", n)`, Qt looks for two markers in the translation: `%1` for the argument and `%n` to select the right plural.
Qt Linguist flags translations that don't have `%n` and `%1`.
But, it makes sense to only have the number once in the translation.


bjorn:
Thanks! I messed that one up indeed, had forgotten about the `%n` when using plural forms. It makes sense, because in general the `%n` tells the translator the location of the number.

PR #2214, opened 2019-10-26
Preferences dialog: Stretch the wheelZoomsByDefault checkbox label

encukou:
If the translation is long, it would stretch the dialog.

This makes the label the same size as the accelerated rendering one.


bjorn:
Thanks for the fix and for paying attention to detail!

PR #2215, opened 2019-10-26
Update Czech translation

encukou:
I missed a few new versions, here's me catching up.

One thing I haven't figured out is what "File→Paste in Place" does (and thus how to best translate it). Could I get a hint?


bjorn:
> I missed a few new versions, here's me catching up.

Thank you for the huge update, impressive work!

> One thing I haven't figured out is what "File→Paste in Place" does (and thus how to best translate it). Could I get a hint?

Is pastes the copied contents in exactly the place where it was copied from. Works for both tile and object selections. It's useful for copying stuff to another layer or another map while not changing its position.

bjorn:
@encukou Do you consider this translation update done or did you want to update the translation for "Paste in Place" still?

Btw, don't worry about the conflicts. I made a few minor changes and removed lots of strings, but I can resolve those.

PR #2218, opened 2019-10-29
Added utf-8 encoding support for script file

HonestQiao:
Added method to auto check whether the script file is UTF-8 encoded.


bjorn:
Alright, so looks like this could mean that now UTF-8 is supported alongside other UTF formats, and finally Latin-1 encoded scripts, as long as they trigger "invalid character" errors. I've tweaked the code a little to simplify it and to avoid a second decoding for UTF-8 files. Thanks!

PR #2219, opened 2019-10-30
Completed Simplified Chinese translation. Done a whole review and testing.

HonestQiao:


bjorn:
What a great update, thanks a lot! I was a bit surprised about the extra words in the About dialog, but I think it's a nice place to explain some of the terminology. :-)

PR #2679, opened 2019-11-23
Fix QVariantMap linker error

mitchcurtis:
https://wiki.qt.io/Coding_Conventions says not to derive from template/tool classes.

The error message was:

tiled.lib(tiled.dll) : error LNK2005: "public: __cdecl QMap<class QString,class QVariant>::~QMap<class QString,class QVariant>(void)" (??1?$QMap@VQString@@VQVariant@@@@QEAA@XZ) already defined in SpriteSequence.cpp.obj
   Creating library C:\dev\tshnm-iso-qt5_14_d-Debug\Debug\libisle.7c81f238\isle.lib and object C:\dev\tshnm-iso-qt5_14_d-Debug\Debug\libisle.7c81f238\isle.exp
C:\dev\tshnm-iso-qt5_14_d-Debug\Debug\libisle.7c81f238\isle.dll : fatal error LNK1169: one or more multiply defined symbols found

### DIFF
--- a/src/libtiled/maptovariantconverter.cpp
+++ b/src/libtiled/maptovariantconverter.cpp
@@ -341,8 +341,8 @@ QVariant MapToVariantConverter::toVariant(const Properties &properties) const
 {
     QVariantMap variantMap;
 
-    Properties::const_iterator it = properties.constBegin();
-    Properties::const_iterator it_end = properties.constEnd();
+    auto it = properties.map().constBegin();
+    const auto it_end = properties.map().constEnd();
     for (; it != it_end; ++it) {
         const QVariant value = toExportValue(it.value(), mDir);
         variantMap[it.key()] = value;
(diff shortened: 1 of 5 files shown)


mitchcurtis:
Note that I didn't test anything other than libtiled and tiledquickplugin as I no longer build all the other stuff.

bjorn:
> Note that I didn't test anything other than libtiled and tiledquickplugin as I no longer build all the other stuff.

Yeah, all checks are failing. :-)

I would propose not to do this by composition since too many other parts of the code need to be changed. Instead, my proposal would be to do:
```c++
using Properties = QVariantMap;
```
And then defining the few extra members that the `Properties` subclass had as stand-alone functions. Something like:
```c++
TILEDSHARED_EXPORT void mergeProperties(Properties &into, const Properties &from);
TILEDSHARED_EXPORT QJsonArray propertiesToJson(const Properties &properties);
TILEDSHARED_EXPORT Properties propertiesFromJson(const QJsonArray &json);
```
And for consistency I think we should do the same with `AggregatedProperties`.

mitchcurtis:
> Yeah, all checks are failing. :-)

You've gotta break some eggs to make an omelette! :D

> And for consistency I think we should do the same with AggregatedProperties.

How would that work? It doesn't derive from QVariantMap; it uses: `QMap<QString, AggregatedPropertyData>`

bjorn:
> How would that work? It doesn't derive from QVariantMap; it uses: `QMap<QString, AggregatedPropertyData>`

Yes, but it also _derives_ from it, which is the problem. If any other lib used libtiled and needed to destroy a `QMap<QString, AggregatedPropertyData>`, it would probably run into the same symbol clashes on its destructor. This case would be pretty rare, hence I suggested to replace that class with `using AggregatedProperties = QMap<QString, AggregatedPropertyData>` for consistency only.

PR #2687, opened 2019-11-29
Added layer tint color

### LINKED ISSUE #2683: Feature: Allow assigning a tint color to a layer
Feature Description:
It would be useful to have a functionality that allows tinting each individual layer with a chosen color, in a similar way to what some game engines do. This could replace the opacity setting for each layer (since a tint color will already have an alpha channel), could be another widget unrelated to the opacity slider or could be a property of the layer. Tint colors in a group layer would stack with the ones of the child layers, in the same manner it works with the opacity setting.

Problem:
Godot, Unity and Unreal (and I suppose many other game engines) provide default shaders that allow tinting sprites with a given color by multiplying the tint color with the fragment shader output color. This is usefull to tweak map layer colors without changing the asset itself and is also useful for effects like having transparency in foreground layers or darkening background layers.

In a personal project I'm using the layer's custom properties to define this tint color which I assign to the tint color of the layer in the engine uppon importing the tmx file. Although this works fine, It would be more convenient to see the layers tinted in tiled instead of having to tint them in the engine and then copying back the color value to tiled.

I'm willing to try to implement this myself and make a pull request but since I'm very new to Qt and I'm also not familiar with tiled's codebase I just couldn't find yet where things are and what I must do.

After some messing with the codebase I stumbled upppon src/libtiled/maprenderer.cpp wich seems to hold the rendering logic. At first I thought I should modify CellRenderer::render to change the tile rendering. I googled a bit to found that there is a QPainter::setCompositionMode which would make the multiplying effect I want. If I understood right, after drawing the tile
```
mPainter->drawPixmap(target, image, source);
```
I should change the composition mode and draw again another thing on top of it.
```
mPainter->drawPixmap(target, image, source);
mPainter->setCompositionMode(QPainter::CompositionMode_Multiply);
QImage tintColor(imageSize.width(), imageSize.height(), QImage::Format_ARGB32);
QPixmap tintPix = QPixmap::fromImage(tintColor);
tintColor.fill(0XFFFFFF00);
mPainter->drawPixmap(target, tintPix, source);
```
But this change simply does not seem have any effect.

It could be because I'm doing the wrong thing in the right place, or the right thing in the wrong place or the wrong thing in the wrong place. Anyways I would appreciate any tip as to how make this change.

Gnumaru:
Initial implementation for tinting a tile layer. the matter of stacking the tint of group layers in the same manner it happens with opacity in the current version of tiled was not handled. Also, tinting non tile layers was also not handled.

Closes #2683

### DIFF
--- a/src/libtiled/isometricrenderer.cpp
+++ b/src/libtiled/isometricrenderer.cpp
@@ -33,6 +33,7 @@
 #include "tile.h"
 #include "tilelayer.h"
 #include "tileset.h"
+#include "objectgroup.h"
 
 #include <QtMath>
 
@@ -304,7 +305,7 @@ void IsometricRenderer::drawTileLayer(QPainter *painter,
     // Determine whether the current row is shifted half a tile to the right
     bool shifted = inUpperHalf ^ inLeftHalf;
 
-    CellRenderer renderer(painter, this);
+    CellRenderer renderer(painter, this, layer->tintColor());
 
     for (int y = startPos.y() * 2; y - tileHeight * 2 < rect.bottom() * 2;
          y += tileHeight)
@@ -372,7 +373,7 @@ void IsometricRenderer::drawMapObject(QPainter *painter,
         const QSizeF size = object->size();
         const QPointF pos = pixelToScreenCoords(object->position());
 
-        CellRenderer(painter, this).render(cell, pos, size,
+        CellRenderer(painter, this, object->objectGroup()->tintColor()).render(cell, pos, size,
                                            CellRenderer::BottomCenter);
 
         if (testFlag(ShowTileObjectOutlines)) {
--- a/src/libtiled/isometricrenderer.cpp
+++ b/src/libtiled/isometricrenderer.cpp
@@ -305,7 +305,7 @@ void IsometricRenderer::drawTileLayer(QPainter *painter,
     // Determine whether the current row is shifted half a tile to the right
     bool shifted = inUpperHalf ^ inLeftHalf;
 
-    CellRenderer renderer(painter, this, layer->tintColor());
+    CellRenderer renderer(painter, this, layer->effectiveTintColor());
 
     for (int y = startPos.y() * 2; y - tileHeight * 2 < rect.bottom() * 2;
          y += tileHeight)
@@ -373,7 +373,7 @@ void IsometricRenderer::drawMapObject(QPainter *painter,
         const QSizeF size = object->size();
         const QPointF pos = pixelToScreenCoords(object->position());
 
-        CellRenderer(painter, this, object->objectGroup()->tintColor()).render(cell, pos, size,
+        CellRenderer(painter, this, object->objectGroup()->effectiveTintColor()).render(cell, pos, size,
                                            CellRenderer::BottomCenter);
 
         if (testFlag(ShowTileObjectOutlines)) {
--- a/src/libtiled/maprenderer.cpp
+++ b/src/libtiled/maprenderer.cpp
@@ -115,6 +115,7 @@ void MapRenderer::drawImageLayer(QPainter *painter,
 {
     Q_UNUSED(exposed)
 
+    // TODO: Support layer tint color
     painter->drawPixmap(QPointF(), imageLayer->image());
 }
 
--- a/src/libtiled/maprenderer.cpp
+++ b/src/libtiled/maprenderer.cpp
@@ -115,8 +115,7 @@ void MapRenderer::drawImageLayer(QPainter *painter,
 {
     Q_UNUSED(exposed)
 
-    // TODO: Support layer tint color
-    painter->drawPixmap(QPointF(), imageLayer->image());
+    painter->drawPixmap(QPointF(), tinted(imageLayer->image(), imageLayer->effectiveTintColor()));
 }
 
 void MapRenderer::drawPointObject(QPainter *painter, const QColor &color) const
--- a/src/libtiled/maprenderer.cpp
+++ b/src/libtiled/maprenderer.cpp
@@ -48,6 +48,35 @@
 
 using namespace Tiled;
 
+static QPixmap tinted(const QPixmap &pixmap, const QColor &color)
+{
+    if (!color.isValid() || color == QColor(255, 255, 255, 255))
+        return pixmap;
+
+    QPixmap resultImage = pixmap;
+    QPainter painter(&resultImage);
+
+    QColor fullOpacity = color;
+    fullOpacity.setAlpha(255);
+    // tint the final color (this will will mess up the alpha which we will fix
+    // in the next lines)
+    painter.setCompositionMode(QPainter::CompositionMode_Multiply);
+    painter.fillRect(resultImage.rect(), fullOpacity);
+
+    // apply the original alpha to the final image
+    painter.setCompositionMode(QPainter::CompositionMode_DestinationIn);
+    painter.drawPixmap(0, 0, pixmap);
+
+    // apply the alpha of the tint color so that we can use it to make the image
+    // transparent instead of just increasing or decreasing the tint effect
+    painter.setCompositionMode(QPainter::CompositionMode_DestinationIn);
+    painter.fillRect(resultImage.rect(), color);
+
+    painter.end();
+
+    return resultImage;
+}
+
 MapRenderer::~MapRenderer()
 {}
 
@@ -223,12 +252,13 @@ static bool hasOpenGLEngine(const QPainter *painter)
             type == QPaintEngine::OpenGL2);
 }
 
-CellRenderer::CellRenderer(QPainter *painter, const MapRenderer *renderer, CellType cellType)
+CellRenderer::CellRenderer(QPainter *painter, const MapRenderer *renderer, const QColor &tintColor, CellType cellType)
     : mPainter(painter)
     , mRenderer(renderer)
     , mTile(nullptr)
     , mIsOpenGL(hasOpenGLEngine(painter))
     , mCellType(cellType)
+    , mTintColor(tintColor)
 {
 }
 
@@ -335,7 +365,7 @@ void CellRenderer::render(const Cell &cell, const QPointF &pos, const QSizeF &si
     const QRectF source(0, 0, fragment.width, fragment.height);
 
     mPainter->setTransform(transform);
-    mPainter->drawPixmap(target, image, source);
+    mPainter->drawPixmap(target, tinted(image, mTintColor), source);
     mPainter->setTransform(oldTransform);
 
     // A bit of a hack to still draw tile collision shapes when requested
@@ -360,7 +390,7 @@ void CellRenderer::flush()
 
     mPainter->drawPixmapFragments(mFragments.constData(),
                                   mFragments.size(),
-                                  mTile->image());
+                                  tinted(mTile->image(), mTintColor));
 
     if (mRenderer->flags().testFlag(ShowTileCollisionShapes)
             && mTile->objectGroup()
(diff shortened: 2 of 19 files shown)


src/libtiled/maprenderer.cpp
```diff
@@ -335,7 +350,17 @@ void CellRenderer::render(const Cell &cell, const QPointF &pos, const QSizeF &si
     const QRectF source(0, 0, fragment.width, fragment.height);
 
     mPainter->setTransform(transform);
-    mPainter->drawPixmap(target, image, source);
+    QPixmap tintedimg = image;
+    if(layer){
```
bjorn: When is layer not a valid pointer? Could it be turned into a reference?

src/libtiled/maprenderer.cpp
```diff
@@ -335,7 +350,17 @@ void CellRenderer::render(const Cell &cell, const QPointF &pos, const QSizeF &si
...
+        QColor tintColor = layer->tintColor();
+        if(tintColor != QColor(255, 255, 255, 255)){
+            tintedimg = QPixmap::fromImage(tinted(QImage(image.toImage())
```
bjorn: Instead of passing `QImage(image.toImage())` here, please take `QPixmap` in the `tinted` function and use `drawPixmap` there instead of `drawImage`. I doubt it matters for performance, but it's less code.

src/libtiled/maprenderer.cpp
```diff
@@ -335,7 +350,17 @@ void CellRenderer::render(const Cell &cell, const QPointF &pos, const QSizeF &si
...
+                                                    ,layer->tintColor()
```
bjorn: Either use the local `tileColor` here or use `layer->tintColor` also in the condition. I would probably remove the local `tintColor`.

src/libtiled/maprenderer.cpp
```diff
@@ -259,7 +274,7 @@ void CellRenderer::render(const Cell &cell, const QPointF &pos, const QSizeF &si
     // The USHRT_MAX limit is rather arbitrary but avoids a crash in
     // drawPixmapFragments for a large number of fragments.
     if (mTile != tile || mFragments.size() == USHRT_MAX)
```
bjorn: In theory we need to flush as well when the layer, or at least the tint color, changed. In practice I don't think a `CellRenderer` instance is ever used with different layers so the tint color won't change. As such, I would actually suggest we add the tint color as constructor parameter to the `CellRenderer` instead.

src/tiled/changelayer.cpp
```diff
@@ -121,6 +121,29 @@ SetLayerOpacity::SetLayerOpacity(Document *document,
...
+          mLayer == o->mLayer))
+        return false;
+
+    mNewTintColor = o->mNewTintColor;
+    return true;
+}
+
+SetLayerTintColor::SetLayerTintColor(Document *document,
+                                 Layer *layer,
+                                 QColor tintColor)
```
bjorn: Please place the constructor before the methods.

src/tiled/propertybrowser.cpp
```diff
@@ -1586,6 +1590,8 @@ void PropertyBrowser::addProperties()
         setExpanded(items(colorProperty).constFirst(), false);
     if (QtProperty *fontProperty = mIdToProperty.value(FontProperty))
         setExpanded(items(fontProperty).constFirst(), false);
+    if (QtProperty *tintColorProperty = mIdToProperty.value(TintColorProperty))
+                setExpanded(items(tintColorProperty).constFirst(), false);
```
bjorn:
```suggestion
        setExpanded(items(tintColorProperty).constFirst(), false);
```

src/libtiled/maprenderer.cpp
```diff
@@ -48,6 +48,21 @@
...
 using namespace Tiled;
 
+// from https://github.com/anak10thn/qt5-tutorials/blob/master/imagetint/imagetint.cpp
+// Tint an image with the specified color and return it as a new image
+QImage tinted(const QImage &image, const QColor &color, QPainter::CompositionMode mode = QPainter::CompositionMode_Screen)
+{
+    QImage resultImage(image.size(), QImage::Format_ARGB32_Premultiplied);
+    QPainter painter(&resultImage);
+    painter.drawImage(0, 0, image);
+    painter.setCompositionMode(mode);
```
bjorn: Since I don't think we're planning to support other composition modes, please just put `QPainter::CompositionMode_Multiply` here.

src/tiled/changelayer.cpp
Gnumaru: done

src/libtiled/maprenderer.cpp
Gnumaru: done

src/libtiled/maprenderer.cpp
Gnumaru: done

src/libtiled/maprenderer.cpp
```diff
@@ -48,6 +48,21 @@
...
+    painter.fillRect(resultImage.rect(), color);
+    painter.end();
+    resultImage.setAlphaChannel(image.alphaChannel());
```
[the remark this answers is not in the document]
Gnumaru: done

src/libtiled/maprenderer.cpp
Gnumaru: Now I'm storing the tintcolor in the cellrenderer as you said, so passing around a pointer or reference to the layer becomes unnecessary.

src/libtiled/maprenderer.cpp
Gnumaru: done

src/libtiled/isometricrenderer.cpp
```diff
@@ -401,6 +401,7 @@ void IsometricRenderer::drawMapObject(QPainter *painter,
             painter->setPen(pen);
             painter->drawRect(rect);
             pen.setStyle(Qt::DotLine);
+            // since now layers will have a tint color, should we remove the object layer color property for objects and use the layer tintcolor here? or since object layers already have a color property, shoul we stick with the name Color and rename tintColor to color? aniways the layer color would have to be provided by the caller of IsometricRenderer::drawMapObject
```
bjorn: It makes sense to keep tint color separate from this color since they do entirely different things.

I believe the layer tint color can be obtained through the object using `object->objectGroup()->tintColor()`, as per previous suggested change.

src/tiled/changelayer.cpp
```diff
@@ -138,7 +166,6 @@ void SetLayerOpacity::setOpacity(qreal opacity)
     emit mDocument->changed(LayerChangeEvent(mLayer, LayerChangeEvent::OpacityProperty));
 }
 
-
```
bjorn: There are two empty lines here because the definitions above and below and for different classes. Please keep it that way. :-)

src/libtiled/orthogonalrenderer.cpp
```diff
@@ -311,7 +311,7 @@ void OrthogonalRenderer::drawTileLayer(QPainter *painter,
     const QTransform savedTransform = painter->transform();
     painter->translate(layerPos);
 
-    CellRenderer renderer(painter, this);
+    CellRenderer renderer(painter, this, layer->tintColor(), CellRenderer::CellType::OrthogonalCells);
```
bjorn: No need to add `CellRenderer::CellType::OrthogonalCells` here since that the default value.

src/tiled/changelayer.cpp
Gnumaru: done

src/libtiled/maprenderer.cpp
```diff
@@ -335,7 +350,17 @@ void CellRenderer::render(const Cell &cell, const QPointF &pos, const QSizeF &si
     const QRectF source(0, 0, fragment.width, fragment.height);
 
     mPainter->setTransform(transform);
-    mPainter->drawPixmap(target, image, source);
+
+    if(mTintColor != QColor(255, 255, 255, 255)){
```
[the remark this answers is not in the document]
Gnumaru: done

src/libtiled/orthogonalrenderer.cpp
```diff
@@ -391,7 +391,7 @@ void OrthogonalRenderer::drawMapObject(QPainter *painter,
 
     if (!cell.isEmpty()) {
         const QSizeF size = object->size();
-        CellRenderer(painter, this).render(cell, QPointF(), size,
+        CellRenderer(painter, this, QColor(255, 255, 255, 255)).render(cell, QPointF(), size,
```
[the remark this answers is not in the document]
Gnumaru: done

src/libtiled/isometricrenderer.cpp
```diff
@@ -372,7 +372,7 @@ void IsometricRenderer::drawMapObject(QPainter *painter,
         const QSizeF size = object->size();
         const QPointF pos = pixelToScreenCoords(object->position());
 
-        CellRenderer(painter, this).render(cell, pos, size,
+        CellRenderer(painter, this, QColor(255, 255, 255, 255)).render(cell, pos, size,
```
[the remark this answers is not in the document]
Gnumaru: done

src/libtiled/isometricrenderer.cpp
Gnumaru: done

src/tiled/changelayer.cpp
```diff
@@ -108,6 +108,7 @@ void SetLayerLocked::swap()
     emit mDocument->changed(LayerChangeEvent(mLayer, LayerChangeEvent::LockedProperty));
 }
 
+
```
bjorn: Ha, thank you for caring about consistency!

Gnumaru: =)

bjorn:
That's really annoying that you had to do this trick with the alpha layer... I wonder why [the multiply composition mode is implemented like that](https://code.woboq.org/qt5/qtbase/src/gui/painting/qcompositionfunctions.cpp.html#_ZL23comp_func_Multiply_implPjPKjiRKT_) rather than just multiplying the alpha channel as well. I'm having trouble thinking of a use-case for this "mixing" of alpha channels:
```c++
static inline int mix_alpha(int da, int sa)
{
    return 255 - ((255 - sa) * (255 - da) >> 8);
}
```
Though I'm probably missing something just because it's not what we are looking for here.

In any case, I've pushed some simplifications, also making sure to ignore the tint color when it is not specified (in which case it is now "invalid" rather than white). This makes the "reset" button in the Properties view trivially work and could simplify the map reader and writer a little, which incidentally is I think the only part still missing in this change (and adjusting the documentation of course).

Gnumaru:
I have just updated my fork with the recent upstream changes.

Gnumaru:
I would say yes to both questions. Image layers are useful to implement static or slow moving, non tiled, backgrounds and tinting them has the same uses as tinting a conventional tile layer. On the other hand, separating in different layers something that in-game seems to be the same thing makes sense for things like fake walls where the character can enter behind the wall in a specific point and then part of the wall becomes transparent, so we can see the secrets behind. In this case, if one wants to tint both tile layers it would be easier to put them in a group and tint the group instead of having to copy the tint color from one layer to the other.

bjorn:
@Gnumaru Would you be interested in implementing those changes as part of this pull request?

Gnumaru:
I'll give it a try.

Gnumaru:
I implemented both the tint color inheritance on groups and the tinting of image layers. Fortunately the changes I had to do were little. I also squashed the commits but only now I realized that you should have reviewed the changes before I squashed them, I'm sorry.

bjorn:
@Gnumaru I assumed that commit 8575836212de8bb63aee524ad6322b9da40fc817 included all your changes and that the squashing didn't include additional changes, and have restored your master branch to that commit (and added a commit message). :-)

Gnumaru:
You assumed right, there were no changes after commit 8575836.

bjorn:
Thank you for this nice addition!

Gnumaru:
I'm really pleased to be able to contribute to Tiled =)

bjorn:
@Gnumaru https://itch.io/post/1144498 :-)

PR #2691, opened 2019-12-07
add ubuntu compiling tips

odys-z:
see last section in README.md


odys-z:
It's only doc modification. The build fail log is :
The command "macdeployqt -verbose=2 install/Tiled.app" exited with 1.
I guess it's checking IOS deploying / compatibility failed. 
What should I do ? Is there any pull request guide line?

bjorn:
I don't really understand why some of these instructions are needed.

1. If it's unclear which `tiled.qbs` should be opened, we can extend the existing mention of that file to include "in the root of the repository".

2. What's "tilelib" and why should it be added as a dependency? If there's a build issue related to a missing dependency please open a pull request for adding it instead of for adding such an instruction. Also, the link goes to documentation about qmake whereas the rest of the instructions use Qbs.

3. You need at least Qt 5.6, and indeed Ubuntu 16.04 ships with 5.5.1. But this is not the compiler! You can indeed install a more recent Qt either through [a PPA](https://launchpad.net/~beineri) or by using the Qt installer (which is already mentioned as an alternative). I would recommend most people to install Qt using its installer, so I'm not sure if we really need this instruction (also since the PPA will require further manual setup, and you have noticed and only partially illustrated with the screenshot).

In general I'm reluctant to populate the readme with very specific instructions since it will just be noise for the majority of readers. Ubuntu 16.04 LTS was superseded by 18.04 LTS and only used by few people, let alone developers.

PR #2699, opened 2019-12-15
Add file watching for template files

### LINKED ISSUE #2676: Reload object templates when they are changed on disk
Object template files are currently not watched for changes, so when the files are changed externally Tiled simply ignores it. More useful behavior would be to automatically reload the template and reflect those changes in any instances of that template.

Phlosioneer:
Whenever a template file changes, all objects that reference that
file are synced.

Closes #2676

### DIFF
--- a/src/libtiled/map.h
+++ b/src/libtiled/map.h
@@ -412,6 +412,11 @@ public:
     QSet<SharedTileset> usedTilesets() const;
 
     /**
+     * Returns a list of MapObjects to be updated by syncing objectTemplate
+     */
+    QList<MapObject*> syncObjectTemplate(const ObjectTemplate *objectTemplate);
+
+    /**
      * Returns a list of MapObjects to be updated in the map scene
      */
     QList<MapObject*> replaceObjectTemplate(const ObjectTemplate *oldObjectTemplate,
--- a/src/libtiled/map.h
+++ b/src/libtiled/map.h
@@ -412,11 +412,6 @@ public:
     QSet<SharedTileset> usedTilesets() const;
 
     /**
-     * Returns a list of MapObjects to be updated by syncing objectTemplate
-     */
-    QList<MapObject*> syncObjectTemplate(const ObjectTemplate *objectTemplate);
-
-    /**
      * Returns a list of MapObjects to be updated in the map scene
      */
     QList<MapObject*> replaceObjectTemplate(const ObjectTemplate *oldObjectTemplate,
--- a/src/libtiled/templatemanager.cpp
+++ b/src/libtiled/templatemanager.cpp
@@ -45,13 +45,17 @@ void TemplateManager::deleteInstance()
 }
 
 TemplateManager::TemplateManager(QObject *parent)
-    : QObject(parent)
+    : QObject(parent),
+      mWatcher(new FileSystemWatcher(this))
 {
+    connect(mWatcher, &FileSystemWatcher::fileChanged,
+            this, &TemplateManager::fileChanged);
 }
 
 TemplateManager::~TemplateManager()
 {
     qDeleteAll(mObjectTemplates);
+    delete mWatcher;
 }
 
 ObjectTemplate *TemplateManager::loadObjectTemplate(const QString &fileName, QString *error)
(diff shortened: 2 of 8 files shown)


src/libtiled/templatemanager.cpp
```diff
@@ -45,13 +45,17 @@ void TemplateManager::deleteInstance()
...
+      mWatcher(new FileSystemWatcher(this))
 {
+    connect(mWatcher, &FileSystemWatcher::fileChanged,
+            this, &TemplateManager::fileChanged);
 }
 
 TemplateManager::~TemplateManager()
 {
     qDeleteAll(mObjectTemplates);
+    delete mWatcher;
```
bjorn: There's no need to delete the watcher explicitly, because you've passed the `TemplateManager` as its parent in the constructor.

src/libtiled/templatemanager.cpp
```diff
@@ -61,14 +65,28 @@ ObjectTemplate *TemplateManager::loadObjectTemplate(const QString &fileName, QSt
...
 }
+
+void TemplateManager::fileChanged(const QString &fileName)
+{
+    ObjectTemplate *objectTemplate = findObjectTemplate(fileName);
+
+    // Most likely the file was removed.
+    if (!objectTemplate)
+        return;
+
```
bjorn: So, the template file changed, and you found a previously loaded template, but then it's time to reload this template, right? I don't see this happening currently.

Probably needs something like:
```c++
auto newTemplate = readObjectTemplate(fileName);
if (newTemplate) {
    objectTemplate->setObject(newTemplate->object());
    emit objectTemplateChanged(objectTemplate);
}
```
In case or error we should probably report it using the `ERROR` function, which will show up in the Console and Issues views.

src/libtiled/map.h
```diff
@@ -411,6 +411,11 @@ class TILEDSHARED_EXPORT Map : public Object
      */
     QSet<SharedTileset> usedTilesets() const;
 
+    /**
+     * Returns a list of MapObjects to be updated by syncing objectTemplate
+     */
+    QList<MapObject*> syncObjectTemplate(const ObjectTemplate *objectTemplate);
```
bjorn: Function should not be necessary, this is already implemented as `MapDocument::updateTemplateInstances`.

src/tiled/documentmanager.h
```diff
@@ -196,6 +196,7 @@ public slots:
     void hideChangedWarning();
 
     void tilesetImagesChanged(Tileset *tileset);
+    void objectTemplateChanged(ObjectTemplate *objectTemplate);
```
bjorn: This function should not be necessary, since each `MapDocument` already connects to `TemplateManager::objectTemplateChanged`.

src/tiled/mapdocument.h
```diff
@@ -180,6 +180,7 @@ class MapDocument : public Document
                          QVector<SharedTileset> *missingTilesets = nullptr,
                          QHash<TileLayer *, QRegion> *paintedRegions = nullptr);
 
+    void syncObjectTemplate(const ObjectTemplate *objectTemplate);
```
bjorn: There's already `updateTemplateInstances` for this.

src/libtiled/templatemanager.cpp
```diff
@@ -61,14 +65,28 @@ ObjectTemplate *TemplateManager::loadObjectTemplate(const QString &fileName, QSt
...
 
     return objectTemplate;
 }
+
+void TemplateManager::fileChanged(const QString &fileName)
+{
+    ObjectTemplate *objectTemplate = findObjectTemplate(fileName);
+
+    // Most likely the file was removed.
+    if (!objectTemplate)
```
bjorn: File removal does not remove the template from `mObjectTemplates`.

Actually at the moment nothing does, so this check shouldn't be needed... but it doesn't hurt either.

src/libtiled/templatemanager.cpp
Phlosioneer: I'm still learing the Qt memory rules and such. I'll remove that.

src/libtiled/templatemanager.cpp
Phlosioneer: Whoops! That was a bad oversight. I was too focused on making all the right integrations.

src/libtiled/map.h
Phlosioneer: updateTileInstances asserts that the before- and after-template instances are different.

Phlosioneer: Nope, I was wrong, that was just dumb. I don't know how I missed that.

src/libtiled/templatemanager.cpp
Phlosioneer: I think it's prudent to keep the check, just in case something else changes in the rest of the codebase.

bjorn: Yeah, like I said it doesn't hurt, but the misleading comment should be removed in any case. :-)

src/libtiled/templatemanager.cpp
```diff
@@ -61,14 +65,34 @@ ObjectTemplate *TemplateManager::loadObjectTemplate(const QString &fileName, QSt
     if (!objectTemplate) {
         auto newTemplate = readObjectTemplate(fileName, error);
 
-        // This instance will not have an object. It is used to detect broken
-        // template references.
-        if (!newTemplate)
+        if (newTemplate) {
```
bjorn: Actually if the template failed to load, but the file exists anyway, we'll probably want to watch it too. Maybe just keep the code as-is and just add something like:

```c++
if (QFile::exists(fileName))
    mWatcher->addPath(fileName);
```

bjorn:
Thanks!

Issue #920, opened 2015-04-16
Optimize minimap updates


bjorn:
When editing large maps the minimap is not performing well enough, causing annoying pauses while the minimap is repainting. Possible fixes:

* Move the updating of the minimap into a thread. This would require the map data and used tilesets to be cloned to prevent changes to it to crash the minimap rendering thread.

* Only redraw the actually changing areas. For most operations this would make the minimap updates really quick since they tend to affect only a small area.

PR #2701, opened 2019-12-18
Map loading speedups

Phlosioneer:
While trying to tackle #920, I had issues loading my test files. So I found these performance improvements in the CSV decoder and the first render of the map.

I wasn't able to make any more progress in 920, so this PR is just the speedups I found on the way.

### DIFF
--- a/src/libtiled/mapreader.cpp
+++ b/src/libtiled/mapreader.cpp
@@ -942,33 +942,45 @@ void MapReaderPrivate::decodeCSVLayerData(TileLayer &tileLayer,
                                           QStringRef text,
                                           QRect bounds)
 {
-    QString trimText = text.trimmed().toString();
-    QVector<QStringRef> tiles = trimText.splitRef(QLatin1Char(','));
-
-    int lengthCheck = bounds.width() * bounds.height();
-
-    if (tiles.length() != lengthCheck) {
-        xml.raiseError(tr("Corrupt layer data for layer '%1'")
-                       .arg(tileLayer.name()));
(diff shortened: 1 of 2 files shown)


Phlosioneer:
Ah, subscript on QStringRef was added Qt 5.7, but we need to support Qt 5.6? I'll switch to using `QStringRef::at` instead.

src/libtiled/mapreader.cpp
```diff
@@ -942,34 +942,46 @@ void MapReaderPrivate::decodeCSVLayerData(TileLayer &tileLayer,
...
+            if (currentIndex >= text.length()) {
+                xml.raiseError(tr("Corrupt layer data for layer '%1'")
+                               .arg(tileLayer.name()));
                 return;
             }
 
+            // Get the next entry.
+            unsigned int gid = 0;
+            while (currentIndex < text.length()) {
+                auto currentChar =  text.at(currentIndex);
```
bjorn:
```suggestion
                auto currentChar = text.at(currentIndex);
```

src/libtiled/mapreader.cpp
```diff
@@ -1349,7 +1361,6 @@ void MapReaderPrivate::readProperty(Properties *properties)
     properties->insert(propertyName, variant);
 }
 
-
```
bjorn: There's two empty lines here because the definitions below are for a different class, so please don't remove one of them. :-)

Phlosioneer:
Removed the map size code.

Issue #1754, opened 2017-09-21
TMX documentation not clear on number of 'can contain' types


dyson:
I'm working on parsing TMX files and I can't figure out from the documentation if 'can contain' objects can be included multiple times or only once. For example under `<imageLayer>` it expressly mentions "A layer consisting of a single image" so I know there can only be one child image. Other objects are not so clear.

Prefixing or appending something to each 'can contain' object that can be included multiple times would make this much clearer.

Fox example an asterisk to mark 1 or many objects of the same type are accepted:
```
Can contain: <properties>, <layer>*, <objectgroup>*, <imagelayer>*, <group>*
```
It is also not clear at all which attributes/'can contain' objects are required as mentioned in #696 .

bjorn:
I'm working a lot on the documentation lately, so while I'm at it I might as well make sure this information is included. Thanks for bringing it up!

dyson:
Thanks heaps bjorn!

I tried to piece the info together from the JSON docs and from the DTD file which does include this information but they were incomplete and I wasn't sure if it was just me reading the docs wrong!

Cheers!

PR #2702, opened 2019-12-18
Clarify "can contain" documentation and error handling

### LINKED ISSUE #696: TMX Format doc: explicitly say which attributes and children are required
https://github.com/bjorn/tiled/wiki/TMX-Map-Format says that the DTD file is "not up-to-date", yet the documentation isn't clear on which elements/attributes are required.

"Can contain" isn't indicative of optionality but children. For example, map.backgroundcolor is optional. That's fine, but indicate that they are children*. The documentation should explicitly say which attributes and child elements are optional for each element.

Phlosioneer:
All "can contain" entries have been split into "can contain at most one"
and "can contain any number". Additionally, there is now a note to
implementors at the top of the specification listing how to handle parsing
errors. (More of a reminder on how to handle XML errors in general, since
TMX is a form of XML.)

Closes #696, #1754

### DIFF
--- a/docs/reference/tmx-map-format.rst
+++ b/docs/reference/tmx-map-format.rst
@@ -263,7 +263,7 @@ number of Wang tiles using these colors.
 -  **name**: The name of the Wang set.
 -  **tile**: The tile ID of the tile representing this Wang set.
 
-Can contain at most one: :ref:`tmx-wangcornercolor`, :ref:`tmx-wangedgecolor`
+Can contain up to 15 (each): :ref:`tmx-wangcornercolor`, :ref:`tmx-wangedgecolor`
 
 Can contain any number: :ref:`tmx-wangtile`
 
@@ -535,7 +535,7 @@ properties.
 
 Can contain at most one: :ref:`tmx-properties`, :ref:`tmx-ellipse` (since
 0.9), :ref:`tmx-point`, :ref:`tmx-polygon`, :ref:`tmx-polyline`, :ref:`tmx-text`
-(since 1.0), :ref:`tmx-image`
+(since 1.0)
 
 .. _tmx-ellipse:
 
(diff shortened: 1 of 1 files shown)


docs/reference/tmx-map-format.rst
```diff
@@ -249,7 +263,9 @@ number of Wang tiles using these colors.
 -  **name**: The name of the Wang set.
 -  **tile**: The tile ID of the tile representing this Wang set.
 
-Can contain: :ref:`tmx-wangcornercolor`, :ref:`tmx-wangedgecolor`, :ref:`tmx-wangtile`
+Can contain at most one: :ref:`tmx-wangcornercolor`, :ref:`tmx-wangedgecolor`
```
bjorn: Actually can contain up to 15 of each.

docs/reference/tmx-map-format.rst
```diff
@@ -515,9 +533,9 @@ properties from the specified template, properties saved with the object
 will have higher priority, i.e. they will override the template
 properties.
 
-Can contain: :ref:`tmx-properties`, :ref:`tmx-ellipse` (since
+Can contain at most one: :ref:`tmx-properties`, :ref:`tmx-ellipse` (since
 0.9), :ref:`tmx-point`, :ref:`tmx-polygon`, :ref:`tmx-polyline`, :ref:`tmx-text`
-(since 1.0), *image*
+(since 1.0), :ref:`tmx-image`
```
bjorn: The image element was included in italics because it used to be supported by the old Java version of Tiled, but the current version of Tiled never supported that. It should probably just be removed since it's unlikely to be relevant for anybody.
```suggestion
(since 1.0)
```

docs/reference/tmx-map-format.rst
Phlosioneer: Did not know that. The number 15 never actually appears in the docs. Added!

bjorn:
Thanks!

PR #2703, opened 2019-12-18
Fix rounding errors in isometric rendering code

### LINKED ISSUE #987: Screen to tile coord
Just to let you know there is a bug in the ISO conversion from screen to tile.

Found this while debugging my own stuff and comparing to tile.

QPointF IsometricRenderer::tileToScreenCoords(qreal x, qreal y) const
{
    const int tileWidth = map()->tileWidth();
    const int tileHeight = map()->tileHeight();
    const int originX = map()->height() * tileWidth / 2;

    return QPointF((x - y) * tileWidth / 2 + originX,
                   (x + y) * tileHeight / 2);
}

Actually working with pure int here loses the conversion. (3.9 is rounded to 3 instead of 4....)

In swift, I had to round before converting to Int. Similar fix should work for Tiled.
return ( Int(round(tileY + tileX)), Int(round(tileY - tileX)))

Pretty late, so sorry for the quick and dirty report.

Phlosioneer:
Converting to ints without rounding was causing integer division to lose
accuracy. The inputs to these functions are qreal and the outputs are
qreal, so we should be using qreals inside.

Closes #987

### DIFF
--- a/src/libtiled/isometricrenderer.cpp
+++ b/src/libtiled/isometricrenderer.cpp
@@ -523,22 +523,22 @@ void IsometricRenderer::drawMapObject(QPainter *painter,
 
 QPointF IsometricRenderer::pixelToTileCoords(qreal x, qreal y) const
 {
-    const int tileHeight = map()->tileHeight();
+    const qreal tileHeight = map()->tileHeight();
 
     return QPointF(x / tileHeight, y / tileHeight);
 }
 
 QPointF IsometricRenderer::tileToPixelCoords(qreal x, qreal y) const
 {
-    const int tileHeight = map()->tileHeight();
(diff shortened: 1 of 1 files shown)


bjorn:
I'm not sure about this change. When halving the tile size, we definitely do not want to end up with fractional values in case the size was odd. That is because Tiled works with pixel graphics and it makes no sense in terms of alignment to start drawing tile images on half pixels.

So, we need to find out what really was the problem #987 is about. Unfortunately no screenshot was attached. Note however that @franksauvag was talking about rounding instead of truncating. In case of odd tile sizes that should make it go up instead of down, but I'm unconvinced that this is really an improvement. It seems arbitrary to me.

Phlosioneer:
Ok.

Issue #2048, opened 2018-12-27
libtiled-java incompatible with Java 11


MinecraftPlaye:
Using libtiled-java with Java 11 gives me the following error:
`java.lang.NoClassDefFoundError: JAXBException `

It seems to be caused by Java 11 removing java.xml.bind (JAXB).

bjorn:
I'd love to also drop JAXB from libtiled-java. But what does Java provide as an alternative?

MinecraftPlaye:
> 
> 
> I'd love to also drop JAXB from libtiled-java. But what does Java provide as an alternative?

Sadly I don't really now if Java provides an alternative at all.
Thought maybe the following links provide helpful information: [JEP 320](http://openjdk.java.net/jeps/320) and [JAXB Java 9 Doc](https://docs.oracle.com/javase/9/docs/api/java.xml.bind-summary.html)

There seemed to be a way to use [JAXB in Java 9 and 10](https://stackoverflow.com/questions/48204141/replacements-for-deprecated-jpms-modules-with-java-ee-apis/48204154#48204154), but it [doesn't work in Java 11 ](https://github.com/eclipse-ee4j/jaxb-api/issues/78) anymore.

bjorn:
I don't know exactly how, since JAXB seems to be still used, but Java 11 compatibility was restored by #1925.

MinecraftPlaye:
> 
> 
> I don't know exactly how, since JAXB seems to be still used, but Java 11 compatibility was restored by #1925.

I'm still affected by this problem. As it seems to work in the mentioned pull request, maybe it's a problem with my small example project I just created with Gradle? I'm not really familiar with Maven.
A minimal project which causes this problem for me: [TiledLevelTest.zip](https://github.com/bjorn/tiled/files/4148764/TiledLevelTest.zip). I'm using Eclipse 2019-12. `gradlew eclipse` will set the project up for eclipse.

bjorn:
@MinecraftPlaye Hmm, while this incompatibility problem is probably fixed, the version of libtiled on Maven Central has not been updated with those changes. Could that be causing problems? Or are you using libtiled-java from source?

MinecraftPlaye:
> 
> 
> @MinecraftPlaye Hmm, while this incompatibility problem is probably fixed, the version of libtiled on Maven Central has not been updated with those changes. Could that be causing problems? Or are you using libtiled-java from source?

I'm using the version from Maven Central.

[...]

PR #2704, opened 2019-12-18
Attempt to fix java tile type bug

### LINKED ISSUE #2128: libtiled-java writeTileSet() didn't write Tile.type attribute
version 1.0.2

Phlosioneer:
This code *should* work, but I'm unable to compile & test it with my
setup. I don't understand how jaxb is setup. The main possible failure
is if the .getTile() function isn't actually generated by jaxb.

Closes #2128 if merged.

### DIFF
--- a/util/java/libtiled-java/src/main/java/org/mapeditor/io/TMXMapWriter.java
+++ b/util/java/libtiled-java/src/main/java/org/mapeditor/io/TMXMapWriter.java
@@ -309,10 +309,17 @@ public class TMXMapWriter {
             // Write tile properties when necessary.
             for (Tile tile : set) {
                 // todo: move the null check back into the iterator?
-                if (tile != null && !tile.getProperties().isEmpty()) {
+                if (tile != null
+                        && (!tile.getProperties().isEmpty()
+                        || !tile.getType().isEmpty()) {
                     w.startElement("tile");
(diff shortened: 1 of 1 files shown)


util/java/libtiled-java/src/main/java/org/mapeditor/io/TMXMapWriter.java
```diff
@@ -527,6 +535,10 @@ private void writeTile(Tile tile, XMLWriter w, String wp) throws IOException {
         w.startElement("tile");
         w.writeAttribute("id", tile.getId());
 
+		if (!tile.getType().isEmpty()) {
+			w.writeAttribute("type", tile.getType());
+		}
```
bjorn: Please make sure to indent with spaces rather than tabs.

bjorn:
> This code _should_ work, but I'm unable to compile & test it with my
> setup. I don't understand how jaxb is setup.

What error are you getting exactly? I personally am having trouble compiling `libtiled-java` as well, but that's probably because my system has Java 11 so I'm running into #2048.

util/java/libtiled-java/src/main/java/org/mapeditor/io/TMXMapWriter.java
Phlosioneer: Whoops! fixed.

PR #2706, opened 2019-12-19
Create separate image element for imagelayer

Phlosioneer:
The image is not parsed the same way as the `image` element in a tileset.

Can only be merged after #2705, to prevent merge conflicts.

### DIFF
--- a/docs/reference/tmx-map-format.rst
+++ b/docs/reference/tmx-map-format.rst
@@ -639,7 +639,19 @@ of the object.
 
 A layer consisting of a single image.
 
-Can contain at most one: :ref:`tmx-properties`, :ref:`tmx-image`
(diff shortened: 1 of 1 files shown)


bjorn:
I wonder if we should really fix the documentation or whether we could change the code to treat the elements the same. The difference is that image layers currently don't support embedded image data?

Phlosioneer:
That, and we also don't support explicit width / height fields.

bjorn:
> That, and we also don't support explicit width / height fields.

Hmm, these attributes are anyway optional and ignored, except for the images used by tilesets in which case it may be used to check if the amount of tile columns changed, to determine whether we may need to adjust tile IDs. Would be no harm in including them also for image layer images.

Phlosioneer:
I'm working on a PR to merge the two image loading methods.

PR #2707, opened 2019-12-19
Speed up first-draw by reducing initial map size

Phlosioneer:
Previously, loading a large map initialized the zoom to arbitrarily small
scales. During my testing, a 3000x3000 map would load in at 0.06% zoom.
Rendering all those tiles made the first draw and any user actions,
including zooming back in, very very slow.

This commit caps the number of tiles on screen to about 4,000 for the first
draw. The user is still free to zoom out, but the app will start out much
more responsive.

Split off from #2701

### DIFF
--- a/src/tiled/mapview.cpp
+++ b/src/tiled/mapview.cpp
@@ -126,9 +126,15 @@ void MapView::fitMapInView()
     if (rect.isEmpty())
         return;
 
-    // Scale and center map to fit in view
+    // Scale and center map to fit in view. For extremely large maps (by pixels), avoid going below
+    // 1% scale. For extremely large maps (by tiles), avoid putting more than 4,000 tiles within the
+    // view. Approximated to 4096 tiles, 64 * 64.
     centerOn(rect.center());
-    setScale(std::min(width() / rect.width(), height() / rect.height()) * 0.95);
+    qreal pixelScale = std::min(width() / rect.width(), height() / rect.height()) * 0.95;
+    auto tileSize = mapScene()->mapDocument()->map()->tileSize();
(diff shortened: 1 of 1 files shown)


Phlosioneer:
Okay, so you disagree with the 4000 tiles bit. I would like there to be at least _some_ limit. 200x200 seems to be the upper limit of responsiveness on my computer. It's laggy but somewhat reasonable.

Remember, this is NOT a limit on how far the user can zoom out, it's just the initial load-in scale. My goal here is to avoid the situation where the user opens a map, and tries to zoom in, but it's already zoomed out far enough that the zoom-in input takes a while to execute.

bjorn:
Yeah, let's just increase the limit instead of 64x64 we can do 256x256 for example. I think it's totally reasonable to keep the UI at least somewhat responsive by not showing the entire map beyond that size.

Phlosioneer:
Updated the size.

bjorn:
I think it's a good change. We'll see whether we get any complaints eventually from people with 512x512 maps saying "fit map in view" isn't working... :-)

Phlosioneer:
Wait, what? I thought that function was only called on initialization. I didn’t see any other uses when I searched. 

If it is a button, we could add a Boolean parameter that’s only true on init.

bjorn:
@Phlosioneer Right, the function was used both to initialize the view and can be triggered from the View menu. But I considered, that with the case you had in mind, you anyway wouldn't want to see the whole map due to potential performance issues. But maybe it makes sense to treat those cases differently.

Phlosioneer:
However you want to handle it. I missed that bit.

Issue #1421, opened 2016-12-28
Support for Metatiles


not-surt:
Metatiles or Chunks are a single tilemap cell indexing a fixed-size block of tiles rather than just a single tile.
Functions both as a convenience for the developer and as a form of memory compression.

Metatiles were common in games on old memory limited tile-based hardware such a the 8/16-bit consoles. eg. Sonic the Hedgehog used 128x128 pixel metatiles composed of 16x16 8x8 pixel tiles.

Implement a _MetaTileSet_ which takes a _TileMap_ and metatile size in the place of the image and tile size of a regular _TileSet_ and draw the metatile inplace of a single tile.
Ideally a _MetaTileSet_ would be able to take a _TileMap_ that in turn uses another _MetaTileSet_ for multiple levels of metatile scale.

This would make Tiled much more friendly toward working within the limitations of older systems.

eg. NES-like use case:
NES uses 8x8 pixel hardware tiles, a game may use 16x16 pixel (2x2 hardware tile) metatiles.
**Hardware TileSet:** Regular TileSet for 8x8 pixel hardware tiles. Draws 8x8 pixel block from image for each tile.
**Metatile TileMap:** Regular TileMap for metatile tile data, using _Hardware TileSet_ for tiles. Each 2x2 tile sector of _Metatile TileMap_ defines a metatile.
**Metatile TileSet:** New MetaTileSet for 16x16 pixel metatiles. Has 2x2 metatile size. Draws 2x2 tile block from _Metatile TileMap_ for each tile.
**Level TileMap**: Regular TileMap for game level, using _Metatile TileSet_ for tiles.

not-surt:
I've just thought now that this could also be used as a means to allow multilayer tiles, if the MetaTileSet's TileMap contains multiple layers.

bjorn:
I think the idea is very interesting, but I'm not sure about how to execute it. In your example use-case, the metatiles in a single metatile-tileset would all need to use the same dimensions. Was that done to keep things simple or because the hardware worked in a similar way? It seems to me that in general, people may expect more flexibility.

Tiled already has a feature to store away chunks of tiles to make placement of multi-tile areas easier, which are called Tile Stamps. But it works only on the editor level, while keeping the tile map in the usual format. Maybe the Tile Stamps feature could be extended somehow to allow placement of a stamp to create a reference to the stamp as metatile in a metatile-tileset. But in that case I'm having some difficulty imagining how the various metatiles used on a map are managed and stored.

[...]

PR #2708, opened 2019-12-20
Support tmx files as readable images

Phlosioneer:
This allows for users to create metatiles, by using a map file as the
source image of a tileset. The same code also allows ANY image to be a
rendered version of a map file, which may allow for interesting use cases.

Maps that can be used this way have some restrictions:
- They must be Orthogonal.
- Object layers are allowed, but they can only contain tile objects.

See discussion in the original issue #1421 for the full pros/cons of this
method.


Phlosioneer:
This PR still needs to document the new feature, and I'm unsure about some of the error handling.

Phlosioneer:
okay, what's going on here
qAsConst is declared in QtGlobal, including that should have fixed it...

src/libtiled/tilesetmanager.cpp
```diff
@@ -57,10 +57,7 @@ TilesetManager::TilesetManager():
 }
 
 TilesetManager::~TilesetManager()
-{
-    // Assert that there are no remaining tileset instances
-    Q_ASSERT(mTilesets.isEmpty());
```
bjorn: We can't just remove this assertion. It's there because the destructor of `Tileset` will attempt to remove the tileset from the tileset manager, re-creating a TilesetManager if it was destroyed before that:

https://github.com/bjorn/tiled/blob/752df93430d2413a86fbb2fb04d726c80e030fa3/src/libtiled/tileset.cpp#L76-L78

Which would then run into this assertion instead

https://github.com/bjorn/tiled/blob/752df93430d2413a86fbb2fb04d726c80e030fa3/src/libtiled/tilesetmanager.cpp#L129-L131

In addition, the assertions simply try to make sure there are no `Tileset` instances "leaking" somewhere. Would it be a problem to clear the cache before destroying the tileset manager? This happens here:

https://github.com/bjorn/tiled/blob/752df93430d2413a86fbb2fb04d726c80e030fa3/src/tiled/mainwindow.cpp#L758

Phlosioneer: Ah, I see. I'll try to work on avoiding the leaked resources then. I didn't really consider the leaking issue, because it's only leaking when the app closes.

Phlosioneer: I've pushed a new version that purges all loaded submap instances at the start of the tileset manager destructor. This ensures that all tileset destructors run before the tileset manager's destructor finishes.

Phlosioneer:
@bjorn friendly reminder to review this when you have time.

bjorn:
@Phlosioneer Yeah, I just didn't get around to it last week. I'll make sure to look at this tomorrow.

Issue #2700, opened 2019-12-18
Image layers cannot load images from Qt resource files


obeezzy:
Images embedded in Qt resource files would produce a null QPixmap when loaded from a TMX file. I tried pushing a change, but I kept getting denied access. (I'm new to this and therefore clueless lol.)

Below is my proposed fix for the issue:

**tiled/src/imagelayer.cpp:**
```cpp
bool ImageLayer::loadFromImage(const QUrl &url)
{
    QString urlAsString;
    if (url.url().startsWith("qrc:/"))
        urlAsString = url.url().replace(QString("qrc:/"), QString(":/"));
    else
        urlAsString = url.toLocalFile();

    return loadFromImage(ImageCache::loadImage(urlAsString), url);
}
```

Phlosioneer:
In the upper-right corner of the github website, while looking at this project, there's a "fork" button. Click that. That will create a copy in your name.

Push your modifications to that copy of the project.

Then return here to this project, look at the code tab, and you'll be prompted to create a "pull request". Click that green button.

For a one-commit change, it'll paste your commit message straight into the pull request, and the name of the pull request will be the name of the branch. Click the green button on that page to submit it. Your changes will now appear in the Pull Request tab, which we can comment on, discuss, etc. You can even continue to push new changes (for example, if we request a change or spot a bug).

Some automated tests will run. Check back a few hours or a day later and see if any of the tests failed. If they did, you can click on them, and see the error logs. The tests compile on multiple platforms (linux, windows, osx) and with different versions of Qt (5.6 and 5.12 at least, possibly more). If there's any failures on those tests, please fix them in your local copy and push the changes. The tests will automatically restart when you push new changes. If you have questions about what the tests say or why they're failing, leave a message on a PR and you'll get a response when someone, probably bjorn, has the time to look at it.

When the tests pass and Bjorn reviews the code, he can merge the pull request. At that point your code is now in the official repository, and you can safely delete your branch, your copy of the project, etc. if you want.

PR's can take some time; a short one goes faster, but it can easily take over a week. Bjorn has another job, and I believe he works one full day per week on tiled. During that time, he's not just reading PR's and issues, he's also coding, so expect the process to take some time.

bjorn:
Thanks for the elaborate introduction @Phlosioneer!

[...]

PR #2711, opened 2019-12-21
Fix issue where images cannot be loaded from Qt resource files

obeezzy:
#2700

### DIFF
--- a/src/libtiled/imagelayer.cpp
+++ b/src/libtiled/imagelayer.cpp
@@ -84,7 +84,7 @@ bool ImageLayer::loadFromImage(const QImage &image, const QString &source)
 
 bool ImageLayer::loadFromImage(const QUrl &url)
 {
-    return loadFromImage(ImageCache::loadImage(url.toLocalFile()), url);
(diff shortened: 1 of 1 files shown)


src/libtiled/imagelayer.cpp
```diff
@@ -84,7 +84,7 @@ bool ImageLayer::loadFromImage(const QImage &image, const QString &source)
 
 bool ImageLayer::loadFromImage(const QUrl &url)
 {
-    return loadFromImage(ImageCache::loadImage(url.toLocalFile()), url);
+    return loadFromImage(ImageCache::loadImage(Tiled::urlToLocalFileOrQrc(url)), url);
```
bjorn:
```suggestion
    return loadFromImage(ImageCache::loadImage(urlToLocalFileOrQrc(url)), url);
```
Just suggesting to remove the namespace, because it's superfluous.

bjorn: Never mind, I see all the other invokations also qualify the namespace, so let's keep it for consistency.

bjorn:
Thanks!

PR #2713, opened 2019-12-24
Combine imagelayer and tileset image tags

Phlosioneer:
Image layer images used separate code, so they didn't support the same
features (like embedding). Follow-up on #2706


bjorn:
@Phlosioneer I've made some small adjustments, mainly to avoid the code duplication. If the change looks fine to you like this as well then we can merge it.

Phlosioneer:
Looks good to me!

PR #2714, opened 2019-12-26
Group some source files into folders

Phlosioneer:
Hopefully the .pro file is updated correctly.

This is a change that I did for my local tinkering, feel free to close if you want to keep a flat project. It was starting to get difficult to locate files.


Phlosioneer:
Ah, I see why we don't do this. It seems QT creator has different header file resolution rules than command-line QBS.

For example, in commanddialog.cpp, both of these compile fine in qt creator (on windows):
```c++
#include "commanddialog.h"
// Or
#include "dialogs/commanddialog.h"
```

But command-line qbs seems to only accept the latter. :(

bjorn:
> Ah, I see why we don't do this. It seems QT creator has different header file resolution rules than command-line QBS.

Hmm, that may be, but it's not the reason I didn't do this. The reasons were more like:

* It would give a false sense of separation, given that qmake would anyway put all .o files for a certain project in the same directory so you can't use folders to resolve potential file name conflicts. You need to use different projects for that. I don't know if this is the case with Qbs.

* I'm generally of the opinion that it creates more work than it saves. I never use the file hierarchy to navigate the code, instead using "jump to definition", "find usages" and writing the file/class/function name in the Locator to get around. I've never missed the option to browse the project in folders, though I've frequently been annoyed by maintaining a folder structure and moving files around in other projects.

Of course, it could be that at some point the project reaches a size that it would be beneficial to use folders. Though as you've noticed such a change would be highly annoying to make, since it requires adjusting all existing includes or adding all the folders to the include path. Also, at work we develop an engine+game with literally tens of thousands of files. While it does have a general module structure and we do use virtual folders to organize things a little within a module, it remains largely useless and the general way of navigation is by methods described above.

Why was it getting difficult to locate files? Are you using the Projects view to locate them? If so, why?

Phlosioneer:
That works well when you know the project. When you’re learning it, you can’t just type the right name in. 

I was trying to find all the classes that touched the js api. There’s `editable*`, `scripted*`, but also others.

Phlosioneer:
To clarify, I make good use of follow-symbol and find-all-references. But that doesn’t help with questions like “what shows the layer list” or “where is the code that runs the new-map window”.

bjorn:
Ah yes, thanks for clarifying. I'm still not sure whether organizing the project in more folders would help answer those questions faster. For most of the UI stuff you can start by searching for a piece of text that occurs in the UI.

Anyway I agree more folders would make some things easier to find, just not sure if it's worth the effort of maintaining them.

> But command-line qbs seems to only accept the latter. :(

Btw, it occurred to me that there must be some other problem going on with the headers, because `#include "commanddialog.h"` will always work if the header file is in the same folder as the cpp file, regardless of the include paths. It must have been related to something else in the build directory being different between the Qt Creator and command-line Qbs cases.

Phlosioneer:
Usually it’s sufficient to just read the file names of all source files. Then I can start to make guesses about where to look. Folders make the “read all file names” plan faster. 

I’ll just have to build a mental map over time.

PR #2716, opened 2019-12-28
Give scripts access to existing file formats

### LINKED ISSUE #2696: Scripting API: Access other file loaders
Currently, if you want to create a plugin that loads a slightly modified .tsx or .tmx file, you have to re-write the entire file parsing algorithm yourself.

It would be very convenient and useful to be able to use another file loader, and then modify that afterwards. For example:

```js
tiled.registerTilesetFormat("My tsx", {
  name: "My TSX modifications",
  extension: "mytsx",

  read: function(filename) {
    // Load this file as if it was a ".tsx" file.
    return tiled.getTilesetFormat("tsx").read(filename);
  },

  write: function(tileset, filename) {
    for (var tile : tileset.tiles) {
      if (tile.width > 32 || tile.height > 32) {
        tile.type = "XL";
      }
    }
    
    // Save this file as if it was a ".tsx" file.
    tiled.getTilesetFormat("tsx").write(tileset, filename);
});
```

This would also help in situations where a custom file is transformed into a normal one: (assuming a command is setup to call `tar`)
```js
tiled.registerTileMapFormat("Archived", {
  name: "Archived Map",
  extension: "tar.gz",

  read: function(filename) {
    // Extract a .tsx file from the archive.
    tiled.executeCommand("unpackMapFromTar", true);
    // Continue with normal reading.
    return tiled.getTileMapFormat("tmx").read(filename);
  },

  write: function(map, filename) {
    // Do normal writing.
    tiled.getTileMapFormat("tmx").write(map, filename);
    // Copy it into the archive.
    tiled.executeCommand("repackMapIntoTar", true);
  }
});

// With a similar loader for tilesets
```

As the examples above show, the only added functions would be `tiled.getTileMapFormat(extension: string)` and `tiled.getTilesetFormat(extension: string)`

Phlosioneer:
Scripts can now call pre-defined map/tileset formats, and can query what
map/tileset formats are available. Closes #2696.

Still need to write documentation for the new APIs.

I wasn't able to get QJSEngine to accept actual ScriptTilesetFormatWrapper
return values, so functions that return those use QObject instead. Same
applies to ScriptMapFormatWrapper.

Suggestions of better names for ScriptTilesetFormatWrapper/ScriptMapFormatWrapper are welcome.

### DIFF
--- a/docs/reference/scripting.rst
+++ b/docs/reference/scripting.rst
@@ -443,27 +443,27 @@ tiled.extendMenu(id : string, items : array | object) : void
 
 .. _script-tilesetFormat:
 
-tiled.tilesetFormat(shortName : string) : TilesetFormatWrapper
+tiled.tilesetFormat(shortName : string) : TilesetFormat
     Returns the tileset format object with the given name, or `undefined` if
     no object was found. Built in options are `tsx`, `json`, and `lua`. See the
     `tilesetFormats` property for more info.
 
 .. _script-tilesetFormatForFile:
 
-tiled.tilesetFormatForFile(fileName : string) : TilesetFormatWrapper
+tiled.tilesetFormatForFile(fileName : string) : TilesetFormat
     Returns the tileset format object that can read the given file, or `undefined`
     if no object was found.
 
 .. _script-mapFormat:
 
-tiled.mapFormat(shortName : string) : MapFormatWrapper
+tiled.mapFormat(shortName : string) : MapFormat
     Returns the map format object with the given name, or `undefined` if no object
     was found. Built in options are `tsx`, `csv`, `gmx`, `json`, `js`, and `lua`.
     See the `mapFormats` property for more info.
 
 .. _script-mapFormatForFile:
 
-tiled.mapFormatForFile(fileName : string) : MapFormatWrapper
+tiled.mapFormatForFile(fileName : string) : MapFormat
     Returns the map format object that can read the given file, or `undefined` if
     no object was found.
 
@@ -571,6 +571,28 @@ Asset.macro(text : string, callback : function) : value
         })
 
     The returned value is whatever the callback function returned.
+
+.. _script-fileformat:
+
+File Format
+^^^^^^^^^^^
+
+Common functionality for file format readers and writers. (Since 1.4)
+
+Properties
+~~~~~~~~~~
+
+.. csv-table::
+    widths: 1, 2
+
+    **canRead** : bool |ro|, Whether this format supports reading files.
+    **canWrite** : bool |ro|, Whether this format supports writing files.
+
+Functions
+~~~~~~~~~
+
+FileFormat.supportsFile(fileName : string) : bool
+    Returns whether the file is readable by this format.
 
 .. _script-grouplayer:
 
@@ -727,35 +749,24 @@ Properties
 
 .. _script-mapformatwrapper:
 
-Map Format Wrapper
-^^^^^^^^^^^^^^^^^^
+Map Format
+^^^^^^^^^^
 
 This is an object that can read or write map files. (Since 1.4)
 
-Properties
-~~~~~~~~~~
-
-.. csv-table::
-    :widths: 1, 2
-
-    **name** : string |ro|, The name of this format.
-    **extension** : string |ro|, "The file extension of this format, or `undefined`
-    if not available. Use `supportsFile` instead."
+Inherits :ref:`script-fileformat`.
 
 Functions
 ~~~~~~~~~
 
-MapFormatWrapper.read(fileName : string) : :ref:`script-map`
+MapFormat.read(fileName : string) : :ref:`script-map`
     Read the given file as a map. This function will throw an error if it
     is not supported.
 
-MapFormatWrapper.write(map : :ref:`script-map`, fileName : string) : string
+MapFormat.write(map : :ref:`script-map`, fileName : string) : string
     Write the given map to a file. This function will throw an error if it
     is not supported. If there is an error writing the file, it will return a
     description of the error; otherwise, it will return "".
-
-MapFormatWrapper.supportsFile(fileName : string) : bool
-    Returns whether the file is readable by this format.
 
 .. _script-mapview:
 
@@ -1365,35 +1376,24 @@ Properties
 
 .. _script-tilesetformatwrapper:
 
-Tileset Format Wrapper
-^^^^^^^^^^^^^^^^^^^^^^
+Tileset Format
+^^^^^^^^^^^^^^
 
 This is an object that can read or write tileset files. (Since 1.4)
 
-Properties
-~~~~~~~~~~
-
-.. csv-table::
-    :widths: 1, 2
-
-    **name** : string |ro|, The name of this format.
-    **extension** : string |ro|, "The file extension of this format, or `undefined`
-    if not available. Use `supportsFile` instead."
+Inherits :ref:`script-fileformat`.
 
 Functions
 ~~~~~~~~~
 
-TilesetFormatWrapper.read(fileName : string) : :ref:`script-tileset`
+TilesetFormat.read(fileName : string) : :ref:`script-tileset`
     Read the given file as a tileset. This function will throw an error if it
     is not supported.
 
-TilesetFormatWrapper.write(tileset : :ref:`script-tileset`, fileName : string) : string
+TilesetFormat.write(tileset : :ref:`script-tileset`, fileName : string) : string
     Write the given tileset to a file. This function will throw an error if it
     is not supported. If there is an error writing the file, it will return a
     description of the error; otherwise, it will return "".
-
-TilesetFormatWrapper.supportsFile(fileName : string) : bool
-    Returns whether the file is readable by this format.
 
 .. _script-tilesetsview:
 
--- a/docs/reference/scripting.rst
+++ b/docs/reference/scripting.rst
@@ -131,11 +131,11 @@ Properties
     **tilesetEditor** : :ref:`script-tileseteditor`, "Access the editor used when editing tilesets."
     **tilesetFormats** : [string] |ro|, "List of tileset format names. Built in
     options are `tsx`, `json`, and `lua` (write-only). Use
-    :ref:`tilesetFormat<script-tilesetFormat>` to get the corresponding format object
+    :ref:`tilesetFormat <script-tilesetFormat>` to get the corresponding format object
     to read and write files. (Since 1.4)"
     **mapFormats** : [string] |ro|, "List of map format names. Built in options
     are `tmx`, `json`, `js`, `gmx` (write-only), and `lua` (write-only). Use
-    :ref:`mapFormat<script-mapFormat>` to get the corresponding format object to
+    :ref:`mapFormat <script-mapFormat>` to get the corresponding format object to
     read and write files. (Since 1.4)"
 
 Functions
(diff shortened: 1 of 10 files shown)


Phlosioneer:
@bjorn Any idea why returning the custom types doesn't work? I did the `Q_DECLARE_METATYPE` stuff and the `qRegisterMetaType` as well. I spent like, 4 hours last night going through the Qt codebase to figure out where the issue was. It turns out that, in the current Qt implementation at least, Q_DECLARE_METATYPE is a no-op for classes that inherit from QObject. qRegisterMetaType has nothing to do with QQmlEngine stuff as far as I can tell... I'm actually able to comment out the qRegisterMetaType for some objects (such as ScriptTextFile) and scripting with it continues to work fine.

I think there's an extra step or condition I'm missing. I just couldn't manage to find it. The code that throws the error eventually checks QPropertyCache, and that was difficult to trace back farther, so I can't tell where along the way it comes up with QMetaType::Undefined for the Q_INVOKABLE's return type.

bjorn:
> Any idea why returning the custom types doesn't work? I did the `Q_DECLARE_METATYPE` stuff and the `qRegisterMetaType` as well.

The other thing that frequently happens to me as well, is that the type is not recognized because it is in a namespace. That is because if you use Q_INVOKABLE or if you use a type in a Q_PROPERTY, the returned type is encoded by string somewhere and later looked up again. In that case, you need to use "Tiled::EditableSomething" for example instead of just "EditableSomething" even if the compiler will understand the type either way.

Phlosioneer:
Thanks, that fixed it! I'll remember that for the future.

Phlosioneer:
This is now blocked on #2720 , to avoid some messy merge conflicts in the scripting API reference.

src/tiled/scriptfileformatwrappers.cpp
```diff
@@ -0,0 +1,128 @@
...
+    : QObject(parent)
+    , mFormat(format)
+{}
+
+QString ScriptMapFormatWrapper::name() const
+{
+    if (auto scriptedFormat = qobject_cast<ScriptedMapFormat*>(mFormat))
+        return scriptedFormat->name();
+    else
+        return mFormat->shortName();
```
bjorn: Why not call this `shortName` and return `mFormat->shortName()` in both cases?

I'm not sure we should introduce the `name` and `extension` as properties. I understand I set them up that way for the scripted file formats, but actually they were just used to make the `nameFilter`. Do you need these properties somewhere? In hindsight I should probably just have made "nameFilter" a property of the scripted file format object instead of "name" and "extension", also because that actually gives more flexibility since it supports mentioning multiple extensions (`ScriptedFileFormat::supportsFile` could then be implemented using `Utils::fileNameMatchesNameFilter`).

src/tiled/scriptfileformatwrappers.cpp
```diff
@@ -0,0 +1,128 @@
...
+    return editable;
+}
+
+QString ScriptMapFormatWrapper::write(EditableMap *editable, QString &filename)
+{
+    auto map = editable->map();
+    if (mFormat->write(map, filename))
+        return QString();
+    else
+        return mFormat->errorString();
```
bjorn: I wonder if it's better to throw a script error here.

src/tiled/scriptfileformatwrappers.cpp
```diff
@@ -0,0 +1,128 @@
...
+    if (auto scriptedFormat = qobject_cast<ScriptedMapFormat*>(mFormat))
+        return scriptedFormat->extension();
+    else
+        return QString();
+}
+
+EditableMap *ScriptMapFormatWrapper::read(QString &filename)
+{
+    auto map = mFormat->read(filename);
+    auto editable = new EditableMap(std::move(map));
```
bjorn: In case of reading error, the returned `std::unique_ptr<Map>` wraps a `nullptr` and no `EditableMap` instance should be created. The reading error should probably be thrown given that the return value is taken (which I guess it also a good reason to change the `write` function error reporting for consistency).

Note that we throw an error by using `ScriptManager::throwError`. This makes sure at least something useful happens when Tiled is compiled against Qt < 5.12 (currently happens Windows XP release (Qt 5.6) and for the snap build (Qt 5.9)).

src/tiled/scriptmodule.cpp
```diff
@@ -282,6 +306,63 @@ QJSValue ScriptModule::registerTool(const QString &shortName, QJSValue toolObjec
     return toolObject;
 }
 
+Tiled::ScriptMapFormatWrapper *ScriptModule::getMapFormat(const QString &shortName) const
+{
+    auto formats = PluginManager::objects<MapFormat>();
```
bjorn:
```suggestion
    const auto formats = PluginManager::objects<MapFormat>();
```
(3 more times below)

src/tiled/scriptmodule.h
```diff
@@ -102,6 +108,11 @@ class ScriptModule : public QObject
     Q_INVOKABLE void registerTilesetFormat(const QString &shortName, QJSValue tilesetFormatObject);
     Q_INVOKABLE QJSValue registerTool(const QString &shortName, QJSValue toolObject);
 
+    Q_INVOKABLE Tiled::ScriptMapFormatWrapper *getMapFormat(const QString &shortName) const;
+    Q_INVOKABLE Tiled::ScriptMapFormatWrapper *getMapFormatByFile(const QString &fileName) const;
+    Q_INVOKABLE Tiled::ScriptTilesetFormatWrapper *getTilesetFormat(const QString &shortName) const;
+    Q_INVOKABLE Tiled::ScriptTilesetFormatWrapper *getTilesetFormatByFile(const QString &fileName) const;
```
bjorn: Naming style: Please remove the "get" prefix from all these methods. And I think `ForFile` would be more suitable than `ByFile`.

src/tiled/scriptfileformatwrappers.cpp
```diff
@@ -0,0 +1,128 @@
...
+    if (auto scriptedFormat = qobject_cast<ScriptedTilesetFormat*>(mFormat))
+        return scriptedFormat->extension();
+    else
+        return QString();
+}
+
+EditableTileset *ScriptTilesetFormatWrapper::read(QString &filename)
+{
+    auto tileset = mFormat->read(filename);
+    auto editable = new EditableTileset(tileset.data());
```
bjorn: `tileset` may be a null shared pointer, and the possible error should probably be thrown.

src/tiled/scriptfileformatwrappers.h
```diff
@@ -0,0 +1,79 @@
...
+class ScriptMapFormatWrapper : public QObject
+{
+    Q_OBJECT
+
+    Q_PROPERTY(QString name READ name)
+    Q_PROPERTY(QString extension READ extension)
+
+public:
+    Q_INVOKABLE ScriptMapFormatWrapper();
+    Q_INVOKABLE explicit ScriptMapFormatWrapper(MapFormat *format, QObject *parent = nullptr);
```
bjorn: Why are these constructors `Q_INVOKABLE`? Should not be needed and the default constructor can be removed as well.

src/tiled/scriptfileformatwrappers.cpp
```diff
@@ -0,0 +1,128 @@
+/*
+ * scriptfileformatwrappers.h
+ * Copyright 2019, Thorbjørn Lindeijer <bjorn@lindeijer.nl>
```
bjorn: I try to do the per-file copyright notices based on who edited those files, though I know, this is laboursome and error-prone to keep up-to-date (and could be automated based on some git tooling, I guess).

In any case you can put your name here and also feel free to add your name to the list if you make significant changes to other files.

src/tiled/scriptfileformatwrappers.cpp
Phlosioneer: I wanted to include them because that makes the wrapper able to access everything in the original object that the user provides. It feels weird to accept a scripted object, and then not have access to the same fields as that object.

I can definitely do without them, just wanted to try to keep things consistent.

src/tiled/scriptfileformatwrappers.cpp
Phlosioneer: See the comment above; we expect errors to be returned as a string in user-supplied write functions, so I made the wrapper do the same.

We can definitely throw an error here instead, just wanted to try to keep things consistent with the user-supplied objects.

src/tiled/scriptfileformatwrappers.cpp
Phlosioneer: I totally forgot std::unique_ptr could be null. I'm still thinking in terms of Rust. Oops!

src/tiled/scriptmodule.h
Phlosioneer: Ok.

Phlosioneer: Still learning the style.

src/tiled/scriptfileformatwrappers.h
Phlosioneer: That was a remnant of a test I did, I'll definitely remove it.

I was trying to figure out why it wouldn't accept the return type, and as part of that I tried making a qMetaObject entry for it and then constructed it via JS.

src/tiled/scriptfileformatwrappers.cpp
Phlosioneer: Another option is to make the properties dynamic; they'd only exist if the backing TilesetFormat is castable to ScriptedTilesetFormat (similarly for MapFormat).

src/tiled/scriptfileformatwrappers.cpp
Phlosioneer: Another option is to only return a string if the backing TilesetFormat is castable to ScriptedTilesetFormat (similarly for MapFormat), and otherwise throw an error?

src/tiled/scriptmodule.cpp
```diff
@@ -282,6 +306,63 @@ QJSValue ScriptModule::registerTool(const QString &shortName, QJSValue toolObjec
     return toolObject;
 }
 
+ScriptMapFormatWrapper *ScriptModule::mapFormat(const QString &shortName) const
+{
+    const auto formats = PluginManager::objects<MapFormat>();
+    for (auto format : formats) {
+        if (format->shortName() == shortName) {
+            auto wrapper = new ScriptMapFormatWrapper(format);
+            QQmlEngine::setObjectOwnership(wrapper, QQmlEngine::JavaScriptOwnership);
```
bjorn: It occurred to me that in this case, explicitly setting the ownership is not needed, because you're returning a QObject without parent from a function, which by default will have JavaScript ownership.

docs/reference/scripting.rst
```diff
@@ -690,6 +725,38 @@ Properties
...
+    :widths: 1, 2
+
+    **name** : string |ro|, The name of this format.
+    **extension** : string |ro|, "The file extension of this format, or `undefined`
+    if not available. Use `supportsFile` instead."
+
+Functions
+~~~~~~~~~
+
+MapFormatWrapper.read(fileName : string) : :ref:`script-map`
```
bjorn: I would not include the "Wrapper" in the name in the documentation, since I consider this just an implementation detail. Similar to how I don't include "Editable" in the documented type names.

src/tiled/scriptfileformatwrappers.cpp
bjorn: I would prefer to do without any special handling for scripted map/tileset formats. And also to replace the `name` and `extension` properties with a `nameFilter` property, to match the `FileFormat` interface.

src/tiled/scriptfileformatwrappers.cpp
bjorn: Let's not implement different behaviors depending on the implementation of the file format. I think throwing an error yields a better default behavior, which can be tweaked by using `try` and `catch` if needed.

Phlosioneer:
The MapFormat/TilesetFormat interface already throws the error if control reaches a non-overrided function. However, thinking about it, that may not be an actual javascript error; it might just be a message in the console? I'll check.

Phlosioneer:
Github won't let me respond to the last two messages you left, so I'll reply here:

> I would prefer to do without any special handling for scripted map/tileset formats. And also to replace the name and extension properties with a nameFilter property, to match the FileFormat interface.

In that case I would rather not expose the nameFilter at all, then. It's not useful for scripts, we already have the supportsFile function.

bjorn:
Nice work, thanks!

Phlosioneer:
Uh, your commits didn't pass the compile tests.

bjorn:
@Phlosioneer Heh, they failed because the pull request was already closed when they started.

PR #2720, opened 2019-12-30
Alphabetize scripting API reference

Phlosioneer:
It's faster to find the objects in the side bar when they're ordered
alphabetically. I wanted to also sort functions and properties, but they
have a consistent ordering, and it wouldn't speed up lookups by much.


bjorn:
Thanks! I wasn't sure myself whether to order these alphabetically, but I agree it's sensible. We may want to introduce categories to still group certain classes together though.

PR #2721, opened 2019-12-30
Fix console warning about Tiled::Issue

Phlosioneer:
It was declared as a metatype, but wasn't registered as a metatype at
runtime.

### DIFF
--- a/src/libtiled/logginginterface.cpp
+++ b/src/libtiled/logginginterface.cpp
@@ -55,6 +55,7 @@ Issue::Issue(Issue::Severity severity,
     , mContext(context)
     , mId(mNextIssueId++)
 {
+    qRegisterMetaType<Issue>();
 }
 
 void Issue::setCallback(std::function<void ()> callback)


bjorn:
Hm, I think I noticed this warning, but not always. Do you know how to reproduce it and what the warning is exactly? I wonder a little why it is necessary, because I haven't needed to use `qRegisterMetaType` for any other type apart from the stuff exposed to scripts.

Phlosioneer:
It occurs when the TLS subsystem _also_ fails to load:

```
QObject::connect: Cannot queue arguments of type 'Tiled::Issue'
(Make sure 'Tiled::Issue' is registered using qRegisterMetaType().)
qt.network.ssl: QSslSocket::connectToHostEncrypted: TLS initialization failed
qt.network.ssl: QSslSocket::connectToHostEncrypted: TLS initialization failed
"TLS initialization failed"
```

Phlosioneer:
It's definitely some kind of transient error. I'm going back through `master`'s history to see if I can get a clean build that reproduces it.

Phlosioneer:
I can't reproduce this issue. :/

bjorn:
Alright, thanks for testing. Maybe it'll show up again and then we can find out why it happens.

PR #2724, opened 2020-01-03
Assign global variables to console evals

Phlosioneer:
Similar to google chrome's debug console. Makes working with intermediate values easier.

### DIFF
--- a/src/tiled/consoledock.cpp
+++ b/src/tiled/consoledock.cpp
@@ -121,8 +121,10 @@ void ConsoleDock::executeScript()
     appendScript(script);
 
     const QJSValue result = ScriptManager::instance().evaluate(script);
-    if (!result.isError() && !result.isUndefined())
-        appendInfo(result.toString());
+    if (!result.isError() && !result.isUndefined()) {
+        auto name = ScriptManager::instance().createTempValue(result);
+        appendInfo(name + QStringLiteral(" = ") + result.toString());
+    }
 
     mLineEdit->clear();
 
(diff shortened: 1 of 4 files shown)


bjorn:
Hmm, is this Chrome specific? I don't see this behavior in the console in Brave, where actually `$0`, `$1` etc. seem to refer to various elements on the page, `$0` being the last clicked one. I'm not sure about the visual noise this is adding and it seems to me preferable to assign the returned values to named variables explicitly.

What I did notice is that `$_` can be used to get the value of the last executed statement. Maybe it's a good compromise?

Phlosioneer:
the problem with assigning it to variables is that when you do that, *nothing* is printed. The value of the expression `var x = 1` is `undefined`, not `1`.

The problem with restricting it to `$_` is that you might need something from a few lines above without realizing you need it at that time. It's annoying to write `var $0 = ...` etc every time I type a line just in case I need it later.

And you appear to be correct, in chrome $0, $1, etc. are the 1st, 2nd, etc. elements you select in the web page. Hm, what interpreter was I thinking of then?

I'd still like this change but I'm open to other ways of accomplishing the same thing. Having this speeds up my script debugging by a lot. What if the feature is enabled but it didn't print out the "$0 = ..." stuff unless a checkmark in the preferences was set?

bjorn:
Maybe we could just make it a little less noisy, by showing it in a darker gray for example?

bjorn:
> the problem with assigning it to variables is that when you do that, _nothing_ is printed. The value of the expression `var x = 1` is `undefined`, not `1`.

Note that while indeed `var x = 1` evaluates to `undefined`, the simpler form `x = 1` does evaluate to `1` and sets a global variable `x`.

But, I think the dimming helped reduce the noise so let's just merge it now. :-)

Phlosioneer:
Ah, I wasn't able to prevent the wrapping. Good idea to use span and &nbsp; together.

PR #2725, opened 2020-01-07
tips for noobs setting up .tiled folder (python scripts)

ZTarget:
I was confused why I couldn't make a folder starting with a dot on windows7 and had to do some searching to solve, hope this will help other noobs figure this out quicker

### DIFF
--- a/docs/manual/python.rst
+++ b/docs/manual/python.rst
@@ -9,7 +9,11 @@ platform.
 For the scripts to get loaded, they should be placed in ``~/.tiled``.
 Tiled watches this directory for changes, so there is no need to restart
 Tiled after adding or changing scripts (though the directory needs to
-exist when you start Tiled).
+exist when you start Tiled). On linux .tiled is considered a hidden file, use 
+Ctrl+H to show hidden files. On Windows 7 to make a .tiled folder (or a folder 
+starting with a dot) you will have to open command promp (cmd.exe) this should 
+start in your home folder by default, then type "mkdir .tiled" without the quotes 
+to create the folder.
 
 There are several `example scripts`_ available in the repository.
 


bjorn:
I think adding this to the docs is a good idea, though I'd put it in its own "note" section.

And in fact, this `~/.tiled` folder is really less than ideal. In the meantime, Tiled supports extensions written in JavaScript and it uses a [more standard folder location](https://doc.mapeditor.org/en/stable/reference/scripting/#scripted-extensions) depending on the operating system. It also creates the folder automatically and in the Preferences there's a button to open a file manager in that folder. It would be nice to support Python scripts in this extensions folder as well.

Note that with all the work done on the JavaScript API, which can do a lot more than the Python plugin API, I'd suggest extending Tiled with JavaScript rather than Python whenever possible. If it's not possible, please let me know why.

ZTarget:
Hmm, yeah that extensions folder would probably be a better location, easier to understand since it gets created, more so on Windows since Linux is easy. But also super easy since you can open it from the preferences like you said.

Python worked really well for me, I'm more of an artist but coming from Godot I was able to make a simple script. It is basically your example script with flipping x,y and xy. Adding 1000 2000 and 3000
since I won’t be going over 999 tiles. This saves to a csv file for Godot, then I have my own script to make the map in Godot and also the csv gets put into an array for old school collision(no shapes).

I’ve never used JavaScript but that may be a good note to add that it can do a lot more than python atm. I’m not much of a writer but something like..
It is strongly recommended that you consider JavaScript for extending Tiled before using Python. JavaScript can currently do more than Python and has better documentation (then have link to scripting)

Another tip for Python and Windows7 I can think of is what I read in the forum, that I should check the box when installing python to add to system path. (Though I didn’t try this without checking the box so not sure what it does)

I hope you keep Python in Tiled in the future even in its current form it seems pretty useful. I’m sure some advanced users will come along and help with updating it eventually.

bjorn:
I've added some notes in the above commit based on your change. Thanks!

> Another tip for Python and Windows7 I can think of is what I read in the forum, that I should check the box when installing python to add to system path. (Though I didn’t try this without checking the box so not sure what it does)

I don't remember explicitly needing to check this. Maybe it's not needed or maybe it used to be checked by default. If it's indeed needed and especially when not the default we should definitely add a note about this.

ZTarget:
I looked into the python path checkbox,(In windows7) I uninstalled python then reinstalled with out the check and I did get the red error for the python plugin, then I uninstalled and reinstalled again with it checked and it turned green. It looks like it is necessary to check the box.
Here is a screengrab of the box if you want see it or use it but you may want to resize it and save as jpg or whatever http://www.ektophase.com/otherstuff/pythonCheck.png 

Here is that forum post I mentioned earlier https://discourse.mapeditor.org/t/error-activating-python-plugin/2346

bjorn:
@ZTarget Alright, thanks for checking this! Indeed we should add a note about this to the docs as well then.

PR #2728, opened 2020-01-09
add go-tiled library

bontibon:


bjorn:
Thanks for the addition and thank you for your work on `go-tiled`!

Just wondering, but since there was already `go-tmx`, can you share a few words on what makes `go-tiled` different or what might help people decide which one to use?

PR #2731, opened 2020-01-13
Create XML scripting API

### LINKED ISSUE #2665: Scripting API for parsing XML documents
ECMAScript 7, the current version of JavaScript supported by Tiled, does not appear to provide an API for parsing XML files. For convenience, a custom API should be added for this.

Potentially we can port over the [Xml Service](https://doc.qt.io/qbs/jsextension-xml.html) from Qbs.

Phlosioneer:
The API is a thin wrapper around Qt's QDomDocument class, with some
ergonomic improvements (lists are mutable, attributes aren't nodes,
single-child text nodes are transparent).

Closes #2665

### DIFF
--- a/docs/reference/scripting.rst
+++ b/docs/reference/scripting.rst
@@ -1678,3 +1678,127 @@ BinaryFile.commit() : void
 BinaryFile.close() : void
     Closes the file. It is recommended to always call this function as soon as
     you are finished with the file.
+
+XmlFile
+~~~~~~~
+
+**Properties**
+
+.. csv-table::
+    :widths: 1, 2
+
+    **root** : XmlFile |ro|, A reference to itself (useful for algorithms)
+    **children** : XmlList |ro|, A special kind of list with all the direct children of this element
+
+**Functions**
+
+new XmlFile(text : string)
+    Creates a new Xml document by parsing `text`.
+
+new XmlFile(source : TextFile)
+    Convenience function; same as `new XmlFile(source.readAll())`
+
+XmlFile.childByName(name : string) : XmlElement
+    Find the first child with the given tag name. Returns null if no
+    child was found.
+
+XmlFile.childrenByName(name : string) : [XmlElement]
+    Find all children with the given tag name.
+
+XmlFile.writeToString() : string
+    Returns this XML document represented as a string.
+
+XmlElement
+~~~~~~~~~~
+
+**Properties**
+
+.. csv-table::
+    :widths: 1, 2
+
+    **root** : XmlFile |ro|, The XML file this element belongs to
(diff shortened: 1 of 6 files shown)


bjorn:
I didn't check yet whether the API matches, but did you  have a look at the Xml Service in Qbs? The code for that is at https://code.qt.io/cgit/qbs/qbs.git/tree/src/lib/corelib/jsextensions/domxml.cpp.

I will have a closer look at this tomorrow.

Phlosioneer:
Among other concerns, the XmlService is a very clunky API. For example, there are no Q_PROPERTIES, and there's no actual way to get a list of children! You have to go firstChild() into nextSibling() repeatedly to iterate over children.

I think my implementation is better, but I'm not a great judge of that obviously. I will note that my current implementation doesn't support CDATA, while XmlService does. That's the only thing it supports that mine doesn't, though; neither have support for namespaces and other stuff.

I think the main usecase for this will be to open and modify tmx/tsx files, though, so I didn't think that was important for a first step.

bjorn:
> Among other concerns, the XmlService is a very clunky API. For example, there are no Q_PROPERTIES, and there's no actual way to get a list of children! You have to go firstChild() into nextSibling() repeatedly to iterate over children.

Hmm, we can introduce `Q_PROPERTY` where possible. Regarding the way of iterating children, in general I'd prefer not to invent a new API where there are existing ones available, and this seems to me only a small inconvenience. On the plus side, the XML DOM API is relatively well known and supported in most programming environments. I agree it's also well known for being a little clunky, but not convinced that the situation is bad enough to provide something rather different.

> I think the main usecase for this will be to open and modify tmx/tsx files, though, so I didn't think that was important for a first step.

Yeah, in the end it doesn't make a huge difference, though once introduced we can't easily change the API anymore. It's easier to take a minimal API like the Xml Service in Qbs and extend it with some convenience later if it turns out there are some real pain points.

Phlosioneer:
I personally hate the XML Dom API. But I suppose it’s more standard than I realize? Ill make a new PR with that code copied over, make some things properties, and we’ll see how it looks. 

I definitely want to keep the “toRaw” options. They’re only for reading but are very nice and useful. So something like “rawChildren” and “rawAttributes”.

bjorn:
> I definitely want to keep the “toRaw” options. They’re only for reading but are very nice and useful. So something like “rawChildren” and “rawAttributes”.

Let's just return a JS array from `element.childNodes()` and a JS object from `element.attributes()`. Neither of these are part of the Xml Service and I think this would be a fine way to add them, since I don't personally see much value in wrapping the `QDomNodeList` and `QDomNamedNodeMap`.

PR #2734, opened 2020-01-15
support converting object reference property to and from int

bontibon:
Users could have previously had custom properties of type int that
referenced objects by ID. Registering a converter for ObjectRef to
and from int should help migration to the new object reference type.

Updates #707

### DIFF
--- a/src/tiled/main.cpp
+++ b/src/tiled/main.cpp
@@ -355,6 +355,9 @@ int main(int argc, char *argv[])
 
     TiledApplication a(argc, argv);
 
+    QMetaType::registerConverter<ObjectRef, int>(&ObjectRef::toInt);
+    QMetaType::registerConverter<int, ObjectRef>(&ObjectRef::fromInt);
+
     a.setOrganizationDomain(QLatin1String("mapeditor.org"));
 #if defined(Q_OS_MAC) || defined(Q_OS_WIN)
     a.setApplicationName(QLatin1String("Tiled"));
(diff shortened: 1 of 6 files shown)


src/tiled/main.cpp
```diff
@@ -355,6 +355,9 @@ int main(int argc, char *argv[])
 
     TiledApplication a(argc, argv);
 
+    QMetaType::registerConverter<Tiled::ObjectRef, int>(&Tiled::ObjectRef::toInt);
+    QMetaType::registerConverter<int, Tiled::ObjectRef>(&Tiled::ObjectRef::fromInt);
```
bjorn: You can drop the `Tiled` namespaces here.

bontibon:
> Maybe while you're at it you could do the same for converting between "string" and "file" properties?

It seems the change for this isn't as trivial as the ObjectRef/int conversion, at least for someone who doesn't know the tiled code very well (the UI doesn't update itself correctly after the conversion). I can perhaps try this in another PR.

bjorn:
> It seems the change for this isn't as trivial as the ObjectRef/int conversion, at least for someone who doesn't know the tiled code very well (the UI doesn't update itself correctly after the conversion).

Yeah, that wasn't so easy to find. It was due to a bug in `VariantPropertyManager::valueType`, which caused `PropertyBrowser::setCustomPropertyValue` to think that the type of the property didn't change, and so it didn't recreate the property.

PR #2735, opened 2020-01-17
Create XML scripting API, second attempt

### LINKED ISSUE #2665: Scripting API for parsing XML documents

Phlosioneer:
Closes #2665

Re-do of #2731

### DIFF
--- a/docs/reference/scripting.rst
+++ b/docs/reference/scripting.rst
@@ -1718,3 +1718,111 @@ BinaryFile.commit() : void
 BinaryFile.close() : void
     Closes the file. It is recommended to always call this function as soon as
     you are finished with the file.
+
+XmlFile
+~~~~~~~
+
+Inherits :ref:`script-xml-node`
+
+**Functions**
+
+new XmlFile(text : string)
+    Creates a new Xml document by parsing `text`.
+
+XmlFile.createNode(type : NodeType [, data : string = ""]) : XmlNode
+    Creates a new node attached to this document. The new node is not in the element
+    tree until it's been added somewhere. `data` is the initial value of XmlNode.text.
+    `type` cannot be NodeType.Root or NodeType.Other.
+
+XmlFile.writeToString(indent : int = 1) : string
+    Returns this XML document represented as a string.
+
+XmlNode
+~~~~~~~~~~
+
+**Properties**
+
+.. csv-table::
+    :widths: 1, 2
+
+    **type** : NodeType |ro|, The type of this node
+    **attributes** : object |ro|, A copy of the attributes on this Element
+    **children** : [XmlNode] |ro|, A copy of the list of all the direct children of this Element or Root.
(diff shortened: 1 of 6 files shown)


Phlosioneer:
Blocked on #2736

docs/reference/scripting.rst
```diff
@@ -1718,3 +1718,111 @@ BinaryFile.commit() : void
...
+
+**Functions**
+
+new XmlFile(text : string)
+    Creates a new Xml document by parsing `text`.
+
+XmlFile.createNode(type : NodeType [, data : string = ""]) : XmlNode
+    Creates a new node attached to this document. The new node is not in the element
+    tree until it's been added somewhere. `data` is the initial value of XmlNode.text.
+    `type` cannot be NodeType.Root or NodeType.Other.
```
bjorn: I'd prefer to have separate `createElement`, `createTextNode` and `createCDATASection` functions for this. Why did you merge them into a single function with type parameter?

src/tiled/scriptxmlfile.cpp
```diff
@@ -0,0 +1,306 @@
...
+    case NodeType::Root:
+        ScriptManager::instance().throwError(QStringLiteral("Cannot use createNode to make XmlFile instances. Use the constructor instead."));
+        return nullptr;
+    default:
+        ScriptManager::instance().throwError(QStringLiteral("Cannot create node type"));
+        return nullptr;
+    }
+}
+
+QString ScriptXmlFile::writeToString(int indent) const
```
bjorn: Did just calling this `toString` conflict with the JavaScript to string conversion?

src/tiled/scriptxmlfile.h
```diff
@@ -0,0 +1,119 @@
...
+    inline ScriptXmlNode *checkIfNull(const QDomNode &node) const;
+
+    QDomNode mNode;
+};
+
+class ScriptXmlFile: public ScriptXmlNode
+{
+    Q_OBJECT
+public:
+    Q_INVOKABLE explicit ScriptXmlFile(const QString &source, QObject *parent = nullptr);
```
bjorn: We have `TextFile` and `BinaryFile` taking a file name as parameter, but now `XmlFile` taking a source string.

I'd prefer to call this `DomDocument` (and the node `DomNode`), to be consistent with Qt API and Qbs Xml Service. I think then also avoiding the "File" postfix makes it acceptable to take a source string in the constructor.

bjorn:
Closing this due to inactivity, and also because I think it's probably better to wrap the `QXmlStreamReader` and `QXmlStreamWriter` instead, which are faster, simpler to use and avoid a dependency on QtXml.

Thanks anyway for your work on this @Phlosioneer! Sorry for change of opinion.

PR #2736, opened 2020-01-17
Added checks for nullptr in script arguments

Phlosioneer:
QJSEngine can pass a nullptr to a `Q_INVOKABLE` or `Q_PROPERTY` function if the argument is either `undefined` or the wrong type.

### DIFF
--- a/src/tiled/editablemap.cpp
+++ b/src/tiled/editablemap.cpp
@@ -497,10 +497,6 @@ void EditableMap::setLayerDataFormat(LayerDataFormat value)
 
 void EditableMap::setCurrentLayer(EditableLayer *layer)
 {
-    if (!layer) {
-        ScriptManager::instance().throwNullArgError(0);
-        return;
-    }
     QList<QObject*> layers;
     if (layer)
         layers.append(layer);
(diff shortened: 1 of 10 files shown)


src/tiled/editablemap.cpp
```diff
@@ -477,6 +497,10 @@ void EditableMap::setLayerDataFormat(LayerDataFormat value)
 
 void EditableMap::setCurrentLayer(EditableLayer *layer)
 {
+    if (!layer) {
+        ScriptManager::instance().throwNullArgError(0);
```
bjorn: This is not an error for this function, see check below.

src/tiled/scriptmanager.cpp
```diff
@@ -260,6 +261,13 @@ void ScriptManager::throwError(const QString &message)
 #endif
 }
 
+void ScriptManager::throwNullArgError(int argNumber)
+{
+    auto message = QCoreApplication::translate("Script Errors",
+                                               "Argument %1 is undefined or the wrong type").arg(argNumber);
```
bjorn: I would opt to skip the local `message` variable.

PR #2742, opened 2020-01-28
Add reference to Liti Engine

Phlosioneer:


bjorn:
This should probably go under "Support by Framework"? Though I guess the distinction is a little bit arbitrary in some cases.

bjorn:
Actually I see LITIengine is [already mentioned](http://doc.mapeditor.org/en/stable/reference/support-for-tmx-maps/#litiengine) among the frameworks.

Phlosioneer:
ah, I missed it.

PR #2758, opened 2020-02-26
Fixed file reading test

oninbo:


bjorn:
Hmm, so the test as running on Travis CI apparently broke with this change. How are you running the tests?

I agree it would be nice to adjust the tests to not be sensitive to the working directory. However, I'm not sure if putting the test files in a resource file is the best way. Alternatively, we can add install rules to the Qbs project to copy these files alongside the binary, so that the test can find them relative to the binary.

PR #2772, opened 2020-03-18
tiledquickplugin: only render visible tiles 

### LINKED ISSUE #2769: Tiled Quick: rendering only visible tiles in isometric maps
Support for isometric maps in tiledquickplugin was added in #1535, but all tiles in the map are rendered, rather than only those tiles within the MapItem's visibleArea.

mitchcurtis:
Fixes #2769

### DIFF
--- a/src/libtiled/hexagonalrenderer.h
+++ b/src/libtiled/hexagonalrenderer.h
@@ -79,6 +79,10 @@ public:
     void drawTileLayer(QPainter *painter, const TileLayer *layer,
                        const QRectF &exposed = QRectF()) const override;
 
+    void drawTileLayer(const TileLayer *layer,
+                       const RenderTileCallback renderTileCallback,
+                       const QRectF &exposed = QRectF()) const override;
+
     void drawTileSelection(QPainter *painter,
                            const QRegion &region,
                            const QColor &color,
--- a/src/libtiled/hexagonalrenderer.h
+++ b/src/libtiled/hexagonalrenderer.h
@@ -80,7 +80,7 @@ public:
                        const QRectF &exposed = QRectF()) const override;
 
     void drawTileLayer(const TileLayer *layer,
-                       const RenderTileCallback &renderTileCallback,
+                       const RenderTileCallback &renderTile,
                        const QRectF &exposed = QRectF()) const override;
 
     void drawTileSelection(QPainter *painter,
--- a/src/libtiled/hexagonalrenderer.h
+++ b/src/libtiled/hexagonalrenderer.h
@@ -80,7 +80,7 @@ public:
                        const QRectF &exposed = QRectF()) const override;
 
     void drawTileLayer(const TileLayer *layer,
-                       const RenderTileCallback renderTileCallback,
+                       const RenderTileCallback &renderTileCallback,
                        const QRectF &exposed = QRectF()) const override;
 
     void drawTileSelection(QPainter *painter,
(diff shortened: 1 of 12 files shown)


mitchcurtis:
Nice, thanks! All works great now.

I just noticed that I have some leftovers in src/tiledquick/qml/main.qml. Will this changes be squashed and so I should just do a fixup commit? Or should I rebase 9f2d7b0 (and I guess force push)?

bjorn:
Yeah, that was a nice cleanup in Tiled Quick and now all orientations are supported. :-)

I wonder a bit about the impact on performance of adding a `std::function` call for each tile, but I did not measure it. A debug build is slow for big maps, but that was already the case. In release mode I did not notice any difference.

> I just noticed that I have some leftovers in src/tiledquick/qml/main.qml. Will this changes be squashed and so I should just do a fixup commit? Or should I rebase [9f2d7b0](https://github.com/bjorn/tiled/commit/9f2d7b0366e8f4bde6e9ad6f85c2e62f3b0a3cf3) (and I guess force push)?

Let's squash this, so please push a cleanup commit. :-)

src/libtiled/hexagonalrenderer.h
```diff
@@ -79,6 +79,10 @@ class TILEDSHARED_EXPORT HexagonalRenderer : public OrthogonalRenderer
     void drawTileLayer(QPainter *painter, const TileLayer *layer,
                        const QRectF &exposed = QRectF()) const override;
 
+    void drawTileLayer(const TileLayer *layer,
+                       const RenderTileCallback renderTileCallback,
```
bjorn:
```suggestion
                       const RenderTileCallback &renderTileCallback,
```
Please change this in all cases. I don't think we want to copy the `std::function`.

mitchcurtis:
> I wonder a bit about the impact on performance of adding a `std::function` call for each tile, but I did not measure it. A debug build is slow for big maps, but that was already the case. In release mode I did not notice any difference.

I figured it was the lesser of two evils, but that was because I was assuming that using an interface would require memory to be allocated... thinking about it a bit more now, passing a helper object thingy that was created on the stack should also work.

bjorn:
> I figured it was the lesser of two evils, but that was because I was assuming that using an interface would require memory to be allocated... thinking about it a bit more now, passing a helper object thingy that was created on the stack should also work.

The `std::function` wrapping a lambda likely takes pretty much the same amount of memory as the helper object would (though indeed it may be heap allocated rather than stack allocated, depending on how big it is). Also, the helper object would need a virtual member function, so it's likely to have similar overhead to the `std::function`. But would need to be measured to be sure. :-)

I think the only way to have really less overhead would be to use a template parameter, but it would mean moving the entire tile layer rendering code into the header as well as to remove the virtual dispatch for `drawTileLayer`.

mitchcurtis:
I've noticed that maps are now shifted half a tile upwards:

```
diff --git a/src/tiledquick/qml/main.qml b/src/tiledquick/qml/main.qml
index d67909ea..5ddfdff6 100644
--- a/src/tiledquick/qml/main.qml
+++ b/src/tiledquick/qml/main.qml
@@ -151,6 +151,12 @@ ApplicationWindow {
                             mapView.width / scale,
                             mapView.height / scale);
                 }
+
+                Rectangle {
+                    anchors.fill: parent
+                    color: "transparent"
+                    border.color: "darkorange"
+                }
             }
         }
     }

```

Screen Shot 2020-03-20 at 5 23 48 pm

bjorn:
> I've noticed that maps are now shifted half a tile upwards. Do you have any idea which part of our recent changes could be causing this?

So, actually only maps using that particular tileset would appear shifted like that. And it's because the tiles are 64 pixels high and have a transparent region at the bottom. In addition, our tile rendering callback is not applying the drawing offset of the tileset. I'll fix that up now. :-)

mitchcurtis:
Thank you Bjorn! Now I can start making some bigger maps! :D

PR #2777, opened 2020-04-07
Don't derive base libtiled classes from QObject

bjorn:
Deriving from QObject makes them rather more heavy-weight than necessary. The only reason they were derived from QObject was to expose them to QML, but so far this functionality is not actually used. Only Tiled::Map was passed on in QML and this use-case was replaced by a small wrapper.

In general we now have "Editable" wrappers around these base objects, which I'd aim to expose in the TiledQuickPlugin somehow, for use by Tiled Quick.

@mitchcurtis Since you may be relying on these classes, please let me know if this would cause you any headaches.

### DIFF
--- a/src/tiledquickplugin/mapitem.h
+++ b/src/tiledquickplugin/mapitem.h
@@ -20,12 +20,13 @@
 
 #pragma once
 
+#include "mapref.h"
+
 #include <QQuickItem>
 
 #include <memory>
 
 namespace Tiled {
-class Map;
 class MapRenderer;
 } // namespace Tiled
 
@@ -41,15 +42,15 @@ class MapItem : public QQuickItem
 {
     Q_OBJECT
 
-    Q_PROPERTY(Tiled::Map *map READ map WRITE setMap NOTIFY mapChanged)
+    Q_PROPERTY(TiledQuick::MapRef map READ map WRITE setMap NOTIFY mapChanged)
     Q_PROPERTY(QRectF visibleArea READ visibleArea WRITE setVisibleArea NOTIFY visibleAreaChanged)
 
 public:
     explicit MapItem(QQuickItem *parent = nullptr);
     ~MapItem();
 
-    Tiled::Map *map() const;
-    void setMap(Tiled::Map *map);
+    MapRef map() const;
+    void setMap(MapRef map);
 
     const QRectF &visibleArea() const;
     void setVisibleArea(const QRectF &visibleArea);
@@ -86,7 +87,7 @@ private:
 inline const QRectF &MapItem::visibleArea() const
 { return mVisibleArea; }
 
-inline Tiled::Map *MapItem::map() const
-{ return mMap; }
+inline MapRef MapItem::map() const
+{ return { mMap }; }
 
 } // namespace TiledQuick
--- a/src/tiledquickplugin/mapref.h
+++ b/src/tiledquickplugin/mapref.h
@@ -1,6 +1,6 @@
 /*
- * tiledquickplugin.cpp
- * Copyright 2014, Thorbjørn Lindeijer <bjorn@lindeijer.nl>
+ * mapref.h
+ * Copyright 2020, Thorbjørn Lindeijer <bjorn@lindeijer.nl>
  *
  * This file is part of Tiled Quick.
  *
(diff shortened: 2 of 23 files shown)


mitchcurtis:
Thanks for the heads up. I'll take a look.

mitchcurtis:
I gotta work by this one usage at a time... the first issue I see is that TiledQuick::MapRef is defined in tiledquickplugin, which means I can't use it from C++.

I could maybe work around that in one place (Level.cpp - mentioning it mostly for my own reference so I don't forget :)) because it looks like I only need the size of the map there. However, beyond MapRef not being exposed by a library, it also has no size properties, and since Map is obviously no longer a QObject, it doesn't have any size properties either.

bjorn:
@mitchcurtis Is your code available somewhere? I'm having some difficulties understanding your problems, and maybe looking at the code would help.

If you need to use `MapRef` from C++, couldn't you just include the `mapref.h` from the `tiledquickplugin` module? Do I assume correctly that your project has classes exposed to QML that are taking a `Tiled::Map*` parameter? To make things easier we could of course move `MapRef` into `map.h`.

Note that I previously tried to just add `Q_GADGET` to `Tiled::Map`, but then I was running into a crash when `MapLoader` emitted `sourceChanged`. I'm not sure why this didn't work.

> However, beyond MapRef not being exposed by a library, it also has no size properties, and since Map is obviously no longer a QObject, it doesn't have any size properties either.

So in your project you're relying on those properties? If we manage to make `Tiled::Map` a gadget then it could again have properties (just no signals). Otherwise, would an `EditableMap` instance be of any help? The biggest challenge there will be to untangle it from the editor in general.

mitchcurtis:
> @mitchcurtis Is your code available somewhere? I'm having some difficulties understanding your problems, and maybe looking at the code would help.

It's not but I can email you the relevant parts if you like?

> If you need to use `MapRef` from C++, couldn't you just include the `mapref.h` from the `tiledquickplugin` module?

For some reason I didn't even consider that because it's a plugin. :x One problem I see there though, is that MapRef is not exported. Somehow I can still use it on macOS though.

> Do I assume correctly that your project has classes exposed to QML that are taking a `Tiled::Map*` parameter?

Yes.

> To make things easier we could of course move `MapRef` into `map.h`.

If I can use it from the plugin then I'm happy. :)

> Note that I previously tried to just add `Q_GADGET` to `Tiled::Map`, but then I was running into a crash when `MapLoader` emitted `sourceChanged`. I'm not sure why this didn't work.

Did you report it? The QML engine guys are quite responsive these days.

> > However, beyond MapRef not being exposed by a library, it also has no size properties, and since Map is obviously no longer a QObject, it doesn't have any size properties either.
> 
> So in your project you're relying on those properties? If we manage to make `Tiled::Map` a gadget then it could again have properties (just no signals). Otherwise, would an `EditableMap` instance be of any help? The biggest challenge there will be to untangle it from the editor in general.

Let me see if I can work something out first... I need a bit more time to play around. :)

mitchcurtis:
I also notice that MapRef::mMap is not initialised to nullptr?

src/tiledquickplugin/mapitem.h
```diff
@@ -41,15 +42,15 @@ class MapItem : public QQuickItem
 {
     Q_OBJECT
 
-    Q_PROPERTY(Tiled::Map *map READ map WRITE setMap NOTIFY mapChanged)
+    Q_PROPERTY(TiledQuick::MapRef map READ map WRITE setMap NOTIFY mapChanged)
```
mitchcurtis, line 45: Can you please add a RESET function? That way I can set map to undefined in QML just to be safe, when a level is exited. The function only needs to set mMap.mMap to nullptr.

src/tiledquickplugin/mapref.h
```diff
@@ -0,0 +1,43 @@
...
+}
+
+namespace TiledQuick {
+
+class MapRef
+{
+    Q_GADGET
+
+public:
+    Tiled::Map *mMap;
```
mitchcurtis, line 36: I think I ran into issues with this not being set to nullptr, as I check whether or not the pointer is valid in a few places.

src/tiledquickplugin/mapref.h
```diff
@@ -0,0 +1,43 @@
...
+
+#include <QObject>
+
+namespace Tiled {
+class Map;
+}
+
+namespace TiledQuick {
+
+class MapRef
```
mitchcurtis, line 31: Should this be exported? It seems I'm able to use it on macOS, which is surprising, but I doubt it would work on e.g. Windows.

bjorn: All its members are inline so I don't think it needs exporting.

src/tiledquickplugin/mapref.h
bjorn: I didn't add it because I never construct a `MapRef` without passing it an existing pointer. I'll add a constructor to force initialization.

src/tiledquickplugin/mapitem.h
bjorn: Yes, that's a good alternative to allowing constructing a null `MapRef` instance from QML!

mitchcurtis:
It seems I can quite easily adapt to the changes. :) Would be nice to have that RESET function though, just to satisfy my paranoia. :D

bjorn:
> It's not but I can email you the relevant parts if you like?

I guess there's no real reason to since it all seems clear now.

> Did you report it? The QML engine guys are quite responsive these days.

No, but I'll be sure to try it again and consider opening a report. The problem is always trying to find a small example that reproduces the issue. :-)

src/tiledquickplugin/mapitem.h
mitchcurtis: I tried assigning null but it gave me an error; probably something to do with it being a gadget, because I know that works with QObject*.

bjorn: Sure, just assigning `null` will not work since it's not a pointer. An alternative to providing an `unsetMap` or `clear` function would be to allow say `mapItem.map = new MapRef()`.

bjorn:
> No, but I'll be sure to try it again and consider opening a report. The problem is always trying to find a small example that reproduces the issue. :-)

@mitchcurtis I'm not sure if it is documented, but it appears to be an unsupported scenario according to [QTBUG-54321](https://bugreports.qt.io/browse/QTBUG-54321). The main problem with supporting Q_GADGET pointers is that there would be no way for the QML engine to know whether the pointer is still valid (as opposed to Q_OBJECT pointers), so currently they are only supported by value. I don't like that it crashes instead of giving some error pointing in this direction (wasting a lot of time for some people), but at least we now know Q_GADGET pointer is a dead-end here.

The alternative solution with the `MapRef` class is one of the possible workarounds (where of course, the QML engine also can't know whether the internal pointer is still valid, but I guess it's fine now because the application chose to do this and any problems resulting from that are now my fault :P).

Long-term, once we have the QObject wrappers exposed in Tiled Quick, this will likely replace the `MapRef` solution.

Issue #2780, opened 2020-04-19
Javascript plugin stdlib


konsumer:
First of all, great work on tiled. I love it, and have used it for years. It pretty much works great for anything that uses tiles I have ever tried it with. Excellent software.

Recently, I was using [tiled-to-godot-export](https://github.com/MikeMnD/tiled-to-godot-export), which uses the javascript plugin API to generate godot map & tile definitions. It works pretty well, but in order to support relative paths, I had to roll my own path function (that is not nearly as good as node's `path.resolve`.)

It'd be really useful for plugin API to have some basic js stdlib type stuff like node's exec/os/path/fs for manipulating paths, spawning processes, introspecting the environment, and manipulating files, directly. In my opinion, keeping it as close as possible to the node API would be really great, since we can go light on documentation, and just point people there, and it will be familiar to experienced javascript devs. "It works like `path.resolve`" would be super-helpful to people like me.

I briefly explored the exposed `Qt` object, hoping it would have something in it, but I couldn't find anything related.

I am making this issue to start a discussion about it, and I am happy to help with anything I can, even though I don't do too much C++/Qt dev, I can get around ok, have a lot of experience with node, and could probly work out how to add support.

How should this look?  Maybe a first step would be to make `path` bindings? If the `posix` and `win32` specific sets are stripped out, you get a pretty small surface, so it might be a useful & relatively easy starting point to expose:

```
path.basename(path[, ext])
path.delimiter
path.dirname(path)
path.extname(path)
path.format(pathObject)
path.isAbsolute(path)
path.join([...paths])
path.normalize(path)
path.parse(path)
path.relative(from, to)
path.resolve([...paths])
path.sep
path.toNamespacedPath(path)
```

konsumer:
With my basic understanding, I am guessing I need to add a class to [script-manager](https://github.com/bjorn/tiled/blob/master/src/tiled/scriptmanager.cpp#L285-L294). How does `Stdlib` global sound for this kind of stuff? like start with `Stdlib.path`, then users can do this:

```js
const { path: { basename } } = Stdlib
console.log(basename(file))
```

which looks very similar to equivalent in node:

```js
const { basename } = require('path')
console.log(basename(file))
```

konsumer:
Or maybe it should go in `ScriptModule`, so it's all available in `tiled` object? I'm really not sure.

bjorn:
Thanks @konsumer for the kind words and the thorough suggestion!

[...]

PR #2782, opened 2020-04-21
Feature 2780 script file info

konsumer:
This adds some file info utils to JS plugin API (#2780)

You can see example usage [here](https://github.com/konsumer/tiled-fileinfodemo/blob/master/export_fileinfo_tilemap.js).

Example output:

```json
{
  "absoluteDir": "/home/konsumer/Documents/otherdev/tiled-example",
  "absoluteFilePath": "/home/konsumer/Documents/otherdev/tiled-example/town1.json",
  "absolutePath": "/home/konsumer/Documents/otherdev/tiled-example",
  "baseName": "town1",
  "birthTime": "2020-04-20T23:20:18.807Z",
  "bundleName": "",
  "caching": true,
  "canonicalFilePath": "/home/konsumer/Documents/otherdev/tiled-example/town1.json",
  "canonicalPath": "/home/konsumer/Documents/otherdev/tiled-example",
  "completeBaseName": "town1",
  "completeSuffix": "json",
  "dir": "/home/konsumer/Documents/otherdev/tiled-example",
  "exists": true,
  "fileName": "town1.json",
  "filePath": "/home/konsumer/Documents/otherdev/tiled-example/town1.json",
  "fileTimeAccess": "2020-04-20T23:20:18.807Z",
  "fileTimeBirth": "2020-04-20T23:20:18.807Z",
  "fileTimeMetaChange": "2020-04-20T23:45:57.840Z",
  "fileTimeModification": "2020-04-20T23:45:57.840Z",
  "group": "konsumer",
  "groupId": 1000,
  "isAbsolute": true,
  "isBundle": false,
  "isDir": false,
  "isExecutable": false,
  "isFile": true,
  "isHidden": false,
  "isNativePath": true,
  "isReadable": true,
  "isRelative": false,
  "isRoot": false,
  "isShortcut": false,
  "isSymLink": false,
  "isSymbolicLink": false,
  "isWritable": true,
  "lastModified": "2020-04-20T23:45:57.840Z",
  "lastRead": "2020-04-20T23:20:18.807Z",
  "makeAbsolute": false,
  "metadataChangeTime": "2020-04-20T23:45:57.840Z",
  "owner": "konsumer",
  "ownerId": 1000,
  "path": "/home/konsumer/Documents/otherdev/tiled-example",
  "permissions": 26180,
  "size": 1424,
  "suffix": "json",
  "symLinkTarget": ""
}
```

### DIFF
--- a/src/tiled/scriptfileinfo.cpp
+++ b/src/tiled/scriptfileinfo.cpp
@@ -117,11 +117,12 @@ QString ScriptFileInfo::filePath(QString file)
 	return fp.filePath();
 }
 
-// QDateTime ScriptFileInfo::fileTime(QString file, QFile::FileTime time)
-// {
-// 	QFileInfo fp = QFileInfo(file);
-// 	return fp.fileTime(time);
-// }
+// https://doc.qt.io/qt-5/qfiledevice.html#FileTime-enum
+QDateTime ScriptFileInfo::fileTime(QString file, uint time)
+{
+	QFileInfo fp = QFileInfo(file);
+	return fp.fileTime((QFile::FileTime)time);
+}
 
 QString ScriptFileInfo::group(QString file)
 {
@@ -261,11 +262,11 @@ QString ScriptFileInfo::path(QString file)
 	return fp.path();
 }
 
-// bool  ScriptFileInfo::permission(QString file, QFile::Permissions permissions)
-// {
-// 	QFileInfo fp = QFileInfo(file);
-// 	return fp.permission(permissions);
-// }
+bool  ScriptFileInfo::permission(QString file, uint permissions)
+{
+	QFileInfo fp = QFileInfo(file);
+	return fp.permission((QFile::Permissions)permissions);
+}
 
 uint ScriptFileInfo::permissions(QString file)
 {
--- a/src/tiled/scriptfileinfo.cpp
+++ b/src/tiled/scriptfileinfo.cpp
@@ -279,12 +279,6 @@ void  ScriptFileInfo::refresh(QString file)
 	return fp.refresh();
 }
 
-void  ScriptFileInfo::setCaching(QString file, bool enable)
-{
-	QFileInfo fp = QFileInfo(file);
-	return fp.setCaching(enable);
-}
-
 qint64  ScriptFileInfo::size(QString file)
 {
 	QFileInfo fp = QFileInfo(file);
--- a/src/tiled/scriptfileinfo.cpp
+++ b/src/tiled/scriptfileinfo.cpp
@@ -1,0 +1,57 @@
+/*
+ * scriptfileinfo.cpp
+ * Copyright 2019, Thorbjørn Lindeijer <bjorn@lindeijer.nl>
+ *
+ * This file is part of Tiled.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the Free
+ * Software Foundation; either version 2 of the License, or (at your option)
+ * any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "scriptfileinfo.h"
+#include <QtCore/qcoreapplication.h>
+#include <QtCore/qdatetime.h>
+#include <QtCore/qdir.h>
+#include <QtCore/qfileinfo.h>
+#include <QtCore/qregexp.h>
+
+#if defined(Q_OS_UNIX)
+#include <cerrno>
+#include <sys/stat.h>
+#include <unistd.h>
+#elif defined(Q_OS_WIN)
+#include <QtCore/qt_windows.h>
+#endif
+
+ScriptFileInfo::ScriptFileInfo(QObject *parent)
+		: QObject(parent)
+{
+
+}
+
+QString ScriptFileInfo::fileName(const QString &fp)
+{
+	int last = fp.lastIndexOf(QLatin1Char('/'));
+	if (last < 0)
+		return fp;
+	return fp.mid(last + 1);
+}
+
+QString ScriptFileInfo::baseName(const QString &fp)
+{
+	QString fn = fileName(fp);
+	int dot = fn.indexOf(QLatin1Char('.'));
+	if (dot < 0)
+		return fn;
+	return fn.mid(0, dot);
+}
--- a/src/tiled/scriptfileinfo.cpp
+++ b/src/tiled/scriptfileinfo.cpp
@@ -264,14 +264,14 @@ QString ScriptFileInfo::path(QString file)
 // bool  ScriptFileInfo::permission(QString file, QFile::Permissions permissions)
 // {
 // 	QFileInfo fp = QFileInfo(file);
-// 	return fp.permission();
+// 	return fp.permission(permissions);
 // }
 
-// QFile::Permissions  ScriptFileInfo::permissions(QString file)
-// {
-// 	QFileInfo fp = QFileInfo(file);
-// 	return fp.permissions();
-// }
+uint ScriptFileInfo::permissions(QString file)
+{
+	QFileInfo fp = QFileInfo(file);
+	return fp.permissions();
+}
 
 void  ScriptFileInfo::refresh(QString file)
 {
--- a/src/tiled/scriptfileinfo.cpp
+++ b/src/tiled/scriptfileinfo.cpp
@@ -19,39 +19,286 @@
  */
 
 #include "scriptfileinfo.h"
-#include <QtCore/qcoreapplication.h>
-#include <QtCore/qdatetime.h>
-#include <QtCore/qdir.h>
-#include <QtCore/qfileinfo.h>
-#include <QtCore/qregexp.h>
 
-#if defined(Q_OS_UNIX)
-#include <cerrno>
-#include <sys/stat.h>
-#include <unistd.h>
-#elif defined(Q_OS_WIN)
-#include <QtCore/qt_windows.h>
-#endif
 
 ScriptFileInfo::ScriptFileInfo(QObject *parent)
 		: QObject(parent)
-{
+{}
 
+QString ScriptFileInfo::absoluteDir(QString file)
+{
+	QFileInfo fp = QFileInfo(file);
+	QDir dir = fp.dir();
+	return dir.absolutePath();
 }
 
-QString ScriptFileInfo::fileName(const QString &fp)
+QString ScriptFileInfo::absoluteFilePath(QString file)
 {
-	int last = fp.lastIndexOf(QLatin1Char('/'));
-	if (last < 0)
-		return fp;
-	return fp.mid(last + 1);
+	QFileInfo fp = QFileInfo(file);
+	return fp.absoluteFilePath();
 }
 
-QString ScriptFileInfo::baseName(const QString &fp)
+QString ScriptFileInfo::absolutePath(QString file)
 {
-	QString fn = fileName(fp);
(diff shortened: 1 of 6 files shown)


konsumer:
Hmm, build locally on QT 5.14.2. But, it looks like the version on travis doens't like a few things. I'll try wrapping it in QT version check.

konsumer:
Ugh, looks like there are a few differences between [qt-5.14]() and [qt-5.9](https://doc.qt.io/qt-5.9/qfileinfo.html). I will try to make a little compatibility-layer that matches latest. It looks like it's a slightly nicer API-shape, anyway.

konsumer:
I put some `QT_VERSION` guards up, and moved things around a little, and it builds. I guess it should be noted in docs that QT < 5.10 will have no `isShortcut()`, as I couldn't find a replacement. Also 'metadataChangeTime' just returns `lastModified` on QT < 5.10, which may be the wrong thing, but it's all I could find.

bjorn:
Alright, so I think it's better to just leave out the functions that don't have an exact equivalent on older Qt versions. I left `FileInfo.isSymbolicLink` because I think the fact that it happens to return true for shortcuts for older versions of Qt will be irrelevant. Seems like the Qt docs are missing some "since 1.4" for the `isShortcut` and `isSymbolicLink` functions.

In any case, thanks a lot for all the work put into this! It is actually a lot more than I had intended to support (I would have just mirrored the rather minimal API of the [FileInfo service in Qbs](https://doc.qt.io/qbs/jsextension-fileinfo.html) to start with).

On the one hand we have a lot of functions, but on the other hand we're still missing some important functionality. In Qbs, we have `FileInfo.relativePath`, `FileInfo.cleanPath` and possibly `FileInfo.from/toNativeSeparators`. These actually map to methods on `QDir` rather than `QFileInfo`, but I think it makes sense to include them here.

konsumer:
> Alright, so I think it's better to just leave out the functions that don't have an exact equivalent on older Qt versions.

Sounds like solid strategy.

> On the one hand we have a lot of functions, but on the other hand we're still missing some important functionality. In Qbs, we have FileInfo.relativePath, FileInfo.cleanPath and possibly FileInfo.from/toNativeSeparators. These actually map to methods on QDir rather than QFileInfo, but I think it makes sense to include them here.

I was also thinking we should have some separator-manipulation stuff. Since I am downgrading `QDir` to strings, we definitely should add it's util functions. I see what you did with dropping `fileTime` in older versions instead of shimming it. I like that.

konsumer:
Is there anything I should add, other than these?

```cpp
QString relativePath(const QString &file)
QString cleanPath(const QString &file)
QString fromNativeSeparators(const QString &file)
QString toNativeSeparators(const QString &file)
```

konsumer:
I added stuff from `QDir`. Should we call it `FIle` instead of `FileInfo` in js?

src/tiled/scriptfileinfo.cpp
```diff
@@ -242,3 +242,74 @@ QString ScriptFileInfo::symLinkTarget(const QString &file) const
 {
     return QFileInfo(file).symLinkTarget();
 }
+
+QString ScriptFileInfo::cleanPath(const QString &file) const
+{
+    return QDir().cleanPath(file);
```
bjorn, line 248: This is a static method, so it doesn't need an instance. It should be `return QDir::cleanPath(file)`.

src/tiled/scriptfileinfo.cpp
```diff
@@ -242,3 +242,74 @@ QString ScriptFileInfo::symLinkTarget(const QString &file) const
...
+}
+
+QString ScriptFileInfo::fromNativeSeparators(const QString &file) const
+{
+    return QDir().fromNativeSeparators(file);
+}
+
+QString ScriptFileInfo::toNativeSeparators(const QString &file) const
+{
+    return QDir().fromNativeSeparators(file);
```
bjorn, line 258: Should be `toNativeSeparators`. Also, these methods are static as well.

src/tiled/scriptfileinfo.cpp
```diff
@@ -242,3 +242,74 @@ QString ScriptFileInfo::symLinkTarget(const QString &file) const
...
+}
+
+bool  ScriptFileInfo::cd(const QString &file)
+{
+    return QDir().cd(file);
```
bjorn, line 263: This is not a very useful binding, with `QDir::cd` just meant to change the directory the QDir instance is pointing to, but we're changing a temporary instance. The only use for this would to check if the directory exists, since that's what it returns, but we already have `exists` for this.

src/tiled/scriptfileinfo.cpp
```diff
@@ -242,3 +242,74 @@ QString ScriptFileInfo::symLinkTarget(const QString &file) const
...
+}
+
+uint  ScriptFileInfo::count(const QString &file) const
+{
+    return QDir(file).count();
```
bjorn, line 268: Let's not include this one.

src/tiled/scriptfileinfo.cpp
```diff
@@ -242,3 +242,74 @@ QString ScriptFileInfo::symLinkTarget(const QString &file) const
...
+}
+
+QString ScriptFileInfo::dirName(const QString &file) const
+{
+    return QDir(file).dirName();
```
bjorn, line 273: This is really the same as `fileName`, so I'd suggest we leave it out.

src/tiled/scriptfileinfo.cpp
```diff
@@ -242,3 +242,74 @@ QString ScriptFileInfo::symLinkTarget(const QString &file) const
...
+}
+
+bool  ScriptFileInfo::mkpath(const QString &file) const
+{
+    return QDir().mkpath(file);
+}
+
+bool  ScriptFileInfo::remove(const QString &file)
+{
+    return QDir().remove(file);
```
bjorn, line 288: This just removes a file, so let's call [`QFile::remove`](https://doc.qt.io/qt-5/qfile.html#remove-1) for it instead.

src/tiled/scriptfileinfo.cpp
```diff
@@ -242,3 +242,74 @@ QString ScriptFileInfo::symLinkTarget(const QString &file) const
...
+}
+
+bool  ScriptFileInfo::removeRecursively(const QString &file)
+{
+    return QDir(file).removeRecursively();
+}
+
+bool  ScriptFileInfo::rename(const QString &oldName, const QString &newName)
+{
+    return QDir().rename(oldName, newName);
```
bjorn, line 298: I think here we should also just use the static [`QFile::rename`](https://doc.qt.io/qt-5/qfile.html#rename-1), which should also work for directories (though the docs don't say so).

bjorn:
> I started making [Qbs4QJS](https://github.com/konsumer/Qbs4QJS) to encapsulate all the functionality of "expose Qbs services in QJSEngine".

This repository should really be described as "Port of Qbs JavaScript extensions to QJSEngine" since currently it suggests it would be a port of the whole Qbs to QJSEngine, which is a challenging problem that they've been struggling with for years already especially considering QScriptEngine is deprecated (and the problem is, that QScriptEngine can still do a bunch of things that QJSEngine can't...).

I do like the idea of putting this in a separate repository in a clean way so that it could be useful to other projects as well.

Is the intention then, to finish that repository and then to use it in Tiled, instead of this PR? I personally find it rather hard to decide whether to just expose all these functions or whether to stay with the rather smaller subset that is also exposed by Qbs. Note that Qbs not only exposes a subset, but also includes a number of useful behaviors (for example, its `File.copy` can copy both files and directories, and the target path will be created automatically). I rather like this simplified API.

konsumer:
Wow, I made a lot of dumb mistakes above. Sorry, I think I wasn't paying enough attention. I will get those cleaned up. I was having a bit of trouble getting things to work statically, and just started using instances everywhere, I will make the changes, above.

> This repository should really be described as "Port of Qbs JavaScript extensions to QJSEngine" since currently it suggests it would be a port of the whole Qbs to QJSEngine, which is a challenging problem that they've been struggling with for years already especially considering QScriptEngine is deprecated (and the problem is, that QScriptEngine can still do a bunch of things that QJSEngine can't...).

Yeh, I just meant the JS services, specifically. I updated the description to your suggestion. I have been doing some reading, and it seems like one of the biggest hurdles was that there wasn't a good way (until fairly recently) to properly expose C++ functions to JS, if I understand it correctly. It seems like that part is mostly worked out. One thing I ran into already is variable-argument-count functions (like `FileInfo.joinPaths`, for example.) In those cases, I wasn't sure how to do it, so I have just used a `QStringList`. I think I will try to do that throughout: get it as close as I can to Qbs, with my limited knowledge, but be ok with any trades that are needed to make it work & useful. Later, I can refine to get it closer if people have better ideas, or I learn new tricks, on my own.

SInce I have no real investment in a total Qbs port (I've never used it before) I'd be happy to diverge completely from it and call the lib something else. I could see a hybrid approach, where I get Qbs copied, but then add some extra stuff that seemed useful. 

I was really surprised there wasn't a "javascript stdlib for basic OS-things" in `QJSEngine`, so that has sort of become my goal with it, partially because I want to totally deck out tiled's js plugin env, but also because I think it would be handy for a wider audience. I could really see myself using `QJSEngine` over node for things, if it had all the basic stuff built-in.

> Is the intention then, to finish that repository and then to use it in Tiled, instead of this PR?

That's what I was thinking, or at least start with the parts that are already useful to TIled. I can start with getting `FileInfo`, `File`, `TextFile`, and `BinaryFile` right, using stuff over here, and look through the Qbs code. I will follow whatever you want to do, though. I am happy to keep working on this PR until it works how you want for Tiled, and then use ideas from it over there, too, if you don't want to use that one directly, I can focus on getting it more complete. My first priority is to get Tiled's js utils working well, so this isn't in an incomplete or broken state, it just seemed like a good idea to spread the love with a shared lib that would be useful to as many people as possible.

I was surprised how easy it was to setup in a nice little CLI and tester javascript file, so I might be able to get the basic thing done faster over there, but I am happy to work in either direction. I like being able to make myself unit-tests, and really focus on the thing I am trying to do, which can be a bit harder when I need to make a plugin for tiled, open tiled, export a map, etc.

Screen Shot 2020-04-22 at 2 17 47 AM

However I do it, I think I should probly seperate `FileInfo` and `File`, like Qbs does, and I like your idea of copying Qbs signature, rather than `QFileInfo` + `QDir`. It seems like it's a slightly more useful set of utilities, even if it might take a bit more work to replicate.

src/tiled/scriptfileinfo.cpp
konsumer: sounds good.

src/tiled/scriptfileinfo.cpp
konsumer: sounds good. dumb error.

src/tiled/scriptfileinfo.cpp
konsumer: sounds good.

src/tiled/scriptfileinfo.cpp
konsumer: sounds good.

src/tiled/scriptfileinfo.cpp
konsumer: sounds good.

src/tiled/scriptfileinfo.cpp
konsumer: sounds good.

src/tiled/scriptfileinfo.cpp
konsumer: sounds good.

konsumer:
I am having some trouble with making the new lib. I think this PR is ready to go, if you want to use it, as it is, now. We could swap it out later, if I work out the issues on the lib.

The issue I am running into now is a nice reusable way to have access to the engine (for throwing errors in the JS runtime without using a global or whatever. For it to act like Qbs, it needs to be a class, not an instance, and there doesn't seem to be a way to get a reference to the engine that the class is running in.

I am considering just dropping the whole idea of re-implementing Qbs services as a bunch of JS classes, and just make a super-useful library that works more like the code here, that makes lots of temporary `QFile`, `QFileInfo`, etc instances. It wasn't such a bad way to go, and seemed much simpler.

konsumer:
Thanks for your help!

Issue #2695, opened 2019-12-12
Scripting API: New image format?


Phlosioneer:
Hey, I'm trying to use the new scripting API to allow tiled to read a new image format for tilesets. However, the API only allows registering handlers for Tileset files and TileMap files.

I'd like to suggest this as a workable API:
```
tiled.registerImageFormat(shortName: string, imageFormat: object): void

The `imageFormat` object is expected to have the following properties:
- name: String
    Name of the format as shown in the file dialog
- extension: String
    File extension used by the format
- read: function(fileName: string): Image
    A function that reads an image from the given file. Can use TextFile or BinaryFile to read the file.
- hasChanged: function(fileName: string, timeOfLastRead: date): bool
    A function that is called periodically to check if the image has updated. This function should be as fast as possible. timeOfLastRead is the timestamp when `read` was last called.
```

bjorn:
What image format are you looking to support? Qt only supports adding custom image formats through C++ plugins and I don't think it even supports even a C++ API to register custom formats. If this is about adding support for a common image format then maybe we could ship another image format plugin.

Alternatively, we could of course route all image loading through a custom function that also takes registered formats from JavaScript into account.

Why the custom `hasChanged` check? As long as the image is simply a file on disk, Tiled already watches it for changes (if it's a tileset image, because individual tile images and image layer images are not yet watched for changes...).

Phlosioneer:
I'm trying to use a generated image as a tileset image. That generated image has two input files. So the normal "watch the file" approach that's built in wouldn't work.

Currently I have to remember to update it whenever _either_ file changes, and I have to perform the generation in a particular order. It's error prone.

Phlosioneer:
As for the image format, in this case I'm just going to make a temporary png file and load that, after running the generator (if needed). So QT doesn't have to bend over backwards for the format, I just need to be able to intercept reads and override the auto-checking behavior.

The exact process in my case is that I'm taking a `.ase` file, exporting it to PNG, and then running a postprocessing script afterwards. If the script or the ase file change, I want the output to update accordingly.

Phlosioneer:
In the more general case, could we allow a script to generate BMP-formatted image data and then "trick" qt into thinking that it's a bmp file?

bjorn:
> Currently I have to remember to update it whenever _either_ file changes, and I have to perform the generation in a particular order. It's error prone.

[...]

PR #2787, opened 2020-04-22
Scripting: Enabled creating images and using them

bjorn:
Now a script can create an image, either directly or by loading from a file or data or based on raw data. Then, the image can be passed to Tileset.loadFromImage or Tile.setImage.

Related to issue #2695

### DIFF
--- a/docs/reference/scripting.rst
+++ b/docs/reference/scripting.rst
@@ -728,6 +728,152 @@ GroupLayer.addLayer(layer : :ref:`script-layer`) : void
     When adding a :ref:`script-tilelayer` to a map, the layer's width and
     height are automatically initialized to the size of the map (since Tiled 1.4.2).
 
+.. raw:: html
+
+   <div class="new">New in Tiled 1.5</div>
+
+.. _script-image:
+
+Image
+^^^^^
+
+Can be used to create, load, save and modify images. Also useful when writing
+an importer, where the image can be set on a tileset or its tiles (see :ref:`Tileset.loadFromImage <script-tileset-loadFromImage>` and :ref:`Tile.setImage <script-tile-setImage>`).
+
+Properties
+~~~~~~~~~~
+
+.. csv-table::
+    :widths: 1, 2
+
+    **width** : int, Width of the image in pixels.
+    **height** : int, Height of the image in pixels.
+    **depth** : int, Number of bits used to store a single pixel.
+    **size** : :ref:`script-size`, Size of the image in pixels.
+    **format** : :ref:`Format <script-image-format>`, Format of the image.
+
+Functions
+~~~~~~~~~
+
+new Image()
+    Constructs an empty image.
+
+new Image(width : int, height : int, format : :ref:`Format <script-image-format>`)
+    Constructs an image of the given size using the given format.
+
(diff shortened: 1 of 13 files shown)


bjorn:
@samhocevar I just finished this up and haven't gotten around to testing it nor documenting it, but this should give you quite a few options for getting image data into Tiled and setting it on tilesets or individual tiles, so hopefully it will help with porting your PICO-8 extension to JS.

With this change, images can be dynamically created from JS in three ways:

* Use `image.setPixel(x, y, index_or_rgb)` to set each pixel directly.
* Use `new Image(data, width, height, format)` to create an image from raw data, see the `Format` enum for the various supported arrangements.
* Use `image.loadFromData(data, format)` to load an image from data in a format like "png" or "bmp".

There are also the methods for modifying the palette.

samhocevar:
For future reference, I provided feedback for this feature in https://github.com/bjorn/tiled/issues/2695#issuecomment-704405170 and resolved current conflicts for this PR in https://github.com/bjorn/tiled/pull/2906

bjorn:
Thanks @samhocevar! Your help is really appreciated and I'll make sure to get around to reviewing this tomorrow.

I'll find out tomorrow but just a quick question, was there anything that stood out while resolving the conflicts? Because I'd prefer to rebase the change instead of merging. Not that it matters much because we can also squash when merging back to `master`.

samhocevar:
Thanks for the reply! A rebase would be harmless; the conflicts just arose in `tiled.pro` and `tiled.qbs` because of the recent addition of `scriptfileinfo.cpp` which was too close to the `scriptfileimage.cpp` entry in the files.

bjorn:
> * `Image.setColorTable(["#123abc"])` does not work properly, because `ScriptImage::setColorTable` explicitly casts the array elements to `UInt`. I have to do `Image.setColorTable([0xff123abc])` instead, which is OK but a bit inconsistent with _e.g._ `Tileset.backgroundColor` which accepts the `"#123abc"` argument.

I've fixed this in 9124a7c88881fbc6df8ae6bd919366c614cc928d.

> * calling `Tileset.loadFromImage()` before `Tileset.setTileSize()` does not work; the order matters here, otherwise Tiled tries to reload a non-existent image.

Yeah, I'm not quite how to handle this, including how to handle undo when setting the image this way, or saving. I guess in case of saving this could lead to storing the image embedded in the file. I'll need to look into this before it can be merged.

bjorn:
@samhocevar For now I just documented that `Tileset.loadFromImage` needs to be called after `Tileset.setTileSize`, since it actually cuts up the image so it needs the tile size first. In addition, once the tileset contains tiles but has no "image" set on it, it is assumed to be a collection of images tileset and on such a tileset you're not allowed to set the tile size (because for such a tileset the tile size is automatically set to the largest tile in the set).

If there are any problems with the current API, please let me know!

bjorn:
@samhocevar I noticed that you currently have the following code:

https://github.com/samhocevar/tiled-pico8/blob/9eb7529d1495988839989611754b72a3bbddced5/pico8.js#L97-L100

But this is the order which doesn't work, right?

Issue #138, opened 2012-01-07
Select tile under mouse in tileset dock


Ablu:
Sometimes I want to find out from which tileset a special tile is. It would be nice to have some shortcut + click which then changes to the right tileset, selects and centers on the tile.

bjorn:
I decided to close this issue, but the current behavior is just to select the tile and with change ed4cc0d35ad9348aa8302f9737719ef871beae14 also to switch tilesets. It doesn't make sure the tile is also visible, however.

Also, this only covers tile layers. There is still no way to select the tile used by a tile object. This could be solved by adding a context menu action.

bjorn:
> It doesn't make sure the tile is also visible, however.

Correction, it appears it does!

PR #2792, opened 2020-05-01
Switch current tileset tab if all selected tiles are from the same tileset

mitchcurtis:
This makes any workflow involving "picking" tiles from the map much easier.

### DIFF
--- a/src/tiled/tilesetdock.cpp
+++ b/src/tiled/tilesetdock.cpp
@@ -401,8 +401,8 @@ void TilesetDock::selectTiles(const QList<Tile *> &tiles)
         else
             currentChanged(currentIndex);
 
-        // If all of the selected tiles are in the same tileset,
-        // switch the current tab to that tileset so the user doesn't need to.
+        // If all of the selected tiles are in the same tileset, switch the
+        // current tab to that tileset.
         if (selections.size() == 1) {
             auto tileset = tiles.first()->tileset()->sharedPointer();
             const int tilesetTabIndex = mTilesets.indexOf(tileset);
--- a/src/tiled/tilesetdock.cpp
+++ b/src/tiled/tilesetdock.cpp
@@ -403,13 +403,7 @@ void TilesetDock::selectTiles(const QList<Tile *> &tiles)
 
         // If all of the selected tiles are in the same tileset,
         // switch the current tab to that tileset so the user doesn't need to.
-        bool allTilesInSameTileset = true;
-        Tileset *firstTileset = tiles.first()->tileset();
-        for (int i = 1; i < tiles.size() && allTilesInSameTileset; ++i) {
-            if (tiles.at(i)->tileset() != firstTileset)
-                allTilesInSameTileset = false;
-        }
-        if (allTilesInSameTileset) {
+        if (selections.size() == 1) {
             auto tileset = tiles.first()->tileset()->sharedPointer();
             const int tilesetTabIndex = mTilesets.indexOf(tileset);
             mTabBar->setCurrentIndex(tilesetTabIndex);
--- a/src/tiled/tilesetdock.cpp
+++ b/src/tiled/tilesetdock.cpp
@@ -401,6 +401,20 @@ void TilesetDock::selectTiles(const QList<Tile *> &tiles)
         else
             currentChanged(currentIndex);
 
+        // If all of the selected tiles are in the same tileset,
+        // switch the current tab to that tileset so the user doesn't need to.
+        bool allTilesInSameTileset = true;
+        Tileset *firstTileset = tiles.first()->tileset();
+        for (int i = 1; i < tiles.size() && allTilesInSameTileset; ++i) {
+            if (tiles.at(i)->tileset() != firstTileset)
+                allTilesInSameTileset = false;
+        }
+        if (allTilesInSameTileset) {
+            auto tileset = tiles.first()->tileset()->sharedPointer();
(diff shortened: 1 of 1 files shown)


src/tiled/tilesetdock.cpp
```diff
@@ -401,6 +401,20 @@ void TilesetDock::selectTiles(const QList<Tile *> &tiles)
...
 
+        // If all of the selected tiles are in the same tileset,
+        // switch the current tab to that tileset so the user doesn't need to.
+        bool allTilesInSameTileset = true;
+        Tileset *firstTileset = tiles.first()->tileset();
+        for (int i = 1; i < tiles.size() && allTilesInSameTileset; ++i) {
+            if (tiles.at(i)->tileset() != firstTileset)
+                allTilesInSameTileset = false;
+        }
+        if (allTilesInSameTileset) {
```
bjorn, line 412: I believe this could be simplified to `selections.size() == 1`.

mitchcurtis:
> Right, this is something I had been considering when I implemented selecting of the captured tiles in the Tilesets view. I didn't do it, because I thought it may be often distracting / unwanted.

Can you elaborate on how would it be distracting or unwanted? My use cases are fairly straightforward, so I'm not aware of how this might affect others.

> The alternative would be to introduce some shortcut or tool for doing this, as well as probably making sure the clicked tile is also visible (as requested in #138).

I don't think a separate tool is necessary or beneficial. A shortcut would be hard to discover and probably remember, too.

> Maybe this should be something you can toggle with a small button in the Tilesets view?

What if it was in the settings/preferences dialog instead? To me it seems like a setting you would change once, so it would just be cluttering up the dock.

bjorn:
> Can you elaborate on how would it be distracting or unwanted? My use cases are fairly straightforward, so I'm not aware of how this might affect others.

I just mean that having this view change while grabbing tiles from the map could be distracting. I think most of the time you grab a tile just because you want to place it somewhere else, not because you want to select its tileset. You may even want to grab a tile to move it aside because you want to place another tile from your currently selected tileset in its place, and then the automatically changing tileset view will disrupt your plan.

It's just a thought though, and I find it hard to outweigh such potential downsides with the benefits, hence I was suggesting making it an option. However, an option would not help much if it's hidden in the Preferences and then you also couldn't easily temporarily disable it.

Anyway, we can simply put the behavior in and see if anybody complains. If that doesn't happen, all is fine. If it does, then we introduce an option. :-)

Could you try if it works fine with my suggested simplification?

mitchcurtis:
> 
> 
> > Can you elaborate on how would it be distracting or unwanted? My use cases are fairly straightforward, so I'm not aware of how this might affect others.
> 
> I just mean that having this view change while grabbing tiles from the map could be distracting. I think most of the time you grab a tile just because you want to place it somewhere else, not because you want to select its tileset. You may even want to grab a tile to move it aside because you want to place another tile from your currently selected tileset in its place, and then the automatically changing tileset view will disrupt your plan.

Hmmm, I see what you mean. I hadn't thought of that...

> It's just a thought though, and I find it hard to outweigh such potential downsides with the benefits, hence I was suggesting making it an option. However, an option would not help much if it's hidden in the Preferences and then you also couldn't easily temporarily disable it.

I still don't think this is something you would want to enable/disable often somehow, but I see your point about it being hard to discover. However, if you think it's useful more often than not, then having it opt-out could be sufficient?

> Anyway, we can simply put the behavior in and see if anybody complains. If that doesn't happen, all is fine. If it does, then we introduce an option. :-)
> 
> Could you try if it works fine with my suggested simplification?

You mean the inline comment? Sure, just wanted to make sure I understood everything first.

mitchcurtis:
Seems to work exactly the same. Not sure how to test when selections.size() > 1; I tried selecting two tiles (from different tilesets) with the selection tool, and when I switch to the stamp tool and right click on one of them, it would take me to that tileset.

bjorn:
> Not sure if I'm testing the case of selections.size() > 1 correctly, but this is how I did it: I tried selecting two tiles (from different tilesets) with the selection tool, and when I switch to the stamp tool and right click on one of them, it would take me to that tileset.

Well, you'd need to grab a stamp of at least two tiles, which are from different tilesets. You can grab multiple tiles by right-click-dragging. :-)

mitchcurtis:
> > Not sure if I'm testing the case of selections.size() > 1 correctly, but this is how I did it: I tried selecting two tiles (from different tilesets) with the selection tool, and when I switch to the stamp tool and right click on one of them, it would take me to that tileset.
> 
> Well, you'd need to grab a stamp of at least two tiles, which are from different tilesets. You can grab multiple tiles by right-click-dragging. :-)

Ha, didn't know you could do that. :) I just tested it and it works as expected:

- Selecting one or more tiles from the same tileset switches to that tileset.
- If you select two or more tiles from different tilesets, the current tab doesn't change (but the tiles are still highlighted in the tileset preview).

bjorn:
Thanks for making this change, @mitchcurtis! We'll see if we need to introduce an option for this.

bjorn:
I just fixed a small bug related to this new behavior (091dba8605a24a76f3ac0f2a76e278a729403ecf). Important part was to check whether the current tileset changed because we're synchronizing the selection, because in that case we should not also change the selection based on the current tileset, because that replaces the just captured stamp.

And actually, now that I think about it I'm not sure if that is ever a good idea even when the user switches tilesets. In some cases it may be useful, but in other cases it may unexpectedly replace your current stamp. But this behavior has been there for a long time so I'll leave it in for now.

PR #2795, opened 2020-05-12
Prevent segfault in MapView::fitMapInView() if no scene is assigned

Tectu:
This is a small fix that prevents a segmentation fault in MapView::fitMapInView() if no scene is assigned to the view.

### DIFF
--- a/src/tiled/mapview.cpp
+++ b/src/tiled/mapview.cpp
@@ -122,7 +122,11 @@ void MapView::setScale(qreal scale)
 
 void MapView::fitMapInView()
 {
-    const QRectF rect = mapScene()->mapBoundingRect();
+    MapScene* scene = mapScene();
+    if (!scene)
+        return;
+
+    const QRectF rect = scene->mapBoundingRect();
     if (rect.isEmpty())
         return;
 


bjorn:
Thank you for the fix! But I'm curious, where did you have a view without scene assigned? There are only three places in Tiled where a MapView is created, and in each of those cases a scene is assigned directly afterwards. Am I missing something?

Tectu:
I'm right now building a small game/application (it's an "experimental" in-between thing) based on `libtiled`. As a first step I need to build a working proof-of-concept as quickly as possible. In order not to waste too much time on the UI I decided to pull in your `MapView`, `MapScene` and `MapDocument` classes.
As such I am not using/building your application (eg. `MainWindow`) itself but only `libtiled` and the corresponding classes to integrate a `QWidget` into an existing Qt application. Therefore, my way of initializing things might be different than what your application does.  The way that the application I'm dealing with is built it must be able to handle cases where no scene and/or no map has been assigned. This made me come across a few possible segmentation faults due to unchecked pointers in `MapView` and `MapScene`. This PR fixes one of those.

Prior to this patch, the following sequence lead to a segfault as `MapView::fitMapInView()` was being called by your code internally without a scene being assigned yet:
```
widget::widget()
{
    // Map scene
    m_map_scene = new Tiled::MapScene(this);

    // Map view
    m_map_view = new Tiled::MapView;
    m_map_view->setScene(m_map_scene);

    // Layout
    QVBoxLayout* layout = new QVBoxLayout;
    layout->addWidget(m_map_view);
    setLayout(layout);
}
```

As such I'd like to state at this point that the code quality of this project is very good. If you'd also separate the base classes for basic user interactions from the actual application code you'd have saved somebody like me a few hours of work tho :p
Of course the separated `libtiled` already makes things easier.

bjorn:
> I'm right now building a small game/application (it's an "experimental" in-between thing) based on `libtiled`. As a first step I need to build a working proof-of-concept as quickly as possible. In order not to waste too much time on the UI I decided to pull in your `MapView`, `MapScene` and `MapDocument` classes.

Ah, that explains a lot. :-)

> Prior to this patch, the following sequence lead to a segfault as `MapView::fitMapInView()` was being called by your code internally without a scene being assigned yet:

In your code you're assigning the scene right after construction, so in this particular example I do not see how `fitMapInView` could be called in between, since there can't be a paint event call happening in the constructor.

> As such I'd like to state at this point that the code quality of this project is very good. If you'd also separate the base classes for basic user interactions from the actual application code you'd have saved somebody like me a few hours of work tho :p

I'm not sure how such a separation would look like. I think in many cases it will be hard to decide whether some class is part of the actual application or not. But, eventually I'll move more stuff to `libtiled` or introduce another library, because I'd like to share functionality between Tiled and the "Tiled Quick". This would include `MapDocument`, but not `MapView` and `MapScene` though.

Talking about Tiled Quick, if you're building a game with Qt did you consider using Qt Quick? The `tiledquickplugin` already does quite fast map rendering based on the Qt Scene Graph.

Tectu:
Well, in the meantime it was decided that the final product requires real 3D capabilities and that a simple isometric tile based solution is not going to make the cut. Therefore, I unfortunately had to move away from Tiled :/

I hope that this PR makes it through anyway and that there's a project in the future where I can come back to this as the code base seems comparably to other similar open source projects rather good.

Keep up the good work!

bjorn:
@Tectu Ah, that's unfortunate. I've merged your patch since I see no harm in it. The check may be superfluous for Tiled itself, but this way it's more future-proof.

Thanks for the comment about the source code. As the maintainer I mostly see the warts and the problems and often experience design struggles while developing new features. :-P

You're welcome to contribute anytime!

PR #2802, opened 2020-05-22
Replaced Corona SDK references with Solar2D

qwebltd:
Corona SDK was recently rebranded to Solar2D so I've renamed the Corona SDK section of the frameworks list, and moved it to retain alphabetical ordering.


bjorn:
That's right, thanks!

PR #2818, opened 2020-06-09
adds Tuesday to libraries list

Q-Bert-Reynolds:
I've been working on a [TMX file importer/exporter for Unity](https://www.youtube.com/watch?v=u42TC3hhnf4) for a few years now. I think it's pretty ok. Please include me in your list.


bjorn:
Please use the `master` branch instead of `snapshot` to propose this addition. I've changed the target branch to `master`, which seems to resolve that issue. Something also went wrong with the merging though, since you're adding back the deprecated Tiled2Unity and the no longer developed UniTMX. I'll fix that up.

In any case, from your video TUESDAY looks like a great option and it's interesting to see you even provide editing functionality in Unity!

Q-Bert-Reynolds:
Terribly sorry about that. I didn't think to check the branch I was pushing from. Thanks for the include.

PR #2821, opened 2020-06-09
Update Czech translation

encukou:


bjorn:
Thank you for the huge update!

PR #2838, opened 2020-06-20
Include QPainterPath where necessary

mitchcurtis:
Apparently an include was removed in a header somewhere in newer
versions of Qt.


bjorn:
Right, I had addressed this in 7551aebcbcb9b3545de7682c85e6d09dee4250c4. :-)

PR #2840, opened 2020-06-21
Pass tilePos to RenderTileCallback

mitchcurtis:
This allows custom tile-rendering code to benefit from the tile position
calculations that the renderers do. It is necessary in order to allow
each tile to be its own QQuickItem.

### DIFF
--- a/src/libtiled/hexagonalrenderer.cpp
+++ b/src/libtiled/hexagonalrenderer.cpp
@@ -270,80 +270,67 @@ void HexagonalRenderer::drawTileLayer(QPainter *painter,
                                       const TileLayer *layer,
                                       const QRectF &exposed) const
 {
-    CellRenderer renderer(painter, this, layer->effectiveTintColor(), CellRenderer::HexagonalCells);
-    auto tileRenderFunction = [&renderer](const Cell &cell, const QPoint &/*tilePos*/, const QPointF &screenPos, const QSizeF &size) {
-        renderer.render(cell, screenPos, size, CellRenderer::BottomLeft);
-    };
-    drawTileLayer(layer, tileRenderFunction, exposed);
-}
-
-void HexagonalRenderer::drawTileLayer(const TileLayer *layer,
-                                      const RenderTileCallback &renderTile,
-                                      const QRectF &exposed) const
-{
-    const RenderParams p(map());
-
-    QRect rect = exposed.toAlignedRect();
-
-    if (rect.isNull())
-        rect = boundingRect(layer->bounds());
+    QRect rect = boundingRect(layer->bounds());
+    if (!exposed.isNull())
+        rect &= exposed.toAlignedRect();
 
     QMargins drawMargins = layer->drawMargins();
-    drawMargins.setBottom(drawMargins.bottom() + p.tileHeight);
-    drawMargins.setRight(drawMargins.right() - p.tileWidth);
+    drawMargins.setBottom(drawMargins.bottom() + map()->tileHeight());
+    drawMargins.setRight(drawMargins.right() - map()->tileWidth());
 
     rect.adjust(-drawMargins.right(),
                 -drawMargins.bottom(),
                 drawMargins.left(),
                 drawMargins.top());
 
+    CellRenderer renderer(painter, this, layer->effectiveTintColor(), CellRenderer::HexagonalCells);
+
+    auto tileRenderFunction = [layer, &renderer, this](QPoint tilePos, const QPointF &screenPos) {
+        const Cell &cell = layer->cellAt(tilePos);
+        if (!cell.isEmpty()) {
+            const Tile *tile = cell.tile();
+            const QSize size = tile ? tile->size() : map()->tileSize();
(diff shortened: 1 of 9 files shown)


mitchcurtis:
ping

bjorn:
Sorry @mitchcurtis, there are a number of things broken in Tiled 1.4.0 that were rather more important to fix and get out a 1.4.1. I should have that out soon so I will have time to review this next week.

mitchcurtis:
OK, was just making sure you hadn't forgotten about me! :D

bjorn:
Right, so this is a good start. I just thought that when we pass in the tile position, we shouldn't need to pass in the cell anymore. In fact, we shouldn't need to pass in the tile layer anymore to the basic algorithm that iterates over the tiles. This does mean we need to move the code around a little, and some code may need to be duplicated for Tiled Quick...

mitchcurtis:
> Right, so this is a good start. I just thought that when we pass in the tile position, we shouldn't need to pass in the cell anymore.

src/libtiled/orthogonalrenderer.cpp and src/libtiled/isometricrenderer.cpp both skip the call to the render function is the cell is empty; wouldn't that be worth keeping? Then, since we've already fetched the cell, it would make sense to pass it in to the render function to avoid it having to fetch it again. In my code, I use the cell to get the tileset.

> In fact, we shouldn't need to pass in the tile layer anymore to the basic algorithm that iterates over the tiles. This does mean we need to move the code around a little, and some code may need to be duplicated for Tiled Quick...

You lost me here. :)

bjorn:
I'm working on changing this around a bit, but couldn't finish it today. Will continue tomorrow.

bjorn:
Alright, let me know what you think of setting it up like this.

I gave it some testing, but since this code is a little tricky I think we need to be careful merging this. Especially the handling of layer position, which probably no longer works the same way. It is tricky to test because I don't think we ever change layer positions in Tiled (likely the only way to test is by loading a file with a layer that has non-0 values for the `x` or `y` attributes).

mitchcurtis:
Hmm, I could be doing something wrong, but adapting to the latest changes breaks the tile rendering for me. I'm not sure how to explain what I see though...

mitchcurtis:
For now I will give up on this approach. The performance doesn't seem that great, and it was sapping my motivation to work on the game. :D

bjorn:
@mitchcurtis Hmm, did you notice performance issues using this callback approach?

mitchcurtis:
> 
> 
> @mitchcurtis Hmm, did you notice performance issues using this callback approach?

Yeah but I think it was a combination of whatever I was doing in my code + using thousands of QQuickItems where previously it was just one big texture.

bjorn:
Alright. I wouldn't expect the callback as such to have a major impact on the performance and would still prefer to go through with this PR, we just need to fix any remaining issues.

mitchcurtis:
I too would prefer to get it to work, as any way I work around my problem will be hacky, but I had issues with https://github.com/bjorn/tiled/pull/2840/commits/05b973f1f50aeb385369fce9917276ef3cc0e098. If I could come up with a minimal example, could you take a look at it?

bjorn:
@mitchcurtis Sure, I only did some minimal testing. If you can show me a problem then I will look into fixing it.

mitchcurtis:
OK, so clone https://github.com/mitchcurtis/quick-tmx-viewer/tree/tiledmapmodel and init the tiled submodule, making sure that it's on drawtilelayer-tilepos. You may also need to change `installBase` to point back to `Tiled Quick`. :D Then, revert https://github.com/bjorn/tiled/commit/05b973f1f50aeb385369fce9917276ef3cc0e098 and play around with the example - it should work as expected. Then, restore https://github.com/bjorn/tiled/commit/05b973f1f50aeb385369fce9917276ef3cc0e098 and try it again. The rendering is immediately different, and if you go to the top left of the map or zoom out, you can see it going all weird.

bjorn:
@mitchcurtis Alright, the problem is that you're assuming that `tilePos` is always a valid tile layer coordinate, but since the renderer is no longer passed the tile layer, it is calling the callback for all positions in the exposed rectangle, which includes positions that are outside the boundaries of your tile layer (even negative coordinates, which can have valid tiles as far as the `TileLayer` class is concerned).

For fixed-size maps, a possible fix would be to do this:

```c++
if (!tileLayer->localBounds().contains(tilePos))
    return;

const int tileIndex = tilePos.y() * tileLayer->width() + tilePos.x();
```

This may be bad for performance reasons if you zoom out a lot, so you may want to limit the visible area you pass to not be larger than the bounding rectangle of your tile layer. I do this in Tiled Quick in `TileLayerItem::updateVisibleTiles` using:
```c++
rect &= mRenderer->boundingRect(mLayer->localBounds());
```
But you still need the above check before you can calculate a valid `tileIndex`.

mitchcurtis:
Thanks a lot for looking into it! I'll give it a try. :)

mitchcurtis:
Since I can't reopen this, I've created a new pull request: https://github.com/bjorn/tiled/pull/2871

PR #2858, opened 2020-07-13
added --show-layer option

varnholt:


varnholt:
@bjorn There you go

src/tmxrasterizer/tmxrasterizer.cpp
```diff
@@ -132,6 +132,11 @@ bool TmxRasterizer::shouldDrawLayer(const Layer *layer) const
     if (mLayersToHide.contains(layer->name(), Qt::CaseInsensitive))
         return false;
 
+    if (!mLayersToShow.empty()) {
+       if (!mLayersToShow.contains(layer->name(), Qt::CaseInsensitive))
```
bjorn, line 136: I wonder if it would make sense to handle group layers, in that naming a group layer should show all its child layers. I think it does, but that should then also apply to the hiding of layers and should be in a difference change.

bjorn:
Thanks a lot!

Issue #752, opened 2014-09-12
object selection with multiple object layers are melting


alwex:
while you have multiple object layer, fit you select one layer, you can drag the object from another layer, this is very anoying as you are forced to disable the overlapping layer to manipulate an object under another layer object.

I think it would be better if the selected object layer is active while editing.

bjorn:
With change 44c02dc2a599abcaad1d3fcde138b63352989931 there are now two alternative options regarding the selection behavior: to either consider only objects from selected layers, or to prefer selecting objects from selected layers. The latter is automatically enabled, when "highlight current layer" is on.

PR #2865, opened 2020-07-15
Added option to change object selection behavior

### LINKED ISSUE #354: Object selection/movement disregards object layer
I have two object layers in my maps, one for my "entities" (player, baddies, etc) and one for metadata regions. Objects on these two layers can overlap but this causes issues as Tiled doesn't properly filter objects by layer for selection and movement. So if I have my Entity layer selected I can still select and move objects on the Metadata layer. This makes it really hard for me to edit my maps since I frequently have metadata regions that overlap entity objects.

bjorn:
Previously when clicking objects it would always take the top-most object from any visible and unlocked layer.

This new option allows changing that such that when available, an object from one of the currently selected layers is preferred. It can also be set to only do this, when the highlighting of current layer option is enabled (the new default).

This would close #354.

### DIFF
--- a/src/tiled/abstractobjecttool.cpp
+++ b/src/tiled/abstractobjecttool.cpp
@@ -79,6 +79,9 @@ static bool isChangedTemplateInstance(MapObject *mapObject)
     return false;
 }
 
+Preference<AbstractObjectTool::SelectionBehavior> AbstractObjectTool::ourSelectionBehavior {
+    "AbstractObjectTool/SelectionBehavior", AbstractObjectTool::PreferHighlightedLayers
+};
 
 AbstractObjectTool::AbstractObjectTool(Id id,
                                        const QString &name,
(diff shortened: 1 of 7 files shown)


bjorn:
I can't continue this patch for now, but a remaining change is to apply this new option also when group-selecting objects.

bjorn:
Another option that may be good to add, is to only pick objects from selected layers. Then this change will resolve issue #752 as well.

bjorn:
I'm a bit hesitant about all the modes this new option could have:

* Select from All Layers (current behavior)
* Select from Selected Layers
* Select from Highlighted Layers
* Prefer Selecting from Selected Layers
* Prefer Selecting from Highlighted Layers (suggested default)

Maybe @eishiya can help with considering which ones could be left out or how to represent it differently. For example, I could also imagine:

* Select from All Layers (current behavior)
* Select from Selected Layers
* Select from Highlighted Layers
[ ] Select from All Layers As Fallback (checkbox to avoid too many options, maybe with better name)

And it could also be an option to:

* Prefer Selecting from Selected Layers, but in "Highlight Current Layer" mode, to select From Selected Layers Only

But how to represent that as an option? :S

bjorn:
Maybe have two options:

Selection Behavior:

* Select from All Layers _(current behavior)_
* Prefer Selected Layers _(suggested default)_
* Select from Selected Layers

Selection Behavior when Highlight Current Layer is enabled: _(better name?)_

* Prefer Selected Layers
* Select from Selected Layers _(suggested default)_

PR #2867, opened 2020-07-18
Add libtiledquick

mitchcurtis:
This moves the C++ types out of tiledquickplugin into a standalone
library so that other projects can link to them.


src/libtiledquick/libtiledquick.qbs
```diff
@@ -0,0 +1,72 @@
...
+    Depends { name: "Qt"; submodules: ["gui", "quick"]; versionAtLeast: "5.6" }
+
+    cpp.cxxLanguageVersion: "c++14"
+    cpp.visibility: "minimal"
+    cpp.defines: [
+        "TILED_QUICK_LIBRARY",
+        "QT_NO_CAST_FROM_ASCII",
+        "QT_NO_CAST_TO_ASCII",
+        "QT_NO_URL_CAST_FROM_STRING",
+        "_USE_MATH_DEFINES",
```
bjorn, line 17: Should also include `QT_DEPRECATED_WARNINGS`, `QT_DISABLE_DEPRECATED_BEFORE=0x050900` and `QT_NO_FOREACH`, or did you have some reason to leave those out?

mitchcurtis: I copied directly from libtiled. I can add them if you want?

bjorn: Right, but the code is copied from `tiledquickplugin`, so please copy the defines from there. :-)

You should remove `_USE_MATH_DEFINES` as well, since there's probably no need for it here.

Issue #1433, opened 2017-01-17
Support animated GIF export


bjorn:
For maps with animated tiles, it would be nice to support exporting to animated GIF format.

Unfortunately, Qt only supports reading animated GIF files, not writing them. But there are other Qt applications that have implemented writing to GIF files using [GIFLIB](http://giflib.sourceforge.net/), like [Drawpile](https://github.com/callaa/Drawpile):

https://github.com/callaa/Drawpile/blob/master/src/client/export/gifexporter.cpp

Feature requested by [@oryxdesignlab on Twitter](https://twitter.com/oryxdesignlab/status/821275958288084993).

sballew:
This is an old feature request, but instead of having Tiled export a GIF, I wonder if it would be easier to update tmxrasterizer to support a parameter for a frame number to rasterize.

This would allow a user to script multiple tmxrasterizer commands to write out a series of frames, and then the user is free to use any GIF creator tool to combine the images together. For example, with `imagemagick`.

Would love to hear @bjorn or anybody's thoughts on this approach. I started going through the source to see how it'd be implemented, but I have little CPP experience.

bjorn:
@sballew I think it's a great idea! For this purpose `TilesetManager::advanceTileAnimations` could be made public, such that it can be called by the `tmxrasterizer` after loading the map. The function takes milleseconds, so the easiest would to have a parameter like `--advance-animations X`, where X is an amount of milliseconds (or maybe you can think of something shorter).

Providing a way to advance a number of frames may be more convenient, but I'm not sure how that would work given that each animation frame has its own duration.

When supporting this also when rendering an entire world, after rendering each map you'd need to call `TilesetManager::resetTileAnimations` and then advance them again after loading the next map.

Would you like to have another go at implementing this? Otherwise I could probably patch this in on Tuesday.

sballew:
Thanks for the guidance @bjorn . Based on your pointers, I put together a PR. It is working for me locally, but I have not yet tested with an animated world.
The PR is here: https://github.com/bjorn/tiled/pull/2868

lightlike:
I would really love to export some kind of Animation file. I tried using a C#-Library to do this myself once but ultimately failed.

I still have some of the Frame-Time and Loop-Duration code (in C# of course). I can post that here if that would help.

lightlike:
I made a small Python tool for those who want to or have no idea how to use the `tmxrasterizer`.
It's pretty basic for the moment but should work for most cases.

[...]

PR #2868, opened 2020-07-18
Add animation advancing to tmxrasterizer

sballew:
Relates to #1433 .

**Objective**
Allow a user to advance animations by a duration prior to rasterizing an image via tmxrasterizer.

**Reasoning**
This is an alternative to GIF support. A user can script multiple tmxrasterizer commands, each advancing the animation. This results in multiple PNG files, which the user can stitch together with any GIF creation tool such as `imagemagick`.

**Notes**
This should work with worlds, as it includes the tile animation reset. However I have not yet tested it on an animated world.

### DIFF
--- a/src/tmxrasterizer/main.cpp
+++ b/src/tmxrasterizer/main.cpp
@@ -130,12 +130,12 @@ int main(int argc, char *argv[])
     }
 
     if (parser.isSet(QLatin1String("advance-animations"))) {
-      bool ok;
-      w.setAdvanceAnimations(parser.value(QLatin1String("advance-animations")).toDouble(&ok));
-      if (!ok || w.advanceAnimations() <= 0.0) {
-        qWarning().noquote() << QCoreApplication::translate("main", "Invalid advance-animations specified: \"%1\"").arg(parser.value(QLatin1String("advance-animations")));
+        bool ok;
+        w.setAdvanceAnimations(parser.value(QLatin1String("advance-animations")).toInt(&ok));
+        if (!ok || w.advanceAnimations() < 0) {
+            qWarning().noquote() << QCoreApplication::translate("main", "Invalid advance-animations specified: \"%1\"").arg(parser.value(QLatin1String("advance-animations")));
             exit(1);
-      }
+        }
     }
 
     return w.render(fileToOpen, fileToSave);
--- a/src/tmxrasterizer/main.cpp
+++ b/src/tmxrasterizer/main.cpp
@@ -77,6 +77,9 @@ int main(int argc, char *argv[])
                           { "show-layer",
                             QCoreApplication::translate("main", "If used only specified layers are shown. Can be repeated to show multiple specified layers only."),
                             QCoreApplication::translate("main", "name") },
+                          { "advance-animations",
+                            QCoreApplication::translate("main", "If used tile animations are advanced by the specified duration."),
+                            QCoreApplication::translate("main", "duration") }
                       });
     parser.addPositionalArgument("map|world", QCoreApplication::translate("main", "Map or world file to render."));
     parser.addPositionalArgument("image", QCoreApplication::translate("main", "Image file to output."));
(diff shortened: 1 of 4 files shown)


src/tmxrasterizer/main.cpp
```diff
@@ -126,5 +129,14 @@ int main(int argc, char *argv[])
         }
     }
 
+    if (parser.isSet(QLatin1String("advance-animations"))) {
+      bool ok;
+      w.setAdvanceAnimations(parser.value(QLatin1String("advance-animations")).toDouble(&ok));
+      if (!ok || w.advanceAnimations() <= 0.0) {
```
bjorn, line 135:
```suggestion
      if (!ok || w.advanceAnimations() < 0) {
```
Suggesting this change because I don't think it's useful to consider passing 0 an error, and the comparison should not be to a double.

src/tmxrasterizer/main.cpp
```diff
@@ -126,5 +129,14 @@ int main(int argc, char *argv[])
         }
     }
 
+    if (parser.isSet(QLatin1String("advance-animations"))) {
+      bool ok;
```
bjorn, line 133: Please fix the indentation (4 spaces).

src/tmxrasterizer/tmxrasterizer.cpp
```diff
@@ -181,6 +183,10 @@ int TmxRasterizer::renderMap(const QString &mapFileName,
         xScale = yScale = mScale;
     }
 
+    if (mAdvanceAnimations > 0) {
+        TilesetManager::instance()->advanceTileAnimations(mAdvanceAnimations);
+    }
```
bjorn, line 188: Please leave out the `{` and `}` for single-line bodies (also below).

src/tmxrasterizer/tmxrasterizer.h
```diff
@@ -49,26 +49,31 @@ class TmxRasterizer
     qreal scale() const { return mScale; }
     int tileSize() const { return mTileSize; }
     int size() const { return mSize; }
+    int advanceAnimations() const {return mAdvanceAnimations; }
```
bjorn, line 52:
```suggestion
    int advanceAnimations() const { return mAdvanceAnimations; }
```

src/tmxrasterizer/tmxrasterizer.h
```diff
@@ -49,26 +49,31 @@ class TmxRasterizer
...
     bool useAntiAliasing() const { return mUseAntiAliasing; }
     bool smoothImages() const { return mSmoothImages; }
     bool ignoreVisibility() const { return mIgnoreVisibility; }
 
     void setScale(qreal scale) { mScale = scale; }
     void setTileSize(int tileSize) { mTileSize = tileSize; }
     void setSize(int size) { mSize = size; }
+    void setAdvanceAnimations(int duration) { mAdvanceAnimations = duration; }
+
```
bjorn, line 61: Please avoid the additional newline.

src/tmxrasterizer/tmxrasterizer.h
```diff
@@ -49,26 +49,31 @@ class TmxRasterizer
...
     void setAntiAliasing(bool useAntiAliasing) { mUseAntiAliasing = useAntiAliasing; }
     void setSmoothImages(bool smoothImages) { mSmoothImages = smoothImages; }
     void setIgnoreVisibility(bool IgnoreVisibility) { mIgnoreVisibility = IgnoreVisibility; }
 
     void setLayersToHide(QStringList layersToHide) { mLayersToHide = layersToHide; }
     void setLayersToShow(QStringList layersToShow) { mLayersToShow = layersToShow; }
+
```
bjorn, line 68: Please don't add a newline here.

sballew:
@bjorn Thanks for the comments and suggestions. I have incorporated all changes. I don't really have a good C++ dev environment set up, so I let a number of those code styles get through, but thank you for catching them.

bjorn:
@sballew Thanks again for this nice idea and for submitting the patch!

PR #2870, opened 2020-07-20
added cute tiled to supported frameworks doc

RandyGaul:
https://github.com/RandyGaul/cute_headers
https://github.com/RandyGaul/cute_headers/blob/master/cute_tiled.h


bjorn:
Thanks for adding it to the list and for your work on this library!

PR #2871, opened 2020-07-22
Pass tilePos to RenderTileCallback

mitchcurtis:
Continues on from https://github.com/bjorn/tiled/pull/2840.

This allows custom tile-rendering code to benefit from the tile position
calculations that the renderers do. It is necessary in order to allow
each tile to be its own QQuickItem.

Discussion about the motivation for this patch can be found here:

https://discourse.mapeditor.org/t/z-ordering-issues-tiledquickplugin/4546

### DIFF
--- a/src/libtiled/hexagonalrenderer.cpp
+++ b/src/libtiled/hexagonalrenderer.cpp
@@ -266,37 +266,6 @@ void HexagonalRenderer::drawGrid(QPainter *painter, const QRectF &exposed,
     }
 }
 
-void HexagonalRenderer::drawTileLayer(QPainter *painter,
-                                      const TileLayer *layer,
-                                      const QRectF &exposed) const
-{
-    QRect rect = boundingRect(layer->bounds());
-    if (!exposed.isNull())
-        rect &= exposed.toAlignedRect();
-
-    QMargins drawMargins = layer->drawMargins();
-    drawMargins.setBottom(drawMargins.bottom() + map()->tileHeight());
-    drawMargins.setRight(drawMargins.right() - map()->tileWidth());
-
-    rect.adjust(-drawMargins.right(),
-                -drawMargins.bottom(),
-                drawMargins.left(),
-                drawMargins.top());
-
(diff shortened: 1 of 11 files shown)


mitchcurtis:
[Here](https://github.com/bjorn/tiled/pull/2840#issuecomment-650818278) you said:
 
> I gave it some testing, but since this code is a little tricky I think we need to be careful merging this. Especially the handling of layer position, which probably no longer works the same way. It is tricky to test because I don't think we ever change layer positions in Tiled (likely the only way to test is by loading a file with a layer that has non-0 values for the `x` or `y` attributes).

When would you change a layer's position?

mitchcurtis:
Quoting from the other pull request again:

> @mitchcurtis Alright, the problem is that you're assuming that `tilePos` is always a valid tile layer coordinate, but since the renderer is no longer passed the tile layer, it is calling the callback for all positions in the exposed rectangle, which includes positions that are outside the boundaries of your tile layer (even negative coordinates, which can have valid tiles as far as the `TileLayer` class is concerned).
> 
> For fixed-size maps, a possible fix would be to do this:
> 
> ```c++
> if (!tileLayer->localBounds().contains(tilePos))
>     return;
> 
> const int tileIndex = tilePos.y() * tileLayer->width() + tilePos.x();
> ```
> 
> This may be bad for performance reasons if you zoom out a lot, so you may want to limit the visible area you pass to not be larger than the bounding rectangle of your tile layer. I do this in Tiled Quick in `TileLayerItem::updateVisibleTiles` using:
> 
> ```c++
> rect &= mRenderer->boundingRect(mLayer->localBounds());
> ```
> 
> But you still need the above check before you can calculate a valid `tileIndex`.

Thanks, that works great! The early return made it a lot faster, and I added the visible area intersection as well, although I didn't notice any visual difference in performance.

Hopefully I can now get my code to a point where it's closer to the performance of TileLayerItem. :)

mitchcurtis:
Ping.

bjorn:
@mitchcurtis This became a nice cleanup of repeated rendering code. :-)

PR #2888, opened 2020-08-19
Wang Improvements

bjorn:
This change is about improving the Wang-related tools to make them more useful. The main goals are:

* To support editing with the blob tileset.
* To support erasing (transitions into nothing).
* To support adjusting the map to make a change work, instead of aborting as soon as a tile can't be found.
* To replace the Terrain Brush (if we also increase the limit on the colors from 15 to 255).

For now, the main change is to unify the colors, so that we don't have "edge" and "corner" colors, but a single color can be used on both edges and corners. This is the first step towards supporting blob tilesets.

See also: [Neither terrain nor Wang can handle a “blob” tileset](https://discourse.mapeditor.org/t/neither-terrain-nor-wang-can-handle-a-blob-tileset/4671)


bjorn:
A short demonstration of how editing with a blob tileset could work:

blob

I imagine the tool will need several editing modes for convenience:

* To change only tile corners.
* To change only tile edges.
* To change either corners or edges based on proximity (shown above).

And the appropriate mode could be selected based on whether the selected color exists on corners, edges or both.

bjorn:
Before this can be merged there are at least the following features still missing in Wang Brush that were present in Terrain Brush:

* Alt modifier for "symmetry painting".
* Shift modifier for drawing lines.

So I'll prioritize those features over other plans that could also be done in a separate pull request.

bjorn:
Line drawing has been added, also in Edge mode:

wang-brush-line-drawing

bjorn:
Fixed the erasing mode again, was a bug in the last commit.

PR #2895, opened 2020-09-19
using CTRL-SHIFT to toggle snapping to fine grid mode

sverx:
Well I don't really know if CTRL-SHIFT is a good idea but since CTRL was to toggle snapping to grid, I thought to keep the idea the CTRL 'toggles' snapping, and with SHIFT you switch to fine grid.
Hopefully it makes sense.

### DIFF
--- a/src/tiled/snaphelper.cpp
+++ b/src/tiled/snaphelper.cpp
@@ -30,8 +30,11 @@ SnapHelper::SnapHelper(const MapRenderer *renderer,
     : mRenderer(renderer)
 {
     Preferences *preferences = Preferences::instance();
-    mSnapToGrid = preferences->snapToGrid();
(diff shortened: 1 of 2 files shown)


bjorn:
Thanks for opening a pull request!

I felt like `Ctrl+Shift` should do something useful as well when you're already in "Snap to Fine Grid" mode, rather than turning off snapping entirely which is already covered by just `Ctrl`. I've updated the patch so that it toggles between "Snap to Fine Grid" and "Snap to Grid" in this case.

Let me know if you see any problems with this change, if not I think this is ready for merging.

sverx:
I don't see any problem, I actually think it's a good idea and I wish it occurred to me in the first place :rofl:

PR #2896, opened 2020-09-22
Add Strong Eraser tool

aganm:
I added a strong eraser tool beside the eraser tool. The strong eraser works exactly like the regular eraser but on all layers at once without having to select them.

image

I present a use case why this is pretty handy:

I have 15 different layers.
There's 1 tile that I want to erase, this tile position is only occupied by 1 layer.
This tile is on another layer than the currently selected one.
How do I do this?

Method 1 >  Search around in my 15 layers to find out where I put this tile, I can't remember... let me hide this one to see if it's there, not there, let me hide this one... When I find where it is, select the layer, select the eraser, and erase the tile in the layer. Unhide all the layers I was hiding to find out where this tile is. Finally, I click the layer I was using previously to resume my previous task.

Method 2 > I can't remember which layer, so I will erase from all layers at once. But to do this, I have to select all my layers: scroll all the way up the layers window, click the first layer, scroll all the way down, shift-click the last layer. Choose eraser tool, erase the tile. Finally, I click the layer I was using previously to resume my previous task.

Strong Eraser Method > Select the Strong Eraser (button or Shift+E). Erase. Done. I can continue with my task.


bjorn:
That's really useful! I wonder though, is there a particular reason you've implemented this as a separate tool, rather than adding handling of the Shift modifier in the existing Eraser tool for the "strong" mode?

aganm:
That would be lack of familiarity with Qt and also because I saw a Qt::NoModifier in eraser.cpp (line 62). I thought that Qt::NoModifier was there to void collisions with other commands.

If it's not the case, handling it with Shift would be a better idea indeed.

src/tiled/eraser.cpp
```diff
@@ -105,7 +107,10 @@ void Eraser::doErase(bool continuation)
     }
     mLastTilePos = tilePos;
 
-    for (Layer *layer : mapDocument()->selectedLayers()) {
+    auto layers = !mStrong ? mapDocument()->selectedLayers()
+                           : mapDocument()->map()->layers();
```
bjorn, line 111: `map()->layers()` only returns the top-level layers for the map, whereas I think you'll want to iterate `map()->tileLayers()`, which also handles group layers (unfortunately that means we need to change the loop a bit since this is not compatibile with `selectedLayers()`. Maybe the loop body could be turned into a lambda.

bjorn (commented):
> I saw a Qt::NoModifier in eraser.cpp (line 62). I thought that Qt::NoModifier was there to avoid collisions with other commands.

Yeah, I saw this too but I have to admit I don't remember why I put that check there. Maybe we'll find out. :-)

aganm:
I implemented the Shift modifier. #2897

PR #2897, opened 2020-09-23
Holding shift with the eraser erases from all layers

aganm:
To implement shift erase, I had to remove the check for Qt::NoModifier so I don't know if that's gonna collide with another command.

### DIFF
--- a/docs/manual/editing-tile-layers.rst
+++ b/docs/manual/editing-tile-layers.rst
@@ -147,6 +147,8 @@ Shortcut: ``E``
 A simple eraser tool. Left click erases single tiles and right click can
 be used to quickly erase rectangular areas.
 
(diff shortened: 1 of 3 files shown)


bjorn:
@aganm Thanks for this improvement! I made a few tweaks. The behavior changed so that other modifiers are now ignored and that you can toggle the strong mode also while erasing. Let me know if it all looks fine, if so we can merge this.

aganm:
Looks good to me!

bjorn:
I found the reason for the `NoModifier` check: 153405db2a3d279ec744cfd6ffb3f5dda341e315

It means, we should at least check that `Ctrl` is not being held, otherwise people trying to use this action will erase instead. :-(

aganm:
That should fix it #2904

PR #2904, opened 2020-10-01
Fix eraser Right Click collision with Ctrl + Right Click select

aganm:


src/tiled/eraser.cpp
```diff
@@ -53,12 +53,20 @@ void Eraser::tilePositionChanged(QPoint tilePos)
 
 void Eraser::mousePressed(QGraphicsSceneMouseEvent *event)
 {
+    /**
+     * Tiled uses Ctrl + Right Click to select tiles at clicked location.
```
bjorn, line 57: Correction: it's used to select layers with tiles at the clicked location, not the tiles.

But, I don't think a comment here is very helpful. As seen in 153405d, there are numerous places that now need to check the modifiers for no apparent reason (essentially all tile tools) and we're not going to put a comment in all these places. The plan is rather to change the way events are handled so that we can remove all these checks again (so, the Ctrl+Right-click action should really be handled before all these tools for example).

aganm: I removed the comment if you want to merge it until then

bjorn:
Thanks!

PR #2905, opened 2020-10-02
Documentation update: libqt5svg5-dev is required on Debian.

samhocevar:


bjorn:
Are you sure this is required for compiling Tiled? What error do you get without this package?

bjorn:
Ah, this is probably required on Debian for the same reason that it was needed on Ubuntu 18.04, since it shipped with a version of Qbs that would fail when it could not find files referenced by the `distribute.qbs` product, even though that product was not being built. With a more recent version of Qbs, this dependency is not needed.

Would it work to install `libqt5svg5` instead?

samhocevar:
You’re right, `libqt5svg5` is enough. It’s `libqsvg.so` that’s needed and I misinterpreted the `.so` extension as being from a `-dev` package whereas it’s just a plugin from the standard lib package.

bjorn:
Alright, could you update the pull request? It seems like you have not given me permission to push a change to your branch.

samhocevar:
> Alright, could you update the pull request?

Done!

> It seems like you have not given me permission to push a change to your branch.

I didn’t know that was standard practice. I’ll remember that next time!

bjorn:
> I didn’t know that was standard practice. I’ll remember that next time!

It's up to you, but enabling this feature allows resolving such small edits without needing a communication roundtrip, which is quite nice. And it makes it easy for me to help out with anything. :-)

PR #2907, opened 2020-10-06
Porting Tiled to Qt 6

bjorn:
With the [Qt 6.0 Alpha](https://www.qt.io/blog/qt-6.0-alpha-released) out it's time to start porting Tiled to Qt 6.

Of course, ideally while keeping it working against Qt 5.6, but more realistically we'll update the minimum required version of Qt to a more recent LTS version if that turns out to be too much effort.


bjorn:
Alright, I accidentally pushed this to `master`...

PR #2912, opened 2020-10-10
Feature 2833 (rotated Wang tiles)

### LINKED ISSUE #2833: Support rotation of Wang tiles
E.g. in this very nice tile set https://forums.rptools.net/viewtopic.php?t=21677 a lot of different terrain types are defined with smooth transitions. Also they work nicely as a Wang tile - but ... you have to (image) rotate some tiles to cover missing edges (the tile set assumes that you rotate the tiles to get complete coverage).
But if you rotate _the images_ you create additional image files (waste space) and loose the connection to the original tile (as tiled supports rotated tiles this also wastes space in the map file (increases the number of symbols up to four times)). 

It would be perfect if either
- there would be a checkbox marking the Wang tile set as rotation compatible (and fix the brush to use rotated tiles at need), or
- the Wang brush would (optionally) fall back to rotated tiles when a matching tile is missing, or
- there would be a context menu command/key to insert a 90° rotated image into the Wang tile set (repeat up to three times to get full coverage for non rotationally equivalent variants) with appropriately rotated Wang colors

I even volunteer to implement this if one of the three proposals is considered feasible and helpful, but I can't promise about the completion date.

PS: I had the feeling that this issue should already have come up multiple times, but I was unable to find any reference. I see it being proposed at https://discourse.mapeditor.org/t/wang-tiles-and-filling-improvements/2339/6 but never seems to have gotten implemented or documented.

cpetig:
Implementation of rotated Wang tiles. Introduces new icons for rotate-on-demand and rotate-for-variation. Implements saving and loading these properties to tile sets. Implements rotated virtual Wang tile generation on loading.
Closes #2833


cpetig:
My current plan for supporting direct use after Wang tile edit would be to regenerate the virtual Wang tile entries whenever the tileset is saved to a file. @bjorn Would you require this before merging to master?

bjorn:
@cpetig It should work without needing to save the tileset, but I have not yet gotten around to looking at your approach in detail and to consider alternatives. I'll look into this tomorrow.

cpetig:
I know that derived (rotated) Wang variants are not recalculated whenever a Wang tile is edited. I thought that recalculating on save might be a reasonable compromise (only need to change a single function) instead of having to recalculate at every change

bjorn:
I've replaced the generated WangTiles approach by iterating the variations on-demand and applying a small penalty to variations in order to keep the "rotate-for-variation" optional. It's slower (even twice as slow at the moment), but I still prefer it since it makes other things easier. Maybe we can still optimize the setting up of the variations a bit later.

There's a few things still to be done before this can be merged:

* When the properties are changed, this should be done through undo commands.
* I'd prefer the options on the WangSet rather than on the Tileset, but on Discord I've had feedback that this could be useful as a per-tile flag as well (whether it can be rotated / flipped).
* Some people will need to be able to toggle these options separately:
    * Can tile be flipped vertically?
    * Can tile be flipped horizontally?
    * Can tile be rotated?
* Since this feels rather WangSet-specific, maybe the above per-tile options could be stored on the WangTile class. The downside of storing it there is that it will need some new UI approach to editing those options.

I'll see if I can help with these things next week, but in the meantime any work or additional feedback is really appreciated!

cpetig:
I still have trouble imagining a tile set where different rules apply to different Wang sets (i.e. I believe that if any image in a set supports vertical flipping (birds eye view typically) all images would allow this). And since the system prefers existing flipped/rotated-WangId tiles over flipping/rotating tiles I see no benefit of making the decision per Wang Set or even per tile.
On another note I assume that permission to flip both horizontally and vertically (180°) would imply permission for 90° rotation (anti-diagonal flipping), so I assume it wouldn't need a separate third property. Of course this only applies to quadratic tiles, as with hexagonal or rectangular tiles 90° rotation would be invalid.

cpetig:
Eishiya convinced me that there is at least one user for per tile settings, see https://discordapp.com/channels/524610627545595904/524610627545595906/777201497770819635 . Also it seems that indeed 180° doesn't imply 90° is valid.
I took a quick look and will make these flags part of WangTile (+inherit) and WangSet (+randomize).

cpetig:
@bjorn Is there a specific reason for defining trivial one statement get functions at the end of the header instead of embedding them inside the class?

cpetig:
I try to figure out whether the rotation permission is a property of a WangTile or a Tile? I assume a WangTile corresponds to an entity which is edited in the Wang editor for one tile, so this would be the correct place to store the information. @bjorn Could you please confirm this understanding as correct?

src/libtiled/mapreader.cpp
```diff
@@ -391,8 +391,15 @@ SharedTileset MapReaderPrivate::readTileset()
                                       tileSpacing, margin);
 
             tileset->setColumnCount(columns);
-            tileset->setCanRotate(atts.value(QLatin1String("canrotate")).toInt());
-            tileset->setAlternateRotation(atts.value(QLatin1String("alternaterotation")).toInt());
+            // recognize older file format for WangTile rotation
```
bjorn, line 394: This is rather pointless since only you and me are using the older file format, right? :-)

src/libtiled/mapreader.cpp
```diff
@@ -716,8 +723,26 @@ void MapReaderPrivate::readTilesetWangSets(Tileset &tileset)
             const QString name = atts.value(QLatin1String("name")).toString();
             const WangSet::Type type = wangSetTypeFromString(atts.value(QLatin1String("type")).toString());
             const int tile = atts.value(QLatin1String("tile")).toInt();
+            const bool randomizeOrientation = atts.value(QLatin1String("randomizeOrientation")).toInt()!=0;
```
bjorn, line 726: I've kept all attributes so far completely lowercase, please also use this style for new attributes.

src/libtiled/mapreader.cpp
```diff
@@ -716,8 +723,26 @@ void MapReaderPrivate::readTilesetWangSets(Tileset &tileset)
...
+            {
+                const bool flipHorizontally = atts.value(QLatin1String("flipHorizontally")).toInt()!=0;
+                const bool flipVertically = atts.value(QLatin1String("flipVertically")).toInt()!=0;
+                const bool flipAntiDiagonally = atts.value(QLatin1String("flipAntiDiagonally")).toInt()!=0;
+                wangSet->setAsNeededFlipHorizontally(flipHorizontally);
+                wangSet->setAsNeededFlipVertically(flipVertically);
+                wangSet->setAsNeededFlipAntiDiagonally(flipAntiDiagonally);
+            }
+            if (alternaterotation || randomizeOrientation)
+                wangSet->setRandomizeOrientation(true);
```
bjorn, line 745: I think it might be more intuitive to negate this and call it "prefer non-transformed tiles".

src/libtiled/mapreader.cpp
```diff
@@ -774,6 +800,16 @@ void MapReaderPrivate::readTilesetWangSets(Tileset &tileset)
                     wangTile.setFlippedVertically(fV);
                     wangTile.setFlippedAntiDiagonally(fA);
 
+                    if (!fInherit) {
+                        const bool canH = tileAtts.value(QLatin1String("can_hflip")).toInt();
+                        const bool canV = tileAtts.value(QLatin1String("can_vflip")).toInt();
+                        const bool canA = tileAtts.value(QLatin1String("can_dflip")).toInt();
```
bjorn, line 806: I would suggest we reuse the `hflip`, `vflip` and `dflip` attributes here, since they were never used before. I mean the user never had a way to set those to `true` anyway.

src/tiled/changewangsetdata.cpp
```diff
@@ -129,6 +129,43 @@ void ChangeWangSetColorCount::redo()
 }
 
 
+ChangeWangSetFlipping::ChangeWangSetFlipping(TilesetDocument *TilesetDocument,
+                                             WangSet *wangSet,
+                                             ChangeType _which,
+                                             bool newValue)
+    : mTilesetDocument(TilesetDocument), mWangSet(wangSet), mWhich(_which),
```
bjorn, line 136: Please remember to call the `QUndoCommand` constructor with the name of the undo action. The name can be generic though, like "Change Wang Set", since I don't think there is much value in being very specific here.

src/tiled/changewangsetdata.h
```diff
@@ -89,6 +89,26 @@ class ChangeWangSetColorCount : public QUndoCommand
     QVector<WangColorChange> mRemovedWangColors;
 };
 
+class ChangeWangSetFlipping : public QUndoCommand
+{
+public:
+    enum ChangeType { flipX, flipY, flipAD, randomFlip };
```
bjorn, line 95: Naming style: `FlipX`, `FlipY`, etc.

bjorn (changes requested):
> @bjorn Is there a specific reason for defining trivial one statement get functions at the end of the header instead of embedding them inside the class?
> Oh, I see it is done differently in WangTile and WangSet. So I assume there is no specific reason.

It just means I'm not feeling overly attached to having them in one place or another, since both have their advantages. The reason to put them outside of the class is simply to keep the class definition more readable.

> I try to figure out whether the rotation permission is a property of a WangTile or a Tile? I assume a WangTile corresponds to an entity which is edited in the Wang editor for one tile, so this would be the correct place to store the information. @bjorn Could you please confirm this understanding as correct?

Yes, though when we're doing this per-WangTile there would need to be some way in the tileset editor to set up which tiles can be flipped in which way. Not sure what's the best approach for now.

I left some inline comments, though I saw you just pushed another change, which is not taken into account yet.

src/libtiled/mapreader.cpp
cpetig: Oh, I know of at least a third person, and more people might have downloaded the tileset from my github repository ... I agree that it should be a temporary measure.

src/libtiled/mapreader.cpp
cpetig: Good point, after looking at these for some days I was less and less able to come up with a good names.

src/libtiled/mapreader.cpp
cpetig: Oh, true, I was overly cautious.

src/libtiled/mapreader.cpp
cpetig: :+1:

src/tiled/changewangsetdata.h
cpetig: :+1:

cpetig:
Dear @bjorn , I completed the code for per WangSet and per WangTile independent flip permissions. But to me it feels that (especially after implementing the undo/redo part per WangTile) overriding flip permissions for a WangTile should be a per Tile property.
Do you agree? I will likely need to recharge my motivation to rewrite half of this patch, though (will take some hours).

bjorn:
@cpetig Well, the way it's implemented now it indeed makes no sense for this stuff to be stored in the WangTile instead of the Tile. But in any case, I think you've done a great job adjusting this patch to the recent discussions. Let's take a short break. I will have time to take a detailed look into the patch again on Tuesday.

src/libtiled/mapreader.cpp
cpetig: On the other hand it feels strange to reuse a(n unused) short name for a rare property.

src/tiled/adjusttileindexes.cpp
```diff
@@ -264,26 +264,26 @@ AdjustTileMetaData::AdjustTileMetaData(TilesetDocument *tilesetDocument)
...
-        }
+//        for (const WangTile &wangTile : wangSet->wangTilesByWangId()) {
+//            if (Tile *fromTile = wangTile.tile()) {
+//                if (Tile *newTile = adjustTile(fromTile)) {
+//                    WangId fromWangId = wangSet->wangIdOfTile(newTile);
+//                    WangId toWangId = wangTile.wangId();
+//                    changes.append(ChangeTileWangId::WangIdChange(fromWangId, toWangId, newTile));
+//                }
+//            }
+//        }
```
cpetig, line 275: I don't recall doing this change, so you probably know best why this should be removed.

cpetig:
@bjorn I tested the per tileset settings and would say that this could be ready (if necessary) for the next release if you hide the per tile flip properties (as they have no implemention yet and would only confuse users). The document/model/changedata classes will probably need some cleanup (especially when it comes to per tile properties).

src/tiled/adjusttileindexes.cpp
bjorn: It's not commented out because it should be removed, it's commented out because my change was still in progress and this code didn't compile anymore. I'll fix it up once we're done with the rest. :-)

bjorn:
@cpetig Thanks for your hard work on this change and for your persistence as we try different approaches! As said, it may take until Tuesday for me to be able to get back to this change, but it's great to work together on this and I'm sure we'll manage to get this merged soon now.

bjorn:
At least two remaining things still need to be addressed:

* Code related to overriding of these options at the tile level should be either fixed and be implemented similarly to how they are implemented at the tileset level, or they should be entirely removed for now. I don't have a strong opinion on which way to go.

* The disabled code in `ChangeTileWangId::changesOnSetColorCount` and `ChangeTileWangId::changesOnRemoveColor` needs to be re-enabled, similarly to how I just re-enabled the code in `AdjustTileMetaData` (565645a1aac2b7be2c5b2fc7d282dc355733f2fa).

cpetig:
FYI: I rebased the branch on the current master for a start.

cpetig:
I implemented JSON im+export and documented the file format changes (although the meaning of the bitset in the JSON file is not documented. Also I turned off the rotation properties for tiles, so that the incomplete feature is not accessible by/visible for users. I will go to bed now.

PS: I get a warning that Q_FLAGS() is obselete and should be replaced by Q_FLAG() see https://doc.qt.io/qt-5/qobject-obsolete.html#Q_FLAGS

docs/reference/tmx-map-format.rst
```diff
@@ -169,7 +169,7 @@ an ``<image>`` tag.
 
 Can contain at most one: :ref:`tmx-image`, :ref:`tmx-tileoffset`,
 :ref:`tmx-grid` (since 1.0), :ref:`tmx-properties`, :ref:`tmx-terraintypes`,
-:ref:`tmx-wangsets` (since 1.1),
+:ref:`tmx-wangsets` (since 1.1), :ref:`tmx-transformations` (since 1.4.4)
```
bjorn, line 172: Should be "since 1.5".

docs/reference/tmx-map-format.rst
```diff
@@ -245,6 +245,23 @@ Can contain any number: :ref:`tmx-terrain`
...
+
+<transformations>
+~~~~~~
+
+- **hflip:** Whether the tiles in this set can be flipped horizontally to
+  stand in for missing patterns (default 0)
+- **vflip:** Whether the tiles in this set can be flipped vertically to
+  stand in for missing patterns (default 0)
+- **rotate:** Whether the tiles in this set can be rotated by 90 degree to
+  stand in for missing patterns (default 0)
```
bjorn, line 258: I'm not sure about the "to stand in for missing patterns", since that seems to imply untransformed tiles are preferred.

src/libtiled/maptovariantconverter.cpp
```diff
@@ -228,6 +228,9 @@ QVariant MapToVariantConverter::toVariant(const Tileset &tileset,
         tilesetVariant[QStringLiteral("imageheight")] = tileset.imageHeight();
     }
 
+    if (tileset.transformationFlags()!=Tileset::NoTransformation)
```
bjorn, line 231: I prefer just `if (tileset.transformationFlags())`, which is clear enough in meaning no flags are set. I needed the `Tileset::NoTransformation` entry mostly for initialization purposes.

src/libtiled/varianttomapconverter.cpp
```diff
@@ -220,6 +220,7 @@ SharedTileset VariantToMapConverter::toTileset(const QVariant &variant)
     const int columns = variantMap[QStringLiteral("columns")].toInt();
     const QString backgroundColor = variantMap[QStringLiteral("backgroundcolor")].toString();
     const QString objectAlignment = variantMap[QStringLiteral("objectalignment")].toString();
+    const unsigned transformations = variantMap[QStringLiteral("transformations")].toUInt();
```
bjorn, line 223: I'm not sure about storing the bitflags as-is in the JSON format. I know we use bitflags in the tile references, but it's a kind of special case to reduce file size. I'd say we should use named properties here.

src/tiled/propertybrowser.cpp
```diff
@@ -886,14 +886,14 @@ void PropertyBrowser::addTileProperties()
...
-    rotateProperty->setEnabled(false);
-    overrideProperty->setEnabled(mTilesetDocument);
+//    QtVariantProperty *overrideProperty = addProperty(OverrideTransformationFlagsProperty, QVariant::Bool, tr("Override Flipping"), groupProperty);
+//    QtVariantProperty *flipHorizontallyProperty = addProperty(AllowFlipHorizontallyProperty, QVariant::Bool, tr("Flip Horizontally"), groupProperty);
+//    QtVariantProperty *flipVerticallyProperty = addProperty(AllowFlipVerticallyProperty, QVariant::Bool, tr("Flip Vertically"), groupProperty);
+//    QtVariantProperty *rotateProperty = addProperty(AllowRotateProperty, QVariant::Bool, tr("Rotate"), groupProperty);
+//    flipHorizontallyProperty->setEnabled(false);
+//    flipVerticallyProperty->setEnabled(false);
+//    rotateProperty->setEnabled(false);
+//    overrideProperty->setEnabled(mTilesetDocument);
```
bjorn, line 896: Should be removed entirely rather than commented out. Would be good to do this in a separate commit, so it can be reverted as a start on implementing it.

bjorn (commented):
Thanks @cpetig, I'll finish up the rest. I've left some inline comments but I will resolve those as well.

bjorn:
Puh, I guess this change is almost ready for merging. I wonder if I'm doing something wrong that makes some changes take such a long time go finish, and to even finish them in a state that leaves more work to be done!

One remaining thing that's nagging me right now, is that the option of whether to allow rotation essentially only allows a single 90-degree clockwise rotation. You need to allow flipping to enable 180 and 270 degree rotations as well. I wanted this flag to say "rotation" rather than "diagonal flip" because I think it's more intuitive and more useful, but I guess if only rotation is selected, it should really also include 180 and 270 degrees. Just need to find a way to work this into `recalculateCells` without making too much of a mess...

I also wonder a bit what should happen when "Rotate" and "Flip Horizontal" is allowed, but not "Flip Vertical". It makes no sense because Rotate180 + FlipHorizontal = FlipVertical. :S

Any opinions welcome, @cpetig!

bjorn (approved):

bjorn:
@cpetig Thanks for a nice collaboration!

cpetig:
@bjorn Same to you! Thank you for programming, maintaining and offering this nice tool.

PR #2918, opened 2020-10-17
update a litte Chinese translation

765643729:


bjorn:
That's great, thanks!

Just note that these strings are from the [Tiled Manual](https://doc.mapeditor.org/en/stable/) and there is currently no Chinese version hosted online, since the translation is very incomplete.

Another issue is that the manual has been changing a lot since these translation files have been updated, so if you're really interested in translating all of this to Chinese, they'd have to be updated first.

765643729:
> That's great, thanks!
> 
> Just note that these strings are from the [Tiled Manual](https://doc.mapeditor.org/en/stable/) and there is currently no Chinese version hosted online, since the translation is very incomplete.
> 
> Another issue is that the manual has been changing a lot since these translation files have been updated, so if you're really interested in translating all of this to Chinese, they'd have to be updated first.

Ok

PR #2927, opened 2020-10-31
Remove executable bit from non-executable files

hoxu:


bjorn:
Thanks for the small cleanup!

Issue #2933, opened 2020-11-08
First GID property is always null when using TMXMapReader


csueiras:
The first gid property is always null when using the TMXMapReader, the older MapReader used to set it in the buildMap method

```java
tileset = readTileset(xmlPath + tileSetSource);
tileset.setFirstgid(firstGid);
tileset.setSource(tileSetSource);
tilesets.set(i, tileset);
```

I'll create a PR to also set this property in the TMXMapReader.

bjorn:
> The first gid property is always null when using the TMXMapReader, the older MapReader used to set it in the buildMap method

Actually, `TMXMapReader` is the older reader, though it appears to be still used and has seen further changes after `MapReader` was introduced in e95050054df80ba871323feccdf73f973a79f875 (as part of #1637).

@mikepthomas Was the `MapReader` class supposed to replace `TMXMapReader`? Is there a reason to have both? Ideally we would only have a single reader class.

mikepthomas:
Completely agree. MapReader was intended to replace TMXMapReader however it is using JaxB simplify the code to load the the map. JaxB was deprecated and removed in JDK 11 (However it is still available as an external dependecy) my preference would be to remove the MapReader I created and use the TMXMapReader

bjorn:
Alright, let's remove `MapReader` again then. :-)

csueiras:
Wouldn't this still be a bit of a fix since otherwise there's no way to access the firstgid property?

Issue #1613, opened 2017-06-13
Ability to re-arrange tilesets in tileset window is broken.


jo26:
In older versions of Tiled, you were able to re-arrange the tilesets in Tiled's tileset window by click-n-dragging on the the tileset's tab. This also restructured the "firstgid" value in the TMX file, and reorganized the map's data accordingly.

As of 1.0.0 and 1.0.1, this feature no longer works. The only way to re-arrange tilesets now is to delete the tileset, and re-add it at the end of the list. This obviously also breaks your maps, and will remove any tile data on the map.

To verify that this feature did work in past versions, I re-downloaded version 0.18.2 and recorded this GIF:
![tiled 18](https://user-images.githubusercontent.com/29412392/27101148-2950be34-5046-11e7-8490-db5940619bd6.gif)

This feature is vital to me, because my game engine expects certain tilesets to be loaded in a certain order (for example: collision_tileset > level_specific_tilesets > tilesets_only_used_wihle_level_editing)

bjorn:
Hmm, this feature was removed because the Tilesets view now displays _all_ loaded external tilesets regardless of which map is selected. Since it no longer reflects just the tilesets used by a single map, I chose to simply order them alphabetically, though that's just for display purposes.

The order in which the tilesets appear in the map file should be irrelevant so I'd rather not introduce additional UI for having this option again. How hard would it be for you to use something else than the tileset order to determine when to load each tileset, like a custom tileset property for example?

jo26:
Well I'm using a plugin that loads TMX files, and I don't have the source to it. I think the plugin simply reads the tilesets in the order that they are listed in the file, incrementally. There are certain tilesets that I skip loading, and I think this is causing errors that didn't manifest before I installed Tiled 1.0 for some reason.

This may be more of an issue on how the plugin I'm using is reading the TMX file, and not necessarily Tiled's fault. Or it could be my own error. I'm gonna play around with it and see if I can find anything. I'll report back if I can figure out what went wrong on my end.

Regardless, I do miss the ability to organize my tilesets how I want. But I suppose it is a minor issue, and I can adjust.

Keep up the great work! Tiled is such a time saver!

alaa-eddine:
Just updated from Tiled 0.18 to 1.0.3 and noticed that the feature is missing it was driving me crazy :D 
in my case, I was using the possibility to rearrange tab to optimize the work.
[...]

PR #2934, opened 2020-11-08
#2933 Set the firstgid property on Tilesets

csueiras:
Fixes issue #2933 

I've also added test coverage for the property being set.

@bjorn I know there's some on going work to release the latest version of the library to Maven central, let me know how can I help out to get these changes into master and released. Thanks!


bjorn:
Hmm, why do you need the tileset's firstGid attribute set? What part of the code relies on it?

I'd prefer if we could remove the firstGid member from the Tileset class instead, because I don't think it belongs there. It is only set when saving a map and only used while loading a map, and for that we have `tilesetPerFirstGid`.

csueiras:
@bjorn isn't the firstgid required to map a global id back to a tileset? I am using this library in a server-side setting btw to parse out maps/tilesets.

bjorn:
@csueiras Normally that happens while loading the map, here:

https://github.com/bjorn/tiled/blob/54d145e28661ca1ed8429ab8b137662b05590d46/util/java/libtiled-java/src/main/java/org/mapeditor/io/TMXMapReader.java#L1028-L1030

which is used from:

https://github.com/bjorn/tiled/blob/54d145e28661ca1ed8429ab8b137662b05590d46/util/java/libtiled-java/src/main/java/org/mapeditor/io/TMXMapReader.java#L780-L787

which is used from:

https://github.com/bjorn/tiled/blob/54d145e28661ca1ed8429ab8b137662b05590d46/util/java/libtiled-java/src/main/java/org/mapeditor/io/TMXMapReader.java#L765-L771

So in the end, you would be working with `Tile` references and should not need to care about mapping any global IDs.

bjorn:
@csueiras With the clarifications above, do you still see a use for exposing the `firstgid` attribute? If you still need it, please let us know why.

duarm:
More than setting the firstgid to an specific value, I would like to reorder each tileset. I have an exclusive tileset for "non-colliding" tiles, setting that to the first tileset allows me to check for collision with `tile < tileset[0].firstgid`. Currently, if you create a new map and arbitrarily start painting, the firstgids would be out of order, and you can't reorder them without destroying the map.

bjorn:
@duarm This PR is a proposed change affecting the Java library for loading Tiled maps. I don't think it has anything to do with your concern, which appears to be about manually setting the firstgid or manually ordering the tilesets. The latter is covered by issue #1613.

Note however, that the firtgids for the tilesets will always be incremental, as defined in the [file format documentation](https://doc.mapeditor.org/en/stable/reference/tmx-map-format/#tileset). So, manually placing a tileset at the start of the list would change its firstgid to 1:

> "If there are multiple `<tileset>` elements, they are in ascending order of their `firstgid` attribute. The first tileset always has a `firstgid` value of 1."

So I don't think your plan would work as such. You'd have to put that tileset first, and then check against the `firstgid` of the second tileset: `tile < tilesets[1].firstgid`. However, I'd still recommend against such an approach, and would rather advise to just find out which tileset the tile is from (as [documented here)](https://doc.mapeditor.org/en/stable/reference/global-tile-ids/#mapping-a-gid-to-a-local-tile-id) and check a custom tile or tileset property to decide whether a tile is colliding. It would be more flexible and more readable that way.

duarm:
Sorry for the wrong issue, yeah, I meant to check for the second tileset, wrong pseudo code

PR #2935, opened 2020-11-11
Change the website to use its own css theme, look more modern

ko2fan:
Changed the website to look more modern. Work in Progress.

TODO:
Dynamically show donation bar.
Dynamically load news.
Update images.
Tweak colours and formatting.


bjorn:
@ko2fan Thank you so much for your work on this!

I actually didn't realize my test with Netlify would lead to an automatic deploy preview, that's pretty neat! I'll see where I can help regarding the dynamic content and finalizing the layout & theme.

bjorn:
@ko2fan What's the purpose of `media-content`? Also, is there just a small mistake somewhere causing it to not be used for the header "Games made with Tiled"? 

Anyway I've been making some changes today, so news and sponsors are dynamic now and I've tweaked the sponsors grid to be a bit more fancy. As usual stuff like this is taking much more time than I had thought, but it's good to make some progress. I think a major job would still be to make the other pages work as well again, like news posts and news archive.

While the hover effect on the sponsor banners works well on desktop, it's not suitable for mobile. Probably on mobile the scale effect and dimmed opacity for non-hovered banners should be disabled. Maybe there is a media query for this?

ko2fan:
@bjorn I had only created a preview of the front page. I figured other pages would follow once you were happy with the front page.
Your changes look good. I am happy to change the sponsors for mobile, I will take a look.

Issue #1211, opened 2016-02-25
Request to Add Support for Enumeration Properties


lazybotter:
One really good option to add to the "Add Property" dialogue would be the ability to add multiple string values that would display as a "Select" or "Dropdown Menu".

So for an example:

I could create an object that has a Type of "Box".
The "Box" could then have a "BoxType" property. "Solid Box, Mystery Box, Danger Box".
Then when creating the object on the map we can just select what "BoxType" we would like the "Box" to be.
This would be great for custom game engines etc when creating the maps.

Thanks! Carl ;-)

bjorn:
Definitely. The main challenge here is that such enumeration properties would need to be based on enumeration types, which would have to be defined before you can define a property of that type. So this means a new section in `objecttypes.xml` for complex type definitions (I could also imagine supporting types that have sub-properties here later) and then coding some way in which these types get used by the property browser.

Starting with just enumeration properties would be a good idea, I just want to be prepared a little for sub-properties later.

lazybotter:
Any movement on this? :-)

lazybotter:
Anyone started on this yet?

bjorn:
I haven't gotten around to this yet. If it is any comfort, for Tiled 0.17 I did get around to several other property enhancements, like adding "file" and "color" types and supporting multi-line values for "string" properties.

This feature is definitely still on the list of things to improve about the custom properties system.

jeremyherbert:
Has any work been done on the UI of this? ie how would one define the enum types? Would you have a separate editor window, much like the object types editor, but one that is used just for defining enums? Or something in the same window? Are enums defined as a simple integer->string mapping like in C, or something else?

I'd be interested in giving this a go, but I am not sure which is the best way to proceed. I'd of course like to try and have the code merged eventually.

bjorn:
@jeremyherbert No work has been done on this. Indeed it would be good to mock up the UI before starting to code this. Thanks for your interest to work on this and to try getting it merged!

> Are enums defined as a simple integer->string mapping like in C, or something else?

I imagine the enum will be defined only as the list of options (strings), and these strings will be used when saving/loading. A downside of that is that you can't easily rename one of the enum options, but it does mean you can add/remove them without causing many issues (and without requiring the user to be careful with value assignment in this case).

> Would you have a separate editor window, much like the object types editor, but one that is used just for defining enums? Or something in the same window?

I'm hope we don't need too many separate windows and am thinking of something like this to start with:

[...]

PR #2941, opened 2020-11-28
Early PR for Custom/Enum Props

svipal:
resolves https://github.com/bjorn/tiled/issues/1211
Hello ! I've started to implement enum props, as discussed in the discord and on the linked issue.

I made an early PR even though not everything is implemented, as suggested by bjorn.
Here's what's not yet in it :
- enum props are the only possible custom props instead of just one subclass
- properties of custom type do not persist between sessions

Here's what works :
- loading/saving custom properties from project
- editing/saving enum props with the editor
- making new properties with thus made "enums"

Let me know what you think !


src/libtiled/customproperties.h
```diff
@@ -0,0 +1,142 @@
...
+    {}
+
+
+    QString name;
+
+    QColor color;
+
+    QStringList values;
+
+    int currentIndex;
```
bjorn, line 74: I don't think this belongs here. We can have a "default value" member of the custom type (though I'd say we pick the first one by default for now), but the actual value should not be stored along with the type.

src/libtiled/customproperties.h
```diff
@@ -0,0 +1,142 @@
...
+
+    static CustomProp fromVariant(const QVariant &variant)
+    {
+        const auto hash = variant.toHash();
+
+        auto read = [&] (const QString &prop) {
+            if (hash.contains(prop))
+                return hash.value(prop);
+
+            QString oldProp = prop.at(0).toUpper() + prop.mid(1);
```
bjorn, line 118: What is this for?

src/tiled/addpropertydialog.cpp
```diff
@@ -46,24 +48,36 @@ AddPropertyDialog::AddPropertyDialog(QWidget *parent)
...
     mUi->typeBox->addItem(typeToName(QVariant::Color),   QColor());
     mUi->typeBox->addItem(typeToName(QVariant::Double),  0.0);
     mUi->typeBox->addItem(typeToName(filePathTypeId()),  QVariant::fromValue(FilePath()));
     mUi->typeBox->addItem(typeToName(QVariant::Int),     0);
+    
     mUi->typeBox->addItem(typeToName(objectRefTypeId()), QVariant::fromValue(ObjectRef()));
     mUi->typeBox->addItem(typeToName(QVariant::String),  QString());
 
+
+    CustomProps customProps = Object::customProps();
```
bjorn, line 63: Should be `const CustomProps &customProps` to avoid reference counting and potential copying.

src/tiled/addpropertydialog.cpp
```diff
@@ -46,24 +48,36 @@ AddPropertyDialog::AddPropertyDialog(QWidget *parent)
...
+
+
+    for (const CustomProp cProp: customProps) {
```
bjorn, line 66: Should be `const CustomProp &cProp` to avoid copying.

src/tiled/addpropertydialog.cpp
```diff
@@ -46,24 +48,36 @@ AddPropertyDialog::AddPropertyDialog(QWidget *parent)
...
+    }
+        
     mUi->buttonBox->button(QDialogButtonBox::Ok)->setEnabled(false);
 
     // Restore previously used type
     mUi->typeBox->setCurrentText(session::propertyType);
 
     connect(mUi->name, &QLineEdit::textChanged,
-            this, &AddPropertyDialog::nameChanged);
+        this, &AddPropertyDialog::nameChanged);
```
bjorn, line 78: Please leave the original alignment intact. The coding style I follow for wrapped parameters aligns them to the opening bracket.

src/tiled/tiled.pro
```diff
@@ -542,23 +547,23 @@ FORMS += aboutdialog.ui \
     tileanimationeditor.ui
 
 icon32.path = $${PREFIX}/share/icons/hicolor/32x32/apps/
-icon32.files += images/32/tiled.png
+icon32.files += images/32x32/tiled.png
```
bjorn, line 550: Seems like a revert of a recent change I made, please undo this part.

src/tiled/variantpropertymanager.cpp
```diff
@@ -199,6 +202,17 @@ QString VariantPropertyManager::valueText(const QtProperty *property) const
             return tr("%1: Object not found").arg(QString::number(ref.id()));
         }
 
+        if (typeId == filePathTypeId()) {
```
bjorn, line 205: Seems you forgot to adjust this code for the `customTypeId()`.

src/tiled/variantpropertymanager.cpp
```diff
@@ -156,14 +159,14 @@ int VariantPropertyManager::alignmentTypeId()
...
 {
-    return qMetaTypeId<DisplayObjectRef>();
+    return qMetaTypeId<UnstyledGroup>();
 }
 
-int VariantPropertyManager::unstyledGroupTypeId()
+int VariantPropertyManager::displayObjectRefTypeId()
 {
-    return qMetaTypeId<UnstyledGroup>();
+    return qMetaTypeId<DisplayObjectRef>();
```
bjorn, line 169: Please don't needlessly reorder these.

src/tiled/varianteditorfactory.h
```diff
@@ -69,12 +69,18 @@ class VariantEditorFactory : public QtVariantEditorFactory
...
+    void enumPropEditTextChanged(const QString &value);
     void objectRefEditValueChanged(const DisplayObjectRef &value);
     void slotEditorDestroyed(QObject *object);
 
     QMap<QtProperty *, QList<FileEdit *> > mCreatedFileEdits;
     QMap<FileEdit *, QtProperty *> mFileEditToProperty;
 
+    // for now there are no other customprops.
+    QMap<QtProperty *, QList<QComboBox *> > mCreatedEnumProps;
+    QMap<QComboBox *, QtProperty *> mEnumPropToProperty;
```
bjorn, line 81: It would be nice if we could delegate widget creation somehow, because a custom type could indeed be more than just an enum but could also be an integer value with custom range for example. The `QtVariantEditorFactory` can already do this, with the only problem that its value needs to be the actual value and not a `CustomValue`, but we need to pass around a `CustomValue` in order to keep track its custom type. Making this possible would require some changes to the `qtpropertybrowser` solution.

src/tiled/objecttypeseditor.h
```diff
@@ -23,7 +23,13 @@
 #include "properties.h"
 
 #include <QDialog>
-
+#include <QColorDialog>
+#include <QCloseEvent>
+#include <QFileDialog>
+#include <QInputDialog>
+#include <QMessageBox>
+#include <QPainter>
```
bjorn, line 31: Seems to me like you need none of these includes, only the `QStyledItemDelegate` include should be added.

bjorn (changes requested):
It's a good start! But I think there's one major issue, which is that you're storing the whole definition of the custom type with each value. I think they should be separated. We can have a "default value" member of the custom type (though I'd say we pick the first one by default for now), but the actual value should not be stored along with the type.

I think we need a separate struct for storing the value, something like:

```c++
struct CustomValue
{
    QVariant value;
    QString type;
}
```
It's going to be a bit confusing, but we would then store this in the `QVariant` storing a property value, and when `value.userType() == qMetaTypeId<CustomValue>()`, we'd be able to look up both the actual value (using `value.value<CustomValue>().value`) and its type. We'd use the type to look up the information about the type so that we can set up appropriate widgets in the `VariantEditorFactory`.

In addition:

* The `toExportValue` function would need to handle custom types.

* `typeToName` will need to take the full `QVariant` as parameter instead of just the type ID, so that it can return the `CustomValue::type` member when appropriate.

* `nameToType` can't just return a type ID anymore. I think the function can be integrated into `fromExportValue`, passing the type as string instead of int to that function. Considering this, it probably makes sense to integrate `typeToName` also with `toExportValue`, probably making it return a `CustomValue` with the `value` member being the value for export and the `type` member the type name to export.

I've left further minor comments inline.

src/libtiled/customproperties.h
svipal: Leftover code I forgot to clean up ~

edin-m:
Is this being worked on?

bjorn:
@edin-m There hasn't been any activity for some months, so I think this change is currently abandoned. Since it's still a priority request from a sponsor I do intend to start work on it soon. Are you asking because you'd like to help complete this feature?

edin-m:
Maybe... I'll see after the components feature.

src/tiled/varianteditorfactory.h
bjorn: I've resolved this issue by moving the translation from `PropertyValue` to the associated property browser type and attributes to a `CustomPropertiesHelper`, which is used by `PropertyBrowser` and `ObjectTypesEditor` to create the properties and convert the values.

PR #2942, opened 2020-12-03
Fix Python plugin linking issue with Python 3.8

kensington:
The default pkg-config module no longer contains -lpython3.8.

Reference: https://docs.python.org/3/whatsnew/3.8.html#debug-build-uses-the-same-abi-as-release-build

### DIFF
--- a/src/plugins/python/find_python.pri
+++ b/src/plugins/python/find_python.pri
@@ -1,6 +1,10 @@
 !contains(DISABLE_PYTHON_PLUGIN, yes) {
     unix {
-        packagesExist(python3) {
+        packagesExist(python3-embed) {
+            HAVE_PYTHON = yes
+            CONFIG += link_pkgconfig
+            PKGCONFIG += python3-embed
+        } else:packagesExist(python3) {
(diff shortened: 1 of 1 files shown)


kensington:
Otherwise, the plugin load fails at runtime:
```
Error: Cannot load library /usr/lib64/tiled/plugins/libpython.so: (/usr/lib64/tiled/plugins/libpython.so: undefined symbol: _Py_FalseStruct)

bjorn:
Thanks!

It would be good to make a similar change to the Qbs build files, which currently uses pkg-config through the following snippet:

https://github.com/mapeditor/tiled/blob/34669e936986cb30239b9c0936e134aa2727c05c/src/plugins/python/python.qbs#L17-L20

I guess it would work by simply including another probe and adding some logic.

bjorn:
I'll have a look at the Qbs project later, in the meantime we can at least have working qmake projects. Thanks again! :-)

bjorn:
Finally resolved for Qbs as well in fa73b7366343608a7bb0c13522b06416fbdcbab8.

PR #2956, opened 2020-12-12
Fix loading extension path from project config

fmoo:
Extension path was getting saved under the `extensionsPath` key
but getting read from the `extensionsFolder` key.

Read the value back from the written `extensionsPath` key.


bjorn:
Thanks @fmoo for finding this error and the fix! It should really go onto the `1.4` branch though, but I think I'll just cherry-pick it there (don't know right now whether there will be a 1.4.4 before 1.5).

PR #2957, opened 2020-12-12
Add `%scriptargs` for Command replacement

fmoo:
When invoking a command from an extension script, there isn't a
great way to specify custom arguments to the command.

Solve this by:
- Adding an extra parameter to the `tiled.executeCommand()` function
- Replacing the contents of %scriptargs in the command definition with that parameter


fmoo:
Per conversation on the discord: a better approach would be to expose an execute command API directly to the scripting layer instead of hopping through the Command presets.

PR #2959, opened 2020-12-12
Allow scripts to extend the object context menu

fmoo:
Presently, there is no way for extension scripts to modify the
context menus shown, e.g., when right-clicking an object.

Resolve this by:
- Add a `QMenu *mCustomObjectMenu` to MainWindow.  Define it as empty, but call registerMenu on it.
- Add a `void addCustomObjectActions(QMenu *menu) const` to MainWindow. This is kinda like the existing addRecentProjectsActions API on it.
- Update `AbstractObjectTool::showContextMenu` to call `MainWindow::Instance().addCustomObjectActions(menu)` before calling exec.

### DIFF
--- a/src/tiled/abstractobjecttool.cpp
+++ b/src/tiled/abstractobjecttool.cpp
@@ -25,6 +25,7 @@
 #include "changemapobject.h"
 #include "changetileobjectgroup.h"
 #include "documentmanager.h"
+#include "mainwindow.h"
 #include "mapdocument.h"
 #include "map.h"
 #include "mapobject.h"
@@ -612,6 +613,8 @@ void AbstractObjectTool::showContextMenu(MapObject *clickedObject,
 
     Utils::setThemeIcon(removeAction, "edit-delete");
     Utils::setThemeIcon(propertiesAction, "document-properties");
+
+    MainWindow::instance()->addCustomObjectActions(&menu);
 
     QAction *action = menu.exec(screenPos);
     if (!action)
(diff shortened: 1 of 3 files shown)


bjorn:
Thanks for this pull request @fmoo! It highlights an important missing feature, but I don't like the approach. This way, we can't control the position where the actions are inserted and it only works for the object context menu. There are a number of other context menus that would be nice to be able to extend as well.

I have a plan for supporting this for all context menus: Instead of directly extending existing `QMenu` instances in `ScriptModule::extendMenu`, we pass the `MenuExtension` object to the `ActionManager`. That way, the `ActionManager` can extend any menus that are already registered as well as those that get registered later, or extend context menus if they just make a call to a function like `ActionManager::applyMenuExtensions(QMenu *menu, Id menuId)`. On script reload, we'd need to make a call like `ActionManager::clearMenuExtensions()`.

To still be able to raise an error when the user tries to extend a non-existing menu, we could allow registering a `nullptr` as menu and adjusting the check accordingly.

Let me know whether you are interested in redoing your patch accordingly. If not I shall implement this approach next week.

fmoo:
Thanks for the reply @bjorn.  I've got my hands pretty full right now and the approach sounds a bit more complicated than I'd have cycles to handle for the near future.

If you're planning to tackle it soon, that's fantastic!  That, combined with the ability to invoke arbitrary commands in extensions will really unlock a huge amount of potential (e.g. enabling tile to directly launch a cutscene editor or database file based on the object properties, etc)

bjorn:
Alright, I've done an initial implementation of the alternative approach at 748f6438acc8edafc39080788a1ff39ccc157bbf (`wip/extendable-context-menus`).

bjorn:
Change submitted as 42430a8e9aecce6859fc07b6ea4abe9a93f56c24 with support for extending a number of context menus.

PR #2961, opened 2020-12-15
Add first draft of tsdoc based documentation

Ablu:
The scripting-interface.ts is currently mostly in the state as provided
by Alula on Discord, I only fixed the syntax for some function
declarations so that they pass as regulat .ts files and added
documentation for `registerMapFormat` as a proof of concept.


bjorn:
Thanks @Ablu and Alula, that's really useful!

Issue #2783, opened 2020-04-21
Allow JS plugin environment to run sub-process


konsumer:
It would be handy if we could spawn a CLI program, for really simple plugins that just wrap some other existing tool. This would allow people to mostly write plugins in whatever language they like, or whatever is already made, as most languages have a way to make a CLI program (stdio/stdout, file reading/writing, etc.)

This is related to #2780 and I'd like to add it in a similar fashion to `ScriptFileInfo`.

Some examples of plugin systems that use CLI programs to do fancy stuff, that we might get inspiration from:

* [inkscape](https://inkscape.org/) has a [nice system](https://inkscape.org/develop/extensions/) where you define meta-data in an XML file. They have a whole system where you can define the UI and it will pass what was selected as options to your CLI program, so you can effectively write super-integrated plugins, with your logic in any language you like (as long as you have the local runtime for it.) This means that python is very common, but for web-related tools I have made a few inkscape extensions in nodejs. In this system, you just directly write to the file, and receive the SVG source on stdin.
* [protoc](https://github.com/protocolbuffers/protobuf) - you write your plugin in whatever language you like and put it in your path. stdin & stdout are in a pre-defined binary format (protobuf, which you need in your language anyway to parse protobuf files) and you output what you want the plugin to write to. It feeds your script a pre-parsed protobuf object, so you just have to be able to read basic protobuf messages to write a plugin.

I imagine in it's simplest form, it would just be a thin wrapper for [QProcess](https://doc.qt.io/qt-5/qprocess.html). So you could something like this (without any async stuff):

```js
/* global tiled, TextFile, Process */

tiled.registerMapFormat('Subprocess Demo', {
  name: 'Subprocess Demo format',
  extension: 'txt',

  write: function (map, fileName) {
    const n = new Process()
    n.start(`ls ${fileName}`)
    n.waitForFinished()
    // some plugins might just write the file directly, themselves
    const out = n.readAllStandardOutput()
    const file = new TextFile(fileName, TextFile.WriteOnly)
    file.write(out)
    file.commit()
  }
})
```

[...]

PR #2966, opened 2020-12-18
Add Scripting Function `tiled.execute(command, arguments)`

fmoo:
Currently, there is no way to pass custom arguments from your
script to commands via `tiled.executeCommand()`.  This change
implements a raw, `tiled.execute()` which takes the executable
and parameters as an argument, and runs that.

Obsoletes #2957

### DIFF
--- a/src/tiled/scriptmodule.cpp
+++ b/src/tiled/scriptmodule.cpp
@@ -498,6 +498,21 @@ void ScriptModule::executeCommand(const QString &name, bool inTerminal) const
     ScriptManager::instance().throwError(QCoreApplication::translate("Script Errors", "Unknown command"));
 }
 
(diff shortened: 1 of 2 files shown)


bjorn:
Alright, so I considered this approach and I think we really should just expose a `Process` API as [available in Qbs](https://doc.qt.io/qbs/jsextension-process.html). It might be a bit more code to use, but it gives a lot more functionality. We can have a look at the code in [Qbs4QJS](https://github.com/4QJS/Qbs4QJS). See also issue #2783.

bjorn:
Ah, I mentioned the wrong issue number. Anyway, this change is obsoleted by the new Process API added in f7445727e988c0afe17e9556fc198cf079ed6fcc.

PR #2977, opened 2021-01-20
Update Czech translation

encukou:


bjorn:
Thank you so much for the update!

PR #2979, opened 2021-01-27
Added export to GameMaker Studio 2

### LINKED ISSUE #1642: GameMaker Studio 2 support
Currently Tiled ships with an export plugin for GameMaker: Studio 1.4. With GameMaker Studio 2, YoYo Games has changed the map format entirely. It is now JSON based instead of XML and the tile layer data is saved in a very similar fashion as Tiled JSON maps.

Even though GameMaker Studio 2 ships with quite a decent tile map editor, it would still be interesting to add support for this new version as well. It would be helpful for those already using Tiled that are porting their game to the newer version, as well as for people who have reason to prefer Tiled for making their maps.

Requested in [this forum post](http://discourse.mapeditor.org/t/game-maker-studio-2-support/2527) and [this forum post](https://discourse.mapeditor.org/t/yy-room-exports/4167).

bjorn:
Closes #1642

Quite a bit of work done, but some tasks remain and no testing was done at all yet.

- [x] Export tile layers as `GMRTileLayer` (single tileset per layer) 
- [x] Export tile objects as `GMRSpriteGraphic` or `GMRGraphic` on `GMRAssetLayer`
- [x] Export of image layers as `GMRBackgroundLayer`
- [x] Export group layers as `GMRLayer`
- [x] Export of instances (objects with type) to `GMRInstanceLayer`
- [x] Support tile layers using multiple tilesets (create a group layer with multiple child tile layers)
- [ ] Export paths (polyline and polygon) as `GMPath` asset and a `GMRPathLayer`
- [x] Export tile layers to asset layers when they use image collection tilesets, or tiles where the size doesn't match the map grid size.
- [x] Export instance creation order
- [x] Handle overriding depth better

### DIFF
--- a/dist/win/installer.wxs
+++ b/dist/win/installer.wxs
@@ -168,6 +168,7 @@
 
                 <File Source="$(var.InstallRoot)\plugins\tiled\tbin.dll" />
                 <File Source="$(var.InstallRoot)\plugins\tiled\tengine.dll" />
+                <File Source="$(var.InstallRoot)\plugins\tiled\yy.dll" />
               </Component>
             </Directory>
             <Directory Id="platformPlugins" Name="platforms">
--- a/src/libtiled/maprenderer.cpp
+++ b/src/libtiled/maprenderer.cpp
@@ -369,7 +369,7 @@ void CellRenderer::render(const Cell &cell, const QPointF &screenPos, const QSiz
     } else if (cell.flippedAntiDiagonally()) {
         fragment.rotation = 90;
 
-        flippedHorizontally = cell.flippedVertically();
+        flippedHorizontally = flippedVertically;
         flippedVertically = !cell.flippedHorizontally();
 
         // Compensate for the swap of image dimensions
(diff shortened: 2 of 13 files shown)


TrunXGIT:
More or less Critical:

- [ ] x and y adjustment of sprites and objects in regard to origin, flipping, scale and rotation-
> just sprites:
> smth. is off with the vertical flipping (depending on the rotation it gets moved to a different wrong position)
> 
> sprites and objects:
> scaling is not part of the x and y (origin) adjustment?
- [ ] Support for views (rectangle objects) and its properties. 

Nice to have:

- [ ] Support for custom properties of objects 
>  (written to the "properties:[]" list of the instances)
> -> always saved as strings
- [ ] Support to overwrite the default sprite name of the image source with a custom property "sprite".
- [ ] Use Map Property "Background Color" (if set) to create a background layer with the accourding color and no image as the backmost layer.
- [ ] Support for custom layer property "depth" and automatic even distribution of the depth values of the layers in-between. (Mimicing GMS2 behaviour)

Issue #2863, opened 2020-07-14
Change tileset definition to avoid renumbering of tiles when image width changed


bjorn:
One problem that's common enough to cause quite some pain is that when a tileset's image width is changed, this causes all the tiles in the tileset to get renumbered. This in turn messes up any tile references from tile layers and tile objects (also affecting object templates) and requires a lot of custom code to fix (a workflow that has always been somewhat broken, see for example issues #1770 and #1851). Even tile references within a tileset definition, like for terrain images, animation frames or Wang sets need adjustment.

I'm considering a solution where adjustment to other assets is entirely avoided, by keeping the tile IDs stable (still pointing to the same location in the image). To achieve this, we can store for each tile its x and y coordinate on the tileset image. Then, when the image has grown in size next time the tileset is loaded, any new tiles will get added to the end of the list, receiving new IDs just like when adding images to an Image Collection tileset.

It would be interesting to store also the width and height of each tile, such that we can at the same time support tilesets that contain tiles of varying sizes (issue #1008), if we add some UI to manage the tiles in a tileset.

One question is what to do when the user changes the parameters of the tileset, like when halving or doubling the tile size. On the one hand, we can't expect this will keep any map using that tileset to still look alright, so it might be OK to just renumber all tiles in this case. On the other hand, this may commonly happen after changing the resolution of the tiles:

* User changes tileset image to a version 2x bigger
* Tiled detects more tiles on the image (actually 4x the amount of tiles), new tile IDs are added
* User changes tileset tile size

Probably, at the last step, Tiled should make an effort to match up tiles based on the old and new size parameters, making sure their IDs stay the same.

bjorn:
As [eishiya pointed out on the forum](https://discourse.mapeditor.org/t/quality-of-life-feature-suggestions/4619/5), in the interest of compatibility it would be a good idea to introduce a new type of tileset for this purpose. This way Tiled can still support the old method where tiles would get renumbered and limit or hide the UI actions, depending on whether the tileset is marked as "spritesheet" or not.

bjorn:
Suggested format for storing the sub-rectangle for each tile:

For image collection tileset:

```xml
<tileset firstgid="1" name="objs" tilewidth="2000" tileheight="2000" tilecount="63" columns="0">
 <grid orientation="orthogonal" width="384" height="332"/>
 <tile id="0" x="0" y="0" width="160" height="192">
  <image width="160" height="192" source="alter.png"/>
 </tile>
```
[...]

PR #2983, opened 2021-02-07
[WIP] Relocate tiles

jmi2k:
**WIP:** The patch works: I think I've fixed all the bugs I could have introduced and it's ready for code review. However, there are still some details I want to fix before merging.

This PR allows tiles in a tileset to be moved around at will. Previously, the tileset order was fixed and couldn't be changed. That was acceptable for prebuilt tilesets but horrendous for collections of images (after you've added some tile, it was appended to the list couldn't be placed elsewhere, which made working with them a nightmare).

### DIFF
--- a/src/libtiled/mapwriter.cpp
+++ b/src/libtiled/mapwriter.cpp
@@ -443,7 +443,8 @@ void MapWriterPrivate::writeTileset(QXmlStreamWriter &w, const Tileset &tileset,
     }
 
     // Write the properties for those tiles that have them
-    for (const Tile *tile : tileset.tiles()) {
+    for (int id : tileset.sortedTileIds()) {
+        const Tile *tile = tileset.findTile(id);
         if (imageSource.isEmpty() || includeTile(tile)) {
             w.writeStartElement(QStringLiteral("tile"));
             w.writeAttribute(QStringLiteral("id"), QString::number(tile->id()));
--- a/src/tiled/tilesetdocument.cpp
+++ b/src/tiled/tilesetdocument.cpp
@@ -347,6 +347,13 @@ void TilesetDocument::removeTiles(const QList<Tile *> &tiles)
     emit tilesetChanged(mTileset.data());
 }
 
+void TilesetDocument::relocateTile(const Tile *tile, int location)
+{
+    mTileset->relocateTile(tile->id(), location);
+    /* TODO: is that the right signal? */
+    emit tilesetChanged(mTileset.data());
+}
+
 void TilesetDocument::setSelectedTiles(const QList<Tile*> &selectedTiles)
 {
     mSelectedTiles = selectedTiles;
--- a/src/tiled/tilesetmodel.cpp
+++ b/src/tiled/tilesetmodel.cpp
@@ -95,9 +95,16 @@ Qt::ItemFlags TilesetModel::flags(const QModelIndex &index) const
     Qt::ItemFlags defaultFlags = QAbstractListModel::flags(index);
 
     if (index.isValid())
-        defaultFlags |= Qt::ItemIsDragEnabled;
+        defaultFlags |= Qt::ItemIsDragEnabled | Qt::ItemIsDropEnabled;
+    else
+        defaultFlags |= Qt::ItemIsDropEnabled;
 
     return defaultFlags;
+}
+
+Qt::DropActions TilesetModel::supportedDropActions() const
+{
+    return Qt::MoveAction;
 }
 
 QStringList TilesetModel::mimeTypes() const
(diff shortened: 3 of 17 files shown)


src/libtiled/mapwriter.cpp
```diff
@@ -443,7 +443,8 @@ void MapWriterPrivate::writeTileset(QXmlStreamWriter &w, const Tileset &tileset,
     }
 
     // Write the properties for those tiles that have them
-    for (const Tile *tile : tileset.tiles()) {
+    for (int id : tileset.sortedTileIds()) {
+        const Tile *tile = tileset.findTile(id);
         if (imageSource.isEmpty() || includeTile(tile)) {
```
bjorn, line 448: This code will only write a `tile` element if either `imageSource.isEmpty()` (when it's an image collection tileset) or when `includeTile` returns true (when the tile has been customized in some way). This will need to be adjusted such that any tile which needs its order to be saved is also included (or possibly better, when the order has been changed in any way, all tiles should be written).

It might even be alright to just remove these checks entirely since there's also a plan to store the location of each tile in the set as per #2863.

Note that the same kind of check happens in other map formats (`maptovariantconverter.cpp` and `luaplugin.cpp`).

src/tiled/tilesetmodel.cpp
```diff
@@ -127,6 +136,50 @@ QMimeData *TilesetModel::mimeData(const QModelIndexList &indexes) const
...
+        return false;
+
+    QByteArray encodedData = data->data(QLatin1String(TILES_MIMETYPE));
+#if QT_VERSION < QT_VERSION_CHECK(6,0,0)
+    QDataStream stream(&encodedData, QIODevice::ReadOnly);
+#else
+    QDataStream stream(&encodedData, QDataStream::ReadOnly);
+#endif
+
+    int sourceId;
```
bjorn, line 158: Got a warning here that `sourceId` might be left uninitialized and still used. Would be better to make sure we only use it when it's set.

src/tiled/tilesetdocument.cpp
```diff
@@ -347,6 +347,13 @@ void TilesetDocument::removeTiles(const QList<Tile *> &tiles)
     emit tilesetChanged(mTileset.data());
 }
 
+void TilesetDocument::relocateTile(const Tile *tile, int location)
+{
+    mTileset->relocateTile(tile->id(), location);
+    /* TODO: is that the right signal? */
```
bjorn, line 353: I think this is currently the right signal, though it isn't optimal or entirely correct. Ideally we'd send a signal before and after the change, such that the model can call `beginResetModel` and `endResetModel` in the appropriate moments. At the same time I'm trying to move away from all the separate signals, instead sending an specialized `ChangeEvent` in a unified `Document::changed` signal. I can help with this part of the change.

src/tiled/tilesetmodel.cpp
```diff
@@ -22,17 +22,20 @@
...
 #include "tiled.h"
 #include "tileset.h"
+#include "tilesetdocument.h"
 
 #include <QMimeData>
 
 using namespace Tiled;
 
-TilesetModel::TilesetModel(Tileset *tileset, QObject *parent):
+TilesetModel::TilesetModel(Tileset *tileset, TilesetDocument *tilesetDocument, QObject *parent):
```
bjorn, line 35: In case this `tileset` and `tilesetDocument` should always match, it would be better to just pass the `tilesetDocument`, since we can obtain the tileset from there.

bjorn (changes requested):
This works quite well already! I've left some inline comments on the implementation.

Apart from the comments, there was a little problem with loading old maps, which only include `tile` elements when the tile was customized in some way. Those tiles will now appear at the start. We need to make sure the order of the tiles is not affected in this case.

I think a possible solution to this would be, to remember the last already found tile index in `Tileset::loadFromImage`, and to insert tiles after this index in `mSortedTileIds` instead of appending them (also incrementing this insertion index for each newly created tile).

src/tiled/tilesetmodel.cpp
jmi2k: Yeah, in the case there is no data embedded in the drop, it would be left uninitialized. The fix was really simple though, I've just didn't see it.

src/tiled/tilesetdocument.cpp
jmi2k: I'll leave it as-is. If that's the best signal for this scenario (given how signal handling is currently implemented) I think it's better to make it consistent with the rest of the code. Signal handling improvements can come later in a separate change.

bjorn:
@jmi2k I've addressed a few of my own comments today and plan to do another review later this week and address the rest of my feedback, hoping to get this merged.

I think it'd also make sense to look into supporting multi-selection, because moving a lot of tiles can get tedious. Please feel free to still help polish up this feature. I'll be off for the next 16 hours or so. :-)

jmi2k:
Thanks, I'm very sorry I couldn't put some time aside to push this forward. I dismissed multi-selection because it could behave in weird ways. Does it preserve the shape? what does it do when reaching a corner? How does it behave when replacing tiles? I was hoping to get a basic version merged before discussing future improvements

bjorn:
Thanks for your help with this change @jmi2k! No worries about not finishing it, I fully understand other things can come up, as also happens for me since I've let #1856 be unresolved for way too long as well.

I've added support for multi-selection in a follow-up change (a794403602c51c3e7208ded187395f752d5eb58a). It does not preserve shapes, it just moves all selected tiles as a sequence to the target location, in what appears to be the order in which they were selected.

Let's hope we're still on time to help @hepari with some tileset reorganization. The feature is now available in the artifacts resulting from the [Build Packages](https://github.com/mapeditor/tiled/actions/workflows/packages.yml) workflow, and should soon (within a few weeks) be part of a Tiled 1.7 release.

hepari:
I'm excited to hear that it will be part of the next release. Thank you for
your hard work!

PR #2987, opened 2021-02-12
Create export-yy.rst

TrunXGIT:
First draft of the GMS2 documentation. 

Not sure if it makes sense to document all the inherit properties without being able to define also the parentRoom and inheritedItemId properties. (And even then I'm not sure if GMS2s inheritance feature can be used properly in combination with Tiled)

### DIFF
--- a/docs/manual/export-yy.rst
+++ b/docs/manual/export-yy.rst
@@ -1,106 +1,115 @@
 .. _gamemaker2-export:
 
 GameMaker Studio 2.3
---------------------
+====================
 
 GameMaker Studio 2.3 uses a JSON-based format to store its rooms, and Tiled
 ships with a plugin to export maps in this format.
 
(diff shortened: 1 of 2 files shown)


bjorn:
Thanks a lot for your help in documenting (and developing) this plugin!

I've generally reorganized the document and adjusted texts. Please let me know if you find any mistakes or things got confusing, and whether it'd be alright to merge this.

bjorn:
Regarding the freely organized assets hierarchy, I just noticed that actually this appears to be a "virtual" hierarchy. Objects are still always stored at `objects/Object/Object.yy` and tilesets are still always stored at `tilesets/Tileset/Tileset.yy`. It's just that inside the file there is the `parent` property which is used to determine where in the asset browser it is displayed.

I think this means that I can remove the note about the path in the section about references to other assets.

PR #2989, opened 2021-02-21
Update pytiled-parser and arcade description in docs

Cleptomania:
I am a developer on both pytiled-parser and Arcade Python libraries.

Both have just finished undergoing re-writes to only suppport the JSON file format as opposed to TMX, so clarifying that here. At time of writing Arcade does not yet have those changes in a release, but will be included in the next version, which I've made note of in the updated description.


bjorn:
Thanks for maintaining these libraries and for keeping these references up to date!

PR #3004, opened 2021-03-18
Tmxmapreader optimization

Sam-Manflame:
Hello, hello, hello. We've optimized TMXMapReader for multithreaded reading.
For example this code runs about 8(!) times faster with optimizations:

```java
List<String> tmxMaps  = ...;
ExecutorService executor = Executors.newFixedThreadPool(NUMBER_OF_EXECUTORS);
List<Future<Map>> mapReadTasks = new ArrayList<>();

// shared tilesets cache
java.util.Map<String, TileSet> cachedTilesets = new ConcurrentHashMap<>();

for(String tmxMap : tmxMaps)
{
    Future<Map> readTask = executor.submit(() -> {
        TMXMapReader reader = new TMXMapReader();
        reader.setCachedTilesets(cachedTilesets);
        reader.settings.reuseCachedTilesets = true;
        return reader.readMap(tmxMap);
    });
    mapReadTasks.add(readTask);
}
```


util/java/libtiled-java/src/main/java/org/mapeditor/io/TMXMapReader.java
```diff
@@ -83,18 +85,28 @@
...
+     * Map of cached tilesets used when {@link TMXMapReaderSettings#reuseCachedTilesets} option is on
+     * TODO: case when multiple tilesets have same name but different sources
+     */
+    private java.util.Map<String, TileSet> cachedTilesets;
+
+    /**
+     * Keeping static map of unmarshaller pools allows to significantly increase map reading speed
+     * when reading maps in multiple threads
+     */
+    private static final java.util.Map<Class<?>, UnmarshallerPool> cachedUnmarshallers = new ConcurrentHashMap<>();
```
bjorn, line 109: I've been considering whether we could possibly get rid of the `cachedUnmarshallers` hash map, since it is only needed because we create a new `Unmarshaller` for each supported class. However, it should be possible to pass all used types to `JAXBContext.newInstance` and create only a single `Unmarshaller` supporting all types. If we do that we'd only need a single `UnmarshallerPool` in the new code. Would you be interested in trying this out, possibly with an eye on performance as well?

util/java/libtiled-java/src/main/java/org/mapeditor/io/TMXMapReader.java
```diff
@@ -219,10 +235,15 @@ private TileSet unmarshalTilesetFile(InputStream in, URL file) throws Exception
             tsNode = tsNodeList.item(0);
             if (tsNode != null) {
                 set = unmarshalTileset(tsNode);
-                if (set.getSource() != null) {
-                    System.out.println("Recursive external tilesets are not supported.");
+
+                // if using cached tilesets - we should already have correct source
+                if (!settings.reuseCachedTilesets) {
```
bjorn, line 240: If we don't ever assign the source when `reuseCachedTilesets` is true, then how would it already have a correct source?

util/java/libtiled-java/src/main/java/org/mapeditor/io/TMXMapReader.java
```diff
@@ -219,10 +235,15 @@ private TileSet unmarshalTilesetFile(InputStream in, URL file) throws Exception
...
+                    if (set.getSource() != null) {
```
bjorn, line 241: I think something is wrong with this check, even apart from it triggering a false warning in case a cached tileset is returned. If a recursive external tileset really happens, I think this code may never get reached and it would instead recurse indefinitely before.

bjorn (commented):
That's a great idea if you need to load multiple maps at once! :-)

I've left some inline comments for potential further improvement. I'm not doing much Java development anymore in general, so I really appreciate your help with improving libtiled-java!

util/java/libtiled-java/src/main/java/org/mapeditor/io/TMXMapReader.java
Sam-Manflame: yeah, great idea

util/java/libtiled-java/src/main/java/org/mapeditor/io/TMXMapReader.java
Sam-Manflame: my bad, fixed it the way I think this should work

util/java/libtiled-java/src/main/java/org/mapeditor/io/TMXMapReader.java
```diff
@@ -162,11 +175,15 @@ private static double getDoubleAttribute(Node node, String attribname, double de
...
     private <T> T unmarshalClass(Node node, Class<T> type) throws JAXBException {
-        UnmarshallerPool unmarshallerPool = cachedUnmarshallers.get(type);
-        if (unmarshallerPool == null) {
-            JAXBContext context = JAXBContext.newInstance(type);
-            unmarshallerPool = new UnmarshallerPool(context);
-            cachedUnmarshallers.put(type, unmarshallerPool);
+        boolean typeIsBound = classesToBeBound.contains(type);
+        if (!typeIsBound) {
+            System.out.printf("Please add %s to the list of bound classes, " +
+                                  "cause dynamic addition slows down perfomance%n", type.getName());
```
bjorn, line 181: I think a missing type in this private method should always indicate an error in the code. We never actually want the unmarshaller to be re-allocated to add types dynamically. So I wonder if we should use an assertion instead. Something like:

```java
assert classesToBeBound.contains(type) : type.getName();
```

Alternatively, do we really need to trigger our own error at all? What happens by default, if an `Unmarshaller` is asked to unmarshall a type that hasn't been passed to the `JAXBContext` instance? If there is some helpful exception then I think that would suffice.

bjorn (commented):
Did you have a look whether using a single unmarshaller with all types has any impact on the performance? I'd expect it to be negligible, but I'm just curious.

Your solution to the recursion check looks good.

I've left one inline comment regarding the dynamic adding of types to the unmarshaller, which I think shouldn't be necessary.

util/java/libtiled-java/src/main/java/org/mapeditor/io/TMXMapReader.java
Sam-Manflame: Catching original JAXBException and adding a little tip seems good
```java
java.lang.RuntimeException: Please, add class to classesToBeBound set if it is not known to this context
	at org.mapeditor.io.TMXMapReader.unmarshalClass(TMXMapReader.java:189)
	at org.mapeditor.io.TMXMapReader.buildMap(TMXMapReader.java:775)
	at org.mapeditor.io.TMXMapReader.unmarshal(TMXMapReader.java:839)
	at org.mapeditor.io.TMXMapReader.readMap(TMXMapReader.java:864)
	at org.mapeditor.io.TMXMapReader.readMap(TMXMapReader.java:881)
	at org.mapeditor.io.TestRead.lambda$main$0(TestRead.java:42)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Caused by: javax.xml.bind.JAXBException: org.mapeditor.core.Map is not known to this context
	at com.sun.xml.bind.v2.runtime.JAXBContextImpl.getBeanInfo(JAXBContextImpl.java:594)
	at com.sun.xml.bind.v2.runtime.unmarshaller.UnmarshallerImpl.getBeanInfo(UnmarshallerImpl.java:569)
	at com.sun.xml.bind.v2.runtime.unmarshaller.UnmarshallerImpl.unmarshal(UnmarshallerImpl.java:318)
	at org.mapeditor.io.TMXMapReader.unmarshalClass(TMXMapReader.java:187)
	... 9 more
```

Sam-Manflame:
Yeah, actually we can have a single unmarshaller, we just need to create a pool of TMXMapReaders when reading maps in parallel instead of unmarshallers pool :smiley:

util/java/libtiled-java/src/main/java/org/mapeditor/io/TMXMapReader.java
```diff
@@ -83,18 +87,39 @@
...
-            | FLIPPED_VERTICALLY_FLAG
-            | FLIPPED_DIAGONALLY_FLAG;
+    public static final long ALL_FLAGS =
+        FLIPPED_HORIZONTALLY_FLAG | FLIPPED_VERTICALLY_FLAG | FLIPPED_DIAGONALLY_FLAG;
+
+    /**
+     * Set of classes to be bound to the {@link JAXBContext}.
+     * @see #unmarshaller
+     */
+    private static final Set<Class<?>> classesToBeBound = Collections.synchronizedSet(new HashSet<>());
```
bjorn, line 97: We can now also avoid the set entirely, right? Now we create set from array and then array again in constructor. Or the list of classes could even be inlined in the constructor.

util/java/libtiled-java/src/main/java/org/mapeditor/io/TMXMapReader.java
```diff
@@ -104,7 +129,13 @@
     /**
      * Constructor for TMXMapReader.
      */
-    public TMXMapReader() {
+    public TMXMapReader()
+    {
+        try {
+            unmarshaller = JAXBContext.newInstance(classesToBeBound.toArray(new Class[0])).createUnmarshaller();
+        } catch (JAXBException e) {
```
bjorn, line 136: When can this happen and what do we gain from catching this exception?

bjorn (commented):
I have two remaining questions about this patch.

util/java/libtiled-java/src/main/java/org/mapeditor/io/TMXMapReader.java
Sam-Manflame: yep

util/java/libtiled-java/src/main/java/org/mapeditor/io/TMXMapReader.java
Sam-Manflame: actually it would be better to make constructor throw exception :thinking:

util/java/libtiled-java/src/main/java/org/mapeditor/io/TMXMapReader.java
```diff
@@ -104,7 +115,10 @@
     /**
      * Constructor for TMXMapReader.
      */
-    public TMXMapReader() {
+    public TMXMapReader() throws JAXBException {
+            unmarshaller = JAXBContext.newInstance(
```
bjorn, line 119: Indentation is a bit off here.

util/java/libtiled-java/src/main/java/org/mapeditor/io/TMXMapReader.java
```diff
@@ -150,13 +164,14 @@ private static double getDoubleAttribute(Node node, String attribname, double de
...
-        if (unmarshaller == null) {
-            JAXBContext context = JAXBContext.newInstance(type);
-            unmarshaller = context.createUnmarshaller();
-            cachedUnmarshallers.put(type, unmarshaller);
+        JAXBElement<T> element;
+        try {
+            element = unmarshaller.unmarshal(node, type);
+        } catch (JAXBException e) {
+            throw new RuntimeException(
+                "Please, add class to classesToBeBound set if it is not known to this context", e);
```
bjorn, line 172: This exception still refers to the no longer existing `classesToBeBound`.

bjorn (commented):
I guess we almost nailed this one, but I found two small issues. :-)

bjorn:
That's great, thanks!

One thing I was wondering about in your original code snippet, is that likely the tilesets will get loaded multiple times. If you load 8 maps at the same time, each using the same tileset, they will each not find the tileset cached and start loading it, right? In this case I normally see some API with a read-write lock that is used to block the other threads while one creates and inserts the cached value, but I'm not sure how this could be implemented here.

Alternatively, the whole thing could be refactored, such that we can first load the maps and then create additional tasks for loading the referenced tilesets. Then of course we can't use direct references to `Tile` instances or at least would need intermediate storing of the tile IDs.

Sam-Manflame:
Yes, such a problem does exist. I think, one of the easiest ways to solve it looks like this:
```diff
diff --git a/util/java/libtiled-java/src/main/java/org/mapeditor/io/TMXMapReader.java b/util/java/libtiled-java/src/main/java/org/mapeditor/io/TMXMapReader.java
index 3a1204ce..4371b012 100644
--- a/util/java/libtiled-java/src/main/java/org/mapeditor/io/TMXMapReader.java
+++ b/util/java/libtiled-java/src/main/java/org/mapeditor/io/TMXMapReader.java
@@ -47,13 +47,15 @@ import java.util.HashMap;
 import java.util.Map.Entry;
 import java.util.StringTokenizer;
 import java.util.TreeMap;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 import java.util.zip.GZIPInputStream;
 import java.util.zip.InflaterInputStream;
 
 import javax.imageio.ImageIO;
 import javax.xml.bind.DatatypeConverter;
 import javax.xml.bind.JAXBContext;
-import javax.xml.bind.JAXBElement;
 import javax.xml.bind.JAXBException;
 import javax.xml.bind.Unmarshaller;
 import javax.xml.parsers.DocumentBuilder;
@@ -99,7 +101,7 @@ public class TMXMapReader {
      * TODO: case when multiple tilesets have same name but different sources
      * @see #setCachedTilesets(java.util.Map)
      */
-    private java.util.Map<String, TileSet> cachedTilesets;
+    private java.util.Map<String, Future<TileSet>> cachedTilesets;
 
     /**
      * Unmarshaller capable of unmarshalling all classes available from context
@@ -274,92 +276,98 @@ public class TMXMapReader {
                 return ext;
             }
         } else {
-            final int tileWidth = getAttribute(t, "tilewidth", map != null ? map.getTileWidth() : 0);
-            final int tileHeight = getAttribute(t, "tileheight", map != null ? map.getTileHeight() : 0);
-            final int tileSpacing = getAttribute(t, "spacing", 0);
-            final int tileMargin = getAttribute(t, "margin", 0);
-
             final String name = getAttributeValue(t, "name");
 
             if (settings.reuseCachedTilesets) {
                 if (cachedTilesets == null) {
                     cachedTilesets = new HashMap<>();
                 }
-
-                set = cachedTilesets.get(name);
-                if (set != null)
-                    return set;
-
-                set = new TileSet();
-                cachedTilesets.put(name, set);
+                return cachedTilesets.computeIfAbsent(name, (_n) -> {
+                    ExecutorService es = Executors.newSingleThreadExecutor();
+                    Future<TileSet> fts = es.submit(() -> processTileset(t, new TileSet(), name));
+                    es.shutdown();
+                    return fts;
+                }).get();
             } else {
                 set = new TileSet();
             }
 
-            set.setName(name);
+            processTileset(t, new TileSet(), name);
 
-            boolean hasTilesetImage = false;
-            NodeList children = t.getChildNodes();
+            return set;
+        }
+    }
 
-            for (int i = 0; i < children.getLength(); i++) {
-                Node child = children.item(i);
+    private TileSet processTileset(Node t, TileSet set, String name) throws Exception
+    {
+        final int tileWidth = getAttribute(t, "tilewidth", map != null ? map.getTileWidth() : 0);
+        final int tileHeight = getAttribute(t, "tileheight", map != null ? map.getTileHeight() : 0);
+        final int tileSpacing = getAttribute(t, "spacing", 0);
+        final int tileMargin = getAttribute(t, "margin", 0);
 
-                if (child.getNodeName().equalsIgnoreCase("image")) {
-                    if (hasTilesetImage) {
-                        System.out.println("Ignoring illegal image element after tileset image.");
-                        continue;
-                    }
+        set.setName(name);
 
-                    String imgSource = getAttributeValue(child, "source");
-                    String transStr = getAttributeValue(child, "trans");
+        boolean hasTilesetImage = false;
+        NodeList children = t.getChildNodes();
 
-                    if (imgSource != null) {
-                        // Not a shared image, but an entire set in one image
-                        // file. There should be only one image element in this
-                        // case.
-                        hasTilesetImage = true;
+        for (int i = 0; i < children.getLength(); i++) {
+            Node child = children.item(i);
 
-                        URL sourcePath;
-                        if (!new File(imgSource).isAbsolute()) {
-                            imgSource = replacePathSeparator(imgSource);
-                            sourcePath = URLHelper.resolve(xmlPath, imgSource);
-                        } else {
-                            sourcePath = makeUrl(imgSource);
-                        }
+            if (child.getNodeName().equalsIgnoreCase("image")) {
+                if (hasTilesetImage) {
+                    System.out.println("Ignoring illegal image element after tileset image.");
+                    continue;
+                }
 
-                        if (transStr != null) {
-                            if (transStr.startsWith("#")) {
-                                transStr = transStr.substring(1);
-                            }
+                String imgSource = getAttributeValue(child, "source");
+                String transStr = getAttributeValue(child, "trans");
 
-                            int colorInt = Integer.parseInt(transStr, 16);
-                            Color color = new Color(colorInt);
-                            set.setTransparentColor(color);
-                        }
+                if (imgSource != null) {
+                    // Not a shared image, but an entire set in one image
+                    // file. There should be only one image element in this
+                    // case.
+                    hasTilesetImage = true;
 
-                        set.importTileBitmap(sourcePath, new BasicTileCutter(
-                                tileWidth, tileHeight, tileSpacing, tileMargin));
-                    }
-                } else if (child.getNodeName().equalsIgnoreCase("tile")) {
-                    Tile tile = unmarshalTile(set, child, xmlPath);
-                    if (!hasTilesetImage || tile.getId() > set.getMaxTileId()) {
-                        set.addTile(tile);
+                    URL sourcePath;
+                    if (!new File(imgSource).isAbsolute()) {
+                        imgSource = replacePathSeparator(imgSource);
+                        sourcePath = URLHelper.resolve(xmlPath, imgSource);
                     } else {
-                        Tile myTile = set.getTile(tile.getId());
-                        myTile.setProperties(tile.getProperties());
-                        //TODO: there is the possibility here of overlaying images,
-                        //      which some people may want
+                        sourcePath = makeUrl(imgSource);
+                    }
+
+                    if (transStr != null) {
+                        if (transStr.startsWith("#")) {
+                            transStr = transStr.substring(1);
+                        }
+
+                        int colorInt = Integer.parseInt(transStr, 16);
+                        Color color = new Color(colorInt);
+                        set.setTransparentColor(color);
                     }
-                } else if (child.getNodeName().equalsIgnoreCase("tileoffset")) {
-                   TileOffset tileoffset = new TileOffset();
-                   tileoffset.setX(Integer.valueOf(getAttributeValue(child, "x")));
-                   tileoffset.setY(Integer.valueOf(getAttributeValue(child, "y")));
-                   set.setTileoffset(tileoffset);
+
+                    set.importTileBitmap(sourcePath, new BasicTileCutter(
+                        tileWidth, tileHeight, tileSpacing, tileMargin));
+                }
+            } else if (child.getNodeName().equalsIgnoreCase("tile")) {
+                Tile tile = unmarshalTile(set, child, xmlPath);
+                if (!hasTilesetImage || tile.getId() > set.getMaxTileId()) {
+                    set.addTile(tile);
+                } else {
+                    Tile myTile = set.getTile(tile.getId());
+                    myTile.setProperties(tile.getProperties());
+                    //TODO: there is the possibility here of overlaying images,
+                    //      which some people may want
                 }
+            } else if (child.getNodeName().equalsIgnoreCase("tileoffset")) {
+                TileOffset tileoffset = new TileOffset();
+                tileoffset.setX(Integer.valueOf(getAttributeValue(child, "x")));
+                tileoffset.setY(Integer.valueOf(getAttributeValue(child, "y")));
+                set.setTileoffset(tileoffset);
             }
-
-            return set;
         }
+
+        return set;
     }
 
     private MapObject readMapObject(Node t) throws Exception {
@@ -988,7 +996,7 @@ public class TMXMapReader {
      * The ability to set cachedTilesets allows cached tilesets to be shared across multiple readers,
      * increasing read speed in a multithreaded environment.
      */
-    public void setCachedTilesets(java.util.Map<String, TileSet> cachedTilesets) {
+    public void setCachedTilesets(java.util.Map<String, Future<TileSet>> cachedTilesets) {
         this.cachedTilesets = cachedTilesets;
     }
 }
```

bjorn:
@Sam-Manflame That doesn't look so bad! It's a bit unfortunate that we need the `Future` as value in the cache, but I understand that if it would just be the `TileSet` and we'd use `computeIfAbsent` then we could only load a single tileset at a time.

I'd accept an API change, but I think we could design it a little differently so we avoid the whole threading and futures in the default implementation. We could provide just an interface for the "TilesetCache" which just needs a function to obtain a tileset. Then instead of `setCachedTilesets` and  `settings.reuseCachedTilesets = true` we'd just do `setTilesetCache(...)`, passing an implementation of that interface that internally has the `java.util.Map<String, Future<TileSet>>`.

PR #3006, opened 2021-03-23
Update chinese translation tiled_zh_CN.ts

kuaima:
fix some translation for better understand

- line 1352, in menu/message, new -> 新建，instead of 创建
- line 906， In menu, world -> 世界，instead of 地图
- line 4126,  add missing translation
- line 4177,  add missing translation


bjorn:
Thanks, that seems very helpful indeed!

PR #3008, opened 2021-03-23
Allow building tiledquick by default

mitchcurtis:
... by setting the BUILD_TILEDQUICK env var to true.


bjorn:
Seems like a nice change to avoid some manual setup for every configuration, thanks!

l3aoNgoc:
Hghj

PR #3011, opened 2021-03-24
Add Fit Map in View action/menu item

mitchcurtis:
Please merge as individual commits.

### DIFF
--- a/src/tiledquick/qml/main.qml
+++ b/src/tiledquick/qml/main.qml
@@ -91,9 +91,12 @@ ApplicationWindow {
             // fits both vertically and horizontally.
             // If we need to upscale, we also want to choose the smaller ratio, as we want
             // both to fit.
-            mapContainer.scale = Math.max(widthRatio, heightRatio)
-            mapContainer.x = (mapView.width / 2) - ((mapItem.width * mapContainer.scale) / 2)
-            mapContainer.y = (mapView.height / 2) - ((mapItem.height * mapContainer.scale) / 2)
+            let scale = Math.min(widthRatio, heightRatio)
+            containerAnimation.stop()
+            containerAnimation.scale = scale
+            containerAnimation.x = (mapView.width / 2) - ((mapItem.width * scale) / 2)
+            containerAnimation.y = (mapView.height / 2) - ((mapItem.height * scale) / 2)
+            containerAnimation.start()
         }
     }
 
--- a/src/tiledquick/qml/main.qml
+++ b/src/tiledquick/qml/main.qml
@@ -59,6 +59,7 @@ ApplicationWindow {
         }
     }
 
+    // File
     Action {
         id: openAction
         text: qsTr("Open...")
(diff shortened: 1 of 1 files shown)


mitchcurtis:
I added that comment since I figure it's easier to just use one branch if I'm gonna be making a bunch of unrelated commits, rather than make one branch for each one. Still not sure how to base a branch on another (unmerged) one. Doesn't seem to be possible with GitHub.

src/tiledquick/qml/main.qml
```diff
@@ -109,11 +138,7 @@ ApplicationWindow {
 
     Item {
         id: mapView
-
-        scale: 1 / Screen.devicePixelRatio
-        width: parent.width * Screen.devicePixelRatio
-        height: parent.height * Screen.devicePixelRatio
```
bjorn, line 115: Actually, this code is necessary precisely because Qt Quick is scaling. It undoes that scaling for the map view, where I explicitly want 1 pixel to be 1 pixel at 100% zoom. We don't want to scale up the user's graphics, just the UI.

src/tiledquick/qml/main.qml
```diff
@@ -76,13 +77,41 @@ ApplicationWindow {
...
+            let heightRatio = mapView.height / mapItem.height
+            // If we need to downscale the map to fit in the view, choose the lesser ratio
+            // because that will result in the largest downscaling, which ensures the map
+            // fits both vertically and horizontally.
+            // If we need to upscale, we also want to choose the smaller ratio, as we want
+            // both to fit.
+            if (widthRatio <= heightRatio)
+                mapContainer.scale = widthRatio
+            else
+                mapContainer.scale = heightRatio
```
bjorn, line 97: Maybe we could just do `mapContainer.scale = Math.min(widthRatio, heightRatio)`. :-)

src/tiledquick/qml/main.qml
mitchcurtis: Ah right. I forgot about that.

src/tiledquick/qml/main.qml
mitchcurtis: Yep, that would be better. I initially had more more than one line in each if statement...

src/tiledquick/qml/main.qml
```diff
@@ -91,9 +91,12 @@ ApplicationWindow {
             // fits both vertically and horizontally.
             // If we need to upscale, we also want to choose the smaller ratio, as we want
             // both to fit.
-            mapContainer.scale = Math.max(widthRatio, heightRatio)
```
mitchcurtis: In my defense... it was 10:30pm. :D

bjorn:
> Still not sure how to base a branch on another (unmerged) one. Doesn't seem to be possible with GitHub.

If this is really necessary, I guess the way to do it would be to just create another branch `bar` based on `foo` and open another PR with `bar`. The commits from `foo` will also show up in the pull request for `bar` of course, but `foo` can still be merged independently and `bar` can be rebased once `foo` is merged.

In practice I think such a case is incredibly rare. Most of the time changes that are worth separate pull requests don't depend on each other. Sometimes there are conflicts, but even this is not a reason to have a dependency (either change will just have to be rebased once the other is merged).

bjorn:
Thanks for this nice addition!

PR #3030, opened 2021-04-19
[Testers needed] Components

edin-m:
Add the ability to use object types as components on an object.

Implemented:

- [x] Add/remove components in Property Dock/Browser
- [x] Use Shift to add/remove multiple components in sequence
- [x] Undo/redo adding components
- [x] Change component's properties
- [x] Undo/redo changing component's properties
- [x] Only map objects supported for now
- [x] Save/Load components to the map file
- [x] History view change component property label
- [x] Being able to load & remove object components that are not defined as object types
- [x] Consolidate add/remove component menus into one
- [x] Accidentally remove the component, undo the action, modified component properties should be correct
- [x] Add support for objects in the collision editor
- [x] Support for tileset tile collision objects
- [x] Component properties of an object type cannot select objects

Future work (in sub-branches):
- Support tile objects and inheritance?
- Scripting/API support
- Support object templates
- Remember the order of components
- A different tool button menu icon
- Make components menu not flicker when holding shift
- Add components to multiple objects
- Modify components for multiple objects
- Add ObjectReferenceItem support
- When adding a component add it into an alphabetically sorted place in the property browser
- Review & Test this & all sub-features
- ~Use the object type color as the property browser bar color (cannot be implemented)~

PR status:
- [x] Initial review
- [x] Working on TODOs
- [x] Code style check
- [x] Fine-tuning
- [x] Testing

### DIFF
--- a/src/libtiled/mapobject.cpp
+++ b/src/libtiled/mapobject.cpp
@@ -361,6 +361,7 @@ MapObject *MapObject::clone() const
     MapObject *o = new MapObject(mName, mType, mPos, mSize);
     o->setId(mId);
     o->setProperties(properties());
+    o->setComponents(components());
     o->setTextData(mTextData);
     o->setPolygon(mPolygon);
     o->setShape(mShape);
@@ -384,6 +385,7 @@ void MapObject::copyPropertiesFrom(const MapObject *object)
     setRotation(object->rotation());
     setVisible(object->isVisible());
     setProperties(object->properties());
+    setComponents(object->components());
     setChangedProperties(object->changedProperties());
     setObjectTemplate(object->objectTemplate());
 }
--- a/src/libtiled/mapreader.cpp
+++ b/src/libtiled/mapreader.cpp
@@ -1243,8 +1243,7 @@ std::unique_ptr<MapObject> MapReaderPrivate::readObject()
             object->setShape(MapObject::Point);
             object->setPropertyChanged(MapObject::ShapeProperty);
         } else if (xml.name() == QLatin1String("components")) {
-            Components c = readComponents();
-            object->setComponents(c);
+            object->setComponents(readComponents());
         } else {
             readUnknownElement();
         }
@@ -1461,37 +1460,33 @@ void MapReaderPrivate::readProperty(Properties *properties)
 
 Components MapReaderPrivate::readComponents()
 {
-  Q_ASSERT(xml.isStartElement() && xml.name() == QLatin1String("components"));
+    Q_ASSERT(xml.isStartElement() && xml.name() == QLatin1String("components"));
 
-  Components commponents;
+    Components commponents;
 
-  while (xml.readNextStartElement()) {
-    if (xml.name() == QLatin1String("component"))
-      readComponent(&commponents);
-    else
-      readUnknownElement();
-  }
+    while (xml.readNextStartElement()) {
+        if (xml.name() == QLatin1String("component"))
+            readComponent(&commponents);
+        else
+            readUnknownElement();
+    }
 
-  return commponents;
+    return commponents;
 }
 
 void MapReaderPrivate::readComponent(Components *components)
 {
-  Q_ASSERT(xml.isStartElement() && xml.name() == QLatin1String("component"));
+    Q_ASSERT(xml.isStartElement() && xml.name() == QLatin1String("component"));
 
-  const QXmlStreamAttributes atts = xml.attributes();
-  QString componentName = atts.value(QLatin1String("name")).toString();
+    const QXmlStreamAttributes atts = xml.attributes();
+    QString componentName = atts.value(QLatin1String("name")).toString();
 
-  while (xml.readNext() != QXmlStreamReader::Invalid) {
-    if (xml.isEndElement()) {
-      break;
-    } else if (xml.name() == QLatin1String("properties")) {
-      Properties props = readProperties();
-      components->insert(componentName, props);
-    } else if (xml.isStartElement()) {
-      readUnknownElement();
+    while (xml.readNextStartElement()) {
+        if (xml.name() == QLatin1String("properties"))
+            components->insert(componentName, readProperties());
+        else
+            readUnknownElement();
     }
-  }
 }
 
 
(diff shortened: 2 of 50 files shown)


src/libtiled/mapreader.cpp
```diff
@@ -1453,6 +1459,42 @@ void MapReaderPrivate::readProperty(Properties *properties)
...
+}
+
+void MapReaderPrivate::readComponent(Components *components)
+{
+  Q_ASSERT(xml.isStartElement() && xml.name() == QLatin1String("component"));
+
+  const QXmlStreamAttributes atts = xml.attributes();
+  QString componentName = atts.value(QLatin1String("name")).toString();
+
+  while (xml.readNext() != QXmlStreamReader::Invalid) {
```
bjorn, line 1485: Why not `while (xml.readNextStartElement()) {` here as well?

src/libtiled/mapobject.cpp
```diff
@@ -358,9 +358,12 @@ void MapObject::flip(FlipDirection direction, const QPointF &origin)
  */
 MapObject *MapObject::clone() const
 {
+    // TODO should I clone components
+    // how are custom properties cloned
```
bjorn, line 362: Properties, and also Components as you've introduced them, are referenced by value and cloning is just done by making a copy, as you've done here.

Of course, Qt containers (like `QMap`) use implicit sharing, which means a copy just increments a reference count, and a real deep copy only happens upon modification. But regarding the cloning that's just an implementation detail (though a useful one, since it makes the cloning cheaper when the clone is never modified).

bjorn:
I'd really appreciate if you'd already make an effort to follow the coding style, in particular to use 4-space indentation (I haven't actually noticed other style differences). Please don't wait with reformatting the code till the end, given the size of the work you're doing here.

src/libtiled/object.h
```diff
@@ -132,13 +132,31 @@ class TILEDSHARED_EXPORT Object
...
+    { return mComponentProperties[componentName]; }
+
     static void setObjectTypes(const ObjectTypes &objectTypes);
     static const ObjectTypes &objectTypes()
     { return mObjectTypes; }
 
 private:
     const TypeId mTypeId;
     Properties mProperties;
+    Components mComponentProperties;
```
edin-m, line 159: TODO: rename to `mComponents`

src/libtiled/mapreader.cpp
edin-m: I mirrored the behaviour of `readProperty()`.

bjorn: Ah, but `readProperty` can't use `readNextStartElement` because it needs to also handle the `xml.isCharacters` case.

src/libtiled/object.cpp
```diff
@@ -145,24 +142,19 @@ QVariantMap Object::resolvedProperties() const
...
-    Properties &props = mComponentProperties[name];
 
-    QMapIterator<QString, QVariant> it(type->defaultProperties);
-    while (it.hasNext()) {
-      it.next();
-      props.insert(it.key(), it.value());
+    if (type) {
+        mComponents[name] = Properties {};
+        Properties &props = mComponents[name];
+        props = type->defaultProperties;
```
bjorn, line 156: This would be the same as `mComponents[name] = type->defaultProperties`. :-)

src/tiled/tilesetdocument.cpp
```diff
@@ -76,6 +76,8 @@ TilesetDocument::TilesetDocument(const SharedTileset &tileset)
     connect(this, &TilesetDocument::propertiesChanged,
             this, &TilesetDocument::onPropertiesChanged);
 
+    // TODO: add components?
```
bjorn, line 79: Properties are handled here because of tile property inheritance. I think components will only need to be added when they are also inherited.

src/tiled/tiled.qbs
```diff
@@ -131,6 +131,8 @@ QtGuiApplication {
         "bucketfilltool.h",
         "capturestamphelper.cpp",
         "capturestamphelper.h",
+        "changecomponents.cpp",
+        "changecomponents.h",
```
bjorn, line 135: New files should also be added to the `src/tiled/tiled.pro` for now, until we can finally drop the qmake project.

bjorn (changes requested):
I love this feature, thanks for your hard work on this!

I've pushed a number of code style tweaks and minor changes, and left a few comments that I think are worth addressing before merging this first change.

I've also set up a new `wip/components` branch, so we can change the target branch for this PR to that.

bjorn:
I forgot to put in the code review what I said on Discord, that I think one `VariantPropertyManager` should suffice.

bjorn:
Alright, I think I'm done reviewing and tweaking this change. Please let me know what you think about merging this to the dev branch.

bjorn (approved):

src/tiled/propertybrowser.cpp
```diff
@@ -549,7 +548,7 @@ void PropertyBrowser::propertyChanged(Object *object, const QString &name)
 void PropertyBrowser::propertiesChanged(Object *object)
 {
     if (objectPropertiesRelevant(mDocument, object))
-      updateCustomProperties();
+        updateCustomProperties();
 }
```
edin-m, line 552: argh :)

edin-m:
> Alright, I think I'm done reviewing and tweaking this change. Please let me know what you think about merging this to the WIP branch.

Looks good. I missed out on a couple of key things. Thanks for the updates.
I think this is a good base for any future work.

bjorn:
Rebasing the changes on latest master before I had updated the `wip/components` branch might not have been such a good idea, because now this PR is a bit messed with all the changes that had been made in the meantime. Something to keep in mind for next time...

But, the state in the repository looks fine to me at least. Please hard-reset your local branch to the new `wip/components` branch to continue this work.

edin-m:
> Rebasing the changes on latest master before I had updated the `wip/components` branch might not have been such a good idea, because now this PR is a bit messed with all the changes that had been made in the meantime. Something to keep in mind for next time...

Maybe wip/3029 should have been rebased onto wip/components after it was rebased onto master to have a clean PR. 

> But, the state in the repository looks fine to me at least. Please hard-reset your local branch to the new `wip/components` branch to continue this work.

Will do. I'll branch off of this one.

PR #3039, opened 2021-04-27
Add a way to reliably disable DBus support

AMDmi3:


bjorn:
Thanks @AMDmi3! May I ask, why do you need to disable DBUS support?

AMDmi3:
> May I ask, why do you need to disable DBUS support?

- I don't use it
- I don't want software to (be able to) communicate in random unexpected ways

> Also, this should probably be added as an option to the Qbs project files as well then:

Unfortunately I'm not familiar with qbs syntax. Again I suggest to use single build system to avoid discrepancies.

bjorn:
> Again I suggest to use single build system to avoid discrepancies.

I'm trying hard to drop the qmake files since years. Now they're only still there to be able to build the snap package, since snapcraft has a qmake plugin but not a Qbs plugin. But eventually I'll figure out how to use Qbs to build the snap. :-)

I'll add the option to the Qbs project.

AMDmi3:
I'd strongly suggest the opposite - to drop qbs, discontinued and widely unsupported system. Most packages (see recipes list on https://repology.org/project/tiled/information) use qmake: rosa, opensuse, freebsd, solus, alpine, guix, gentoo, macports, nix, alt, fedora (only arch seem to use qbs). Some of these have infrastructure support for qmake, but not qbs:

https://github.com/NixOS/nixpkgs/blob/093c7395732496848b7c89eb0210151a8744ed92/pkgs/applications/editors/tiled/default.nix#L1
https://github.com/macports/macports-ports/blob/e924f19761a6fadcb96580156228306306d6a8bc/devel/tiled/Portfile#L4

Dropping qmake would complicate things for a lot of downstream package maintainers.

AMDmi3:
Also note that a lot of repositores which package tiled don't even package qbs:

bjorn:
Yeah, regarding infrastructure support I'm experiencing that first-hand with the snap package. But Qbs is actually still in active development and it's a great tool, as opposed to qmake which is a royal pain to get anything done with.

I think your list is misleading because it contains many ancient versions of distributions that actually do package Qbs today. Also, if Qbs is not packaged separately, it will likely still be available as part of the Qt Creator package.

The reason all but Arch Linux use qmake is just for historical reasons. They never had a reason to switch. But I have a lot of reasons to drop qmake in favour of Qbs. Also the Qt company has pretty much deprecated qmake in favour of cmake, so if I'd drop Qbs it would be in favour of cmake and not qmake.

AMDmi3:
Swithing to CMake would be the best solution. But leaving no alternative to qbs would hurt the project IMO.

PR #3040, opened 2021-04-28
Added a bunch of definitions, still not all. But I was missing these.

MrMasterplan:
refined TiledObject, MapObject, TileMap
added Font, TileLayer, TileLayerEdit, Tileset, Tile
added declarton of WangSet, color, SelectedArea

I tested my added declarations by using them in another project that needs them.

Change-Id: Ice5a64696293de467abc064b16e0096c7224c825

### DIFF
--- a/docs/scripting-doc/index.d.ts
+++ b/docs/scripting-doc/index.d.ts
@@ -52,6 +52,8 @@ interface ObjectRef {
     id: number;
 }
 
+
+
 interface Menu {
   /**
    * ID of a registered action that the menu item will represent.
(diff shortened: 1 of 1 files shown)


bjorn:
Wow, thanks a lot! I didn't realize so much was still missing. I had a quick look over it and it all seems fine.

> Change-Id: Ice5a64696293de467abc064b16e0096c7224c825

I assume this line is irrelevant, and possibly an artifact of using Gerrit somewhere?

MrMasterplan:
Yes, I think you are right. An artefact of the way my IDE is set up. just ignore it.

PR #3041, opened 2021-04-28
Cleanup for strict, some fixes, Process class.

MrMasterplan:
List of changes in this file:
- refine Menu item type
- fix BinaryFile interface
- untangle tileToPixel definitons
- add a few any types to allow compilition in strict mode
- add process class

I use this file in my project https://github.com/MrMasterplan/rwklevelfiles and can report that I can transpile with these changes.

### DIFF
--- a/docs/scripting-doc/index.d.ts
+++ b/docs/scripting-doc/index.d.ts
@@ -2264,7 +2264,7 @@ declare class Process {
 
   /**
    *   Executes the program at filePath with the given argument list and blocks until the process is finished. If an error occurs (for example, there is no executable file at filePath) and throwOnError is true (the default), then a JavaScript exception will be thrown. Otherwise, -1 will be returned in case of an error. The normal return code is the exit code of the process.
-    * @param filePath
+   * @param filePath
    * @param arguments
    * @param throwOnError
    */
--- a/docs/scripting-doc/index.d.ts
+++ b/docs/scripting-doc/index.d.ts
@@ -1,4 +1,4 @@
-declare const __filename: string;
+// declare const __filename: string; // collides with nodejs types
 
 interface rect {
   /**
@@ -53,8 +53,7 @@ interface ObjectRef {
 }
 
 
-
-interface Menu {
+interface MenuAction {
   /**
    * ID of a registered action that the menu item will represent.
    */
@@ -64,13 +63,17 @@ interface Menu {
    * ID of the action before which this menu item should be added
    * (optional).
    */
-  before: string;
+  before?: string;
+}
 
+interface MenuSeparator {
   /**
    * Set to `true` if this item is a menu separator (optional).
    */
   separator: boolean;
 }
+
+type Menu = MenuAction|MenuSeparator
 
 interface FilePath {
   url: string;
@@ -243,12 +246,17 @@ declare class BinaryFile {
   public readAll(): ArrayBuffer;
 
   /**
-   * Commits all written data to disk and closes the file. Should be called when writing to files
-   * in WriteOnly mode. Failing to call this function will result in cancelling the operation,
-   * unless safe writing to files is disabled.
+   * Writes data into the file at the current position.
    * @param data
    */
   public write(data: ArrayBuffer): void;
+
+  /**
+   * Commits all written data to disk and closes the file. Should be called when writing to files
+   * in WriteOnly mode. Failing to call this function will result in cancelling the operation,
+   * unless safe writing to files is disabled.
+   */
+  public commit(): void;
 
   /**
    * Closes the file. It is recommended to always call this function as soon as you are finished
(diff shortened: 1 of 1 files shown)


bjorn (commented):
That's great, thanks for your attention to detail!

> I use this file in my project https://github.com/MrMasterplan/rwklevelfiles and can report that I can transpile with these changes.

I just read about it and took a short look at the game. What a great idea to make it possible to use Tiled to edit those levels! The built-in level editor is so limiting that I wonder what the developers of the game have used, actually.

What does it mean to "transpile" with these changes? You mean your TypeScript code can be translated to JavaScript code to be executed by Tiled?

MrMasterplan:
Yes, sorry I thought that was standard language. Exactly. The typescript compiler can convert my entire project to .js files. I actually use webpack to crunch my entire extension down into a single .js file which is loaded by Tiled, including new file type, a custom tool, and a menu item. I am still writing documentation, so in the future the usage will be clearer. 

It was a really fun project since I reverse engineered the binary file format with zero documentation. I spent a lot of time staring at hex dumps. :)

MrMasterplan:
The code that interacts with Tiled is in these files in case you want to have a look at how I wrote my extensions:

https://github.com/MrMasterplan/rwklevelfiles/blob/rewrite/src/dotkitty/index.ts
https://github.com/MrMasterplan/rwklevelfiles/blob/rewrite/src/kittypaint/index.ts
https://github.com/MrMasterplan/rwklevelfiles/blob/rewrite/src/kittymenu/index.ts

bjorn:
> Yes, sorry I thought that was standard language.

I'm sure it is, but I'm not very familiar with TypeScript. :-)

> The code that interacts with Tiled is in these files in case you want to have a look at how I wrote my extensions:

It looks like the kittypaint tool is doing something very similar to the Terrain Brush. Was there some reason that you couldn't rely on that existing tool?

bjorn:
@MrMasterplan Btw, regarding `getExtBasePath`, couldn't you rely on `__filename`? I'd also recommend using a `.tiled-project` with its `extensions` directory rather than installing the extension globally.

MrMasterplan:
Thank you for your feedback on my code!!
Regarding `__filename`. I tried debug-logging that and it just comes out as `/index.js` which is less than useless. My extension is actually packaged as `kitty.js` maybe it's something to do with webpack.

I'll look into the use of `.tiled-project`. What I like about the global install is that I can know the path of the tilesets in advance. `.kitty` files don't contain or reference tilesets. I open default tilesets on loading the `.kitty` file (https://github.com/MrMasterplan/rwklevelfiles/blob/rewrite/src/dotkitty/index.ts#L30) and match cell byte-values to tiles from there. In a local installation I'm not sure how I could point at the tilesets for opening.

The Terran tool, I have to say I overlooked it. I can see now that yes, `kittypaint` does the same thing. It could have been useful to find the literature on the blob tiles before I ended up figuring it all out by myself.  There is just one thing about the terrain tool that I couldn't figure out right now: how can you mark a tile as being terrain itself while surrounded by emptiness (only the middle is set)? This last part is needed if you want to paint individual tiles like the default images for the paint styles here: http://robotwantskitty.com/tutorial/images/tutor5.jpg In my attempt right now I couldn't set it and then the terrain painter could never paint less than two tiles at a time.

MrMasterplan:
btw. I sent you a linkedIn invite.

bjorn:
> I tried debug-logging that and it just comes out as `/index.js` which is less than useless. My extension is actually packaged as `kitty.js` maybe it's something to do with webpack.

Hmm, yeah sounds like there might be some interference. Tiled sets it to the absolute path of the file it is evaluating. Did you have a look at whether it might be getting modified inside of the generated script?

> What I like about the global install is that I can know the path of the tilesets in advance. `.kitty` files don't contain or reference tilesets.

Does it work when you use `ext:rwk` as the base path, or even just `ext:`? This might work because the extensions directory, as well as the directory of each extension, are generally made available with this "ext:" prefix using `QDir::setSearchPaths`.

> how can you mark a tile as being terrain itself while surrounded by emptiness (only the middle is set)?

This is not possible at the moment and one of the things that would indeed still be nice to handle, but I'm not entirely sure how.

MrMasterplan:
> This is not possible at the moment and one of the things that would indeed still be nice to handle, but I'm not entirely sure how.
I can see that all the literature that you reference from the Terrain set documentation also does not consider the lonely tile to be valid. I wonder why!? Well for now I will stick with my kittypaint tool.

`ext:` works! That is great. I was also able to fix `__filename` in webpack. 

One more question. What are the conditions to make the custom tool icon work? Setting the icon property on the tool changes the tool button to gray, but does not display my icon. Does it have to be a certain format or pixel size? I tried with one of my tiles which is a 40x40 pixel `.png`.

bjorn:
> One more question. What are the conditions to make the custom tool icon work?

Hmm, should be no special conditions. A 40x40 pixel PNG should work fine, and relying on the `ext:` prefix. The https://github.com/mapeditor/tiled-extensions repository features some icons. Do they work for you? (they're SVG, but that shouldn't matter).

PR #3042, opened 2021-04-29
docs: Extended TypeScript API definitions

bjorn:
@MrMasterplan, I noticed there was a "TiledImage" class and namespace being defined, but actually Tiled registers this class as just `Image`:
```c++
globalObject.setProperty(QStringLiteral("Image"), mEngine->newQMetaObject<ScriptImage>());
```
It could be that "TiledImage" was chosen on purpose, so I've opened a PR for this. Do you see any issues with this change?

### DIFF
--- a/docs/scripting-doc/index.d.ts
+++ b/docs/scripting-doc/index.d.ts
@@ -1176,7 +1176,6 @@ interface frame {
    */
   duration : number
 }
-interface Image {}
 
 declare class Tile extends TiledObject{
   /**
@@ -1240,7 +1239,7 @@ declare class Tile extends TiledObject{
    * Warning: This function has no undo and does not affect the saved tileset!
    * @param image
    */
-  setImage(image : Image) : void
+  setImage(image : TiledImage) : void
 }
 
 declare class Layer extends TiledObject {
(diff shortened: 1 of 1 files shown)


bjorn:
@MrMasterplan I've extended this PR with the additions I've just made (boy did the scripting API become massive!). I think it should all be fine now since I was able to generate the documentation based on it, but I'd appreciate a quick review from your side.

Also, of course I had to revert the initial rename from `TiledImage` to `Image`. Would be nice if there was a way to disable the parsing of DOM API. Since it is not available in Tiled scripts, there is actually not a conflict...

MrMasterplan:
My code doesn't use the Image type and I see no issue with this. Since `TiledImage` was never actually exposed in `tiled`, this can't be a breaking change for anyone.

bjorn:
> My code doesn't use the Image type and I see no issue with this. Since `TiledImage` was never actually exposed in `tiled`, this can't be a breaking change for anyone.

Well, it was a breaking change for me, since it broke the generation of the documentation:

> Turns out Image class is already defined in TypeScript in
> node_modules/typescript/lib/lib.dom.d.ts. If we want to use this name
> we'd have to find some way to avoid including the DOM lib.

PR #3045, opened 2021-04-29
Tsconfig change allows Image to be used

MrMasterplan:
Setting some compiler options disables loading the DOM definitions.
I found the solution here: https://stackoverflow.com/questions/52850309/how-to-disable-global-variable-from-lib-dom-d-ts

### DIFF
--- a/docs/scripting-doc/index.d.ts
+++ b/docs/scripting-doc/index.d.ts
@@ -887,7 +887,7 @@ interface AspectRatio {}
 
 interface TransformationMode {}
 
-declare namespace TiledImage {
+declare namespace Image {
   const Format_Invalid: number;
   const Format_Mono: number;
   const Format_MonoLSB: number;
@@ -927,7 +927,7 @@ declare namespace TiledImage {
  *
  * @since 1.5
  */
-declare class TiledImage {
+declare class Image {
   /**
    * Width of the image in pixels.
    */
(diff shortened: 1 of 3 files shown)


MrMasterplan:
I am still testing this. Don't merge yet.

MrMasterplan:
I have tested this some more and I can compile my entire project with this state of the project. Yes, it will be incompatible with any libraries that make use of DOM definitions, and yes, people might need to figure out that they need specific `tsconfig.json` settings to make this work. I feel that this is more correct since it now describes the actual interface in the scripting API.
A way to address the learning curve is to add a complete working typescript project that makes use of these definitions, so people can copy that and go from there.
An alternative solution to this entire PR is to rename the class in the scripting API to avoid this conflict.

bjorn:
Thanks for patching this up!

> Yes, it will be incompatible with any libraries that make use of DOM definitions, and yes, people might need to figure out that they need specific `tsconfig.json` settings to make this work. I feel that this is more correct since it now describes the actual interface in the scripting API.

Of course libraries that use DOM definitions were already incompatible.

> A way to address the learning curve is to add a complete working typescript project that makes use of these definitions, so people can copy that and go from there.

Either that or just to extend the `README.md` with some TypeScript-specific instructions.

> An alternative solution to this entire PR is to rename the class in the scripting API to avoid this conflict.

I don't think we'll ever actually support the DOM Image class, so I'll stick to using this name for now.

PR #3049, opened 2021-05-01
Fix flickering when adding/removing components.

edin-m:
Fix flickering the menu when adding/removing components.


bjorn (approved):
Very clever!

PR #3057, opened 2021-05-05
Add components context menu option

edin-m:


src/tiled/propertiesdock.cpp
```diff
@@ -78,6 +78,12 @@ PropertiesDock::PropertiesDock(QWidget *parent)
     connect(mActionAddProperty, &QAction::triggered,
             this, &PropertiesDock::openAddPropertyDialog);
 
+    mActionAddComponent = new QAction(this);
+    mActionAddComponent->setEnabled(false);
+    mActionAddComponent->setIcon(QIcon(QLatin1String(":/images/16/add.png")));
+    connect(mActionAddComponent, &QAction::triggered,
+            this, &PropertiesDock::openAddComponentMenu);
```
bjorn, line 85: Hmm, this is not usually how a menu works. I'd suggest we just do something like this instead:
```c++
mActionAddComponent->setMenu(mButtonComponents->menu());
```
And we should probably just call this menu item "Components" instead of "Add Component".

Unfortunately the behavior of not hiding the menu on Shift is a little weird with submenus though, since the main context menu will hide.

edin-m: In this case, when the context menu action is triggered, the component menu will be opened at the position of the context menu itself. I think this is unusual yes, but rather useful and intuitive behavior. Shift-assisted behavior is still present if needed. 
Using submenus requires a lot of mouse movement and it doesn't have shift-assisted behavior.

bjorn: > Using submenus requires a lot of mouse movement and it doesn't have shift-assisted behavior.

Puh, I'm not sure it takes "a lot of mouse movement" to open a submenu. It is also the case, that this way we force the user to click on the "Add Component" menu item, whereas with a submenu a hover suffices to open the submenu. I'm really not sure if the non-standard behavior is beneficial here.

The Shift-assisted behavior would still be there though, it would just close the parent menu, but that also happens with the current implementation.

edin-m: 👍

PR #3059, opened 2021-05-07
Add components support for all types of objects

edin-m:
Add components support for all types of objects

Add components support for tmx/json


tests/maptovariantconverter/maptovariantconverter.pro
```diff
@@ -0,0 +1,22 @@
...
+!win32:!macx:!cygwin {
+    QMAKE_RPATHDIR += \$\$ORIGIN/../../lib
+
+    # It is not possible to use ORIGIN in QMAKE_RPATHDIR, so a bit manually
+    QMAKE_LFLAGS += -Wl,-z,origin \'-Wl,-rpath,$$join(QMAKE_RPATHDIR, ":")\'
+    QMAKE_RPATHDIR =
+}
+
+# Input
+SOURCES += test_maptovariantconverter.cpp
```
bjorn, line 22: It's great that you're adding test cases for this new functionality! I've been very bad at making sure any functionality is property tested...

However, I would say we shouldn't bother adding qmake projects for the tests. In fact I've only been keeping the qmake projects around since the Tiled snap is still built with qmake, but it is otherwise no longer used (except for packaging for various distributions). I've just pushed a change that makes the tests easily runnable using Qbs (334cb15b7b19c5db5fb9f0b710197a5d9844daeb).

src/libtiled/varianttomapconverter.cpp
```diff
@@ -966,7 +981,12 @@ bool VariantToMapConverter::readTileLayerData(TileLayer &tileLayer,
 Properties VariantToMapConverter::extractProperties(const QVariantMap &variantMap) const
 {
     return toProperties(variantMap[QStringLiteral("properties")],
-                        variantMap[QStringLiteral("propertytypes")]);
+            variantMap[QStringLiteral("propertytypes")]);
```
bjorn, line 984:
```suggestion
                        variantMap[QStringLiteral("propertytypes")]);
```
Please leave the existing indentation style intact. :-)

src/libtiled/mapreader.cpp
```diff
@@ -297,6 +297,8 @@ std::unique_ptr<Map> MapReaderPrivate::readMap()
             mMap->addLayer(std::move(layer));
         else if (xml.name() == QLatin1String("properties"))
             mMap->mergeProperties(readProperties());
+        else if (xml.name() == QLatin1String("components"))
+            mMap->mergeComponents(readComponents());
```
bjorn, line 301: Since we now support properties and components everywhere, it probably makes sense to introduce a helper function like `readCommonObjectElement`, so we can change this to:
```c++
else if (!readCommonObjectElement(object))
    readUnknownElement();
```

src/libtiled/maptovariantconverter.cpp
```diff
@@ -396,6 +418,7 @@ QVariant MapToVariantConverter::toVariant(const WangColor &wangColor) const
     colorVariant[QStringLiteral("probability")] = wangColor.probability();
     colorVariant[QStringLiteral("tile")] = wangColor.imageId();
     addProperties(colorVariant, wangColor.properties());
+    addComponents(colorVariant, wangColor.components());
```
bjorn, line 421: Maybe we should just change `addProperties` to `addPropertiesAndComponents(QVariantMap &variantMap, const Object &object)`?

src/libtiled/object.cpp
```diff
@@ -40,6 +40,19 @@ ObjectTypes Object::mObjectTypes;
...
 
+void Object::mergeComponents(const Components &components)
+{
+    QMapIterator<QString, Properties> it(components);
+    while (it.hasNext()) {
+        it.next();
+        auto const &name = it.key();
+        auto const &value = it.value();
+
+        // overwrites everything
```
bjorn, line 51: I don't think this comment is necessary. If you want to explain the behavior of `mergeComponents` you can add a doc comment there. But I think this short explanation is misleading, since it only overwrites the values of the properties defined in `value`, while keeping any other property values.

src/libtiled/object.h
```diff
@@ -88,6 +88,14 @@ class TILEDSHARED_EXPORT Object
     void mergeProperties(const Properties &properties)
     { Tiled::mergeProperties(mProperties, properties); }
 
+    /**
+     * Merges \a components with the existing components.
+     *
+     * Non-existing components are added.
+     * Existing components are overwritten.
```
bjorn, line 95: This sounds to me like the whole component would be replaced, while instead they are merged. I would in this case vote for letting the code speak for itself rather than trying to describe its behavior in detail here.

src/libtiled/varianttomapconverter.cpp
```diff
@@ -184,6 +185,20 @@ Properties VariantToMapConverter::toProperties(const QVariant &propertiesVariant
     return properties;
 }
 
+Components VariantToMapConverter::toComponents(const QVariant &componentsVariant) const
+{
+    Components components;
+
+    for (QVariant component : componentsVariant.toList()) {
```
bjorn, line 192: When using a range-based for with a Qt container, it is important to make sure it is `const` otherwise a non-const iterator is used which can (and will in this case) cause a detach (copy of the implicitly shared data). In addition, we should iterate with `const QVariant &` to avoid additional reference counting:
```suggestion
    const auto componentsVariantList = componentsVariant.toList();
    for (const QVariant &component : componentsVariantList) {
```

src/libtiled/varianttomapconverter.cpp
```diff
@@ -184,6 +185,20 @@ Properties VariantToMapConverter::toProperties(const QVariant &propertiesVariant
...
+        QVariantMap componentMap = component.toMap();
```
bjorn, line 193:
```suggestion
        const QVariantMap componentMap = component.toMap();
```
It is important to declare this map as `const`, otherwise `componentMap[QStringLiteral("name")]` will cause a detach.

bjorn (changes requested):
Nice to see components work everywhere! I've added a few inline comments about the implementation.

tests/maptovariantconverter/maptovariantconverter.pro
edin-m: Alright, I'll remove .pro files.
Anyway, I was thinking there should be one test project for libtiled with one test suite running tests of lots of classes. That way test_mapreader test_mapwriter etc... would be in one test project and run at once.

src/libtiled/varianttomapconverter.cpp
edin-m: oops

src/libtiled/mapreader.cpp
edin-m: I think the existing solution is much more readable. There isn't a lot of bilerplate while `readCommonObjectElement` is kind of ambiguous.

src/libtiled/maptovariantconverter.cpp
edin-m: I suggest against `addPropertiesAndComponents` or alike. It would be a type of function doing two inherently different things. That's why I put read/write components calls next to read/write/etc properties calls all throughout the code.

src/libtiled/object.h
edin-m: Other methods in this interface have the description so I followed along. Let me know what to do.

edin-m:
@bjorn Can you take a look at why the builds fail?
It seems it has something to do with qbs and not the code itself.

bjorn:
> @bjorn Can you take a look at why the builds fail?
> It seems it has something to do with qbs and not the code itself.

We need to increment the version of Qt Creator it tries to install, since the current version is no longer available. I've already fixed this up on master, so the `wip/components` branch will either need a merge from master or a rebase.

bjorn (approved):

PR #3061, opened 2021-05-12
Components multi-object selection

edin-m:
Add the ability to select multiple objects and assign, deassign, and change components and their properties.

### DIFF
--- a/src/tiled/changecomponents.cpp
+++ b/src/tiled/changecomponents.cpp
@@ -56,20 +56,20 @@ RemoveComponent::RemoveComponent(Document *document,
 {
     setText(QCoreApplication::translate("Undo Commands", "Remove Component (%1)")
             .arg(componentName));
+
+    mProperties.reserve(mObjects.size());
+    for (int i = 0; i < mObjects.size(); i++)
+        mProperties.append(mObjects[i]->componentProperties(mComponentName));
 }
 
 void RemoveComponent::undo()
 {
-    for (int i = 0; i < mObjects.size(); i++)
+    for (int i = 0; i < mObjects.size(); ++i)
         mDocument->addComponent({ mObjects.at(i) }, mComponentName, mProperties.at(i));
 }
 
 void RemoveComponent::redo()
 {
-    mProperties.reserve(mObjects.size());
-    for (int i = 0; i < mObjects.size(); i++)
-        mProperties.append(mObjects[i]->componentProperties(mComponentName));
-
     mDocument->removeComponent(mObjects, mComponentName);
 }
 
@@ -91,7 +91,7 @@ SetComponentProperty::SetComponentProperty(Document *document,
 
     mOldValues.reserve(objects.size());
 
-    for (int i = 0; i < objects.size(); i++) {
+    for (int i = 0; i < objects.size(); ++i) {
         Object *object = objects.at(i);
         Properties &props = object->componentProperties(componentName);
         mOldValues.append(props[mPropertyName]);
@@ -100,7 +100,7 @@ SetComponentProperty::SetComponentProperty(Document *document,
 
 void SetComponentProperty::undo()
 {
-    for (int i = 0; i < mObjects.size(); i++) {
+    for (int i = 0; i < mObjects.size(); ++i) {
         Object *object = mObjects.at(i);
         QVariant oldValue = mOldValues.at(i);
         mDocument->setComponentProperty(object, mComponentName, mPropertyName, oldValue);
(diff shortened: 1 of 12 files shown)


src/tiled/propertybrowser.cpp
```diff
@@ -551,16 +552,13 @@ void PropertyBrowser::propertiesChanged(Object *object)
...
+void PropertyBrowser::componentAdded(QList<Object *> objects, const QString &)
 {
     addComponents();
 }
 
-void PropertyBrowser::removeComponent(Object *object, const QString &componentName)
+void PropertyBrowser::removeComponent(QList<Object *> objects, const QString &componentName)
 {
-    if (object != mObject)
-        return;
```
bjorn, line 562: Hmm, we can't just remove this check. Now `PropertyBrowser::removeComponent` removes the components from the properties view regardless of which objects are selected. This can be wrong especially in case of "redo" with different objects selected. I'll look into fixing that.

src/tiled/propertybrowser.cpp
```diff
@@ -551,16 +552,13 @@ void PropertyBrowser::propertiesChanged(Object *object)
         updateCustomProperties();
 }
 
-void PropertyBrowser::componentAdded(Object *, const QString &)
+void PropertyBrowser::componentAdded(QList<Object *> objects, const QString &)
```
bjorn, line 555: Please always remember to use `const &` when passing implicitly shared containers around. Passing them by value generally means it'll need to increment and decrement the reference count, and risks causing detach when a non-const override is used.

src/tiled/changecomponents.cpp
```diff
@@ -21,78 +21,100 @@ static Properties objectTypeProperties(const QString &name)
...
                                  const QString &componentName,
                                  QUndoCommand *parent)
     : QUndoCommand(parent)
     , mDocument(document)
-    , mObject(object)
+    , mObjects(objects)
     , mComponentName(componentName)
 {
-    setText(QCoreApplication::translate("Undo Commands", "Remove Component"));
+    QString caption = QCoreApplication::translate("Undo Commands", "Remove Component (%1 objects)");
```
bjorn: I'm replacing the object count with the component name here.

bjorn: Ah, I see you've already pushed a change for this.

bjorn (approved):

bjorn:
Change looks fine to me now! Let me know if you see any other issues or whether this change should be ready for merging. :-)

src/tiled/changecomponents.cpp
```diff
@@ -21,78 +21,96 @@ static Properties objectTypeProperties(const QString &name)
...
+    , mObjects(objects)
     , mComponentName(componentName)
 {
-    setText(QCoreApplication::translate("Undo Commands", "Remove Component"));
+    setText(QCoreApplication::translate("Undo Commands", "Remove Component (%1)")
+            .arg(componentName));
+
+    mProperties.reserve(mObjects.size());
+    for (int i = 0; i < mObjects.size(); i++)
+        mProperties.append(mObjects[i]->componentProperties(mComponentName));
```
edin-m, line 62: Good catch.

src/tiled/changecomponents.h
```diff
@@ -34,7 +34,7 @@ class RemoveComponent : public QUndoCommand
 {
 public:
     RemoveComponent(Document *document,
-                    Object *object,
+                    const QList<Object *> &objects,
```
edin-m, line 37: damn, I keep missing adding const &, as much as I try not to.

src/tiled/propertybrowser.cpp
```diff
@@ -551,26 +551,30 @@ void PropertyBrowser::propertiesChanged(Object *object)
...
-    }
+        // remove and delete component group property
+        delete mComponents.take(name);
 
-    // remove and delete component group property
-    delete mComponents.take(componentName);
+    } else if (!Object::commonComponents(mDocument->currentObjects()).contains(name)) {
+        // object on current but no longer common, then it needs to be disabled
+        if (auto componentProperty = mComponents.value(name))
+            componentProperty->setEnabled(false);
```
edin-m, line 576: I haven't found a scenario where this would execute since I think the menu will have the item disabled. Anyways, good to have it.

edin-m:
> Change looks fine to me now! Let me know if you see any other issues or whether this change should be ready for merging. :-)

LGTM

src/tiled/propertybrowser.cpp
bjorn: This is relevant when the action is triggered via undo/redo, which can happen when a different set of objects is selected.

PR #3071, opened 2021-05-29
Add scripting support for components

edin-m:
* Add scripting support for components

e.g. tiled.activeAsset.currentLayer.objects[1].addComponent('boss', { hp: 77 })

* addComponent, removeComponent, changeComponentProperty
* Add object `commonComponents` tests

### DIFF
--- a/src/tiled/document.cpp
+++ b/src/tiled/document.cpp
@@ -229,7 +229,7 @@ void Document::addComponent(const QList<Object *> &objects, const QString &name,
             object->addComponent(name, properties);
     }
 
-    if (objects.size() > 0)
+    if (!objects.isEmpty())
         emit componentAdded(objects, name);
 }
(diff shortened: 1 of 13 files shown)


src/tiled/document.cpp
```diff
@@ -151,15 +151,17 @@ void Document::addComponent(const QList<Object *> &objects, const QString &name,
             object->addComponent(name, properties);
     }
 
-    emit componentAdded(objects, name);
+    if (objects.size() > 0)
```
bjorn, line 154:
```suggestion
    if (!objects.isEmpty())
```
But why did you add this check? It's never expected to be an empty list, right?

bjorn:
Puh, I didn't realize that resolving a conversation before the commit would remove it from the batch...

PR #3073, opened 2021-06-01
offsetx and offsety methods added to Python plugins api

sverx:
(not sure why this PR contains all previous commits, it seems I can't cherry pick here)


bjorn:
> (not sure why this PR contains all previous commits, it seems I can't cherry pick here)

Yeah, I don't think you can do anything about this on GitHub. In your local repository, you'd need to rebase your last change onto my latest master branch, which already contains the other commits (but merged together as 10c943f975fe428503937572fd2fbb269b9d5011).

Once rebased, you can force-push to your `master` branch to update this PR with just that change.

src/plugins/python/tiledbinding.py
```diff
@@ -331,6 +331,8 @@ def _decorate(obj, *args, **kwargs):
 cls_layer.add_method('asObjectGroup', retval('Tiled::ObjectGroup*',reference_existing_object=True), [])
 cls_layer.add_method('asImageLayer', retval('Tiled::ImageLayer*',reference_existing_object=True), [])
 cls_layer.add_method('asGroupLayer', retval('Tiled::GroupLayer*',reference_existing_object=True), [])
+cls_layer.add_method('offsetx', 'double', [])
+cls_layer.add_method('offsety', 'double', [])
```
bjorn, line 335: This change as such is insufficient. You'll need to actually run the `tiledbinding.py` script to update the `pythonbind.cpp` file.

In addition, the `Layer` class does not have `offsetx` and `offsety` methods. It has `QPointF offset() const;`. Probably what we need to add here is `cls_layer.add_method('offset', 'QPointF', [])`.

bjorn (changes requested):

src/plugins/python/tiledbinding.py
sverx: I see. I thought the build process would regenerate that file for me :frowning_face: 
Please cancel this, I'll see if I can do better.

bjorn: I've pushed the necessary changes.

bjorn:
Next time, just be sure to base your changes on the latest `master` branch in the main repository. Another good practice is to first create a branch, so that you don't end up opening the pull request from your own `master`, but rather from a branch specific to the pull request.

Thanks for your help with this addition!

sverx:
I'm learning. One day I'll open a proper PR, hopefully :sweat_smile: 
Thank you! :heart:

Issue #3080, opened 2021-06-09
Keyboard shortcut to select all layers


samw3:
Often while editing I'd like to select all layers so I can use the right-click brush grab feature, but currently there isn't a keyboard shortcut to do this and requires a "move to layers, click, shift-click, move to tiles" maneuver which really slows me down at times.

UliAbo:
Very good suggestion! I could also imagine pressing "CTRL+A" like in Windows, while hovering with the MouseCursor above the "Layers" window, or alternatively after I have clicked inside it.

UliAbo:
Another alternative: I would like even more to do a right click inside the "Layers" window, where currently there is only the context menu showing "New". There, I can imagine the following entries:
- Select all Layers
- Select all TileLayers
- Select all ObjectLayers

That would help me a lot!

bjorn:
@UliAbo Ah, I forgot to close this issue when merging #3081. I agree it might be better for Ctrl+A to work context-sensitive, currently the global shortcut always takes precedence, except when a text editing widget is selected.

Regarding the more specific actions to select all layers of a certain type, I can see that being useful as well sometimes. As a first step I tried implementing a "Select All Tile Layers" action in script:

```js
function collectTileLayers(mapOrGroupLayer, tileLayers = []) {
    for (let i = 0; i < mapOrGroupLayer.layerCount; ++i) {
        const childLayer = mapOrGroupLayer.layerAt(i)
        if (childLayer.isTileLayer)
            tileLayers.push(childLayer)
        else if (childLayer.isGroupLayer)
            collectTileLayers(childLayer, tileLayers)
    }
    return tileLayers
}

const action = tiled.registerAction("SelectAllTileLayers", function(action) {
    const map = tiled.activeAsset
    map.selectedLayers = collectTileLayers(map)
})

action.text = "Select All Tile Layers"

tiled.extendMenu("LayerView.Layers", [
    { action: "SelectAllTileLayers" },
]);
```

Except there is some issue with the Layers view not always reflecting the selected layers correctly (but at least they are selected...).

Anyway I could consider replacing the current "Select all Layers" action with a menu similar to the new layer menu, providing options to select either all layers or only a specific type of layer.

UliAbo:
> Anyway I could consider replacing the current "Select all Layers" action with a menu similar to the new layer menu, providing options to select either all layers or only a specific type of layer.

Yes I would totally love that!

PR #3081, opened 2021-06-10
Adds a menu item with keyboard shortcut for selecting all layers.

samw3:
Related to issue #3080, a feature I requested.  This adds a menu item with keyboard shortcut for selecting all layers.  I'm not a QT developer, so any comments and pointers will be happily accepted.  Thanks for making Tiled.

### DIFF
--- a/src/tiled/mapdocumentactionhandler.cpp
+++ b/src/tiled/mapdocumentactionhandler.cpp
@@ -29,6 +29,7 @@
 #include "documentmanager.h"
 #include "erasetiles.h"
 #include "grouplayer.h"
+#include "layermodel.h"
 #include "map.h"
 #include "mapdocument.h"
 #include "mapobject.h"
@@ -114,6 +115,9 @@ MapDocumentActionHandler::MapDocumentActionHandler(QObject *parent)
     mActionSelectNextLayer = new QAction(this);
     mActionSelectNextLayer->setShortcut(Qt::CTRL + Qt::Key_PageUp);
 
+    mActionSelectAllLayers = new QAction(this);
+    mActionSelectAllLayers->setShortcut((Qt::CTRL) + Qt::Key_B);
+
     mActionMoveLayersUp = new QAction(this);
     mActionMoveLayersUp->setShortcut((Qt::CTRL | Qt::SHIFT) + Qt::Key_Up);
     mActionMoveLayersUp->setIcon(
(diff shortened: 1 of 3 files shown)


src/tiled/mapdocumentactionhandler.cpp
```diff
@@ -114,6 +115,9 @@ MapDocumentActionHandler::MapDocumentActionHandler(QObject *parent)
     mActionSelectNextLayer = new QAction(this);
     mActionSelectNextLayer->setShortcut(Qt::CTRL + Qt::Key_PageUp);
 
+    mActionSelectAllLayers = new QAction(this);
+    mActionSelectAllLayers->setShortcut((Qt::CTRL) + Qt::Key_B);
```
bjorn, line 119: Why did you choose <kbd>Ctrl+B</kbd>? Did you check if there's an equivalent shortcut in other applications with layers, like Krita, GIMP or Photoshop?

src/tiled/mapdocumentactionhandler.cpp
```diff
@@ -829,6 +849,7 @@ void MapDocumentActionHandler::updateActions()
                                                    [] (Layer *layer) { return layer->canMergeDown(); }));
     mActionSelectPreviousLayer->setEnabled(hasPreviousLayer);
     mActionSelectNextLayer->setEnabled(hasNextLayer);
+    mActionSelectAllLayers->setEnabled(true);
```
bjorn, line 852: Since the enabled state never changes, we can just leave out this line.

src/tiled/mapdocumentactionhandler.cpp
```diff
@@ -706,6 +713,19 @@ void MapDocumentActionHandler::selectNextLayer()
...
 
+void MapDocumentActionHandler::selectAllLayers()
+{
+    if (!mMapDocument)
+        return;
+    QList<Layer *> layersToSelect;
+
+    for (Layer *layer : mMapDocument->map()->allLayers()) {
+        layersToSelect.append(layer);
+    }
```
bjorn, line 724: For single-line bodies, please leave out the braces.

bjorn (commented):
Nice addition! I've left some small comments inline.

src/tiled/mapdocumentactionhandler.cpp
samw3: I wasn't sure what to pick, but checking similar apps is a great idea.  It looks like Krita and GIMP do not have shortcuts for this, but Photoshop uses Ctrl+Alt+A.  How does that sound?

bjorn: I considered <kbd>Ctrl+Alt+A</kbd> as well, so let's go with that!

src/tiled/mapdocumentactionhandler.cpp
samw3: Will do.

src/tiled/mapdocumentactionhandler.cpp
samw3: Will do

bjorn:
Thanks!

PR #3086, opened 2021-06-27
[components] Add component support to object templates

edin-m:


src/libtiled/mapobject.cpp
```diff
@@ -428,6 +428,9 @@ void MapObject::syncWithTemplate()
 
     if (!propertyChanged(MapObject::VisibleProperty))
         setVisible(base->isVisible());
+
+    // maybe component properties can be detachable (long future)
+    mergeComponents(base->components());
```
bjorn, line 433: Hmm, I don't think this will work. Each time `syncWithTemplate` is called, we essentially copy all the components from the template? What about when the user has changed one of those properties on the instance? They'll be overwritten on loading the file, as well as when a template is edited.

Note how the properties are not copied from the template. If a property is set on an instance, it means it overrides a potential value inherited from the template. I think it could work similarly for components. Of course, the problem is that when adding a component, we'd need to add an empty map instead of one filled with the properties of the component. The Properties view would still know which properties to display, because it can look that up from the component definition, similar to how it displays properties based on the object type.

tests/mapobject/test_mapobject.cpp
```diff
@@ -0,0 +1,44 @@
...
+    props["hp"] = 50;
+    obj.addComponent("component1", props);
+    ObjectTemplate templateObj;
+    templateObj.setObject(&obj);
+
+    MapObject mapObject;
+    mapObject.setObjectTemplate(&templateObj);
+
+    mapObject.syncWithTemplate();
+    QCOMPARE(mapObject.componentProperties("component1")["hp"], 50);
```
bjorn, line 30: This appears to cause:
```
undefined reference to `bool QTest::qCompare<QVariant, int>(QVariant const&, int const&, char const*, char const*, char const*, int)'
```
when compiling against Qt 5.6 with MinGW 4.9 (the WinXP build). Maybe replace `50` with `QVariant(50)`.

tests/mapobject/test_mapobject.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+    obj.setComponentProperty("component1", "hp", QVariant(55));
```
bjorn, line 32: Here we don't need `QVariant(55)` and can just write `55`, because the conversion to `QVariant` happens implicitly.

src/libtiled/mapobject.cpp
edin-m: I thought object template properties would overwrite instances'.

src/libtiled/mapobject.cpp
```diff
@@ -428,6 +428,16 @@ void MapObject::syncWithTemplate()
...
     if (!propertyChanged(MapObject::VisibleProperty))
         setVisible(base->isVisible());
+
+    QMapIterator<QString, Properties> it(base->components());
+    while (it.hasNext()) {
+        it.next();
+        if (!hasComponent(it.key())) {
+            addComponent(it.key(), Properties {});
+            Properties& properties = componentProperties(it.key());
+            Tiled::mergeProperties(properties, it.value());
```
bjorn, line 438: This is not very different from what the `mergeComponents` method was doing before, right?

What I meant to say is, we shouldn't actively copy down properties from the template to the instance, because the existence of the property on the instance means it overrides the value from the template. At least, that's what it currently means for the properties, and I think it would be a good idea to use the same approach with the components.

We'd need something like the `Object::resolvedProperties`, but then `Object::resolvedComponents`, which returns the components of the object, including inherited components and the values of their properties. That's what we display in the Properties view (grayed out values, to indicate when they are inherited vs. when they are overridden).

edin-m: No, previously it copied from template to instance no questions asked.
Now, it's copying only if the instance does not have the component. (iterates the components).

I didn't plan on adding properties inheritance for components, I'm not sure how much work is that...

bjorn: Ok, then it's fine if this is making an initial copy for now. It can be removed when we do add support for inheritance to components in the future. The code can then be simplified to:
```c++
        if (!hasComponent(it.key()))
            addComponent(it.key(), it.value());
```

edin-m:
Do you think I should rebase wip/components onto master? I'm not sure how much it has deviated...

bjorn:
> Do you think I should rebase wip/components onto master? I'm not sure how much it has deviated...

I've merged `master` into `wip/components` and rebased your branch onto it.

PR #3105, opened 2021-07-13
Fix copy tool failing on some staggered maps

Lobachevskiy:
Fix https://github.com/mapeditor/tiled/issues/2874

Stamp tool has code that handles staggered maps, however it doesn't trigger if the stamp and the map have mismatch in their stagger index. In the case of copying it was always initialized to the default values instead of the values of the map copied from.

In my opinion some refactoring should take place, like forcing these parameters to be in the constructor, to avoid this same issue in other places.

### DIFF
--- a/src/tiled/clipboardmanager.cpp
+++ b/src/tiled/clipboardmanager.cpp
@@ -152,6 +152,8 @@ bool ClipboardManager::copySelection(const MapDocument &mapDocument)
                 selectionBounds.height(),
                 map->tileWidth(), map->tileHeight());
     copyMap.setRenderOrder(map->renderOrder());
+    copyMap.setStaggerAxis(map->staggerAxis());
+    copyMap.setStaggerIndex(map->staggerIndex());
 
     bool tileLayerSelected = std::any_of(selectedLayers.begin(), selectedLayers.end(),
                                          [] (Layer *layer) { return layer->isTileLayer(); });


bjorn (approved):
Great find! I agree some refactoring to prevent such issues could definitely help, but for now it's great to have this issue fixed!

bjorn:
> In my opinion some refactoring should take place, like forcing these parameters to be in the constructor, to avoid this same issue in other places.

So, I agreed and I took a jab a it in PR #3109.

I didn't think adding further parameters to the constructor was a good idea, since it also tends to be error-prone and affecting a lot of code. So instead I introduced a `Map::Parameters` structure, which should make it easier to add/remove stuff later on. I've also added a `Map::EditorSettings` structure which contains things we'd generally not want to copy. What do you think of this change, @Lobachevskiy?

> In the case of copying it was always initialized to the default values instead of the values of the map copied from.

While doing the refactoring, I noticed the fix in this PR is actually incomplete, it just shifts the problem to the other 50% of the cases (the problem now happens on odd indices instead of even indices). This is because, when we copy an area from a map that has "stagger index" set to odd, but we start at an odd index, the resulting copy should actually get its "stagger index" set to even.

This is what the `CaptureStampHelper` does in the following lines:

https://github.com/mapeditor/tiled/blob/98cfe7b307efa7139c57552c3f6db806defea41f/src/tiled/capturestamphelper.cpp#L73-L84

Potentially another small refactoring should be done, such that the `CaptureStampHelper` can also be used in `ClipboardManager::copySelection` and `stampFromContext` (`tilestampmanager.cpp`), so that we don't need to duplicate this logic.

Lobachevskiy:
> Potentially another small refactoring should be done, such that the CaptureStampHelper can also be used in ClipboardManager::copySelection and stampFromContext (tilestampmanager.cpp), so that we don't need to duplicate this logic.

Perhaps refactor this out, maybe even into the `Map` itself? A method that returns part of the `Map`, stagger-adjusted.

bjorn:
> Perhaps refactor this out, maybe even into the `Map` itself? A method that returns part of the `Map`, stagger-adjusted. @bjorn

For now the refactor was only done halfway (d082031556545b14e51f2d37d8ccf28a546b909e unifies the two other locations) and I just put this code in two places now (0f7804cc29bf04a16078e16295ad687f3775bff0). I've also decided to make the code a bit shorter after all, as you suggested in https://github.com/mapeditor/tiled/pull/3109#discussion_r676605642. :-)

PR #3123, opened 2021-08-25
Moved object types into the project

bjorn:
This change moves the storage of object types from a separate file into the project. The main advantage being to simplify the user experience (avoiding the need to manage an additional file, as well as getting rid of the the "global" vs. "overridden by project file" distinction).

When a loaded project referred to an object types file, its types are now automatically imported to the project. This is not done for the global object types file however, so instructions are probably needed for those who need to import the file manually (documenting the location of this file).

When the object types file was changed outside of Tiled, it was automatically reloaded. Automatic reloading still needs to be implemented for the project.

The new setup is not ideal for those who were generating the object types file. A potential solution would be to add a scripting API for registering object types.

Feedback welcome!

### DIFF
--- a/src/tiled/mainwindow.cpp
+++ b/src/tiled/mainwindow.cpp
@@ -1038,7 +1038,6 @@ void MainWindow::initializeSession()
     bool projectLoaded = !session.project.isEmpty() && project.load(session.project);
 
     if (projectLoaded) {
-        Preferences::instance()->setObjectTypesFile(project.mObjectTypesFile);
         ProjectManager::instance()->setProject(std::move(project));
         updateWindowTitle();
         updateActions();
@@ -1522,7 +1521,6 @@ bool MainWindow::switchProject(Project project)
         prefs->addRecentProject(project.fileName());
     }
 
-    prefs->setObjectTypesFile(project.mObjectTypesFile);
     ProjectManager::instance()->setProject(std::move(project));
 
     restoreSession();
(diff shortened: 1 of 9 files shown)


eishiya:
I find it very convenient to have multiple Projects share the same object definitions file, as my Object Types are what my engine will accept and the default property values are the defaults the engine uses, the definitions do not vary per game. I would like the option to have multiple projects continue to use the same Object Types definitions, so that updates to it will affect all my projects using that engine.

I think the current approach of being able to choose what definitions file your project uses is great for my scenario. However, if possible, I think Tiled should start with local definitions by default, and provide the option to select an external file instead.

All that said, I imagine my use case is somewhat unusual. The reason my definitions change and need to be updated on a semi-regular basis across multiple projects is because I am still working on the engine in conjunction with the games (which stress different aspects of it, hence it not just being one project). I imagine most people will either be working with a stable version, or will at least have a fixed engine version per project, in both cases, they'd just import their object properties once and not have to worry about sync across projects.
As such, I do not expect to have my needs catered to perfectly. I can deal with exporting/importing instead, especially if an option to *replace* the definitions is provided, to better handle cases were properties are removed or renamed. Even better would be an option to append new Object types and replace existing Object types, but I imagine having this option would just confuse most people, and it may better be implemented via scripting. (Ditto for enums! I'd love this Append+Replace ability for enums especially...)

bjorn:
Superceded by #3376, since the property types were already stored in the project.

PR #3125, opened 2021-08-30
Fix inherited parallax values causing layers to display incorrectly

### LINKED ISSUE #3124: Incorrect preview position with inhereted parallax values
## Summary

* The parallax preview doesn't produce the correct "effective" parallax factor when group parallax inheritance is involved.
* The UI selection box and object name is, however, in the proper position.
* Tiled 1.7.2, Windows 10 x64

## Overview

The docs state:
> When the parallax scrolling factor is set on a group layer, it applies to all its child layers. The effective parallax scrolling factor of a layer is determined by multiplying the parallax scrolling factor by the scrolling factors of all parent layers.

However, I'm getting incorrect preview behavior, as demonstrated in this screenshot:

image

_Camera at X = 0_

image

_Camera at X = 4_

The camera is translated to the right by four tiles, so the 1 tile moves 4 tiles, the 0.5 tile moves 2 tiles, and the 0.25 tile moves 3 tiles. However, I would expect the 0.5*0.5 tile (a group set to 0.5, with an image layer within set to 0.5) to look like the 0.25 tile, but it actually looks like the -0.25 tile!

It seems, however, that the hover UI selection and name behaves correctly, and is offset from the real thing:
image

## Reproduction

Here is a file containing the screenshotted test, as well as several other tests (demonstrating the incorrect behavior with parallax values above 1, in 0 to 1, in -1 to 0, and below -1):

[test_parallax.zip](https://github.com/mapeditor/tiled/files/7075996/test_parallax.zip)

This file _doesn't_ test mismatched group/layer parallax values (one at 0.8 and another at 0.5, for example), or the behavior of more than one level of inheritance.

LilithSilver:
Fixes #3124.

Before, parallax was causing group layers to be offset as well as non-group layers. Now, only non-group layers have the parallax offset, so that they can properly use their effective inherited parallax value.

Visually tested on the scene I included in the issue. Also, tested with a 2-layer inheritance: 0.8 * 0.5 * 0.5 = 0.2.

### DIFF
--- a/src/tiled/mapitem.cpp
+++ b/src/tiled/mapitem.cpp
@@ -266,14 +266,7 @@ void MapItem::updateLayerPositions()
 
     for (LayerItem *layerItem : qAsConst(mLayerItems)) {
         const Layer &layer = *layerItem->layer();
-        QPointF parallaxOffset = mapScene->parallaxOffset(layer);
-
-        // If this is a group layer, we don't want to apply any parallax -- that'll be
-        // taken care of by the effective parallax offset calculated for the leaf layers.
-        // If we don't do this, it doubles up bad.
-        if (layer.layerType() == Layer::GroupLayerType) parallaxOffset = QPointF(0, 0);
-
-        layerItem->setPos(layer.offset() + parallaxOffset);
+        layerItem->setPos(mapScene->layerItemPosition(layer));
     }
 
     if (mDisplayMode == Editable) {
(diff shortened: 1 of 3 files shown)


bjorn:
Good catch and nice fix!

Since the layer item position is actually set in three places (on creation, when updating all positions and when responding to changes in layer properties), I've introduced a helper function with the group layer check.

I considered whether it would be possible to use the local parallax offset instead of the effective one, but it seems difficult to obtain the same behavior that way. Could it be that the current behavior is not intuitive at all? Maybe parallax factors should be summed up instead of multiplied (after subtracting 1). I wonder if anybody would be relying on the multiplication behavior already...

LilithSilver:
Sounds good, thank you! 

Personally, I think multiplication is pretty intuitive. It's exactly how I would expect it to work -- something going slower in the background should go even slower if I made their parent go slower, and vice versa. I think where it gets confusing is with negative values -- a negative times a negative is a positive! -- but negative values aren't super useful, and the user can always make sure their group structure accounts for that. Or perhaps a multiplication strategy like `-2 * -0.5 = - 1` is in order, for those; I'm not sure. 

It's also pretty easy to implement with multiplication; I just added it to my renderer for my game and it really wasn't bad!

bjorn:
@LilithSilver Thanks for sharing your thoughts! One thing that bothered me about the multiplication, is that once you put a group on factor 0, you can't make any of its child layers move, but I guess that's unlikely to bother anyone in practice. I think alternatively, it could behave like any value set on a child layer would override the value of the parent rather than multiply by it.

PR #3128, opened 2021-09-06
docs: Update pytiled-parser and Arcade links

Cleptomania:
Some links have changed for these projects recently so just a quick update to them in the docs.


bjorn:
@Cleptomania Thanks for your help!

PR #3163, opened 2021-10-19
[WIP] Set collision editor bg color to tileset bg color

### LINKED ISSUE #3070: Tileset background colour should affect Tile Collision Editor
The Tile Collision Editor currently always uses the theme's default panel background colour, even if the Tileset has a background colour set. One of the main uses for the Tileset background colour is to make tiles more visible regardless of theme colour, and it's a shame to be missing that in the Tile Collision Editor, where readability and precision are often very important.

I think the background colour of the Tile Collision Editor should be the Tileset background colour, and it should display the theme's default colour only when no background colour is set.

Lyqst:
Setting collision editor background color from tileset background color is working. However, when changing the tileset background color, the collision editor stays the same color until a new tile is opened.

Something I noticed is that when you change the tileset background, the tile that is open on the collision is unselected, and the collision editor stays open without a tile, which I'm thinking might be the reason the background is not updated.

Fixes #3070

### DIFF
--- a/src/tiled/mapscene.cpp
+++ b/src/tiled/mapscene.cpp
@@ -292,6 +292,11 @@ void MapScene::refreshScene()
 
 void MapScene::updateDefaultBackgroundColor()
 {
+    if(mOverrideBackgroundColor.isValid()){
+        setBackgroundBrush(mOverrideBackgroundColor);
+        return;
+    }
+
     mDefaultBackgroundColor = QGuiApplication::palette().dark().color();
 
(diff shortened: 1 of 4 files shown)


bjorn:
Sorry for the long silence! I've given it another review and some testing. While I didn't find any issues, I changed the code a bit to make it more consistent (using only override color, and not both override color and map background color) and potentially handle some edge-cases better (by moving the actual updating of the background color to a single place). Thanks a lot for your help with this feature!

PR #3167, opened 2021-10-28
Update Translation Korean

chopmozzi:
I supplemented the existing translation file `tiled_ko.ts`.
Also, I maintained the unity with the existing translation as much as possible.
example1
I checked the typos as much as possible.
I hope it helps you.

### DIFF
--- a/translations/tiled_ko.ts
+++ b/translations/tiled_ko.ts
@@ -2,24 +2,474 @@
 <!DOCTYPE TS>
 <TS version="2.1" language="ko_KR">
 <context>
+    <name>QtPointFPropertyManager</name>
+    <message>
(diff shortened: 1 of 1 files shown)


bjorn:
Thank you for this amazing effort! The Korean translation was previously less than 1/3rd done, so great respect for getting it all the way to 100%!

Btw, I plan to make a Tiled 1.8 release this month. About two weeks before a new release I usually announce a string freeze so that the translations can be fully updated at the time of release. Let me know if you'd like to get notified about this (or subscribe to updates [on Weblate](https://hosted.weblate.org/projects/tiled/translations/), which I think can also send out an email when I update the source file).

PR #3182, opened 2021-11-21
Update Traditional Chinese translation

SiderealArt:


bjorn:
Thanks for another good looking update!

Issue #2962, opened 2020-12-15
Implement in-game viewport preview


bjorn:
In general it would be useful if you could quickly see what part of a level the player will see while playing the game. Currently, one could use a rectangle object for this, but it would be nice if there was direct support for this. Advantages of a built-in feature would include:

* Automatically center when view is resized.
* Nice visual representation, for example darkening the area outside of the view.
* Could support optional "save area" that might become visible depending on various screen resolutions / ratios.

eishiya:
Additional benefits:

* This could be a Project or editor property, meaning that it wouldn't be wasting space in the TMX files (this data is not needed in-engine)
* Easier to make a keyboard shortcut to toggle this the preview
* If it's a Project/editor property, adjusting it once would adjust it for all related maps. Viewports are usually game-wide, afterall.

PR #3205, opened 2021-12-08
Add "Repeat ..." properties to Image Layers

krukai:
Introduce the boolean properties "Repeat Horizontally" and "Repeat Vertically" to Image Layers. This mirrors GamerMaker's "Horizontal Tile" and "Vertical Tile" options.
The YY plugin will export these properties accordingly. For backwards compatibility however, the respective custom properties "htiled" and "vtiled" take precedence over the built-in ones.

I am currently still tinkering with the drawing to actually behave as described in the commit message. I may also have to touch the changes for the bounds again. There is also some room for more consistency I still want to tackle.
Finally, I am planning to move these two new properties above "Offset X", if that feels right to you as well.

Until then, feel free to review the rest. I hope I did not overlook anything serious; I only started working with the source code yesterday.

Cheers!


src/libtiled/imagelayer.cpp
```diff
@@ -37,7 +37,9 @@
 using namespace Tiled;
 
 ImageLayer::ImageLayer(const QString &name, int x, int y):
-    Layer(ImageLayerType, name, x, y)
+    Layer(ImageLayerType, name, x, y),
+    mRepeatHorizontally(false),
+    mRepeatVertically(false)
```
bjorn, line 42: Please place the initialization at the declaration.

src/libtiled/imagelayer.h
```diff
@@ -114,6 +114,26 @@ class TILEDSHARED_EXPORT ImageLayer : public Layer
      */
     bool isEmpty() const override;
 
+    /**
+     * Returns true if the image of this layer repeats horizontally.
+     */
+    bool repeatHorizontally() const { return mRepeatHorizontally; }
```
bjorn, line 120: Let's go with a shorter forms like in CSS, `repeatX` and `repeatY`.

src/libtiled/mapreader.cpp
```diff
@@ -1142,6 +1142,9 @@ std::unique_ptr<ImageLayer> MapReaderPrivate::readImageLayer()
     auto imageLayer = std::make_unique<ImageLayer>(name, x, y);
     readLayerAttributes(*imageLayer, atts);
 
+    imageLayer->setRepeatHorizontally(atts.value(QLatin1String("hrepeat")).toInt());
+    imageLayer->setRepeatVertically(atts.value(QLatin1String("vrepeat")).toInt());
```
bjorn, line 1146: I know other attributes are already not consistent (we have `parallaxx` but `hflip`), I'd still prefer to change this to `repeatx` and `repeaty`.

src/libtiled/maprenderer.cpp
```diff
@@ -82,7 +82,30 @@ MapRenderer::~MapRenderer()
...
+        return bounds;
+
+    QRect mapBounds = QRect(QPoint(), map()->size());
+
+    LayerIterator iterator(map(), Layer::TileLayerType);
+    while (TileLayer *tileLayer = static_cast<TileLayer*>(iterator.next()))
+        mapBounds = mapBounds.united(tileLayer->bounds());
+
+    if (mapBounds.size() == QSize(0, 0))
+        mapBounds.setSize(QSize(1, 1));
```
bjorn, line 97: What about calling `mapBoundingRect`?

src/libtiled/mapwriter.cpp
```diff
@@ -669,6 +669,12 @@ void MapWriterPrivate::writeLayerAttributes(QXmlStreamWriter &w,
                          colorToString(layer.tintColor()));
     }
 
+    if (layer.layerType() == Layer::ImageLayerType) {
+        auto &imageLayer = static_cast<const ImageLayer&>(layer);
+        w.writeAttribute(QStringLiteral("hrepeat"), QString::number(imageLayer.repeatHorizontally()));
+        w.writeAttribute(QStringLiteral("vrepeat"), QString::number(imageLayer.repeatVertically()));
+    }
```
bjorn, line 676: Let's just put that in `writeImageLayer`. I think the only reason I did this strange handling of some tile layer attributes was because I cared about its `width` and `height` to stay together with the `x` and `y` attributes. :-)

src/tiled/changeimagelayerproperties.cpp
```diff
@@ -33,7 +33,9 @@ ChangeImageLayerProperties::ChangeImageLayerProperties(
         MapDocument *mapDocument,
         ImageLayer *imageLayer,
         const QColor &color,
-        const QUrl &newSource)
+        const QUrl &newSource,
+        bool newRepeatH,
+        bool newRepeatV)
```
bjorn, line 38: That's a bit too many constructor parameters, which is starts to become annoying to use. I think the `ChangeImageLayerProperties` class could be changed to use the pattern used by `ChangeMapProperty`.

src/tiled/editableimagelayer.cpp
```diff
@@ -71,6 +75,37 @@ void EditableImageLayer::setImage(ScriptImage *image, const QUrl &source)
...
+        asset()->push(new ChangeImageLayerProperties(doc,
+                                                     imageLayer(),
+                                                     transparentColor(),
+                                                     imageSource(),
+                                                     repeatHorizontally,
+                                                     repeatVertically()));
+    } else {
+        imageLayer()->setRepeatHorizontally(repeatHorizontally);
+        if (!imageSource().isEmpty())
+            imageLayer()->loadFromImage(imageSource());
```
bjorn, line 90: Why would we call `loadFromImage` again?

bjorn (changes requested):
Thanks for working on this feature! I did no testing yet, but left some comments on the changes. :-)

src/tiled/editableimagelayer.cpp
krukai: Because we copy and paste too much. Fixed.

src/libtiled/maprenderer.cpp
krukai: It seems to me that the method is not implemented for maprenderer. It is for others renderers, though.

src/libtiled/mapreader.cpp
krukai: Done.

src/libtiled/imagelayer.h
krukai: Done.

src/libtiled/imagelayer.cpp
krukai: Done.

src/libtiled/mapwriter.cpp
krukai: I wanted to do that originally, but it seemed to me that `writeImageLayer` only writes attributes to the image tag. I can have a look again later.

krukai:
I resolved most of the review comments.
Originally, I wanted to move the properties above Offset X and Y because that would feel natural for Image Layers, IMHO. However, that would require to insert properties into the overall layer group of properties. I am not sure if we want to do that.

I also still have to tinker with the actual drawing. It works fine until offsets are set, but then the drawing rectangle would have to be adjusted to fill the space created by the offset within the overall bounds. I am not entirely sure yet where to make the adjustments, but I will look into it further.

src/libtiled/maprenderer.cpp
bjorn: Yes, this is a pure virtual function, but you can still call it. It is not possible to instantiate a `MapRenderer` because of this and other pure virtual functions. All instances will be from one of its subclasses, and the virtual function call will end up at the appropriate implementation (which is why function is virtual).

src/libtiled/mapwriter.cpp
bjorn: That would mean you have placed the code after the `w.writeStartElement(QStringLiteral("image"));` line. You should place it right after the `writeLayerAttributes(w, imageLayer);` call.

bjorn:
> Originally, I wanted to move the properties above Offset X and Y because that would feel natural for Image Layers, IMHO. However, that would require to insert properties into the overall layer group of properties. I am not sure if we want to do that.

In the end the attribute order does not matter in XML, and I think it's fine to have the repeat attributes follow after the parallax ones.



> I also still have to tinker with the actual drawing. It works fine until offsets are set, but then the drawing rectangle would have to be adjusted to fill the space created by the offset within the overall bounds. I am not entirely sure yet where to make the adjustments, but I will look into it further.

Yes, to handle the offset you'd need to call [`QBrush::setTransform`](https://doc.qt.io/qt-5/qbrush.html#setTransform) with a transform created with [`QTransform::fromTranslate`](https://doc.qt.io/qt-5/qtransform.html#fromTranslate).

bjorn:
Ah, handling the offset will be a little more complicated, because currently that's done by simply moving the `ImageLayerItem` around.

Hmm, is it really important that when repetition is used, the layer is clipped to the map boundaries? It's rather inconsistent with the way image layers have worked so far, and actually nothing else is clipped to the map boundaries. If it would be alright if the layer expands beyond the map boundaries then we'd just need to increase the bounding rect of the `ImageLayerItem` to be practically infinite.

krukai:
> In the end the attribute order does not matter in XML, and I think it's fine to have the repeat attributes follow after the parallax ones
I was referring to the properties panel. My bad.

>Hmm, is it really important that when repetition is used, the layer is clipped to the map boundaries? It's rather inconsistent with the way image layers have worked so far, and actually nothing else is clipped to the map boundaries. If it would be alright if the layer expands beyond the map boundaries then we'd just need to increase the bounding rect of the ImageLayerItem to be practically infinite.
Would infinity actually be acceptable here or would that screw with other features relying on the bounds?

While I see your point, I would prefer if it was cut off at the overall bounds at least optionally, assuming we would not have to jump through too many hoops to achieve that. Otherwise, the overall view would get extremely noisy and it might be difficult to see at a glance where the actual level bounds begin.
But I realize while typing this that arguing for allowing an inconsistency here is pretty close to arguing for a "show me what the player would see" mode, which would include cutting off images with an offset but no repetition as well. This would also help when fiddling with parallax parameters.
So perhaps this would be a feature on its own and not within scope here.

bjorn:
> Otherwise, the overall view would get extremely noisy and it might be difficult to see at a glance where the actual level bounds begin.

Right, I think visualizing the map boundaries more effectively is a different problem. Maybe darkening the area outside of the map would help, for example.

> But I realize while typing this that arguing for allowing an inconsistency here is pretty close to arguing for a "show me what the player would see" mode, which would include cutting off images with an offset but no repetition as well. This would also help when fiddling with parallax parameters.
> So perhaps this would be a feature on its own and not within scope here.

Yeah, this really sounds like issue #2962.

Btw, another option to consider, is to add a repetition mode to _tile objects_. They already have a size that would naturally be used to clip the repetition. Generally image layers are mostly an inferior version of the tile objects, with their only advantage being (at least until this PR) that you did not have to first add the image to a tileset.

src/libtiled/mapwriter.cpp
krukai: Done.

src/libtiled/maprenderer.cpp
krukai: I get what you mean now, sorry. But I think this is irrelevant now.

src/tiled/changeimagelayerproperties.cpp
krukai: Done.

krukai:
Added scripting support (I hope). Everything should work as expected with "infinite" repeating now, I hope.

I also refactored the Use the same pattern as `ChangeImageLayerProperties` as requested, albeit in a separate commit. I am also leaving renaming the class to singular to another commit as otherwise the diff would not be very useful. If you want to squash any of that, please let me know.

That being said, is there anything else to take care of? Do my changes warrant modifying some header comment or anything?

src/tiled/changeimagelayerproperties.cpp
```diff
@@ -32,41 +32,87 @@ using namespace Tiled;
...
+    case ImageSourceProperty: {
+        const QUrl source = mImageLayer->imageSource();
+        mImageLayer->setSource(mImageSource);
+        mImageSource = source;
 
-    if (mUndoSource.isEmpty())
-        mImageLayer->resetImage();
-    else
-        mImageLayer->loadFromImage(mUndoSource);
+        if (mImageSource.isEmpty())
```
bjorn, line 96: Hmm, there's a bug here in that this checks the old `mImageSource`. I was testing the change and when I tried to set the image on the layer it would reset the image instead of load the one I just set. This block should happen before assigning `source` to `mImageSource` (and since `loadFromImage` sets the image URL as well, we can actually just remove the call to `setSource`).

bjorn (changes requested):
I found some minor issues and a bug that need addressing, but I think this will be the last review! :-)

src/libtiled/maprenderer.cpp
```diff
@@ -115,7 +126,10 @@ void MapRenderer::drawImageLayer(QPainter *painter,
 {
     Q_UNUSED(exposed)
 
-    painter->drawPixmap(QPointF(), tinted(imageLayer->image(), imageLayer->effectiveTintColor()));
+    painter->save();
+    painter->setBrush(tinted(imageLayer->image(), imageLayer->effectiveTintColor()));
```
bjorn, line 130: Oh, one more thing, we need also `painter->setPen(Qt::NoPen)`, otherwise it's rendering a black outline around the layer.

krukai:
I implemented all review comments, updated the documentation, and added support for JSON and Lua export.
Please just squash everything into one commit; I just kept the latter commits separate for easier reviewing and at this point, I do not think it makes much sense to separate anything into individual commits.

If there is anything else, let me know!

bjorn:
@krukai Thanks a lot for implementing this very welcome feature!

krukai:
Thank you! My bad for missing the changed filename in the comments.

PR #3209, opened 2021-12-15
Add Parallax Origin property to Map Properties

krukai:
Independent from the plans for viewports, giving the user means to change the parallax offset (as in a "starting position") would help greatly to accommodate differences between Tiled's approach and that of a user's game project.

Please note that this is still WIP; the PR was created for enabling discussion (and bug fixing).


src/libtiled/mapreader.cpp
```diff
@@ -270,6 +270,17 @@ std::unique_ptr<Map> MapReaderPrivate::readMap()
...
     mapParameters.staggerIndex = staggerIndexFromString(staggerIndex);
 
+    QPointF parallaxOffset;
+    bool ok;
+    const qreal parallaxOffsetX = atts.value(QLatin1String("parallaxoffsetx")).toDouble(&ok);
+    if (ok)
+        parallaxOffset.setX(parallaxOffsetX);
+    const qreal parallaxOffsetY = atts.value(QLatin1String("parallaxoffsety")).toDouble(&ok);
+    if (ok)
+        parallaxOffset.setY(parallaxOffsetY);
```
bjorn, line 280: I'd suggest just `mapParameters.parallaxOffset.setY(parallaxOffsetY);`.

src/tiled/propertybrowser.cpp
```diff
@@ -618,6 +618,8 @@ void PropertyBrowser::addMapProperties()
 
     staggerIndexProperty->setAttribute(QLatin1String("enumNames"), mStaggerIndexNames);
 
+    addProperty(ParallaxOffsetProperty, QMetaType::QPointF, tr("Parallax Offset"), groupProperty);
```
bjorn, line 621: You also need to set the value of this property in `PropertyBrowser::updateProperties`.

bjorn (changes requested):

eishiya:
Rather than "parallax offset", which to me suggests the calculated draw offset based on parallax, I think the property should be called something like "parallax origin" or something else that suggests that it's a reference point.

src/tiled/propertybrowser.cpp
krukai: Done.

src/libtiled/mapreader.cpp
krukai: Done.

bjorn:
I've changed the parallax origin to get subtracted, rather than added, to the view center. It seemed more intuitive to me, since I interpret the parallax origin to be a point on the map. Does that make sense?

I've also fixed the immediate updating of the layer positions when you change the property, by connecting the `MapScene` to the `Document::changed` signal and checking for this property.

The change looks fine to me now, but before we merge this we need to again update the documentation:

* Add property to `index.d.ts`.
* Add property to JSON and TMX format documentation, including their changelog.
* Update the https://doc.mapeditor.org/en/latest/manual/layers.html#parallax-reference-point section in the documentation. It  now needs to mention the parallax origin, how it is relative to the top-left of the map's bounding box, and after that can mostly replace "map’s origin" with "parallax origin".

krukai:
That should be all, I think. Please have a final look.
Cheers!

PR #3222, opened 2021-12-24
docs: Update pytiled-parser and Arcade description

Cleptomania:
pytiled-parser and Arcade both previously only supported JSON maps, but have recently been updated to add support for TMX as well. This PR just removes the note of not having TMX support from them.


bjorn:
Thanks! :-)

PR #3237, opened 2022-01-16
AdjustTileMetaData also updates tile types

### LINKED ISSUE #3236: Auto-adjusting tile references when a tileset has grown doesn't update tile types and properties
### Bug description

When Tiled detects that a tileset image has changed size, it offers the option to update all the references to it to reflect the new tile IDs.

However, if tiles in the spritesheet had types and/or properties defined, they remain assigned to the old tile IDs they were assigned to previously, which after the re-enumeration is now incorrect.

### Steps to reproduce:
1. Create a small tileset with 2x2 tiles:
spritesheet
 
2. Add a type to the first tile of the second row (id=2).
Screenshot 2022-01-15 01 43 36

3. Resize the tileset image to be 4x2 tiles:
spritesheet

4. When prompted in Tiled, accept to auto-update the tile references.
image

5. Check that the tile you previously given the property is now tile id=4, but the type is still assigned to tile id=2.
image

albertvaka:
When a tileset image has changed size and existing tile IDs are auto-adjusted, also update tile types in addition to other metadata.

Fixes #3236

### DIFF
--- a/src/tiled/adjusttileindexes.cpp
+++ b/src/tiled/adjusttileindexes.cpp
@@ -177,7 +177,7 @@ AdjustTileMetaData::AdjustTileMetaData(TilesetDocument *tilesetDocument)
 
     auto applyMetaData = [&](Tile *toTile,
                              const Properties &properties,
-                             const QString& type,
+                             const QString &type,
                              qreal probability,
                              std::unique_ptr<ObjectGroup> objectGroup,
                              const QVector<Frame> &frames)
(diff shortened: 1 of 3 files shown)


src/tiled/adjusttileindexes.cpp
```diff
@@ -175,6 +177,7 @@ AdjustTileMetaData::AdjustTileMetaData(TilesetDocument *tilesetDocument)
 
     auto applyMetaData = [&](Tile *toTile,
                              const Properties &properties,
+                             const QString& type,
```
bjorn, line 180:
```suggestion
                             const QString &type,
```

bjorn (approved):
This is just perfect, thanks!

PR #3271, opened 2022-02-13
Add missing <errno.h> include for errno

AMDmi3:


bjorn:
For some reason I hadn't run into this missing, but it seems reasonable, thanks!

PR #3275, opened 2022-02-17
Improves InputStream handling

theNestruo:
This commit improves InputStream handling in the `libtiled-java` artifact:

- Ensures that file and URL InputStreams are properly closed
- Ensures consuming from BufferedInputStreams (for performance reasons)
- Centralizes .gz (GZIP) support

### DIFF
--- a/util/java/libtiled-java/src/main/java/org/mapeditor/io/MapReader.java
+++ b/util/java/libtiled-java/src/main/java/org/mapeditor/io/MapReader.java
@@ -33,11 +33,9 @@ package org.mapeditor.io;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
-import java.net.URL;
 import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-import java.util.zip.GZIPInputStream;
 
 import javax.xml.bind.JAXBContext;
 import javax.xml.bind.JAXBElement;
@@ -49,6 +47,7 @@ import javax.xml.stream.XMLStreamException;
 
 import org.mapeditor.core.Map;
 import org.mapeditor.core.TileSet;
+import org.mapeditor.util.StreamHelper;
 
 /**
  * The standard map reader for TMX files. Supports reading .tmx, .tmx.gz and
@@ -82,17 +81,10 @@ public class MapReader {
         int fileSeparatorIndex = filename.lastIndexOf(File.separatorChar) + 1;
         String xmlPath = makeUrl(filename.substring(0, fileSeparatorIndex));
 
-        String xmlFile = makeUrl(filename);
-
-        URL url = new URL(xmlFile);
-        InputStream is = url.openStream();
-
         // Wrap with GZIP decoder for .tmx.gz files
-        if (filename.endsWith(".gz")) {
-            is = new GZIPInputStream(is);
+        try (InputStream in = StreamHelper.openStream(filename)) {
+            return readMap(in, xmlPath);
         }
-
-        return readMap(is, xmlPath);
     }
 
     /**
@@ -113,9 +105,9 @@ public class MapReader {
      * @throws java.io.IOException if any.
      */
     public TileSet readTileset(String filename) throws IOException {
-        String xmlFile = makeUrl(filename);
-        URL url = new URL(xmlFile);
-        return readTileset(url.openStream());
+        try (InputStream in = StreamHelper.openStream(filename)) {
+            return readTileset(in);
+        }
     }
 
     private Map buildMap(Map map, String xmlPath) throws IOException {
@@ -147,7 +139,7 @@ public class MapReader {
     private <T> T unmarshal(InputStream in, Class<T> type) {
         try {
             XMLInputFactory factory = XMLInputFactory.newInstance();
-            XMLEventReader reader = factory.createXMLEventReader(in);
+            XMLEventReader reader = factory.createXMLEventReader(StreamHelper.buffered(in));
 
             JAXBContext context = JAXBContext.newInstance(type);
             Unmarshaller unmarshaller = context.createUnmarshaller();
--- a/util/java/libtiled-java/src/main/java/org/mapeditor/io/TMXMapReader.java
+++ b/util/java/libtiled-java/src/main/java/org/mapeditor/io/TMXMapReader.java
@@ -34,7 +34,6 @@ package org.mapeditor.io;
 import java.awt.Color;
 import java.awt.geom.Ellipse2D;
 import java.awt.geom.Path2D;
-import java.awt.geom.Point2D;
 import java.awt.geom.Rectangle2D;
 import java.awt.image.BufferedImage;
 import java.io.ByteArrayInputStream;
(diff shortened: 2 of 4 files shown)


util/java/libtiled-java/src/main/java/org/mapeditor/io/TMXMapReader.java
```diff
@@ -260,14 +271,15 @@ private TileSet unmarshalTileset(Node t, boolean isExternalTileset) throws Excep
         if (source != null) {
             source = replacePathSeparator(source);
             URL url = URLHelper.resolve(xmlPath, source);
-            InputStream in = url.openStream();
-            TileSet ext = unmarshalTilesetFile(in, url);
+            try (InputStream in = StreamHelper.openStream(url)) {
```
bjorn, line 274: This seems to be the first time we rely on try-with-resources, introduced in Java 7. Does the library already specify at least Java 7 is required somewhere?

util/java/libtiled-java/src/main/java/org/mapeditor/io/MapReader.java
```diff
@@ -82,17 +81,10 @@ public Map readMap(String filename) throws IOException {
         int fileSeparatorIndex = filename.lastIndexOf(File.separatorChar) + 1;
         String xmlPath = makeUrl(filename.substring(0, fileSeparatorIndex));
 
-        String xmlFile = makeUrl(filename);
-
-        URL url = new URL(xmlFile);
-        InputStream is = url.openStream();
-
         // Wrap with GZIP decoder for .tmx.gz files
```
bjorn, line 84: Let's remove this comment now.

util/java/libtiled-java/src/main/java/org/mapeditor/util/StreamHelper.java
```diff
@@ -0,0 +1,140 @@
...
+	/**
+	 * @param in the {@link InputStream} to wrap with a {@link BufferedInputStream}
+	 * @return a {@link BufferedInputStream} wrapping the {@code in},
+	 * the same {@code in} if it already was a {@link BufferedInputStream},
+	 * or {@code null} if {@code in} was {@code null}
+	 */
+	public static BufferedInputStream buffered(InputStream in) {
+
+		// (sanity check)
+		if (in == null) {
```
bjorn, line 132: Let's not do this additional check. The function will return `null` in any case, since `in instanceof BufferedInputStream` would be `false` and `BufferedInputStream.class.cast(in)` will handle the `null` case fine.

util/java/libtiled-java/src/main/java/org/mapeditor/util/StreamHelper.java
```diff
@@ -0,0 +1,140 @@
...
+	/**
+	 * @param in the {@link InputStream} to wrap with a {@link GZIPInputStream}
+	 * @return a {@link GZIPInputStream} wrapping the {@code in},
+	 * the same {@code in} if it already was a {@link GZIPInputStream},
+	 * or {@code null} if {@code in} was {@code null}
+	 */
+	public static GZIPInputStream ungzip(InputStream in) throws IOException {
+
+		// (sanity check)
+		if (in == null) {
```
bjorn, line 114: Same here, the function will behave the same way if we remove this check.

util/java/libtiled-java/src/main/java/org/mapeditor/util/StreamHelper.java
```diff
@@ -0,0 +1,140 @@
...
+	 * Opens an {@link InputStream} for reading from the specified location,
+	 * automatically uncompressing data in the GZIP file format if required.
+	 * @param location The filename, path or URL to read
+	 * @return the input stream for reading from the specified location,
+	 * or {@code null} if the location is neither {@link #isUrl(String)} nor {@link #isPathname(String)}
+	 */
+	public static InputStream openStream(String location) throws IOException {
+
+		// (sanity check)
+		if ((location == null) || location.length() < 0) {
```
bjorn, line 56:
```suggestion
		if (location == null || location.length() < 0) {
```

util/java/libtiled-java/src/main/java/org/mapeditor/util/StreamHelper.java
```diff
@@ -0,0 +1,140 @@
...
+import java.io.BufferedInputStream;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URL;
+import java.util.zip.GZIPInputStream;
+
+public class StreamHelper {
+
+	private static final String GZIP_EXTENSION = ".gz";
```
bjorn, line 40: Please adjust this file to use 4-space indent instead of tabs.

util/java/libtiled-java/src/main/java/org/mapeditor/util/StreamHelper.java
```diff
@@ -0,0 +1,140 @@
...
+	 * @return {@code true} if the filename, path or URL has GZIP extension (ignoring case),
+	 * {@code false} otherwise
+	 */
+	public static boolean isGzip(String location) {
+
+		if (location == null || location.length() < GZIP_EXTENSION_LENGTH) {
+			return false;
+		}
+		final int offset = location.length() - GZIP_EXTENSION.length();
+        return location.regionMatches(true, offset, GZIP_EXTENSION, 0, GZIP_EXTENSION_LENGTH);
```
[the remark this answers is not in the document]
theNestruo: `.regionMatches(...)` has an `ignoreCase` parameter, but `.endsWith(...)` does not (and performs a case-sensitive check).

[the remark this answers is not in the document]
bjorn: Ah, alright! Indeed, we'll want this check to be case-insensitive.

util/java/libtiled-java/src/main/java/org/mapeditor/io/TMXMapReader.java
theNestruo: Yes, at `org.mapeditor:tiled-parent` pom.xml:
```xml
<maven.compiler.source>1.8</maven.compiler.source>
```
However, I can reorganize the code to properly close the streams without relying on try-with-resources if you prefer.

bjorn: No, I much prefer the new feature, thanks! :-)

util/java/libtiled-java/src/main/java/org/mapeditor/io/MapReader.java
theNestruo: Ok

util/java/libtiled-java/src/main/java/org/mapeditor/util/StreamHelper.java
theNestruo: Ok, sorry

util/java/libtiled-java/src/main/java/org/mapeditor/util/StreamHelper.java
theNestruo: Actually, it is the other way around: `new GZIPInputStream(in)` will cause a `NullPointerException`.
I can change these sanity checks by `Objects.requireNonNull(...)`, if you prefer. That would change the behaviour to throw a `NullPointerException`.

bjorn: You're right, let's just leave them be. I don't prefer `Objects.requireNonNull` if the behavior would then be the same as without.

theNestruo:
Hi @bjorn ! Thank you for making Tiled!
I think moving the java libraries to a separate repository is a good idea, indeed!

I don't have too much free time but, also, the amount of Java library-related issues seems to be low, so... if my sparse dedication sounds good enough, I'll be happy to help :)

I was planning to write an internal command line tool, and that's why I started looking at `libtiled-java`. At the same time, I'll try to tackle some of the issues; expect a couple of pull requests during the next couple of weeks.

PR #3298, opened 2022-03-13
docs: Update list of 3rd-party libraries

hendursaga:
python-tmx is no longer maintained, but the homepage links to
https://ulvl.github.io/ which has support for reading TMX.

pytmxloader appears to be the best/most up-to-date fork of
https://code.google.com/archive/p/pytmxloader/

Note that I have no idea how to "update" the i18n locales - it might be good to add a section to the docs about that, if needed.

### DIFF
--- a/docs/reference/support-for-tmx-maps.rst
+++ b/docs/reference/support-for-tmx-maps.rst
@@ -78,6 +78,10 @@ C#/.NET
    `tiled-xna <https://github.com/zachmu/tiled-xna>`__ and
    `TmxCSharp <https://github.com/gwicksted/TmxCSharp>`__.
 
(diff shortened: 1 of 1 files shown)


bjorn:
Thanks a lot for helping to update this list!

> Note that I have no idea how to "update" the i18n locales - it might be good to add a section to the docs about that, if needed.

Right, normally that is only done right after string freeze, so no need to worry about that now. But it would still make sense to document the commands. I've used these last time:
```
make gettext
sphinx-intl update -p _build/gettext -l de -l fr
```

bjorn:
> But it would still make sense to document the commands. I've used these last time:

Done in ec45294d.

PR #3311, opened 2022-03-24
docs: Add irrTiled to list of 3rd-party libraries

hendursaga:
Just came across this, wasn't even looking for it.


bjorn:
Nice addition, thanks!

PR #3320, opened 2022-04-01
Add German translation to AppData

JakobDev:


bjorn:
Thanks, that sounds great! :-)

PR #3346, opened 2022-05-02
docs: Changed File script API to a namespace so it can be accessed in TypeScript

jonny64bit:
Hey. I was struggling to call `File.directoryEntries` in typescript due to `File` currently being an interface. Looks like this should be a namespace instead. I've tested the methods on this namespace version from typescript and all look fine.


bjorn:
Many thanks! I'll make a new release soon. :-)

bjorn:
I guess the same should be done for `FileInfo` then?

PR #3347, opened 2022-05-03
docs: Changed FileInfo script API to a namespace so it can be accessed in TypeScript 

jonny64bit:
FileInfo also need to be changed to a namespace.

Related: #3346


bjorn:
Thanks! :-)

PR #3352, opened 2022-05-07
%worldfile as variable in command.cpp

Pixel-Nori:
I know that this is basically nothing, but in the game that I'm working on, this would make it way easier to make levels and worlds, so I decided to try to add it.
Basically this is just like the %mapfile variable, but for the tiled worlds


bjorn:
Good addition, thanks!

Issue #1008, opened 2015-06-29
Isometric tile sheet with different sized tiles


Spanfile:
I have a tile sheet of isometric tiles, where the tiles aren't all exactly the same size and aren't placed like tiles would be, neatly in rows and columns. For example, the "first" tile in the sheet is at  coordinates 596, 85 and is 99 pixels wide and 85 pixels tall. The next tile is at 0, 1270 and is 133 pixels wide and 127 pixels tall. Next to the sheet, I have an XML file which defines all the tile positions and sizes by their names so extracting them from the sheet is just like loading them as individual images.

Currently, Tiled wants all the tiles in the sheet to be the exact same size and be placed neatly in rows and columns in the sheet, which means I can't use my tile sheet in Tiled.

Would it be possible to add a way to read tile positions and sizes from some kind of file, such as an XML file like I currently have? It can be a proprietary Tiled file, following a Tiled format, it shouldn't be too hard converting between my XML and it. I would think it would work just like using a collection of images, but have the images come from a sheet, rather than individual files.

brenolf:
:+1: for this feature

Glissando:
Support for texture atlases is something I've wanted for a while. I've been using multiple tilesets based on size as a workaround.

bjorn:
Alright, I've added it to the Tiled roadmap where you can also vote on it: https://trello.com/b/yl3PAtN0/tiled-roadmap

One worry I have with this feature is that people will pack their images using for example TexturePacker and then use the result in Tiled. When done this way, Tiled has currently no way of numbering the tiles in a way that would remain stable in the event that the texture atlas is re-packed after images have been added/removed. For this use case, the user is expected to use the individual images in Tiled and only use the packed texture in the game, as an optimization.

The reason I still want to support this feature is mainly because it is often convenient for an artist to put related tiles on the same image, even if they have different sizes. In this case, there should be a way to tell Tiled where to find those tiles and to update that information when the image changes,,without invalidating tile references.

bjorn:
Also asked for [on the forum](http://discourse.mapeditor.org/t/working-with-texture-atlas-in-tiled/588/8?u=bjorn).

pga-avionics:
And here again :+1: 

https://discourse.mapeditor.org/t/make-tileset-working-with-texture-atlas/3661/5

zyend36:
And asked here again :+1: 

https://discourse.mapeditor.org/t/make-tileset-working-with-texture-atlas/3661/4

[...]

PR #3359, opened 2022-05-17
Use Tile::imageRect instead of cutting out tiles

bjorn:
Now that we're storing a sub-rectangle for each tile, we can use it to avoid actually cutting out the tiles from the tileset image.

Changing of a tile's image rect now also has its own undo command, and the tile image rect property is visible also for tiles from tileset-image based tilesets, though not editable for now.

### DIFF
--- a/src/libtiled/imagecache.cpp
+++ b/src/libtiled/imagecache.cpp
@@ -39,39 +39,6 @@
 
 namespace Tiled {
 
-bool TilesheetParameters::operator==(const TilesheetParameters &other) const
-{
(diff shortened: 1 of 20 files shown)


bjorn:
@devnewton Since this is a follow-up on your PR #3339, I'd welcome any feedback. Apart from further simplifying the code, this is another step in the direction of resolving #1008.

devnewton:
The code looks fine. When multi sized tiles will be available, I can adapt my tileset to test it.

Issue #3100, opened 2021-07-07
Automap: Allow explicitly specifying Empty and Wildcard tiles


eishiya:
**tl;dr**: Add special tiles that users can use to represent Empty, Any Tile, or Wildcard (empty or any tile) in their Automap rules, to make creating automapping rules more intuitive.

--

A common task when creating Automapping rules, especially for sidescrollers, is to make rules that match certain arrangements of various tiles and empty tiles. In the simplest scenario, such as a grass-top tile that replaces an empty tile that is above a grass-base tile, this isn't very difficult to achieve - *if* you remember that `StrictEmpty` is a thing. However, using StrictEmpty means that you've now given up the ability to match any non-empty tile in that rule. It's also just difficult to figure out (it still trips me up after years of making Automapping rules!), and it's not very intuitive.

Similarly, leaving a tile empty to denote "any non-empty tile" is not very intuitive.
And when you need a rule that needs to combine wildcards and empty tiles, e.g. "change empty tile on top of a Grass tile into Flower, but only if there is a Tree two tiles away", it gets very messy.

It would make creating Automap rules more intuitive and simpler if Empty Tiles, Any Tile (non-empty wildcards), and Full Wildcards (matching empty *and* non-empty) could be specified explicitly in Automap rules, with visual representations. This poses a challenge since Automap rules are just regular map files with a few special properties.

I've described them in the context of input layers so far, but the Empty Tile could also be used on output layers to explicitly erase a tile, which is currently challenging to do. The wildcards probably don't make much sense on output layers and could be disregarded.

I can think of two options to implement this:
- A special user-defined layer that contains the three tiles, in order. This would be even better as some special properties, but Tiled has no way to reliably set a tile as the value of a special property. If only 1 or 2 tiles are in this layer, then no tile will ever match the remaining match meaning(s).
   - This option has the benefit of letting users easily define their own visuals. However, in most cases, I suspect this is more of a burden than a bonus.
   - This option doesn't do much to make Automapping more accessible to new users, since there's still a lot of setup required to get these special tiles working, and it may lead to confusion if users aren't sure whether defining these tiles is required.
   - The special layer would have to be added to every single rule file, adding tedium to an already tedious process. The docs encourage one rule per file in scenarios where order matters, so for many users,t hat's a lot of files.

- A special "automap" mode you can enable when creating maps that adds a special helper tileset with Empty, Any Tile, and Wildcard "tiles".
[...]

PR #3373, opened 2022-05-24
Added built-in Automapping Rules Tileset and action to add it

bjorn:
Main problem with this built-in tileset is that it is hardcoded to tiles of size 32x32. Before merging this, we need a solution to make it adjust to other tile sizes. One idea would be to add a tile rendering mode option to the tileset, that can be set to "stretch to cell" or "fill cell".

Part of issue #3100


bjorn:
Hmm, there appears to be still a repainting issue caused by the Tile Render Size option:

image

Issue #3001, opened 2021-03-15
Scripting API: Allow creating more complex and persistent GUI


eishiya:
Currently, scripting provides access to alerts, prompts, and confirmation modal windows. Technically, this is enough for any configuration the user may need to do for a script, but it makes for a very inconvenient user experience for most things.

Scripting could be more useful if users could be given more convenient control over a script's configuration via child windows that can contain several widgets, with a variety of widgets available, and styled according to the user's theme. Recommended widget options:
- Label (just a plain bit of text, used to label other widgets, but this should be separate rather than forced on each widget, as sometimes the label itself can be used for output, or to label groups of related widgets)
- Text input
- Slider (returns a value between 0 and 1 inclusive, or perhaps takes min/max/step parameters)
- Multiline text (can be input, but probably more useful as output so the user can copy generated text)
- Dropdown
- Checkbox

Less important, nice-to-have widgets:
- Separator (with optional text, for labelling sections)
- Radio button groups (can technically be implemented script-side with checkboxes)
- File selector (lets the user select a file path, any actual opening/parsing/writing would be done by the script)
- Image (shows an `Image`, useful for diagrams, graphs, and any other non-standard UI)
- Color (just like the Color properties in Tiled, shows a color's hex value, a preview of it, and allows selecting a new color by opening a Select Color modal)

Being able to set tooltips on the widgets (or just Labels?) would also be useful.

For Tools and some kinds of actions, persistent windows that can remain open while a user does other things would be a very powerful tool, as well. Examples:
- A random tree generator Tool could have a configuration panel that remains open as long as the tool is active to let the use set the type and size of the tree.
- Provide access to different modes of a Tool using buttons or checkboxes as an alternative to using modifier keys, to give your hand a rest.
- A toggleable Action could open/close a persistent window that shows a warning if you put certain tiles on the wrong layer, forgot to set certain custom properties, or do something else your engine doesn't allow. This would be more convenient and compact than a list of warnings in the Issues panel. (For now I don't think this can actually be done due to the limited signals available, but hopefully we'll get more editor-focused signals in the future.)

For an example of GUI scripting done this way, check out Dialogs in Aseprite's scripting API: https://github.com/aseprite/api/blob/master/api/dialog.md#dialog
[...]

PR #3384, opened 2022-06-14
Add Dialog class for use in scripting

dogboydog:
Allow a customizable dialog to be created via scripting as requested in #3001. Allows script writers to get more pieces of information from the user. The dialogs can be persistent or you can connect to a signal such as `finished()` and only move on with your script after a dialog is closed. 

Goals: 

- [x]     Label (just a plain bit of text, used to label other widgets, but this should be separate rather than forced on each widget, as sometimes the label itself can be used for output, or to label groups of related widgets). Also allows multiline text -- just include \n in your label text and specify maxWidth = true. 
- [x]     Text input 
- [x]  Text edit - multi line input (allows rendering HTML) implemented
- [x] Separator (with optional text, for labelling sections) addSeparator()
- [x]    Slider (returns a value between 0 and 1 inclusive, or perhaps takes min/max/step parameters) addSlider()
- [x]    Dropdown: addComboBox()
- [x]    Checkbox: addCheckBox()
- [x] dialog.clear(): Now working
- [x] File selector (lets the user select a file path, any actual opening/parsing/writing would be done by the script). 
- [x] Color  like the Color properties in Tiled, shows a color's hex value, a preview of it, and allows selecting a new color by opening a Select Color modal)
- [x] Typescript doc (I believe this is now complete but let me know any omissions)

Stretch goals:

* [x] addImage(): add an image to the dialog which can be updated - Implemented. 
* [x] Ability to disable automatic grouping of same-type widgets so that you can mix widget types in some cases. Implemented -- set newRowMode to achieve this. 
Example scripts to use this functionality:
https://github.com/dogboydog/tiled-dialog-scripts


src/tiled/scriptdialog.cpp
```diff
@@ -0,0 +1,330 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "scriptdialog.h"
+#include "scriptmanager.h"
+#include <QProcessEnvironment>
```
bjorn, line 24: I don't think you need that include here.

src/tiled/scriptdialog.cpp
```diff
@@ -0,0 +1,330 @@
...
+
+ScriptDialog::ScriptDialog(): ScriptDialog(QString(), 450, 450)
+{
+    
+}
+ScriptDialog::ScriptDialog(const QString &title, const int width=450, const int height=450)
+ : QDialog(MainWindow::maybeInstance())
+{
+    if (title.isEmpty()){
+        setTitle(QString::fromUtf8("Script"));
```
bjorn, line 40: That should be translatable:
```suggestion
        setTitle(tr("Script"));
```

src/tiled/scriptdialog.cpp
```diff
@@ -0,0 +1,330 @@
...
+      setTitle(title);
+    }
+    setAttribute(Qt::WA_DeleteOnClose);
+    m_widgetNumber = 0;
+    m_lastWidgetTypeName.clear();
+
+    setMinimumSize(width, height);
+    setSizePolicy(QSizePolicy::Expanding, QSizePolicy::Expanding);
+    initializeLayout();
+    QMetaObject::connectSlotsByName(this);
```
bjorn, line 51: Which slots are you connecting here?

src/tiled/scriptdialog.cpp
```diff
@@ -0,0 +1,330 @@
...
+    QSize oldSize = size();
+    QDialog::resize(width, height);
+    QSize newSize = QSize(width, height);
+    m_gridLayout->setGeometry(QRect(15,15, width-30, height-30));
+    m_gridLayoutWidget->setGeometry(QRect(15,15, width-30, height-30));
+    setMinimumWidth(width);
+    setMinimumHeight(height);
+    adjustSize();
+    QResizeEvent event = QResizeEvent(newSize, oldSize);
+    resizeEvent(&event);
```
bjorn, line 92: Hmm, this whole function should just be calling `Dialog::resize(width, height)` and be done with it. The way to make the internal grid layout widget resize along with the dialog is by putting it in a layout and setting that layout on the dialog.

Likely, it would be enough to set the `m_gridLayout` as the layout of the dialog, in which case we don't need `m_gridLayoutWidget` at all.

src/tiled/scriptdialog.cpp
```diff
@@ -0,0 +1,330 @@
...
+            addNewRow();
+        }
+    }
+    else
+        addDialogWidget(label);
+    return label;
+}
+Tiled::ScriptDialogWidget *ScriptDialog::addSeparator()
+{
+    return addSeparator(QString());
```
bjorn, line 125: You should be able to just declare a default in the declaration, like `labelText = QString()`, instead of this separate overload.

src/tiled/scriptdialog.cpp
```diff
@@ -0,0 +1,330 @@
...
+    addNewRow();
+    m_gridLayout->addLayout(m_rowLayout, m_rowIndex,0, 1, -1); // span entire
+    QLabel * separatorLabel;
+    if(!labelText.isEmpty()){
+        separatorLabel= newLabel(labelText);
+        separatorLabel->setWordWrap(false);
+        m_rowLayout->addWidget(separatorLabel, 1);
+    }
+    line = new QFrame(m_gridLayoutWidget);
+    line->setObjectName(QString::fromUtf8("line%1").arg(m_widgetNumber));
```
bjorn, line 139: It's not needed to set an object name.

bjorn (commented):
Some initial comments.

I'm not sure about the addition of `ScriptDialogWidget` and think I'd rather see the labels and widgets added separately than automatically tagging a label onto every widget. But maybe you could explain why you added this.

src/tiled/scriptdialog.cpp
dogboydog: Should I remove all instances of setObjectName?

src/tiled/colorbutton.h
```diff
@@ -32,14 +32,15 @@ namespace Tiled {
...
 
 public:
     ColorButton(QWidget *parent = nullptr);
-
+    Q_INVOKABLE void setToolTip(const QString& toolTip);
     QColor color() const { return mColor; }
     void setColor(const QColor &color);
 
-signals:
+Q_SIGNALS:
```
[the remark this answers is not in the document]
dogboydog: I will have to retry it, but I believe before I changed this,  referencing `colorChanged.connect` from my script crashed Tiled

src/tiled/scriptdialog.cpp
bjorn: Probably, unless you're using those names somewhere, but I don't think so.

src/tiled/colorbutton.h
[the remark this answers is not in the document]
bjorn: Hmm, that should not happen because of this, cause these macros are synonyms.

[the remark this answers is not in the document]
dogboydog: I will be sure to re-test

dogboydog:
> Some initial comments.
> 
> I'm not sure about the addition of `ScriptDialogWidget` and think I'd rather see the labels and widgets added separately than automatically tagging a label onto every widget. But maybe you could explain why you added this.

Thanks. I can change it to be that way potentially but let me explain the thinking based on my conversations with @eishiya . A lot of these widgets have no  label built in to them. This style allows the user to add a label and widget in one call. Then if you wanted to change the text of a label afterwards, it would be stored together with the actual widget. The other aspect is that  currently the way the automatic layout works is that if you change widget type,  it will automatically go to the next row, but if you use  the same widget type multiple times, even with adding labels, it will group them in the same row. 

The alternative approach would be making it so that labels can inherently  be mixed  with any other widget type without wrapping to the next row.  Right now the way it works is that all labels except those attached to a ScriptDialogWidget are always placed in the first column, and then optionally span the entire width of the dialog.  I think it's fine to always require you to call addLabel()  but  I wonder what users would think.  This would change the API to remove all of the label text parameters (except in cases where the widget itself has a label like QCheckBox) and remove the ScriptDialogWidget concept

eishiya:
The idea behind attaching a label to every widget comes from the Aseprite Dialog API. It avoids the need to think about labels as an additional element, it's just "I can add a widget, and it can have a label", and as a bonus, it's just more convenient since you can set the label in the widget's constructor. The labels are optional, but for ease of implementation, all of them have labels, they're just left blank when not needed.

I also feel that this distinction between row labels (i.e. Label Widgets) and widget labels (the ones attached to widgets) makes it easier to understand the automatic layout system\*, with how it groups like widgets into rows and puts a line break between different widget types. Heck, I'd even consider renaming the labels to RowLabel or something to make this clearer.

\* The automatic layout system is also something we yoinked from Aseprite. This automatic grouping/splitting makes it easy to put together an appealing Dialog without thinking too much. It naturally tends to split unrelated things between rows while keeping related things together, and like widgets tend to align more nicely when automatically scaled to fit in a row. Perhaps others might want more flexibility, and it might be cool to provide that at some point, but I think a basic system like this that just lets you append the widgets you need and focus on your script logic is great to have.

src/tiled/colorbutton.h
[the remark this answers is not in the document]
dogboydog: Something else must have created my earlier problem. I changed it back to `signals:`

src/tiled/scriptdialog.cpp
dogboydog: Removed resize()

src/tiled/scriptdialog.h
```diff
@@ -0,0 +1,99 @@
...
+private:
+    int m_rowIndex = 0;
+    int m_widgetsInRow = 0;
+    int m_leftColumnStretch = 1;
+    // stretch as much as we can so that the left column looks as close to zero width as possible when there is no content
+    int m_rightColumnStretch = 9999;
+    QGridLayout *m_gridLayout;
+    QLabel *newLabel(const QString& labelText);
+    void initializeLayout();
+    QWidget *m_rowLayoutWidget;
```
bjorn, line 82: Unused variable.

src/tiled/scriptdialog.h
```diff
@@ -0,0 +1,99 @@
...
+{
+    Q_OBJECT
+
+public:
+    Q_INVOKABLE ScriptDialog(const QString &title = QString());
+    ~ScriptDialog() override;
+
+    bool checkForClosed() const;
+
+    Q_INVOKABLE void setTitle(const QString &title);
```
bjorn, line 59: I don't really see the point of this one, since there is already the `windowTitle` property.

src/tiled/scriptdialog.h
```diff
@@ -0,0 +1,99 @@
...
+namespace Tiled {
+class ScriptDialog : public QDialog
+{
+    Q_OBJECT
+
+public:
+    Q_INVOKABLE ScriptDialog(const QString &title = QString());
+    ~ScriptDialog() override;
+
+    bool checkForClosed() const;
```
bjorn, line 57: Unimplemented, probably should be removed?

src/tiled/colorbutton.h
```diff
@@ -32,10 +32,10 @@ namespace Tiled {
 class ColorButton : public QToolButton
 {
     Q_OBJECT
+    Q_PROPERTY(QColor color READ color WRITE setColor NOTIFY colorChanged)
 
 public:
     ColorButton(QWidget *parent = nullptr);
-
```
bjorn, line 38: Please don't remove this newline. :-)

src/tiled/scriptdialog.cpp
```diff
@@ -0,0 +1,319 @@
...
+    }
+    colorButton = new ColorButton(this);
+    addDialogWidget(colorButton);
+    return colorButton;
+}
+
+
+ScriptDialog::~ScriptDialog()
+{
+    close();
```
bjorn, line 236: What's the point of calling `close` upon destruction?

bjorn (commented):
Some new comments. :-)

dogboydog:
Thanks, I'll work on these changes.

src/tiled/scriptdialog.cpp
```diff
@@ -0,0 +1,319 @@
...
+        else {
+            delete item;
+        }
+    }
+}
+QLabel *ScriptDialog::addHeading(const QString &text, bool fillRow)
+{
+    QLabel * label;
+    addNewRow();
+    label = newLabel(text);
```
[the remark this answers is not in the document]
dogboydog: I think I just accidentally left them separate after the initialization was in an if statement before. Fixed in the version I'm going to push

src/tiled/scriptdialog.cpp
```diff
@@ -0,0 +1,319 @@
...
+    ColorButton *colorButton;
+    checkIfSameType("ColorButton");
+    moveToColumn2();
+    if(!labelText.isEmpty()){
+        QLabel *colorLabel = newLabel(labelText);
+        addDialogWidget(colorLabel);
+    }
+    colorButton = new ColorButton(this);
+    addDialogWidget(colorButton);
+    return colorButton;
```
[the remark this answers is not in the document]
dogboydog: Good idea. Only place it wouldn't work is for things like the color button where I'm editing the size policy to change it from the default after I call addDialogWidget(). But I can do this change soon

src/tiled/scriptdialog.cpp
```diff
@@ -0,0 +1,319 @@
...
+    m_widgetsInRow = 0;
+    m_lastWidgetTypeName.clear();
+    m_rowLayout = new QHBoxLayout();
+}
+
+void ScriptDialog::moveToColumn2()
+{
+    if (m_widgetsInRow == 0){
+        m_widgetsInRow = 1;
+    }
```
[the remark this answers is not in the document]
dogboydog: I think I fixed all instances of this

src/tiled/scriptdialog.cpp
```diff
@@ -0,0 +1,319 @@
...
+    m_lastWidgetTypeName = QString::fromUtf8(widgetTypeName);
+    return isSameType;
+}
+void ScriptDialog::addNewRow()
+{
+    m_rowIndex++;
+    m_widgetsInRow = 0;
+    m_lastWidgetTypeName.clear();
+    m_rowLayout = new QHBoxLayout();
```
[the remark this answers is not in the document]
dogboydog: Done thanks

src/tiled/scriptdialog.h
```diff
@@ -0,0 +1,99 @@
...
+    Q_INVOKABLE Tiled::FileEdit *addFilePicker(const QString &labelText = QString());
+    Q_INVOKABLE Tiled::ColorButton *addColorButton(const QString &labelText = QString());
+    Q_INVOKABLE void clear();
+    Q_INVOKABLE void addNewRow();
+private:
+    int m_rowIndex = 0;
+    int m_widgetsInRow = 0;
+    int m_leftColumnStretch = 1;
+    // stretch as much as we can so that the left column looks as close to zero width as possible when there is no content
+    int m_rightColumnStretch = 9999;
```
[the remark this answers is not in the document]
dogboydog: Made them static variables and left = 0, right =1

src/tiled/scriptdialog.cpp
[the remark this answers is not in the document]
dogboydog: Added the change to make it return the pointer. Reduces the size of a lot of the methods. thanks

src/tiled/scriptdialog.cpp
```diff
@@ -0,0 +1,270 @@
...
+    line->setFrameShape(QFrame::HLine);
+    line->setFrameShadow(QFrame::Sunken);
+    m_rowLayout->addWidget(line, rightColumnStretch); // higher stretch
+    addNewRow();
+    return line;
+}
+QLineEdit *ScriptDialog::addTextInput(const QString &labelText, const QString &defaultValue)
+{
+    checkIfSameType("QLineEdit");
+    moveToRightColumn();
```
bjorn, line 116: It seems all calls to `checkIfSameType` are followed by `moveToRightColumn`, so you might consider whether it makes sense to include `moveToRightColumn` in `checkIfSameType`.

src/tiled/scriptdialog.cpp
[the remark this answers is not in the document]
bjorn: Yeah, that's better. I noticed you added casts though, and we actually don't need them, at least not when the function is called from JS. We can just return `QWidget*` also from say `addTextInput` and it will behave exactly the same. In the TypeScript types we should still use the more specific type.

src/tiled/scriptdialog.cpp
```diff
@@ -0,0 +1,270 @@
...
+    }
+    QSlider *horizontalSlider = new QSlider(this);
+    horizontalSlider->setOrientation(Qt::Horizontal);
+    return (QSlider *)addDialogWidget(horizontalSlider);
+}
+
+QCheckBox *ScriptDialog::addCheckBox(const QString &labelText, bool defaultValue)
+{
+    checkIfSameType("QCheckBox");
+    moveToRightColumn();
```
bjorn, line 149: Actually, could these calls be merged into `addDialogWidget`? We don't need to mention the class name, since `addDialogWidget` could obtain that through `widget->metaObject()->className()`. In fact I guess we don't need to compare class names, but could just compare the `QMetaObject*`.

dogboydog:
Sneak preview of  my WIP  Image widget functionality demonstrating dynamically updating an image in a dialog. 

https://user-images.githubusercontent.com/9920963/174457679-d4fdc99f-8130-4f94-b931-d2a57415413a.mp4

dogboydog:
Just pushed the "NewRowMode" property for dialogs and image widgets, and updated the example scripts repository to use the new features. 
image

dogboydog:
Ran my example scripts on Mac. 

image
image
image

PR #3407, opened 2022-07-10
Make Terrain Brush fill full tiles mode toggle-able

fjnlpy:
Hi, I have tried to address https://github.com/mapeditor/tiled/issues/3082. I hope it is useful.

A few things I wasn't sure about:
- I don't really understand the translations so I didn't add any for the new text I added.
- I didn't make an icon for the new toolbar tool I added; I just re-used an icon fromsomething else. But it's not really appropriate, so I need some guidance about which icon to pick.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 ### Unreleased
 
 * Added support for SVG 1.2 / CSS blending modes to layers (#3932)
+* Added button to toggle Terrain Brush to full tile mode (by Finlay Pearson, #3407)
 * Added export plugin for Remixed Dungeon (by Mikhael Danilov, #4158)
 * Added "World > World Properties" menu action (with dogboydog, #4190)
 * Scripting: Added API for custom property types (with dogboydog, #3971)
--- a/src/tiled/abstracttileselectiontool.cpp
+++ b/src/tiled/abstracttileselectiontool.cpp
@@ -69,13 +69,13 @@ AbstractTileSelectionTool::AbstractTileSelectionTool(Id id,
     mActionGroup->addAction(mIntersect);
 
     connect(mReplace, &QAction::triggered,
-            [this] { mSelectionMode = mDefaultMode = Replace; });
+            this, [this] { mSelectionMode = mDefaultMode = Replace; });
     connect(mAdd, &QAction::triggered,
-            [this] { mSelectionMode = mDefaultMode = Add; });
+            this, [this] { mSelectionMode = mDefaultMode = Add; });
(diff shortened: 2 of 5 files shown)


bjorn:
@fjnlpy This is great, thanks for working on this!

Indeed we'll need a proper icon for this. I guess some schematic, either with a full tile highlighted or we could revert the logic and have an icon that indicates a smaller change will be made.

No need to worry about the translations, I'll update those when we reach "string freeze", usually 1-2 weeks before the release.

I'm not sure it makes sense to set <kbd>L</kbd> as a shortcut for this toggle. Mostly because there is already Ctrl as temporary toggle, we could just leave it empty. Users who use the toggle a lot can set the shortcut themselves.

fjnlpy:
@bjorn Thanks for responding. I removed the shortcut key and setting one from the preferences menu seems to be working ok. I'm no good at visuals and all the other icons look very nice so I wouldn't want to risk making an icon that isn't very good -- is there someone else who can make a better icon?

eishiya:
Last month, Bjorn asked about making an icon for this on the Tiled Discord, but it doesn't look like he ever shared the result with you:
TerrainToggle
This is an SVG, so it would go into the /scalable/ directory.

fjnlpy:
Hi, sorry for my delay in getting back to this, I will add the SVG and resolve conflicts soon :)

fjnlpy:
I've added the icon, seems to be working locally for me. I initially had to add to `tiled.qrc`, but after merging master I see that this file no longer exists and the icon now seems to display without it. I'm not sure what the story is there but hopefully somebody who knows about Qt can keep me right ;)

bjorn:
> I've added the icon, seems to be working locally for me.

Thanks, I'll check it out!

> I initially had to add to `tiled.qrc`, but after merging master I see that this file no longer exists and the icon now seems to display without it. I'm not sure what the story is there but hopefully somebody who knows about Qt can keep me right ;)

Yes, the creation of a resource file is taken care of by the build system since 1c1e33cae58ac728f62d5af96091afe6e86a1b85, so adding the file to the `src/tiled/resources` folder suffices. :-)

bjorn:
Hmm, I think the icon could use a bit less detail maybe:

image

I also think that while holding Ctrl to toggle the current state, it should be reflected on the button. We have another case like this for the selection modes of the `AbstractTileSelectionTool`.

bjorn:
@fjnlpy Well, sorry for taking so long with finishing this PR! I've reverted part of my own change again so we can reflect the temporary toggle by modifier on the tool bar button's checked state.

Thanks for this fine contribution!

PR #3421, opened 2022-07-22
Docs: Layer width and height are for tilelayers only

Zer0-One:


bjorn:
That's right, thanks!

Issue #3424, opened 2022-07-25
JSON tilemap load ignores repeat-x/y


ksdhans:
Loading a tilemap in *.tmj ignores the repeat x/y settings for layers

**To Reproduce**
1. Export a tilemap to *.tmj that has layers with repeat-x set for an image layer
2. Load the reulting *.tmj file in Tiled, and check if the repeat has been preserved
3. Look at the *.tmj file in a text editor to confirm that "repeatx":true, was included for that layer

**Expected behavior**
The JSON tilemap should look identical to the original *.tmx. Everything appears to be loaded correctly, except for the "repeatx":true layer attribute (which is in the JSON file, I checked)

**Specifications:**
 - Tiled Version: 1.9.0

PR #3428, opened 2022-08-03
Fixed "repeatx" and "repeaty" of ImageLayer for json

JeneLitsch:
I was able to reproduce the bug #3424 and made an attempt to fix the issue.
I added the necessary code to load the "repeatx" and the "repeaty" property of ImageLayer correctly from tmj/json.


bjorn:
That's great, thanks!

I've simplified the code a little because `toBool` will return `false` anyway when the variant is null.

JeneLitsch:
Your welcome :)

PR #3450, opened 2022-08-13
MapToVariantConverter: Write out the correct version for Tilesets

eishiya:
After the 1.8 compatibility mode was added in 1.9, `toVariant(Map)` was updated to use `FileFormat::versionString()`, but `toVariant(Tileset)` was not and was still always writing out "1.8".


bjorn:
Nice catch, and thanks for the patch!

eishiya:
It was Clepto on Discord who noticed this issue, not me. I'm just the one who had time to go digging for the cause sooner xP

PR #3466, opened 2022-09-01
Update tiled_ru.ts

solovyovs:
Several translation enhancements

### DIFF
--- a/translations/tiled_ru.ts
+++ b/translations/tiled_ru.ts
@@ -50,7 +50,7 @@
     <message>
         <location filename="../src/tiledapp/main.cpp" line="+453"/>
         <source>Export syntax is --export-map [format] &lt;source&gt; &lt;target&gt;</source>
-        <translation>Синтаксис команды экспорта такой: --export-map [format] &lt;source&gt; &lt;target&gt;</translation>
+        <translation>Синтаксис команды экспорта: --export-map [format] &lt;source&gt; &lt;target&gt;</translation>
     </message>
     <message>
         <location line="-303"/>
@@ -60,7 +60,7 @@
     <message>
         <location line="+11"/>
         <source>Non-unique file extension. Can&apos;t determine correct export format.</source>
-        <translation>Не уникальное расширение файла. Невозможно определить формат экспорта.</translation>
+        <translation>Неуникальное расширение файла. Невозможно определить формат экспорта.</translation>
     </message>
     <message>
         <location line="+7"/>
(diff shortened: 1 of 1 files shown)


translations/tiled_ru.ts
```diff
@@ -7537,13 +7537,13 @@ Please select specific format.</source>
...
         <source>Change Image Rect</source>
-        <translation type="unfinished"></translation>
+        <translation>Изменить прямоугольник изображения</translation>
     </message>
     <message>
         <location filename="../src/tiled/adjusttileindexes.cpp" line="-127"/>
         <location line="+80"/>
         <source>Adjust Tile Indexes</source>
-        <translation>Отрегулировать индексы тайлов</translation>
+        <translation>Настройте индексы тайлов</translation>
```
Gerold103, line 7546: Probably better use imperative here - *"Настроить"*.

Gerold103 (commented):
Hi! I was asked to make a review, so here are some comments. Most of them are about other places which also could be fixed alongside:
* There is this line (163): `<translation>Минимизировать экспортированный файл, убрав ненужный пробел</translation>`. I suppose it means whitespace as plural? So it would be *"убрав ненужные пробелы"*, right?
* The line 184: `<translation>Отобразить список поддерживаемых форматов экспорта</translation>`. In some other places you changed *"Отобразить"* to *"Показать"*. Shouldn't we do the same here?
* The line 194: `<translation>Экспортировать карту или набор тайлов с отсоединёнными от них шаблонами</translation>`. Sounds a little weird. I suppose the meaning was *"Экспортировать карту или набор тайлов без шаблонов"*? Or could rephrase *"Экспортировать карту или набор тайлов без включения шаблонов"*. Right?
* The line 273: `<translation>Параметры команды:</translation>`. I would omit the word *"команды"*. But up to you.
* The line 465: `<translation>Файлы правил Автокарты (*.txt)</translation>`. I am not sure what Автокарты means here. Is it literally "automatic map"? Or was it meant as auto-reflection of some sort? Shouldn't it then be *"Файл правил авто-разметки"*?
* The line 475: `<translation>JSON типов объектов (*.json)</translation>`. The conjugation seems a bit off. Shouldn't it be *"JSON-типы объектов"*? I don't know the context, but the original looks strange right now regardless of context. Same on the line 480.
* The line 671: `<translation>Отображать &amp;имена объектов</translation>`. Looks like it should be *"Отобразить"*.
* On the lines 676 and 681 the `&amp;` in the middle of the words looks wrong. Shouldn't it be before the words, like in the English version?

And now I've noticed that the file has almost 8k lines, so I will better stop here :)

bjorn:
Thank you so much for these improvements @solovyovs, and thank you @Gerold103 for the review!

@solovyovs Would you have time to extend this PR to cover additional texts and suggestions from Vladislav?

bjorn:
@Gerold103 Unfortunately it seems there will not be any updates from @solovyovs. Could you make a call, whether this is an overall improvement or not as-is, so we can either merge or close this?

solovyovs:
Hi, yes, it's overall improvement.

Gerold103:
It is better than before. So yes, could be called an improvement.

bjorn:
Alright, thanks a lot guys! :-)

PR #3476, opened 2022-09-22
add warning about QT 5 index change signal on qtcombobox

dogboydog:
Add note about problem with current index signal on Dialog's QComboBox

### DIFF
--- a/docs/scripting-doc/index.d.ts
+++ b/docs/scripting-doc/index.d.ts
@@ -472,7 +472,7 @@ declare namespace Qt {
     /**
      * Signal emitted when the user selects a different option. Provides the index
      * into the list of values for the new selection.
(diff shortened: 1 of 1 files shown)


docs/scripting-doc/index.d.ts
```diff
@@ -472,6 +472,8 @@ declare namespace Qt {
     /**
      * Signal emitted when the user selects a different option. Provides the index
      * into the list of values for the new selection.
+     * @warning Most builds of Tiled use Qt5, where this signal is bugged and `index` is the chosen text rather than the index number.
+     *          When using this event, instead of using `index`, you can get the updated index using {@link QComboBox.currentIndex}.
```
bjorn, line 476:
```suggestion
     * @warning When Tiled is compiled against Qt 5 (which includes most current releases), the `index` parameter is the chosen text rather than the actual index.
     *          Use {@link QComboBox.currentIndex} to get the selected index.
```

PR #3488, opened 2022-10-05
Add tileset tile x/y draw offset.

Chase-san:
This is a modern(ish) version of #1323. I just want to do the minimum needed to get this merged in, whatever that minimum might be. I want to avoid making the PR overly complex, so it is very bare bones at the moment.

This functionality, or something like it is important for a game I am working on.


bjorn:
@Chase-san I appreciate that you're picking up this change again! I will have a look at this tomorrow, but at a quick glance I could not see much difference from the changes at #1323 so it'd probably be fair to look into merging that PR instead. Is there anything in this PR that the older one didn't cover?

bjorn:
> The enums are not saved or encoded into a file at any point right? Seen that be an issue before, which is why I put the new values at the end.

Right, these enums are only used within the application. :-)

What remains to be fixed before this can be merged, is fixing the rendering artifacts that can appear due to the tile not occupying the area expected by for example the brush preview. We may also need a more specific change signal than `tileImageSourceChanged`.

bjorn:
@Chase-san Sorry, I ran out of time and today again I ran out of time, but regarding the artifacts, the steps I can reproduce it with are:

* Take one of the example maps, for example `desert.tmx`
* Open its tileset
* Change the origin of one of the tiles such that the tile shifts to the top-left (say, set origin to 10,10)
* Go back to the map
* Select the tile with the changed origin
* Hover the map with the brush preview
* Notice that the preview leaves artifacts of the tile when it changes position

There is already code to avoid such artifacts, which takes into account the tile offset set on a tileset. This will need to be adjusted, so that also individual origins on tiles are taken into account.

bjorn:
@Chase-san No worries, I'm currently looking into fixing this up.

bjorn:
Still to do:

* Support writing and reading tile origin for the JSON format (`maptovariantconverter.cpp` and `varianttomapconverter.cpp`)
* Support writing out tile origin in Lua export (`luaplugin.cpp`)
* Support tile origin in GameMaker 2 export (`yyplugin.cpp`)

bjorn:
> How usable is this commit at this juncture? Per-tile offset / origin is important to me too so if the progress of this feature is on hold, I'll be happy just using something minimal instead.

Unfortunately I had to take a rather long (over 2 months) break from Tiled development recently, and it's a bit hard to tell for me which state this was in. There were some issues raised (see my previous comment), but it's been a while and I need to get back into it, play around with the current state and then decide what way to go forward.

RomainMazB:
This is **the** missing feature, [first asked](https://github.com/mapeditor/tiled/issues/871) 9 years ago 😢.

Hope this will we merged soon.

stoner-jack:
Thanks for this.

- when exporting, the origin properties are lost in `tsj`.
- the properties are reserved in `lua` and `tsx`

Will keep the notes here as I test

6thpath:
Is there any updates on this?

carlislefox:
Just popping in to see how this is going, this is the final piece of the puzzle for Tiled meeting all of my needs so I've got my fingers and toes crossed that this moves forwards this year. I'm working around this using properties and handling the offsets in-engine but it'd be really nice to have this natively supported for when working in the editor.

PR #3490, opened 2022-10-05
[PR] Impl Basic functionality for LPS project on original Tiled

yuwenhuisama:
This PR has completed features as follow:
1. Json5 support for Tiled.
2. Directly drag/drop file to fill fileedit's path widget.
3. Export filepath to file with relative path instead of absolute path.
4. Event layer can only be created under `Event Layers` group.
5. Draw layer can only be created under `Render Layers` group.


bjorn:
Hey, some of these changes, like drag & drop support for `FileEdit`, look interesting to merge into Tiled. Would you be interested in opening more specific pull requests?

Other changes, like the addition of the two group layers by default, could alternatively be done using a custom script function hooked up to the [`tiled.assetCreated`](https://www.mapeditor.org/docs/scripting/modules/tiled.html#assetCreated) signal.

In general I guess some things might not be possible without forking Tiled, but I'd be interested to help you achieve any desired customizations without the need to maintain a Tiled fork, unless you prefer that approach of course, which you're welcome to do as well.

yuwenhuisama:
Hi @bjorn , sorry for just saw your reply now, these changes are for my own project initially. But if these features are useful for Tiled (the drag & drop and json5 support), I'm glad to make a PR for these features.

bjorn:
@yuwenhuisama Sorry for getting back late as well... For PR the most obvious thing would be the drag and drop support for `FileEdit`. I'd still welcome this contribution.

I'm not sure what the JSON5 support means. It seems there's only the addition of the ".json5" extension in the project model?

PR #3493, opened 2022-10-06
Fix compilation on Windows (via Qt Creator)

lhiginbotham:
As an aspiring contributor to this project, I downloaded Qt 6 and cloned the repo and right off the bat whenever trying to build the project I hit these two missing includes, resulting in compilation failure.

### DIFF
--- a/src/tiled/scriptmanager.cpp
+++ b/src/tiled/scriptmanager.cpp
@@ -59,6 +59,7 @@
 #include <QFile>
 #include <QQmlEngine>
 #include <QStandardPaths>
+#include <QStringDecoder>
 #if QT_VERSION < QT_VERSION_CHECK(6,0,0)
 #include <QTextCodec>
 #endif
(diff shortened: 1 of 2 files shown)


lhiginbotham:
@bjorn and no worries, I was able to get setup, build the code, and run tiled on Windows Subsystem for Linux

bjorn:
> whelp, on second look, seems like we cannot use Qt 6 with this project

Actually Tiled builds fine with Qt up to 6.3.2, but it seems a few new includes are needed with the just released Qt 6.4, probably due to some header dependency cleanups on the Qt side.

bjorn:
In the meantime I've pushed the necessary changes for Qt 6.4 (4d4663eed392c9be3efdf025919b5d868397fe44), so you can use latest Qt if you want. Just keep in mind that the minimum required Qt is still at 5.12 for now. :-)

PR #3502, opened 2022-10-21
Allow a tile's object group (collision objects) be cleared via script

### LINKED ISSUE #3495: Can't set Tile.objectGroup back to null
**Describe the bug**

I can't find a way to remove an ObjectGroup from a Tile.  That is, to set Tile.objectGroup back to null after setting it.  Either I'm just not finding it, or a method or something is missing.

The best I can currently do is either loop through and remove all objects from the group, or assign it a new empty group.

**To Reproduce**

```
tile.objectGroup = new ObjectGroup()
tile.objectGroup = null
```

**Expected behavior**

> The ObjectGroup associated with the tile in case collision shapes were defined. Returns null if no collision shapes were defined for this tile.

If it can be null (which it can), I should be able to set it back to null, either directly or via method.

**Specifications:**
 - OS: Windows 10
 - Tiled Version: 1.9.2

lhiginbotham:
- fixes #3495 by handling the case where a user calls `tile.objectGroup = null`, allowing the user to clear out assigned collision shape objects

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 ### Unreleased
 
 * Scripting: Fixed painting issues after changing TileLayer size (#3481)
+* Scripting: Allow assigning null to Tile.objectGroup (by Logan Higinbotham, #3495)
 * Defold plugin: Allow overriding z value also when exporting to .collection (#3214)
 * Qt 6: Fixed invisible tileset tabs when only a single tileset is open
 * Fixed compile against Qt 6.4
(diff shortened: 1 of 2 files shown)


lhiginbotham:
FYI @bjorn and @juanitogan, tested it out and it seems to work. Let me know if there are any unforeseen issues with this 😃

bjorn:
@lhiginbotham Yes, patch looked fine, thanks!

PR #3511, opened 2022-10-23
Fix point label placement to stick near the point object

### LINKED ISSUE #3400: Name label for point object is not aligned properly
**Describe the bug**
Since 1.9, and the change to the size of the point object, the name label is positioned at the top of the old point object size and not at the top of the new smaller point object.

**To Reproduce**
Create a new pointer object, shortcut I.
Give it a name. 
See how the label is not positioned as they are for other objects with name labels.

**Media**
pointer

**Specifications:**
 - OS: KDE Neon (Ubuntu 20.04)
 - Tiled Version: 1.9.0

lhiginbotham:
- fixes #3400 which notes that the labels for points are attached to a world coordinate, rather than a screenspace coordinate
	- the issue came about when points were changed from having a world-fixed size to now having a screen-fixed size, such that the annotation remains the same size when changing the zoom level of the map editor
 - one minor issue with this fix is that the painterScale used seems to be one "frame" behind whenever we are re-rendering due to a zoom level change; this corrects itself on the next non-cached repaint (for example, by panning the map editor)
- I have a prototype of a different fix where this one-frame lag doesn't occur but a different issue occurs; the bounding box of the label does not match where it is actually rendered and thus Qt graphics does not allow it to repaint after something was drawn over top of it

### DIFF
--- a/src/tiled/objectselectionitem.cpp
+++ b/src/tiled/objectselectionitem.cpp
@@ -199,16 +199,28 @@ void MapObjectLabel::syncWithMapObject(const MapRenderer &renderer)
     QPointF screenPos = renderer.pixelToScreenCoords(mObject->position());
     QRectF bounds = mObject->screenBounds(renderer);
 
-    // Adjust the bounding box for object rotation
-    bounds = rotateAt(screenPos, mObject->rotation()).mapRect(bounds);
+    if (mObject->shape() == MapObject::Point) {
+        // Use screenspace offset, since the Point annotation rescales with the zoom-level
+        mScreenspaceOffset = { 0, -bounds.height() };
 
-    // Center the object name on the object bounding box
-    QPointF pos((bounds.left() + bounds.right()) / 2, bounds.top());
+        if (auto mapScene = static_cast<MapScene*>(scene()))
+            screenPos += mapScene->absolutePositionForLayer(*mObject->objectGroup());
 
-    if (auto mapScene = static_cast<MapScene*>(scene()))
-        pos += mapScene->absolutePositionForLayer(*mObject->objectGroup());
+        setPos(screenPos);
+    } else {
+        mScreenspaceOffset = { 0, 0 };
 
-    setPos(pos);
+        // Adjust the bounding box for object rotation
+        bounds = rotateAt(screenPos, mObject->rotation()).mapRect(bounds);
+
+        // Center the object name on the object bounding box
+        QPointF pos((bounds.left() + bounds.right()) / 2, bounds.top());
+
+        if (auto mapScene = static_cast<MapScene*>(scene()))
+            pos += mapScene->absolutePositionForLayer(*mObject->objectGroup());
+
+        setPos(pos);
+    }
 
     if (mBoundingRect != boundingRect) {
         prepareGeometryChange();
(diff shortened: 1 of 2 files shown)


lhiginbotham:
FYI @bjorn, @smjnab

This fixes the issue such that the label tracks the Point object annotation better with zoom level changes.  Do note the issue that I mentioned about the scale factor being "one frame behind" during zoom-triggered repaints.  Video of that:

[label.webm](https://user-images.githubusercontent.com/14078392/197367120-d13e10eb-1854-4843-9118-a3f168168285.webm)

Maybe this fix is good enough for now since it vastly improves the previous case, especially if we do not have any ideas on how to fix the one frame lag.

src/tiled/objectselectionitem.cpp
```diff
@@ -199,16 +199,21 @@ void MapObjectLabel::syncWithMapObject(const MapRenderer &renderer)
     QPointF screenPos = renderer.pixelToScreenCoords(mObject->position());
     QRectF bounds = mObject->screenBounds(renderer);
 
-    // Adjust the bounding box for object rotation
-    bounds = rotateAt(screenPos, mObject->rotation()).mapRect(bounds);
+    if (mObject->shape() == MapObject::Point) {
+        // Use screenspace offset, since the Point annotation rescales with the zoom-level
+        setPos(screenPos - QPointF{ 0, bounds.height() / renderer.painterScale()});
```
bjorn, line 204: The problem is that the position set with `setPos` will be scaled with the view, even if this item will not. Previously this is what we wanted, but now that we don't, we should not use `setPos` to center the label above the object but rather apply an offset when rendering the label (and to the bounding rect of this item). Of course, this should still only be done for point objects, since for all other objects we do want the position of the label to be affected by the scaling.

lhiginbotham: @bjorn how does this look now?  In my other branch I used a similar approach but didn't realize that the repaint was misbehaving because the bounding box was not also translated.

src/tiled/objectselectionitem.cpp
```diff
@@ -229,13 +238,18 @@ void MapObjectLabel::updateColor()
...
+    return mBoundingRect.translated(mScreenspaceOffset).adjusted(0, 0, 1, 1);
 }
 
 void MapObjectLabel::paint(QPainter *painter,
                            const QStyleOptionGraphicsItem *,
                            QWidget *)
 {
+    if (mObject->shape() == MapObject::Point) {
+        painter->save();
+        painter->translate(mScreenspaceOffset);
```
lhiginbotham, line 250: @bjorn technically since we always have other shapes set to { 0, 0 } offset, we could remove the guards on `mObject->shape() == MapObject::Point`, but keeping the guards does save us from needlessly saving and restoring the painter transformation matrices.

bjorn: To make it a bit more future-proof, you could check `!mScreenspaceOffset.isNull()` instead of the object shape.

lhiginbotham: @bjorn done - thanks!

src/tiled/objectselectionitem.cpp
```diff
@@ -199,16 +199,25 @@ void MapObjectLabel::syncWithMapObject(const MapRenderer &renderer)
...
-    if (auto mapScene = static_cast<MapScene*>(scene()))
-        pos += mapScene->absolutePositionForLayer(*mObject->objectGroup());
+        // Adjust the bounding box for object rotation
+        bounds = rotateAt(screenPos, mObject->rotation()).mapRect(bounds);
+
+        // Center the object name on the object bounding box
+        QPointF pos((bounds.left() + bounds.right()) / 2, bounds.top());
+
+        if (auto mapScene = static_cast<MapScene*>(scene()))
+            pos += mapScene->absolutePositionForLayer(*mObject->objectGroup());
```
bjorn, line 217: This call is here to take into account layer offsets as well as the parallax offset. It should be taken into account also for point objects.

lhiginbotham: @bjorn ah okay, don't know how I missed this.  I tried it out with and without the `absolutePositionForLayer` added, it is working better now after your suggestion.  Thanks!

bjorn (approved):
Patch worked fine now, good work!

Since all rendering was done using `mBoundingRect` and `mTextPos`, I've decided to simplify things a little and embed the offset in those variables instead of using an additional member and the translations. :-)

PR #3512, opened 2022-10-24
Fix selection highlight issue for unordered layer lists set by script

### LINKED ISSUE #3451: Scripting: Assigned map.selectedLayers displays incorrectly
When assigning to `TileMap.selectedLayers`, if the assigned array has its layers in any order other than ascending by layer ID and doesn't contain every layer in the map, the layers will be selected, but the Layers panel will only highlight the bottom-most layer. I confirmed that the layers were selected via Highlight Current Layer, and via `tiled.activeAsset.selectedLayers`.

Here for example, I selected all "Invalid" layers via script, in order from bottom to top (the same order they appear in `map.layers`). The layers are sorted by their ID (which I've added to their layer name, for convenience), so everything is highlighted correctly:
image
But here, I've switched layers 2 and 3, so they're no longer in ascending order, and thus only the bottom-most selected layer is highlighted (but all three layers are actually selected):
image

Sorting the layers prior to assigning to `map.selectedLayers` works to correctly highlight all layers regardless of order:
```js
invalidLayers.sort(function(a,b) {return a.id - b.id;});
map.selectedLayers = invalidLayers;
```
Selecting every layer in the map also highlights the layers correctly, regardless of their order in the array.

lhiginbotham:
- fixes #3451 where layer selection highlighting was bugging out if the newly set selection was not ordered by layer ID
- this seems to be an event interplay issue with the following sequence:
	- script sets the layer selection to some array
	- first step in handling that is to set the currently selected layers to be the full array, triggering LayerView::selectedLayersChanged to select the whole range
	- second step is to make sure that the mCurrentLayer variable is pointing to an object within that array selection, which in this bad case it is not, so we end up calling setSelectedLayers with one item, triggering LayerView::selectedLayersChanged again with a single item and effectively clearing out the selection we had just highlighted
- fix this by doing the second step before the first step

### DIFF
--- a/src/tiled/mapdocument.cpp
+++ b/src/tiled/mapdocument.cpp
@@ -281,11 +281,11 @@ void MapDocument::switchCurrentLayer(Layer *layer)
 
 void MapDocument::switchSelectedLayers(const QList<Layer *> &layers)
 {
-    setSelectedLayers(layers);
-
     // Automatically make sure the current layer is one of the selected ones
     if (!layers.contains(mCurrentLayer))
         setCurrentLayer(layers.isEmpty() ? nullptr : layers.first());
+
+    setSelectedLayers(layers);
 }
 
 /**


lhiginbotham:
FYI @bjorn @eishiya let me know what you think about this change (you will be a better judge of this fix than I am, a newcomer 😄)

PR #3513, opened 2022-10-25
Add more "select all" capabilities for the tileset view and layer view

### LINKED ISSUE #3444: "Select All" in Tilesets panel
Sometimes, it is useful to select *all* the tiles in a tileset and place them as a brush. The only way to do this currently seems to be to click+drag across the entire tileset in the Tilesets panel. This works fine for small tilesets, but gets very time-consuming (if zoomed in) or imprecise (if zoomed out) for large tilesets.

This would be made much easier if the right-click menu in the Tilesets panel had a "Select All Tiles" option. There's probably no need to set a default shortcut for this action, as it's not a frequently-needed action ~~and because ctrl+shift+A is already taken among the defaults~~.

On a somewhat related note: There is already a similar option for layers, `ctrl+alt+A` or Layer > Select All Layers. It would be nice if this was accessible via the right+click menus in the Layers panel as well, as one's cursor will tend to be on the layers panel if they're thinking about selecting a bunch of layers.

lhiginbotham:
This PR fixes #3444 by adding a way to select all the tiles in the tileset view (useful for treating an entire tileset as a conglomerate stamp) and by adding the pre-existing "select all layers" action to the right-click context menu of the layers list.

### DIFF
--- a/src/tiled/layerdock.cpp
+++ b/src/tiled/layerdock.cpp
@@ -452,10 +452,10 @@ void LayerView::contextMenuEvent(QContextMenuEvent *event)
         menu.addAction(handler->actionDuplicateLayers());
         menu.addAction(handler->actionMergeLayersDown());
         menu.addAction(handler->actionRemoveLayers());
-        menu.addSeparator();
     }
     
     // independent of whether index is valid but maintaining positioning
+    menu.addSeparator();
     menu.addAction(handler->actionSelectAllLayers());
 
     if (proxyIndex.isValid()) {
--- a/src/tiled/layerdock.cpp
+++ b/src/tiled/layerdock.cpp
@@ -453,6 +453,12 @@ void LayerView::contextMenuEvent(QContextMenuEvent *event)
         menu.addAction(handler->actionMergeLayersDown());
         menu.addAction(handler->actionRemoveLayers());
         menu.addSeparator();
+    }
+    
+    // independent of whether index is valid but maintaining positioning
+    menu.addAction(handler->actionSelectAllLayers());
+
+    if (proxyIndex.isValid()) {
         menu.addAction(handler->actionMoveLayersUp());
         menu.addAction(handler->actionMoveLayersDown());
         menu.addSeparator();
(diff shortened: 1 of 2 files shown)


lhiginbotham:
FYI @bjorn @eishiya please take a look, thanks!

lhiginbotham:
> Nitpick: The Tileset Editor, which this edit does not touch, and the Tileset panel (TilesetView), which this edit does change, are two different things, please don't confuse them. The Tileset Editor is where you _edit_ Tilesets (add terrains, animations, etc), and the Tilesets panel is the one in the Map Editor, used for selecting tiles to put in the map. The Tileset Editor already has a working Select All (for some reason, it's greyed out in the Edit menu, but ctrl+A works just fine).
> 
> Unfortunately GitHub isn't playing nicely today and I can't download the auto-build to try, so I'll have to leave the real feedback for bjorn.

Fixed all erroneous references to "tileset editor" and am now calling it "tileset view", thanks!

src/tiled/layerdock.cpp
```diff
@@ -453,6 +453,7 @@ void LayerView::contextMenuEvent(QContextMenuEvent *event)
         menu.addAction(handler->actionMergeLayersDown());
         menu.addAction(handler->actionRemoveLayers());
         menu.addSeparator();
+        menu.addAction(handler->actionSelectAllLayers());
```
bjorn, line 456: It seems to me like it could be useful if this action showed up even if the clicked index is not valid. To keep it in the same position in the menu, I'd suggest to end the `if` block here, and just check `proxyIndex.isValid()` again for the rest of the menu.

src/tiled/tilesetview.cpp
```diff
@@ -814,6 +814,9 @@ void TilesetView::contextMenuEvent(QContextMenuEvent *event)
     connect(toggleGrid, &QAction::toggled,
             prefs, &Preferences::setShowTilesetGrid);
 
+    QAction *selectAllTiles = menu.addAction(tr("Select &All"));
```
bjorn, line 817: I'd prefer "Select &All Tiles" here, to make this action a little more specific.

lhiginbotham: @bjorn done!

src/tiled/layerdock.cpp
lhiginbotham: @bjorn done!  Feel free to remove the comment if you wish - just a heads up in case anyone is wondering why there are two blocks for the same condition

PR #3526, opened 2022-11-15
Add a user preference to disable the zoom limiting on Fit Map In View

NworbLegin:
- The discussion here https://github.com/mapeditor/tiled/issues/3524
- In summary: The ability to zoom to fit has been limited in code for large maps because of historic performance reasons. This may be unnecessary now, so having the ability to make this optional for the user might be preferable.
- An additional check box has been added to the preferences.ui view and preferencesdialog.cpp
- The parameter is exposed within mapview.h as noZoomLimitEnabled
- The new parameter is used in mapview.cpp to enable or disable the use of scale limiting

### DIFF
--- a/src/tiled/mapview.cpp
+++ b/src/tiled/mapview.cpp
@@ -59,6 +59,7 @@ using namespace Tiled;
 
 Preference<bool> MapView::ourAutoScrollingEnabled { "Interface/AutoScrolling", false };
 Preference<bool> MapView::ourSmoothScrollingEnabled { "Interface/SmoothScrolling", true };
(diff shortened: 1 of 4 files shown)


bjorn:
As I commented at https://github.com/mapeditor/tiled/issues/3524#issuecomment-1336504743, I would like to avoid adding an option for this, so I'm closing this for now. I think after some testing we can decide whether loosen the limitation or remove it.

Related issue is #3582, though this patch currently doesn't help there because it only disables the limit for the "Fit Map in View" action. And there too I would prefer not to make it an option.

PR #3560, opened 2023-02-02
Restored Tiled 1.8 file format compatibility by default

bjorn:
Since Tiled 1.9, the "type" attribute was written out as "class", for consistency with the new "class" attribute added to other elements. This change affected the TMX, JSON and Lua formats. For compatibility, a "Compatibility Version" option was introduced for projects, which could be used to restore compatibility with Tiled 1.8.

This change has unfortunately caused considerable confusion, and many existing libraries and frameworks supporting Tiled files still do not support "class" even half a year after this change. Renaming this attribute just for consistency was clearly a mistake.

This change restores the compatibility with Tiled 1.8 by default, using "class" only in new places but reverting back to the older "type" attribute for map objects and tiles. This is not consistent, but compatibility is more important than consistency.

For projects that came to rely on the Tiled 1.9 approach of using "class" consistently, a new "Tiled 1.9" compatibility mode has been introduced.

Finally, new projects no longer default to the latest version, but rather to the current version. This means that should there ever be a reason to break compatibility in the future, such breakage will not affect existing projects (when they are using a .tiled-project file at least).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 ### Unreleased
 
+* Restored Tiled 1.8 file format compatibility by default (#3560)
 * Added action search popup on Ctrl+Shift+P (with dogboydog, #3449)
 * Added file system actions also for tileset image based tilesets (#3448)
 * Fixed new layer names to be always unique (by Logan Higinbotham, #3452)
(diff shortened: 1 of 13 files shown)


bjorn:
TODO:

- [ ] Adjust TMX and JSON formats documentation and their changelogs.

PR #3562, opened 2023-02-03
AutoMapping: Skip locked layers when applying rules

### LINKED ISSUE #3544: Automapping output should be ignored if the output layer is locked
There could be a boolean added to the map or layer properties of a rule file that would override this, but I think the default should be locked layers are ignored, the opposite of what currently happens.

The current workaround is to duplicate the layer I don't want altered, run the automap, then delete the new output layer that was changed and rename my copy back to the correct layer name.

bjorn:
Added "SkipLocked" option, which can be set to false to apply rules to locked layers anyway.

I'm still a little undecided whether it might be better to call this option "IgnoreLock" and default is to `false` instead (which would have the same behavior, just different name / value).

Closes #3544


eishiya:
I think "IgnoreLock" (or maybe "IgnoreLayerLock"?) with a default value of `false` is the easiest to understand. Plus, I think `false` is generally a good default value for bools.

I am concerned about changing the default behaviour - the current behaviour is "SkipLocks" false / "IgnoreLock" true, isn't it? I would not be negatively impacted by this change, but I can't speak for others.

bjorn:
> I think "IgnoreLock" (or maybe "IgnoreLayerLock"?) with a default value of `false` is the easiest to understand. Plus, I think `false` is generally a good default value for bools.

I agree about defaulting to `false` and I'll go for "IgnoreLock" then.

> I am concerned about changing the default behaviour - the current behaviour is "SkipLocks" false / "IgnoreLock" true, isn't it? I would not be negatively impacted by this change, but I can't speak for others.

Yeah, I was weary of the breakage too, but I think this is different from the file format breakage we've struggled with in Tiled 1.9, because it's easy to adjust and adding the needed property will not impact team members who may still be on an older version of Tiled.

I also think it's worth the small break to have more intuitive behavior by default. That AutoMap will modify locked layers is basically a bug and never really intended. I think nobody considered the AutoMapping code when the layer locking feature was added (neither #734 nor #1627 mention AutoMapping).

bjorn:
> and adding the needed property will not impact team members who may still be on an older version of Tiled.

Well, I realized they will get an "unknown property" warning, though I think that's only a minor annoyance.

PR #3563, opened 2023-02-03
Added Windows layout management option

podsvirov:
The new `windowsLayout` project option is enabled by default on Windows.
If this opotion is disabled manually, installed files will follow the Unix-like FHS (Filesystem Hierarchy Standard).

*UPDATE*: I am using these changes as a patch to [add](https://github.com/msys2/MINGW-packages/pull/15464) the `tiled` package in the [MSYS2](https://www.msys2.org/) distribution.
It will be great if similar changes are accepted into the code base of the original project.

### DIFF
--- a/qbs/imports/TiledPlugin.qbs
+++ b/qbs/imports/TiledPlugin.qbs
@@ -37,7 +37,7 @@ DynamicLibrary {
     Group {
         qbs.install: true
         qbs.installDir: {
-            if (qbs.targetOS.contains("windows"))
+            if (project.windowsLayout)
                 return "plugins/tiled"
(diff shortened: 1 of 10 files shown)


bjorn (approved):
That looks like a fine option. Thanks for packaging Tiled as part of MSYS2!

bjorn:
I've rebased after hopefully fixing the compile issue.

podsvirov:
The package has been added to the distribution.  Users can choose and install any of the available binary versions of the [tiled](https://packages.msys2.org/base/mingw-w64-tiled) package.

PR #3572, opened 2023-02-13
Windows: Remove unnecessary TLS plugins and OpenSSL DLLs

bjorn:
Since Qt 6.2, there is a schannel backend TLS plugin which uses Windows API rather than OpenSSL, which should suffice for our purposes.

Also adjusted distribute.qbs to ship these plugins for Windows and macOS.

### DIFF
--- a/dist/distribute.qbs
+++ b/dist/distribute.qbs
@@ -190,6 +190,27 @@ Product {
     }
 
     Group {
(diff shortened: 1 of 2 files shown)


bjorn:
Closing this in favor of including the change in #3567.

PR #3573, opened 2023-02-13
Qt 6: Fixed visibility of status bar News button

bjorn:
Unfortunately, a regression in Qt 6.3 is causing the hiding of status bar widgets when switching editors to hide the news button.

Fix proposed in https://codereview.qt-project.org/c/qt/qtbase/+/460302

### DIFF
--- a/src/tiled/mainwindow.cpp
+++ b/src/tiled/mainwindow.cpp
@@ -33,7 +33,6 @@
 #include "actionsearch.h"
 #include "addremovetileset.h"
 #include "automappingmanager.h"
-#include "commandbutton.h"
 #include "commandmanager.h"
 #include "consoledock.h"
 #include "documentmanager.h"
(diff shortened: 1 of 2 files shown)


src/tiled/mainwindow.cpp
```diff
@@ -2501,6 +2493,12 @@ void MainWindow::currentEditorChanged(Editor *editor)
         statusBar()->removeWidget(widget);
     mEditorStatusBarWidgets.clear();
 
+    // Somehow, in Qt 6 the above hiding of widgets can hide the news button
+#if QT_VERSION > QT_VERSION_CHECK(6, 0, 0)
```
bjorn, line 2497: Not that it matters in practice since nobody is going to build Tiled against Qt 6.0.0, and I don't actually know since when this bug exists, but I meant to check `>=` here.

PR #3579, opened 2023-02-19
docs: Fix QDoubleSpinBox step value property name

jzischkau:
While working on an extension, I noticed the `tickInterval` property was not working as expected on a QDoubleSpinBox.  I checked the docs and the name is currently `singleStep`.

See https://doc.qt.io/qt-5/qdoublespinbox.html#singleStep-prop

### DIFF
--- a/docs/scripting-doc/index.d.ts
+++ b/docs/scripting-doc/index.d.ts
@@ -548,8 +548,7 @@ declare namespace Qt {
      */
     maximum: number;
     /**
-     * The amount that each tick left or right on the slider will
-     * change the value.
+     * This property holds the step value.
      */
     singleStep: number;
     /**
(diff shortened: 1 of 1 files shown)


docs/scripting-doc/index.d.ts
```diff
@@ -551,7 +551,7 @@ declare namespace Qt {
      * The amount that each tick left or right on the slider will
      * change the value.
```
bjorn, line 552: Obviously a copy/paste issue from the `QSlider`. This comment needs to be adjusted as well, could be just the short form "This property holds the step value."

jzischkau: Thanks - I didn't notice that!  Updated the comment per your suggestion.

bjorn:
Thanks for noticing this and for the fix!

PR #3599, opened 2023-03-03
Fix mistake in Search Actions where the shortcut Ctrl+Shift+P was marked as translatable 

dogboydog:
Found by @eishiya


bjorn (approved):
Thanks!

PR #3605, opened 2023-03-08
Document setStyleSheet for script Dialog widgets

dogboydog:
I didn't realize that this method  existed but it seems you can do a lot with this 

Example:
```js
 var style = `
    background-color : #0f1314;
    color : #a9a0f5;
    `;
    dialog.setStyleSheet(style);
```
Can produce: 
image

I found this because a user was asking if the foreground and background colors could be changed for dialogs. 
I added links to the documentation but I am not sure if there are limitations of what stylesheet features can be used versus what is documented in the Qt doc.


eishiya:
Rather than "adjust the stylesheet", I think better wording would be "set the stylesheet", since the entire stylesheet is replaced.

docs/scripting-doc/index.d.ts
```diff
@@ -4328,4 +4333,9 @@ declare class Dialog {
    * the provided width, it will scale itself up automatically.
    */
   minimumHeight: number;
+  /**
+   * Set the style sheet for the entire dialog.
+   * @param styleSheet The text of your style sheet. See https://doc.qt.io/qt-6/stylesheet.html and https://doc.qt.io/qt-6/stylesheet-examples.html for more information
+   */
+  setStyleSheet(styleSheet: string) : void;
```
bjorn, line 4340: Rather than documenting `setStyleSheet` here, I think we could derive `Dialog` from `Qt.QWidget`. The `minimumWidth` and `minimumHeight` properties could be moved there as well, since they're `QWidget` properties after all and should work on all widgets you can add to the dialog as well.

bjorn:
> I added links to the documentation but I am not sure if there are limitations of what stylesheet features can be used versus what is documented in the Qt doc.

Most of that stuff should be supported. However, I'm in the process of implementing a custom style for Tiled, and unfortunately the QStyle documentation notes:

> **Warning:** Qt style sheets are currently not supported for custom QStyle subclasses. We plan to address this in some future release.

(and I think this warning has been there for many years, since QStyle barely changed since Qt 4)

docs/scripting-doc/index.d.ts
bjorn: Oh, btw, I think this should just be documented as a `styleSheet : string` property.

dogboydog: Done thanks

bjorn (approved):

PR #3606, opened 2023-03-09
docs: Added new documentation about Automapping

bjorn:
Written by @eishiya, converted from HTML to MyST by @bjorn. :-)


eishiya:
In the credits section, it might be good to word "Some examples on this page use art from [...]" more specifically: "The Sidescroller Details example uses art from [...]"

My original page had vague wording because it's from my tip sheet that used several tilesets, some of them for multiple examples each. Here, we have just one tileset we should credit, and it's used for only one example, so it's easy to be specific.

PR #3610, opened 2023-03-10
Automapping docs: Improve styling, minor wording fix

eishiya:
Changed property names to **bold** from `code`, so that it's clearer what's a value or text you type and what's a property.


bjorn (approved):
Looks good to me, thanks!

PR #3615, opened 2023-03-12
Added object support to Godot exporter

Skrapion:
Now supporting objects in the Godot 4 exporter initially added in pull #3550.

Video showing it off here: https://www.youtube.com/watch?v=3UfOOdepDDw

### DIFF
--- a/src/plugins/tscn/tscnplugin.cpp
+++ b/src/plugins/tscn/tscnplugin.cpp
@@ -240,7 +240,8 @@ struct CustomDataLayer
 // Remove any special chars from a string
 static QString sanitizeSpecialChars(QString str)
 {
-    return str.replace(QRegularExpression("[^a-zA-Z0-9]"), "");
+    static QRegularExpression sanitizer("[^a-zA-Z0-9]");
+    return str.replace(sanitizer, "");
 }
 
 // For collecting information about the tilesets we're using
@@ -340,19 +341,19 @@ static void findUsedTilesets(const TileLayer *layer, AssetInfo &assetInfo)
 // Search an object layer for all object resources and save them in assetInfo
 static void findUsedObjects(const ObjectGroup *objectLayer, AssetInfo &assetInfo)
 {
+    QRegularExpression resPathValidator("^res://(.*)\\.tscn$");
+
     for (const MapObject *object : objectLayer->objects()) {
         auto resPath = object->resolvedProperty("resPath").toString();
 
-        if (resPath.isEmpty()) {            
-            if (resPath.isEmpty()) {
-                Tiled::WARNING(TscnPlugin::tr("Only objects with the resPath property will be exported"),
-                    Tiled::JumpToObject { object });
-                continue;
-            }
+        if (resPath.isEmpty()) {
+            Tiled::WARNING(TscnPlugin::tr("Only objects with the resPath property will be exported"),
+                Tiled::JumpToObject { object });
+            continue;
         }
         
         QRegularExpressionMatch match;
-        if (!resPath.contains(QRegularExpression("^res://(.*)\\.tscn$"), &match)) {
+        if (!resPath.contains(resPathValidator, &match)) {
             Tiled::ERROR(TscnPlugin::tr("resPath must be in the form of 'res://<filename>.tscn'."),
                 Tiled::JumpToObject { object });
             continue;
@@ -362,7 +363,7 @@ static void findUsedObjects(const ObjectGroup *objectLayer, AssetInfo &assetInfo
         int uniqueifier = 1;
         QString id = baseName;
 
-        // Create the objectId map such that every key and every value is unique.
+        // Create the objectId map such that every resPath has a unique ID.
         while (true) {
             // keys() is slow. If this becomes a problem, we can create a reverse map.
             auto keys = assetInfo.objectIds.keys(id);
@@ -372,9 +373,14 @@ static void findUsedObjects(const ObjectGroup *objectLayer, AssetInfo &assetInfo
                 break;
             }
 
+            // The key already exists with the right value
             if (keys[0] == resPath)
                 break;
 
+            // The baseName is based off of a file path, which is unique by definition,
+            // but because we santized it, paths like res://ab/c.tscn and res://a/bc.tscn
+            // would both get santitized into the non-unique name of abc, so we need to
+            // add a uniqueifier and try again.
             ++uniqueifier;
             id = baseName + QString::number(uniqueifier);
         }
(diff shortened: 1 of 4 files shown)


src/plugins/tscn/tscnplugin.cpp
```diff
@@ -182,6 +190,52 @@ static void findUsedTilesets(const TileLayer *layer, AssetInfo &assetInfo)
...
+        
+        QRegularExpressionMatch match;
+        if (!resPath.contains(QRegularExpression("^res://(.*)\\.tscn$"), &match)) {
+            Tiled::ERROR(TscnPlugin::tr("resPath must be in the form of 'res://<filename>.tscn'."),
+                Tiled::JumpToObject { object });
+            continue;
+        }
+        
+        QString baseName = sanitizeSpecialChars(match.captured(1));
+        int uniqueifier = 1;
```
bjorn, line 215: Since the object ID should always be unique for a given map, would it make sense to use it instead of an incremental uniqueifier?

src/plugins/tscn/tscnplugin.cpp
```diff
@@ -693,6 +774,35 @@ bool TscnPlugin::write(const Map *map, const QString &fileName, Options options)
...
+
+            device->write(formatByteString(
+                "[node name=\"%1\" parent=\".\" instance=ExtResource(\"%2\")]\n",
+                sanitizeQuotedString(name),
+                sanitizeQuotedString(assetInfo.objectIds[resPath]))
+            );
+
+            auto pos = object->position();
+            pos.setX(pos.x() + object->resolvedProperty("originX").toFloat());
+            pos.setY(pos.y() + object->resolvedProperty("originY").toFloat());
```
bjorn, line 798: It seems like it would make sense to take into account the "Object Alignment" tileset property here, which provides an easier way of changing the object origin (for tile objects), which can also keep the position of the object in Tiled the same as the position of the object in Godot.

"Object Alignment" can be used when one of its values matches the origin in Godot, and all objects in the same tileset share the same alignment.

bjorn (commented):
Looks great, thanks for continuing work on this plugin!

I've left a number of comments.

Lexpeartha:
Also it might be useful to have custom properties to rename node added as object (which I really think should be default node's name by default).
Another thought is maybe having a properties allowing to directly add child nodes aswell, for example `VelocityComponent` or `HealthComponent` which can be useful if game is built on principles of composition, but I think this delves more into discussion whether or not this is something that should be done in the engine itself or not

Lexpeartha:
@Skrapion @bjorn any news on this PR?

Blucky87:
@bjorn is there any movement on this PR?

bjorn:
@Lexpeartha @Blucky87 The original author hasn't responded to my feedback and I'm not sure if they are still interested in finishing this change. I could try to do it myself, but I am not very familiar with Godot.

If you need object support when using Tiled maps in Godot 4 now, I would suggest you give https://github.com/Kiamo2/YATI a try instead. If there is a reason to prefer Tiled's export plugin, please let me know!

src/plugins/tscn/tscnplugin.cpp
Skrapion: The ID should be unique per res path, not per object. I've updated the comments to reflect that.

Skrapion:
Finally got back to this! Thanks for your review, and your patience, @bjorn.

src/plugins/tscn/tscnplugin.cpp
Skrapion: Much simpler! (The workflow; not the code ;) )

Done.

src/plugins/tscn/tscnplugin.cpp
```diff
@@ -574,6 +648,10 @@ static void writeTileset(const Map *map, QFileDevice *device, bool isExternal, A
             sanitizeQuotedString(it.key()),
             sanitizeQuotedString(it->id)));
     }
+
+    if (!isExternal)
+        writeExtObjects(device, assetInfo);
```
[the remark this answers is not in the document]
Skrapion: I'll move it out of the function call. I think I just put it there because I was trying to match Godot's files, which put the Texture2D resources first, but the order doesn't actually matter.

src/plugins/tscn/tscnplugin.cpp
```diff
@@ -932,6 +1018,93 @@ bool TscnPlugin::write(const Map *map, const QString &fileName, Options options)
...
+
+                if (alignment == Alignment::Unspecified) {
+                    if (tileset->orientation() == Tileset::Orientation::Isometric)
+                        alignment = Alignment::Bottom;
+                    else
+                        alignment = Alignment::BottomLeft;
+                }
+            } else {
+                alignment = Alignment::BottomLeft;
+            }
```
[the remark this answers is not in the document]
Skrapion: No good reason; I just didn't know it existed :)

src/plugins/tscn/tscnplugin.cpp
```diff
@@ -1037,9 +1037,67 @@ bool TscnPlugin::write(const Map *map, const QString &fileName, Options options)
...
+                break;
+            case Alignment::BottomLeft:
+            case Alignment::Bottom:
+            case Alignment::BottomRight:
+                pos.setY(object->position().y() - object->height()/2);
+                break;
+            case Alignment::Unspecified:
+                // Suppress warning
+                break;
+            }
```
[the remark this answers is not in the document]
Skrapion: I also think Tiled::alignmentOffset should work. I'll have a go at it.

And yes, objects in Godot are centre aligned, and I don't believe there's any way to pick any other alignment.

src/plugins/tscn/tscnplugin.cpp
Skrapion: I made this change in the last commit. I could take it or leave it. The pros:

* This reduces the line found from 14 to 8.
* This is more efficient.

The cons:

* The generated files are less human readable.
* It makes the files less git-friendly. For instance, given:

```
[node name="Object8" parent="." instance=ExtResource("object_resource_0")]
position = Vector2(320, 10)

[node name="Object9" parent="." instance=ExtResource("object_resource_1")]
position = Vector2(125, 29.5)

[node name="Object10" parent="." instance=ExtResource("object_resource_2")]
position = Vector2(172, 29.5)
```

If you delete the first object from the map, then every object afterwards will change, because object_resource_1 will be renamed to object_resource_0, etc.

With the old code, it would remove Object8 and leave it at that.

Skrapion:
Okay, this should be good to go as soon as CI completes. But whether we use commit 6bde51c or dd6eb6d requires a judgement call. See [the review](https://github.com/mapeditor/tiled/pull/3615/commits/0caa9e55efb1624ccb3212fd462d1a6c545e7fa6#r1496776448)

src/plugins/tscn/tscnplugin.cpp
bjorn: I liked the simplification, but I think you're concerns are valid and likely the previous approach will be preferred by some.

bjorn:
@Skrapion Thanks a lot for finishing this nice addition!

PR #3618, opened 2023-03-14
General Brush/Tool fixes

AtlaStar:
Fixes an issue with the tile selection tool where trying to place a copied selection on staggered indices would cause misalignment, and fixes a misalignment issue when using corner/mixed terrain sets in a staggered isometric map.

### DIFF
--- a/src/tiled/tileselectiontool.cpp
+++ b/src/tiled/tileselectiontool.cpp
@@ -46,8 +46,9 @@ TileSelectionTool::TileSelectionTool(QObject *parent)
 
 void TileSelectionTool::tilePositionChanged(QPoint)
 {
+
     if (mSelecting)
-        brushItem()->setTileRegion(selectedArea());
+        brushItem()->setTileRegion(QRegion(selectedArea()));
 }
 
 void TileSelectionTool::updateStatusInfo()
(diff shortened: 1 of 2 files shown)


AtlaStar:
Tested some of the issues noted in the StaggeredRenderer class to see what still does/does not exist after this PR is applied, although some of those issues could have been solved prior (didn't fully test) as it appears that some of those issues may have been related to the fixed brushes.

* Fill Tool: seems to work fine
* Stamp Tool: right click selections work fine. Line and circle tool still need work. Selections made in the tileset window are offset and do not have parity with non-staggered isometric.
* Rectangle Select Tool: appears to work fine
* Map Axis and stagger index change: Appears that if changing from a Y to X axis, the current stamp selection isn't modified properly. Tested going from Y to X with an odd stagger index, but the stamp reflects the change as if the stagger index were also even. Thus switching to an even stagger index as well causes the saved selection to match the translated location it was copied from, hinting at where to look into a fix.
* Map Resize and Offset actions: unable to discern the original issue, but possibly related to the above?

If there is anymore work that should be done before the PR is accepted, or if it is preferred to attempt to solve some of these existing issues first, just let me know.

bjorn:
I'm closing this since these issues should be fixed by #3628 and #3624.

Issue #2902, opened 2020-09-29
Add scripting API for the project


bjorn:
Scripts should be able to access the files in the project and the project properties.

Split off from #1665.

jonny64bit:
So i could do with this feature for my use case. 

I want to be able to open every map in a project and run a set of Actions which automate a whole bunch of stuff.  Mostly auto mapping.

For the current work around im having to hardcode the map paths into the triggering action.

````ts
/// <reference types="@mapeditor/tiled-api" />

const automapAllMaps = tiled.registerAction("AutomapAllMaps", function(action) {
    tiled.trigger("CloseAll");

    processMap("X:\\Source\\MangoDungeon\\Mango\\Assets\\Tiled\\Maps\\Test.tmx");
});

function processMap(path:string) {
    tiled.open(path);
    tiled.trigger("AutomapExtra");
    tiled.trigger("Close");
}

automapAllMaps.text = "Automap All Maps";
automapAllMaps.checkable = false;

tiled.extendMenu("Edit", [
    { action: "AutomapAllMaps" }
]);`

bjorn:
> For the current work around im having to hardcode the map paths into the triggering action.

Just one thing to note, there is API available to access [a list of files in a folder](https://www.mapeditor.org/docs/scripting/interfaces/File.html#directoryEntries), so you may only need to hardcode the project path instead of the paths to each map.

Since you're applying AutoMapping to all files in your project, I hope you noticed the [Tiled 1.9 Alpha](https://github.com/mapeditor/tiled/releases/tag/v1.8.90) release that greatly improves the speed of AutoMapping. Any feedback regarding AutoMapping is very welcome.

solerante:
from #3527 - I also have a use case for this. I would like to get the path to the current project. I'm working on an extension to edit maps for Cataclysm: Dark Days Ahead. It involves importing tilesets and mapping the game's IDs to each sprite and preparing maps with data from the game files.

I generate tilesets, maps, and supplemental files to the project folder which I prompt from the user and then store in a config file to autofill next time the command is run for slight convenience.

eishiya:
I think I've asked for this elsewhere, but I think it makes sense being included here: the Project would be a very convenient place for scripts to store script configuration, which at present has to be either set every time the user restart Tiled, hard-coded by the user, or stored in config files that clutter the user's directories. To this end, I'd like to be able to set Custom Properties or something like them on the project, both via scripts and via the GUI.

[...]

PR #3622, opened 2023-03-17
Expose basic properties of Tiled projects to scripting

dogboydog:
related to #2902

### DIFF
--- a/docs/scripting-doc/index.d.ts
+++ b/docs/scripting-doc/index.d.ts
@@ -1069,7 +1069,13 @@ declare class TiledObject {
  * the assets belonging to that project.
  */
 declare class Project {
-
+/**
+ * A project-specific directory where you can put Tiled extensions.
+ *  It defaults to simply extensions, so when you have a directory called “extensions” alongside
+ *  your project file it will be picked up automatically.
+ *  The directory is loaded in addition to the global extensions.
+ */
+  extensionsPath: string;
 }
 /**
  * Defines the font used to render objects which have {@link MapObject.shape}
--- a/docs/scripting-doc/index.d.ts
+++ b/docs/scripting-doc/index.d.ts
@@ -1076,6 +1076,14 @@ declare class Project {
  *  The directory is loaded in addition to the global extensions.
  */
   extensionsPath: string;
+  /**
+   * Path to a file where object custom types are stored.
+   */
+  objectTypesFile: string;
+  /**
+   * Path to the file where automapping rules are stored for this project.
+   */
+  automappingRulesFile: string;
 }
 /**
  * Defines the font used to render objects which have {@link MapObject.shape}
--- a/docs/scripting-doc/index.d.ts
+++ b/docs/scripting-doc/index.d.ts
@@ -3585,9 +3585,10 @@ declare namespace tiled {
   export let activeAsset: Asset | null;
 
   /**
-   * Currently opened project, or `null` if no project is open.
+   * Currently opened project. If no project is open, the properties of the project
+   * will be blank.
    */
-  export let project: Project | null;
+  export let project: Project;
   /**
    * List of currently opened {@link Asset | assets}.
    */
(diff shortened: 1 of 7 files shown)


docs/scripting-doc/index.d.ts
```diff
@@ -1064,6 +1064,31 @@ declare class TiledObject {
...
+ * A project-specific directory where you can put Tiled extensions.
+ *  It defaults to simply extensions, so when you have a directory called “extensions” alongside
+ *  your project file it will be picked up automatically.
+ *  The directory is loaded in addition to the global extensions.
+ */
+  extensionsPath: string;
+  /**
+   * Path to a file where object custom types are stored.
+   */
+  objectTypesFile: string;
```
bjorn, line 1082: Ah, you'll find this property is always an empty string, and there is no way to set it in the UI anymore, so this can be removed.

It only exists for compatibility reasons. Any types referenced here will be loaded and merged into `project.propertyTypes` in `ProjectManager::setProject`, after which the property is cleared.

src/tiled/editableproject.h
```diff
@@ -0,0 +1,50 @@
...
+    Q_OBJECT
+    Q_PROPERTY(QString extensionsPath READ extensionsPath)
+    Q_PROPERTY(QString objectTypesFile READ objectTypesFile)
+    Q_PROPERTY(QString automappingRulesFile READ automappingRulesFile)
+    Q_PROPERTY(QStringList folders READ folders)
+
+    QString extensionsPath() const;
+    QString objectTypesFile() const;
+    QString automappingRulesFile() const;
+    QStringList folders() const;
```
bjorn, line 39: Let's make these access functions public by moving them below the constructor. There is no reason to keep them private.

src/tiled/editableproject.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+EditableProject::EditableProject(Project *project, QObject *parent)
+    : QObject(parent),
+      mProject(project)
+{
+}
+QString EditableProject::extensionsPath() const
+{
+   QString path;
+   if (mProject != nullptr)
```
bjorn, line 33: These `nullptr` checks are not necessary.

docs/scripting-doc/index.d.ts
dogboydog: Okay. So i'll remove it from the index.d.ts and from the c++ code.

src/tiled/editableproject.cpp
dogboydog: Thanks, removed

src/tiled/editableproject.h
dogboydog: Done

dogboydog:
Eishiya and I weren't sure if it makes sense to add the project file path to `tiled.project.fileName` since there is already `tiled.projectFilePath`

src/tiled/scriptmodule.cpp
```diff
@@ -651,6 +651,10 @@ void ScriptModule::currentDocumentChanged(Document *document)
     emit activeAssetChanged(document ? document->editable() : nullptr);
 }
 
+EditableProject *ScriptModule::project() const
+{
+    return new EditableProject(&ProjectManager::instance()->project(), MainWindow::maybeInstance());
```
bjorn, line 656: Why is the parent of the `EditableProject` the `MainWindow`? I think that will mean that these instances are never deleted, and essentially leaking every time the `project` property is accessed.

I think we have two options:

* We instantiate one `EditableProject` in the `ScriptModule`, owned by the `ScriptModule`. We can do this on-demand, remember the instance in a member variable so that we can return the same instance the next time.
* We set ownership to the JS engine, in which case it can be cleaned up by the garbage collector.

I think I'd prefer the first option in this case, because we only need one instance. The `Project*` is never going to be different after all, since it always points to the `Project` member of the `ProjectModel`.

dogboydog: Just my misunderstanding. I'll try your first option.

PR #3624, opened 2023-03-20
Tweaked handling of Hexagonal (Staggered) maps by Terrain Brush

### LINKED ISSUE #3617: Terrain sets do not work properly with staggered isometric maps
As the title says, a misalignment occurs when using the terrain brush causing the tiles to not properly conform to the expected shape while using a corner set.

1. Open the isometric_grass_and_water.tmx
2. Change the map properties to Isometric (staggered)
3. Try to use the premade corner set with your Terrain brush.

This is what happens
image

This is what should happen
image

 Tiled Version: built from master, using Qt 6.2.4
 
 Issue appears to be related to the `WangBrush::mouseMoved` function, as for corner and mixed sets it tries to perform some additional logic when a staggered renderer is being used. Simply removing that logic fixes the issues with the corner set, and appears to fix the issues with a mixed set as well (with limited testing, likely requires more)

bjorn:
Now Hexagonal (Staggered) maps are treated the same way as Isometric (Staggered) maps, which might be better than treating them as orthogonal maps. It at least fixes the behavior when "hex side length" is 0, but it does not mean hexagonal maps are actually supported by the Terrain Brush.

Closes #3617

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 ### Unreleased
 
 * Fixed object labels to adjust to application font changes
+* Improved Terrain Brush for Hexagonal (Staggered) maps with side length 0 (#3617)
 * Qt 6: Increased the image allocation limit from 128 MB to 1 GB (#3616)
 
 ### Tiled 1.10.0 (10 March 2023)
--- a/src/tiled/bucketfilltool.cpp
+++ b/src/tiled/bucketfilltool.cpp
@@ -23,17 +23,14 @@
 
 #include "bucketfilltool.h"
 
-#include "addremovetileset.h"
 #include "brushitem.h"
 #include "tilepainter.h"
-#include "tile.h"
 #include "tilelayer.h"
 #include "mapdocument.h"
-#include "painttilelayer.h"
-#include "staggeredrenderer.h"
 #include "stampactions.h"
 
-#include <QApplication>
+#include <QCoreApplication>
+#include <QUndoStack>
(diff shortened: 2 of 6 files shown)


NEWS.md
```diff
@@ -1,6 +1,7 @@
 ### Unreleased
 
 * Fixed object labels to adjust to application font changes
+* Tweaked handling of Hexagonal (Staggered) maps by Terrain Brush (#3617)
```
eishiya, line 4: I think this wording might mislead people who only read the changelog into thinking Terrains actually work with hex maps. I think it would be good to make it clear here that this does not add general hexagonal Terrain support, since the only immediately useful feature of this change is that side length 0 maps work nicely.

bjorn: Yeah, I thought the same thing, but could not think of a wording that would make that clear, which would not be overly verbose. It might be better to just leave out the changelog entry?

eishiya: I think it's best to include changes that affect behaviour, even if they're not meaningful to most people.
Maybe like this?
> Improved Terrain Brush for isometric-like Hexagonal (Staggered) maps with side length 0.

bjorn: I've left out the "isometric-like" since I think it's still too verbose. I hope adding "with side length 0" is specific enough, heh.

eishiya: I included it in there just to emphasise how much of an edge-case this is xP

PR #3625, opened 2023-03-20
Fixed grid rendering for odd Hex Side Length values

### LINKED ISSUE #3623: Hexagonal grid with odd side lengths is not drawn correctly
When drawing the grid on a "Hexagonal (Staggered)" map, if the side length is an odd number, the grid is drawn incorrectly. Here are screenshots of such a grid on a map with 16x16 tiles and hex side length 1:
Stagger axis Y:
image
Stagger axis X:
image

**Expected** appearance (faked by doubling all the sizes, i.e. 32x32 with side length 2, and zooming out):
Stagger axis Y:
image
Stagger axis X:
image

Even side lengths look just fine, it's just odd ones that are weird.

bjorn:
Closes #3623

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 ### Unreleased
 
 * Fixed object labels to adjust to application font changes
+* Fixed grid rendering for odd Hex Side Length values (#3623)
 * Improved Terrain Brush for Hexagonal (Staggered) maps with side length 0 (#3617)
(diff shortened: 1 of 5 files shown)


bjorn:
I noticed there is a bit of a more serious issue which is probably worth fixing while we're at it, which is that grid rendering can lock up on infinite maps when a negative side length is used that makes the rows/columns overlap.

Issue #3626, opened 2023-03-20
Screen pan optionally with Space + Leftclick


UliAbo:
**AsIs:**
- Panning screen in map requires holding down just "Space". 

**ToBe:**
- Option in preferences: Require click and drag for panning with Space.

Would love to have that (like it's default in Aseprite in Photoshop and also in Unity)! :)

PR #3637, opened 2023-03-28
Make panning with Space require pressing a mouse button as well

bjorn:
At the same time, you can now also use Space panning with the tileset view.

### DIFF
--- a/src/tiled/mainwindow.cpp
+++ b/src/tiled/mainwindow.cpp
@@ -62,7 +62,6 @@
 #include "resizedialog.h"
 #include "scriptmanager.h"
 #include "sentryhelper.h"
-#include "stylehelper.h"
 #include "templatesdock.h"
 #include "tileset.h"
 #include "tilesetdock.h"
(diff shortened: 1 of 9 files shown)


bjorn:
Issues to look into based on feedback from @eishiya:

- [ ] Update the preview when the user releases Space, since the mouse can be in a different position.
- [ ] Hide the preview while in panning mode.

bjorn:
I've decided to try a different approach than hiding the preview and updating its position. Now the view stays interactive (hover and tools still update) while <kbd>Space</kbd> is pressed and only becomes non-interactive when actually panning.

bjorn:
I noticed one more annoying side-effect of this change. Now it is no longer possible to pan the view while painting tiles or moving objects, since it requires releasing the mouse button and then re-pressing it to start the pan, and then releasing again and re-pressing to resume what you were doing.

In addition, pressing <kbd>Space</kbd> while you're already doing something else with the mouse makes the "open hand" cursor appear, but without a way to trigger the panning since the mouse button is already pressed, this would be undesirable.

I think it would be nice to keep the existing behavior, where panning immediately activates and pauses the current operation, when <kbd>Space</kbd> is pressed while already holding the left mouse button.

bjorn:
> I managed to somehow keep panning after releasing space, but can't repeat it. Clicking again fixed it, so it wasn't a huge problem.

You mean after releasing both space and the mouse button, or did you still have the mouse button pressed? In the latter case it's intended that you keep panning.

> MapView and TilesetView behave inconsistently: If I told space while selecting with LMB in the Map Editor, my selection will stop updating until I let go of space, but in the Tileset Editor and Tilesets panel, the selection will keep updating as I pan.

I'll see if I can find some fix for this.

> Perhaps this PR should be linked to #3626?

Definitely, linked now. Thanks for reminding about that one!

eishiya:
> You mean after releasing both space and the mouse button, or did you still have the mouse button pressed? In the latter case it's intended that you keep panning.

After releasing the mouse button and space, but since I can't repeat it, it's possible I misremembered.

bjorn:
Not happy with all the duplication, I've decided to introduce a `PanningViewHelper` class that is shared between the `MapView` and `TilesetView`. The issue with the tileset view selection still getting updated should also be gone now.

Issue #3652, opened 2023-04-09
Include custom tile properties in .tscn (Godot 4) export


keharriso:
**Is your feature request related to a problem? Please describe.**
When exporting a tilemap as a .tscn (Godot 4) file, the custom tile properties of the tileset aren't present in the output.

**Describe the solution you'd like**
I would like to see the tile properties be added to the .tscn file as a custom property layer of type Dictionary. Supported property types like strings, numbers, and booleans would automatically be added to the generated .tscn file under the `sub_resource type="TileSetAtlasSource"` tag.

*I am working on implementing this feature myself, but would also appreciate any insight from people with a better understanding of the project.*

bjorn:
Implemented in #3653.

PR #3653, opened 2023-04-10
Add custom tile properties to Godot .tscn export format

keharriso:
This is my attempt at adding feature #3652.

Tile properties are exported as custom property layer 0, named "Properties".

Allows code like: `get_custom_data("Properties").get(<Property>)`

Feedback welcome.

### DIFF
--- a/src/plugins/tscn/tscnplugin.cpp
+++ b/src/plugins/tscn/tscnplugin.cpp
@@ -216,9 +216,9 @@ static VariantType variantType(const QVariant &value)
 
     default:
         if (value.userType() == filePathTypeId()) {
-            // todo
+            return TYPE_STRING;
         } else if (value.userType() == objectRefTypeId()) {
-            // todo
+            return TYPE_INT;
         } else if (value.userType() == propertyValueId()) {
             const auto propertyValue = value.value<PropertyValue>();
             if (propertyValue.type()->isClass()) {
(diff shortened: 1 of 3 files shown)


src/plugins/tscn/tscnplugin.cpp
```diff
@@ -454,6 +454,28 @@ static void writeTileset(const Map *map, QFileDevice *device, bool isExternal, A
                         device->write(formatByteString("%1/transpose = true\n", tileName));
 
                     foundCollisions |= exportTileCollisions(device, tile, tileName, alt);
+
+                    // Custom properties
+                    const Properties &properties = tile->properties();
+                    const auto keys = properties.keys();
```
bjorn, line 460: Instead of allocating an array for the keys we could iterate the properties directly by using:
```c++
QMapIterator<QString,QVariant> it(tile->properties());
while (it.hasNext()) {
    auto value = it.next().value();
    // access it.key() here
}
```
Also, we may want to use `tiled->resolvedProperties()`, if we also want to get the property values that the tile may have inherited from its class.

bjorn (commented):
Overall it looks fine to me, but it would be good if @Skrapion would also have a look since he wrote this plugin.

Also, I personally can't really judge the Godot part since I don't know if there would be any other ways of associating custom data with tiles.

keharriso:
I implemented your feedback and added support for custom classes and enums.

bjorn:
> Allows code like: `get_custom_data("Properties").get(<Property>)`

Just wondering, but this double call seems a bit superfluous. Could we also make `get_custom_data(<Property>)` work? Or would it be undesirable to add a layer for each property?

bjorn:
Just for your information, I'm currently working on making the changes to achieve the above change in the export, but I don't know if I will be able to finish it today.

bjorn:
@keharriso I think I managed to get it to work. It would be great if you could review my change and do some testing as well. :-)

There are some todo's left in the code regarding support for QColor, FilePath and ObjectRef properties. It would be great to handle those in case they have any equivalent in Godot (QColor would certainly have, and FilePath could probably be written as string).

bjorn:
@keharriso Thank you very much for getting this change started! I think it turned out quite well now so I've updated the news / docs and merged this for Tiled 1.10.2!

Since you've got some experience with this plugin and with Godot, maybe you could help get #3615 finished as well? @Skrapion has been silent since opening the PR so I'm not sure we can still expect feedback from them, but it's another important feature of the exporter that needs testing and polish.

PR #3665, opened 2023-04-18
docs: Added missing MapObject constructor

bjorn:
The constructor overload taking a shape was missing. It appears to have been available since Tiled 1.3, when the scripting API was added.


bjorn:
It would be nice to add a GitHub Action that actually builds the scripting documentation, which would help avoid syntax errors and such.

PR #3666, opened 2023-04-18
Fixed the entry limit for enums used as flags to be 31

### LINKED ISSUE #3658: wrong export if enum as flags have more than 30 values
**Describe the bug**
If add more than 30 values (31 or 32) to enum when checked "Allow multiple values" (act ass flags) exported to json values are wrong.
There is possible overflow on 30 bit.

**Question**
Why only 32 values if we have today 64 bit computers? Is there any masochist here who programs games on 32 bit windows? 😏 

**To Reproduce**
- create enum
- check "Allow multiple values"
- add 31 or 32 values 
- assign this enum to object
- check some values
- export to json
- see exported file

**Expected behavior**
right export or unable to add more than 30 values as flags

**Specifications:**
 - OS: Windows 10
 - Tiled Version: 1.10.1

bjorn:
Due to the use of a 32-bit int to store the flags, the maximum number of flags is actually 30 rather than 32. When more entries are added, the editing of enum values breaks entirely.

The translations have been updated to use a parameter for the maximum number of flags in an enum, to make it easier to adjust this value when it might be increased in the future.

Closes #3658

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,6 +11,7 @@
 * Scripting: Added WangSet.effectiveTypeForColor
 * Fixed crash when changing file property of custom class (#3783)
 * Fixed loading of invalid color properties (#3793)
+* Fixed handling of enum values with 31 flags and fixed the applied limit (#3658)
 * Fixed object preview position with parallax factor on group layer (#3669)
 * Fixed hover highlight rendering with active parallax factor (#3669)
 * Fixed updating of object selection outlines when changing parallax factor (#3669)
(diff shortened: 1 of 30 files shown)


bjorn:
As noted by @eishiya, the expected maximum number of supported flags is actually 31, so we still need to do some research on why the problems already start at 31 flags. The documentation added in 2a8fc6e9153643eee61d3638ace0458ff5e62970 should also be clarified, since contrary to what I wrote in the commit message, the value of the 31st flag should actually be 1073741824 (2^30), given that the 1st flag has value 1 (2^0).

PR #3667, opened 2023-04-19
Project properties

dogboydog:
Projects now support custom properties via scripting and UI thanks to collaboration with @bjorn  

Example scripts: https://gist.github.com/dogboydog/87e294ab14c17d6e2188713de9c36d15

Fix https://github.com/mapeditor/tiled/issues/2903

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 ### Unreleased
 
+* Added support for setting custom properties on the project (#2903)
 * Removed Space and Ctrl+Space shortcuts from Layers view to avoid conflict with panning (#3672)
 * Fixed object preview position with parallax factor on group layer (#3669)
 * Fixed hover highlight rendering with active parallax factor (#3669)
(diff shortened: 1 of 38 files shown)


dogboydog:
Thanks for the fixes. I did some testing on this version. 

Some example scripts: 
https://gist.github.com/dogboydog/87e294ab14c17d6e2188713de9c36d15

Running ProjectReadProperties.js is able to log all properties as expected
image

Running AddProjectProperties works while Tiled is running:

image

 but then closing Tiled and reopening shows that the properties were not saved when the properties were modified via script ( I think this is where you were considering hooking into the undo stack to automatically save the project?)
image

Deleting properties from the UI worked  and persisted the deletion after restart

I was also able to add properties of custom enum & class types from my existing project
image

bjorn:
> but then closing Tiled and reopening shows that the properties were not saved when the properties were modified via script ( I think this is where you were considering hooking into the undo stack to automatically save the project?)

Good catch, I'll have a look into that today. Thanks for the thorough testing!

dogboydog:
Your last change worked for me. I ran the script  on a project with no properties and then closed Tiled without opening the Project Properties popup. When I restarted, the new properties were still there.

PR #3680, opened 2023-05-06
doc: Mention OCaml tmx library

fishyfriend:
Here's a [new OCaml library](http://github.com/fishyfriend/tmx) for reading TMX data files.

Thanks,
Jacob


bjorn:
Looks great, thanks for making this library!

PR #3688, opened 2023-05-11
GitHub Actions: Updated builds to Qt 6.5

bjorn:


bjorn:
Seems we're missing a new dependency on Linux:
```
[qt/stdout] Deploying dependencies for ELF file /opt/Qt/6.5.0/gcc_64/plugins/platforms/libqxcb.so 
[qt/stdout] ERROR: Could not find dependency: libxcb-cursor.so.0 
```

Issue #3707, opened 2023-05-17
Add native support for Apple Silicon


Majirefy:
Since Tiled is now complied in QT 6, I believe it's good time to add native support for Apple Silicon.

Now Tiled is in x86 on macOS, shall we make it a universal app with native support both for x86(Intel) and Apple Silicon Macs?

bjorn:
Yes, good idea! Is this something you can help with? Otherwise I'll have a look into it next week. I just don't have this hardware myself so I won't be able to test it.

bjorn:
Here are the commands being used to compile Tiled for macOS:

https://github.com/mapeditor/tiled/blob/062145f77b5489601a8c16677b0a7df5a40b704a/.github/workflows/packages.yml#L208-L234

What would need to be added is something like:

* The Apple Silicon binary release of Qt would need to be installed as well. This may be done already, if the Qt binaries for macOS ship as universal in the first place. It seems like this is the case, if "ARM64" refers to Apple Silicon in the files at https://mirror.accum.se/mirror/qt.io/qtproject/online/qtsdkrepository/mac_x64/desktop/qt6_650/qt.qt6.650.clang_64/.
* Qbs would need to be somehow instructed to build a universal app. [This page](https://doc.qt.io/qbs/qbs-target-macos.html) has some information about multiple architectures which might apply.

iplanetcn:
Hope to support apple silicon as soon as possible.

bjorn:
There is now a universal build available at https://github.com/mapeditor/tiled/actions/runs/5393521019 (the macOS 10.14+ artifact). It is not signed + notarized for now, but any testing would be really appreciated!

PR #3754, opened 2023-05-21
macOS: Try creating an x64_64 + arm64 fat binary

bjorn:
Issue #3707


xentripetal:
Tested latest build artifact on arm64 mac and confirmed working

Issue #3654, opened 2023-04-10
Switching tilesets overwrites current brush


eishiya:
When switching tilesets in the Tilesets panel, the brush is set to the currently selected tiles in that tileset. This overwrites any currently set brush, and if the tileset previously had multiple tiles selected, as happens when painting with a larger brush, it's overwritten by the arrangement of those tiles in the tileset, which is rarely useful.

![brushoverwrite](https://user-images.githubusercontent.com/70545646/230807244-4065a2a2-a9f8-4599-a1a6-f37e00ca32b1.gif)

**To Reproduce**
This requires that at least two tilesets are available in the Tilesets panel.
1. Sample some tiles from a map into a brush, or choose a saved stamp. For the clearest results, this brush should consist of tiles that aren't all together in the tileset.
2. Go to another tileset tab.
3. Observe that the brush you just spent time making has been overwritten by whatever tiles were selected in that tileset.
4. Switch back to the previous tileset.
5. Observe that your brush is *again* changed, this time to the relevant tiles from your sampled brush, but as they occur in the tileset rather than your original brush.

**Expected behaviour**
Switching to a different tileset should not modify the brush if doing so would be destructive or would be likely to produce a garbage brush.
I can think of two concrete ways to make this decision:
- Simple: Do not modify the current brush if it has an area larger than 1x1 or has more than one layer.
This would prevent complex brushes from being overwritten, but would also result in "false positives" - for example, if I select a 3x3 tree in Tileset A and switch to Tileset B where previously had a 5x5 house selected, I probably wouldn't mind having my tree overwritten by that house, and then having the house overwritten by the tree when I switch back to Tileset A. It would also overwrite 1x1 brushes that have been flipped/rotated, but this is not a big loss.
- Complex: Do not modify the current brush if the tile(s) in the newly selected tileset and/or in the current brush were selected by means other than manually selecting from a tileset or sampling a single tile from a map/selecting a single-tile, single-layer stamp. Brushes with multiple layers or using tiles from multiple tilesets should never be overwritten.
This would prevent the aforementioned false positives, because it would allow Tiled to distinguish between multi-tile arrangements selected from the tileset by the user (=probably sensible) from multiple tiles that are selected from being in some other arrangement in a map or stamp.
This one would probably not overwrite any brush that has been flipped, for better or worse.
This one can be further improved by counting map selections and saved stamps that are identical to the tiles' arrangement in the tileset as direct selections.

[...]

PR #3768, opened 2023-06-29
Share the selection between Tilesets view and Tileset Editor

bjorn:
Now `TilesetDocument` owns a single `TilesetModel` and associated `QItemSelectionModel`, which are used by the `TilesetView` instances in the `TilesetDock` as well as in the `TilesetEditor`.


bjorn:
@eishiya noted on Discord that it would be good to address issue #3654 before making a change like this, especially if it's non-optional.

bjorn:
Patch was broken during one of the rebases. A helpful message was printed to the console when opening a tileset in the Tileset Editor:
```
QAbstractItemView::setSelectionModel() failed: Trying to set a selection model, which works on a different model than the view.
```
Of course we needed to use the model from the `TilesetDocument`.

PR #3774, opened 2023-07-04
Improve WangBrush and TileLayerWangEdit behavior with empty color

bjorn:
Mostly removing the special handling for erasing and changing the way filling is implemented. Should also improve the new TileLayerWangEdit behavior when erasing.


bjorn:
Not sure what went wrong here, but GitHub provided me only with a rebase button that executed directly, even though I wanted to squash the commits. I've force-pushed the squashed version now.

PR #3777, opened 2023-07-05
add FileEdit.isDirectory and fix @link tags in scripting doc

dogboydog:


bjorn:
While you're at it, could you also document the `FileEdit.filter` property? :-)

Something like:

> When specified, only files that match the filter are shown. The filter is given in a format like "Images (*.png *.xpm *.jpg)". If you want multiple filters, separate them with ';;', for example:
> ```
>  "Images (*.png *.xpm *.jpg);;Text files (*.txt);;XML files (*.xml)"
> ```

bjorn:
Thanks! :-)

PR #3778, opened 2023-07-06
Provide feedback when the Terrain Brush can't find some tiles

bjorn:
* These locations are now marked in red

* The text "Missing terrain transition" appears on the status bar


bjorn:
Example of what this currently looks like:

image

This change only affects the Terrain Brush, but might also be useful in the Terrain Fill Mode. However, it relies on a custom `WangBrushItem`, which is currently only used by the `WangBrush`.

PR #3779, opened 2023-07-07
Scripting: Added Geometry interface with line and ellipse helpers

bjorn:
Providing access from scripts to the existing line and ellipse rendering functions.


eishiya:
In the comment for ellipseRegion(rect), it may be a good idea to specify whether the rect represents a bounding rect, or the origin and radius.

eishiya:
When using `Geometry.ellipseRegion(rect)` with a `Qt.rect(x,y,0,0)`, the result is a 2x2 rectangle with x,y at its bottom right corner. I would expect an ellipse with width or height 0 to be either an empty region, or to replace the 0 width and/or height with 1.

This doesn't affect the version that takes numbers, since that takes inclusive points for the rect, and thus cannot make a rectangle with width or height 0.

PR #3782, opened 2023-07-08
Add file open / save prompts in scripting API

### LINKED ISSUE #3537: Scripting: Provide a way to prompt user for a file or directory
### Problem
Sometimes scripts need to prompt the user for a file. There are three main ways to do this in Tiled currently:
- `tiled.prompt()` lets the user enter the path as just a string. It's very inconvenient since there's no file selector, but has the advantage that the script will automatically wait until the user is done.
- Read a custom property from some relevant asset in Tiled. File custom properties let the user choose the file with a file picker and let the user save their preference in the asset., but they have to be set up before the script runs, which isn't always appropriate, and not all scripts deal with already-open assets.
- Create a Dialog and add a FilePicker widget to it. This is usually the best option (and can be combined with the previous one when appropriate), but has one major disadvantage: Tiled doesn't wait for Dialogs to finish, the function that creates and shows the Dialog will finish long before the Dialog does. This makes Dialogs inappropriate for functions that have to return something and need to wait for the input before they do, such as `MapFormat.read()` or some action that has to be complete before the user is allowed to do anything else.

There is currently no way to prompt the user for a file with a proper file picker that also allows the script to block until input is received. Blocking would be convenient for many kinds of scripts, and is necessary for the `read()` methods of custom formats.
Using a Dialog or any kind of user prompting is of course not ideal for `read()` in general since these methods may be called via CLI, but it would be convenient to have the option when the function is called in a context that allows it, so that opening files can be silent in CLI (and fail if additional configuration is required) and more interactive in the GUI.

### Solutions
I can think of two possible solutions to this:
- Provide something like `tiled.promptForFile()` that prompts specifically for a file or directory and blocks until the user is done, similar to how `tiled.prompt()` and `tiled.confirm()` work.
- Add the option to make Dialogs block after they're shown and until they're done. The actual handling of the Dialog would still be done via signals like at present, but after the signals for the dialog have been handled, control flow would return to just after `Dialog.show()` was called, allowing the script to use any values that were set in the signal handlers. Blocking Dialogs should not allow the user to interact with other Tiled UI until they are done, just like `tiled.prompt()` and its siblings.

I think allowing Dialogs to be blocking would be the better solution, as it would be useful for much more than just files. For example, a blocking Dialog can be used to configure the work of an Action without letting the user call any other actions until it's done. However, `tiled.promptForFile()` is a lot simpler compared to setting up a whole Dialog and signal callbacks, so it's possible that it's worth implementing both.

dogboydog:
Fix #3537 

I didn't write the scripting doc yet but I figured I could put up the C++ code for people to try & for review before I do that.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,7 @@
 * Added support for setting custom properties on the project (#2903)
 * Removed Space and Ctrl+Space shortcuts from Layers view to avoid conflict with panning (#3672)
 * Scripting: Added API for editing tile layers using terrain sets (with a-morphous, #3758)
+* Scripting: Added file dialog API (with dogboydog, #3782)
 * Scripting: Support erasing tiles in Tool.preview and TileMap.merge
 * Scripting: Added WangSet.effectiveTypeForColor
 * Fixed object preview position with parallax factor on group layer (#3669)
--- a/docs/scripting-doc/index.d.ts
+++ b/docs/scripting-doc/index.d.ts
@@ -183,6 +183,7 @@ interface size {
  * polygon array and then assign it to {@link MapObject.polygon}.
  */
 type Polygon = point[];
+
 /**
  * A string used to show only certain types of files when prompting the user to select a file path.
  * 
@@ -195,6 +196,7 @@ type Polygon = point[];
  * ```
  */
 type FileFilter = string;
+
 /**
  * The value of a property of type 'object', which refers to a
  * {@link MapObject} by its ID.
@@ -3941,7 +3943,7 @@ declare namespace tiled {
    * 
    * Returns the absolute path of the chosen directory, or an empty string if the user cancels the dialog. 
    */
-  export function promptForDirectory(defaultDir?: string,  title?: string): string;
+  export function promptDirectory(defaultDir?: string, title?: string): string;
   
   /**
    * Shows a dialog which asks the user to choose one or more existing files.
@@ -3950,7 +3952,7 @@ declare namespace tiled {
    * 
    * Returns an array of the absolute paths of the chosen files, or an empty array if the user cancels the dialog. 
    */
-  export function promptOpenMultipleFiles(defaultDir?: string, filters?: string, title?: string): string[];
+  export function promptOpenFiles(defaultDir?: string, filters?: FileFilter, title?: string): string[];
   
   /**
    * Shows a dialog which asks the user to choose an existing file.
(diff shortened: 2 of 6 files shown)


dogboydog:
Example of making a "choose directory" UI using a Dialog 

```js
var ChooseDirectoryDialog = {};

ChooseDirectoryDialog.testPromptAction = tiled.registerAction("ChooseDirectory", function (action) {

    var dialog = new Dialog();
    dialog.windowTitle = 'Choose a Directory';
    dialog.newRowMode =  Dialog.ManualRows;
    var dirDisplay =dialog.addTextInput('');
    dirDisplay.minimumWidth = 200;
    dirDisplay.enabled = false;
    var dirButton = dialog.addButton('Browse');
    dirButton.clicked.connect(()=>{
       var chosenDir = tiled.promptForDirectory('', 'Directory for Dialog');
       if (chosenDir != ''){
           dirDisplay.text = chosenDir;
           
       }
    });
    dialog.exec();
});
ChooseDirectoryDialog.testPromptAction.text = "ChooseDirectory";

tiled.extendMenu("Edit", [
    { action: "ChooseDirectory", before: "SelectAll" },
    { separator: true }
]);

```

image


I did notice just now, if you have one of these file / directory dialogs up and then you reload scripts, Tiled seems to crash. I ran via command line to see if I could see a message and just got `Segmentation fault`.  Not sure how to prevent that though. I think it could be from the file dialogs sticking around,  then when they close due to the scripting reload, we can't return the chosen path since all the scripting state was reinitialized.

eishiya:
There used to be a similar issue with the existing prompts, because they were trying to return to a context that no longer existed. Commit https://github.com/mapeditor/tiled/commit/994492e92db8906395ded0bfbfac8312586c0e45 fixed it, maybe there's something of use you can find in there?

dogboydog:
Thanks, I added `ScriptManager::ResetBlocker blocker;` to the beginning of promptForDirectory locally, then reloaded scripts with 'open directory' dialog open, but Tiled still seems to crash when the dialog is canceled.

Here is where it looks to be crashing: 

image

bjorn:
> but Tiled still seems to crash when the dialog is canceled.

This was due to the `ScriptDialog::exec` also opening a modal dialog that should block script reloads. I've pushed a fix for that.

I've done a few renames. I know @eishiya preferred "promptOpenMultipleFiles", but it really sounds too verbose to me, so I moved it back to "promptOpenFiles". I've removed the "For" in "promptForDirectory", since even though it reads fine, none of the other new "prompt" methods use "For" even though they technically could.

I almost added the "Name" postfix as used in the Qt API, but was undecided. Why did you go for `promptOpenFile` as opposed to `promptOpenFileName`?

dogboydog:
Just my opinion, but putting "name" on the end sounds awkward to me. So that's why I didn't include it.

eishiya:
The documentation for this change doesn't mention that the new methods are `@since 1.10.2` xP

I wanted to make the fix myself, but even though this has been merged into master already, I couldn't find the methods in master to add it.

Issue #3756, opened 2023-05-25
Scripting: Tool function updateEnabledState() no longer usable


TehCupcakes:
**Describe the bug**
As of 1.10.0, one can no longer use `updateEnabledState()` to override when a tool should be enabled. This is likely a regression introduced with the changes to accommodate `targetLayerType`. While this covers the most common method of enabling/disabling a tool, there are still more complex requirements which call for a way to hook into layer changes to determine enabled status.

**To Reproduce**
1. Create an extension which calls `tiled.registerTool()` and has `updateEnabledState()` defined in the Tool definition.
2. Inside `updateEnableState()`, set `this.enabled = true;`
3. Observe that the tool is always disabled, regardless of current layer or other extension settings.

**Expected behavior**
Tool extensions should be able to dynamically set enabled status inside `updateEnabledState()` when either:

- No `targetLayerType` is defined
- The current layer matches the `targetLayerType`

**Specifications:**
 - OS: Windows 10
 - Tiled Version: 1.10.1

**NOTES**

- This might have been introduced by the change to call `AbstractTileTool:updateEnabledState()` instead of `AbstractTool:updateEnabledState()`? ([diff](https://github.com/mapeditor/tiled/commit/afe4758fca0b97d3bcd192c879c9a4cf39b37d3c#diff-cdbdd2f43f8ee8a05a22f5bad84460b07f3dcacd8ae15640bf90278ef66e0fbcL307))
- My current workaround is to disable in the `mousePressed()` event if an incorrect layer is selected, since this tool happens to be related to tile placement. But it would be better to disable it sooner when the layer is selected.
- The types for `tiled.registerTool` include the full `Tool` interface, which has no optional properties. I think the types could be improved to indicate most fields are optional. As it is, I had to ignore a Typescript error in order to avoid defining dozens of properties/functions on my tool.

bjorn:
> * The types for `tiled.registerTool` include the full `Tool` interface, which has no optional properties. I think the types could be improved to indicate most fields are optional. As it is, I had to ignore a Typescript error in order to avoid defining dozens of properties/functions on my tool.

Hmm, how does this work? Maybe this means we can't simply reuse the `Tool` interface for the argument to `registerTool`? These properties are not optional in the `ScriptedTool` class they are documenting, and marking them as such might result in other TS issues when trying to use these properties? Also, does it give errors even for `readonly` properties like `map` and `selectedTile`, which are never meant to be provided by the user?

[...]

PR #3800, opened 2023-08-02
Fixed Tool.updateEnabledState behavior and related changes

bjorn:
See #3756

### DIFF
--- a/docs/scripting-doc/index.d.ts
+++ b/docs/scripting-doc/index.d.ts
@@ -989,7 +989,7 @@ declare class TiledObject {
   /**
    * The asset this object is part of, or `null`.
    */
-  readonly asset: Asset;
+  readonly asset: Asset | null;
 
   /**
    * Whether the object is read-only.
@@ -1306,13 +1306,13 @@ declare class MapObject extends TiledObject {
    * Layer this object is part of (or `null` in case of a standalone
    * object).
    */
-  layer: ObjectGroup;
+  layer: ObjectGroup | null;
 
   /**
    * Map this object is part of (or `null` in case of a
    * standalone object).
    */
-  readonly map: TileMap;
+  readonly map: TileMap | null;
 
   /**
    * Constructs a new map object, which can be added to an {@link ObjectGroup}.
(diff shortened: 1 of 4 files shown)


bjorn:
> * `map` is defined as possibly `null` in the `Tool` interface. When would this be null? I thought the toolbar only shows when a map is active. If it is always defined, removing the null saves me an if condition.

I changed it because I realized it would sometimes be `null`. If there is no map open, then it will always be `null` here:
```js
var tool = tiled.registerTool(name, toolDefinition);
tool.map; // null
```
And when closing the last map, it will also be `null` during the call to `mapChanged` (I realize now this function's parameters also need `| null`). The definitions are currently missing `| null` in a bunch of places, because I originally assumed that like in Java and C#, any object reference is allowed by be `null`.

That said, in most cases, there will be a map, and most functions will not need to check this variable because they can't be called while the `tool.map` is `null`. I'm not sure if there's any way to communicate this, or whether this is reason to just leave out the `| null` even if that's not entirely correct?

> * The ergonomics of `updateStatusInfo()` and `updateEnabledState()` could be a bit awkward. Initially I thought I should return a boolean, but when that did nothing I realized they are void functions and I have to set the internal property directly. Thoughts?

These functions were already present in the `AbstractTool` and  `AbstractTileTool` classes in C++, where they were mostly introduced for some convenience and code sharing between tools. It might be easier to have functions that return updated values, but I'm not sure what they should be called then, because they can't be called the same way as the property getters. I personally would prefer to remove these calls entirely, possibly replacing them with more specific calls similar to `mapChanged`, like `selectedLayersChanged`, so that it is more clear in response to what we are updating those properties, but I think it needs a small refactoring and not just a change in the scripting API.

TehCupcakes:
Ah, I see. That all makes sense. The most accurate type is the best; even if it's rarely `null`, it should indicate that it could be to prevent runtime errors. TS forces us to check if it is null even in situations we know it won't be, but the extra safety is worth it. 🙂

bjorn:
> The only alternative I could think of is that you could have types that extend from `Tool` and use those in the functions instead, which would set certain properties as non-null if we know from context they won't be null then. But it's probably not worth introducing extra types for that.

Hmm, right, I guess we could have something like:
```ts
interface ActiveTool extends Tool {
  /**
   * Currently active tile map.
   */
  readonly map: TileMap;
}
```
And then use `this: ActiceTool` for the callbacks that can only happen for active tools. But yeah, I'm not sure if it's worth the clutter, especially since it also affects the generated documentation.gi

PR #3811, opened 2023-08-23
Show parent layer names in "Move Object to Layer" menu

### LINKED ISSUE #3454: Show "Move Object To Layer" context menu as a tree when Object Layers are inside Group Layers
### Problem
We have a map where we have a **Group Layer** for each **"Floor"** of a building. Inside each group there is a set of object and tile layers, like **"Walls"**, **"Floor"**, **"Enemies"**. When we right click on a selected object to **Move Object To Layer**, the layers are shown in a flattened hierarchy so there are multiple **"Walls"**, etc. layers and it's not clear which one is for which floor.

### Suggested Solution
It would be better if the **Move Object To Layer** pull right menu displayed a tree where further sub-menus are Group Layer names and leaf nodes are the Object Layer names under those group layers.

### Current Workaround
As a workaround we're currently adding a floor number to each layer name, e.g. **"Walls4"** for the walls on floor 4. It works but it would be nicer to be able to select the **"Walls"** layer under the **"Floor4"** group layer.

This image shows the problem of the flattened menu and the workaround we're using to add numbers to layer names:
image

Here are our test files if it helps to see the issue and see any potential fix:
[Example.zip](https://github.com/mapeditor/tiled/files/9363137/Example.zip)

bjorn:
Also reversed the order of the layers, to match the order in the Layers view.

image

Closes #3454


eishiya:
While this is much better than not showing anything, I'd strongly prefer a less verbose "tree view" like I suggested in https://github.com/mapeditor/tiled/issues/3454#issuecomment-1218143735, with indentation for nested layers. For large number of layers especially, I feel it would be easier to navigate the list if it superficially resembles the layers list, with "lines" for the groups.

bjorn:
Now it looks like that:

image

PR #3821, opened 2023-09-19
Changed StampBrush to do rectangle-erase on right-click

bjorn:
Feature requested by elvisish, but can't be merged like this. Instead, the mouse interaction in general would need to be made more configurable.

This change currently leaves no way to capture a stamp from the map.

### DIFF
--- a/src/tiled/stampbrush.cpp
+++ b/src/tiled/stampbrush.cpp
@@ -26,6 +26,7 @@
 #include "map.h"
 #include "mapdocument.h"
 #include "mapscene.h"
+#include "painttilelayer.h"
 #include "stampactions.h"
 #include "tile.h"
 #include "tilestamp.h"
@@ -33,6 +34,7 @@
 #include "wangfiller.h"
 
 #include <QAction>
+#include <QCoreApplication>
 #include <QToolBar>
 #include <QVector>
 
@@ -143,7 +145,8 @@ void StampBrush::mousePressed(QGraphicsSceneMouseEvent *event)
                 break;
             }
             return;
-        } else if (event->button() == Qt::RightButton && event->modifiers() == Qt::NoModifier) {
+        } else if (event->button() == Qt::RightButton && (event->modifiers() == Qt::NoModifier ||
+                                                          event->modifiers() == Qt::ShiftModifier)) {
             beginCapture();
             return;
         }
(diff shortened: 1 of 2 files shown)


eishiya:
I'm curious why this behaviour couldn't be added as e.g. Alt + right click, without a need for additional user configuration? When using right click to, none of the modifier keys seem to do anything except break drag behaviour, so they seem free to use.

(I find the right-click sampling behaviour *very* useful, so of course I hope to see it stay xP)

bjorn:
> I'm curious why this behaviour couldn't be added as e.g. Alt + right click, without a need for additional user configuration?

Hmm, you're right, Alt+Right-click is still free for all the relevant tools. It's worth considering.

elvisish:
> > I'm curious why this behaviour couldn't be added as e.g. Alt + right click, without a need for additional user configuration?
> 
> Hmm, you're right, Alt+Right-click is still free for all the relevant tools. It's worth considering.

Because my tilemapping muscle-memory instinct is to just right-click, I'd never be able to convert to having to remember ot press alt as well (and if it were a configurable option, I'd also never choose alt+)

eishiya:
I'm certainly not against more options xP And if there's an option, maybe the other behaviour can be available with alt. That way, those of us with right-click sample muscle memory from using Tiled can also be happy, *and* access the new feature easily.

bjorn:
@elvisish I've fixed the clearing of the stamp on right-click erase. New builds will soon be available at https://github.com/mapeditor/tiled/actions/runs/6309399827?pr=3821.

elvisish:
> @elvisish I've fixed the clearing of the stamp on right-click erase. New builds will soon be available at https://github.com/mapeditor/tiled/actions/runs/6309399827?pr=3821.

Thankyou!

bjorn:
@elvisish I've updated this PR with all the changes that happened since October last year, just in case you wanted the benefit of any new features / fixes in addition to this change.

The change has also gotten a lot smaller, thanks to a small refactoring as part of the addition of "Cut Stamp" on Shift+Right-Click in #3963. If you would be able to get used to it, this new "Cut Stamp" action on Shift+Right-Click might be an alternative for you, though on the other hand, it does also change the stamp to the area that was cut, which might not be desired.

New builds should soon be available at https://github.com/mapeditor/tiled/actions/runs/9548746873?pr=3821 (though I see the macOS builds were cancelled due to the macOS-11 environment being deprecated).

elvisish:
> @elvisish I've updated this PR with all the changes that happened since October last year, just in case you wanted the benefit of any new features / fixes in addition to this change.
> 
> 
> 
> The change has also gotten a lot smaller, thanks to a small refactoring as part of the addition of "Cut Stamp" on Shift+Right-Click in #3963. If you would be able to get used to it, this new "Cut Stamp" action on Shift+Right-Click might be an alternative for you, though on the other hand, it does also change the stamp to the area that was cut, which might not be desired.
> 
> 
> 
> New builds should soon be available at https://github.com/mapeditor/tiled/actions/runs/9548746873?pr=3821 (though I see the macOS builds were cancelled due to the macOS-11 environment being deprecated).

Thanks, I'll give that a try, I'm not sure if it's possible to rebind the keys yet so right-click can be redefined for any purpose?

bjorn:
> I'm not sure if it's possible to rebind the keys yet so right-click can be redefined for any purpose?

Unfortunately not. If you need pure erasing on right-click, you'd need to use a build from this pull request.

Issue #3713, opened 2023-05-17
request: reset all settings to default


pixelartgirl:
I'm Pixel Art Girl and I come from Pixel Land, where all people have autism!
I was send to earth to make Tiled more user friendly

**Is your feature request related to a problem? Please describe.**
Currently there is no way to reset all settings to their default

**Describe the solution you'd like**
Make option to  reset all settings to their default

pixelartgirl:
PUT THIS SHIT UP YOUR FUCKING ASS

PR #3828, opened 2023-10-15
Add reset to all settings to default except plugins.

0xchatware:
Implements #3713.

### DIFF
--- a/src/tiled/preferencesdialog.cpp
+++ b/src/tiled/preferencesdialog.cpp
@@ -187,11 +187,13 @@ PreferencesDialog::PreferencesDialog(QWidget *parent)
         this->restoreToDefault(QString::fromUtf8(
             "Interface/(?!(ApplicationStyle|BaseColor|SelectionColor|UseCustomFont|CustomFont))"));
         this->restoreToDefault(QString::fromUtf8("Install/(DisplayNews|CheckForUpdates)"));
+        this->restoreToDefault(QString::fromUtf8("AbstractObjectTool/"));
     });
 
-    connect(mUi->defaultTheme, &QPushButton::clicked, this, [=] {
+    connect(mUi->defaultTheme, &QPushButton::clicked, preferences, [preferences, this] {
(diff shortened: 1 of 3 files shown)


0xchatware:
However, I don't know how to translate the button "Reset to default" to other languages.

eishiya:
Thanks for implementing this! Consider calling it "**reset** to default" instead of "restore to default" like it is now. When looking for such an option, that's usually the keyword I'm looking for, whereas "restore" is for when I'm looking for data that has been deleted rather than changed. You yourself even called it "reset to default" in your comment.

As for translations, I think display strings in the .ui file automatically get passed through the translation system and no special mark-up is needed. As far as I can tell form other PRs, you also don't need to add it to any list of strings or anything, that seems to be handled automatically. So, as far as I can tell, you're all set! The actual translations are provided by other contributors and are usually only done after each new Tiled version's string freeze.

0xchatware:
Thank you very much for your comment and for clarifications. I agree with you, and that it is better to rename the buttons to "Reset" instead of "Restore". Also, it is very nice that the languages can be translated automatically.

bjorn:
I agree there should be a way to reset all the settings, but I'm not sure if we need it to work per-page, because:

* The pages are rather arbitrary, so it will reset some arbitrary subset of the settings.
* It makes it harder to reset everything, since you now need to do it on each page.
* Not all preferences are set in the Preferences dialog. Many preferences are simply toggles in various places in the UI, like whether to show the grid, whether parallax is enabled, dock widget layouts, etc. Likely there should be a way to reset these as well (I know layout has its own action)?

I personally think what #3713 was about is to provide an action that basically resets Tiled to the state of first launch. I would suggest to just clear everything using `prefs->clear()` and then suggest the user performs a restart. It would probably be a good idea to confirm the user really wants this, too.

bjorn:
@0xchatware I forgot to ask, would you be interested to adjust your PR to work as I described above? Also, I'm sorry for not giving feedback earlier.

Issue #3647, opened 2023-04-05
Unable to build Tiled on Windows in Debug mode


bjorn:
When building Tiled in Debug mode on Windows, the following error pops up:

```
:-1: error: Cannot install files
'C:/Projects/build-tiled-Desktop_Qt_6_4_2_MSVC2019_64bit-Debug/Debug_Desktop__49d835fd610c6acf/tiled.1920fdf8/tiled.pdb' and
'C:/Projects/build-tiled-Desktop_Qt_6_4_2_MSVC2019_64bit-Debug/Debug_Desktop__49d835fd610c6acf/libtiled.080f31f8/tiled.pdb'
to the same location
'C:/Projects/build-tiled-Desktop_Qt_6_4_2_MSVC2019_64bit-Debug/Debug_Desktop__49d835fd610c6acf/install-root/tiled.pdb'.
If you are attempting to install a directory hierarchy, consider using the qbs.installSourceBase property.
```

This happens as a result of 46c447422209929b14e6622dec51942d5bbe09b1, since before that the debug information files were not being installed at all.

Apparently `tiled.exe` and `tiled.dll` have the same separate debug information file `tiled.pdb`, they can't install to the same location. We'd need to either resolve the naming conflict or disable installing debug information files (though they are likely necessary for debugging to work properly).

azagaya:
Changing the `targetName` in libtiled.qbs to something like "libtiled" instead of "tiled" seems to fix this, and debug works with that change. Is there any problem with that to your knowledge?

bjorn:
@azagaya I guess there isn't really a problem with it, apart from that rather uncommon resulting DLL name "libtiled.dll". And of course such a rename will need a small adjustment in the installer. I have no objections to this solution.

bjorn:
Resolved by #3832.

PR #3832, opened 2023-10-17
updated qbs file for libtiled to solve naming conflict

azagaya:
Solves #3647


bjorn:
Thanks! Judging by the CI it does not appear to be causing any issues (though the builds should still be tested). However, I'm not particularly fond of the duplicate "lib" prefix on Linux (and macOS):
```
linking liblibtiled.so [libtiled]
```
So I think we should limit this change to only the Windows builds. We could probably do this by setting the [`dynamicLibraryPrefix`](https://doc.qt.io/qbs/qml-qbsmodules-cpp.html#dynamicLibraryPrefix-prop) to "lib" instead of changing the `targetName`:
```
cpp.dynamicLibraryPrefix: "lib"
```

azagaya:
I updated with the suggestion!

bjorn:
I've just tested the AppImage for now and it appears to be working fine. Thanks for resolving this issue!

PR #3837, opened 2023-10-29
fix typo in scripting doc for Dialog.exec()

dogboydog:
I added doc for this in a past PR, but it wasn't showing on the site: 
https://www.mapeditor.org/docs/scripting/classes/Dialog.html#exec


bjorn:
Thanks for noticing and for the fix! :-)

dogboydog:
Credit to eishiya for noticing :)

Issue #3630, opened 2023-03-21
Scripting: Unexpected results when using Image for tiles


eishiya:
These are really several issues, but they're all about roughly the same thing: embedding image data directly into tilesets via scripting.

1. `Tileset.loadFromImage(image)` creates an Image Collection in which every tile has the same source image, but uses different subrects of it. This is quite unexpected, I'd have thought this would set the *Tileset's* source image instead. Saving a Tileset created this way to a file duplicates the entire image (base64-encoded) for each tile. I'm guessing this is because there is no source image path and Tiled treats every tileset with no image path as an Image Collection, but this is one case where it should not.
   - @djedditt Reported a related issue where large embedded tilesets created this way would freeze Tiled when copy+pasting tiles from them. I suspect this is due to the large image being duplicated for each copy of the tile, and then potentially duplicated to the preview brush.
2. `Tile.setImage()` requires that the tile's `imageRect` be set first. If it's left at the default (presumably 0,0,0,0), setImage() seems to fail, and the tile is not written out when the Tileset is saved (in TSX; see (5) below).
3. The documentation for `Tile.setImage()` warns that it "does not affect the saved tileset!". It does though, and the image data is embedded correctly (in TSX, anyway). What it *doesn't* affect is Tiles that have a source image path, as far as I can tell. So, `setImage()` cannot be used to *modify* tiles that were created normally. I have not looked into what happens if one tries `setImage()` on a tile that was created with `setImage()` :]
4. Because `Tile.setImage()` does not affect normal tiles, it means it cannot be used to *replace* the image of a tile. This means that if I want to create a script that embeds all the images of a tileset, I have to either create a whole new tileset and put the tiles there, or create new tiles and delete the original ones. Either way, tile IDs are likely to change, breaking existing maps. Some way to override an existing tile's image with `setImage()` would be nice.
5. TSX files will store the embedded images as base64 data, but TSJ files ditch them entirely. This means TSJ files lack a feature that TSX files have, when they should have exact feature parity. In addition, there seems to be no way for scripted formats to access images set via `Tileset.loadFromImage` and `Tile.setImage`, so they'll have the same problem as TSJ.

Edit: I also experimented with `ImageLayer.setImage` and ran into several issues:
- The map view is not updated on `setImage`, you have to change document tabs and back to see the result.
- If no source path is set, the image is *not* embedded into the file, even in TMX.
[...]

PR #3841, opened 2023-11-06
TMX format: Support embedded images on the tileset

bjorn:
Previously, embedded images were only supported per-tile. Due to the recent change to use sub-rects rather than physically cutting up the tileset, when you called `Tileset.loadFromImage(image)`, the entire tileset image would get duplicated for each tile.

Now the image is embedded once at the tileset level instead.

Also, the `Tileset.margin` and `Tileset.tileSpacing` properties are now writable in the scripting API.

Addresses part of #3630.


eishiya:
The documentation for the XML and JSON formats should probably be amended to mention the possibility of the image source being a base-64 data URL. Currently, this doesn't seem to even be described well for image collection tiles, where this has been a feature for a while. The XML documentation page mentions that an image can have `format` and `data`, but links to the description of <data>, which deals solely with tile layer data and not with image data. Is this format/data thing even accurate?

bjorn:
> The XML documentation page mentions that an image can have `format` and `data`, but links to the description of `<data>`, which deals solely with tile layer data and not with image data. Is this format/data thing even accurate?

Yes, this is accurate. Embedded images are not stored as a base-64 data URL, but re-use the ["data" element](https://doc.mapeditor.org/en/stable/reference/tmx-map-format/#data) also used by tile layer data. The documentation for the "data" element should be updated to mention this, though.

The JSON format still does not support embedded images. I will look into supporting it and then of course will need to adjust the documentation.

eishiya:
Gotcha. It'd be good to mention the use for images in the <data> section, even if it's just to say that the data may be binary image data in the case of <data> in <image> elements. Presumably, CSV encoding isn't supported for image data.

PR #3854, opened 2023-12-08
Point rotation

IoriBranford:
Some use cases:

- give points visible directions/angles
- place multiple points at the same position without overlapping

image

### DIFF
--- a/src/libtiled/mapobject.h
+++ b/src/libtiled/mapobject.h
@@ -422,7 +422,7 @@ inline bool MapObject::hasDimensions() const
  * Returns true if this object can be rotated.
  */
 inline bool MapObject::canRotate() const
-{ return mShape != Point; }
+{ return true; }
 
 inline bool MapObject::isTileObject() const
 { return !mCell.isEmpty(); }
--- a/src/tiled/mapobjectitem.cpp
+++ b/src/tiled/mapobjectitem.cpp
@@ -97,8 +97,7 @@ void MapObjectItem::syncWithMapObject()
     }
 
     setVisible(mObject->isVisible());
-    setFlag(QGraphicsItem::ItemIgnoresTransformations,
-            mObject->shape() == MapObject::Point);
+    setFlag(QGraphicsItem::ItemIgnoresTransformations, false);
 }
 
 void MapObjectItem::setIsHoverIndicator(bool isHoverIndicator)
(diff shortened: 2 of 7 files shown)


eishiya:
I don't think it makes sense for *points* to have a direction. Points are points, they are not lines or vectors.

I do like the idea of having a way to clearly display multiple points at a location, but I feel this would be better solved by smarter handling of the labels than by giving points rotation.

IoriBranford:
I know mathematically "directed point" isn't a thing, but looking at a map editor as a creativity tool, I'd think enabling more ways to be creative with shapes is worth it.

I could add a choice of point appearances. For starters the choices would be either the current appearance, or a dot with attached arrowhead for the "directed point" case. Later on this option could be expanded with more point shapes you could use for classification and personalization.

I think I see what you mean by handling the labels, maybe stacking them vertically when the points overlap but that doesn't help with clicking a specific one unless we allow clicking its label to select it.

eishiya:
I think vectors would be better as a new object type because they have *two* things Points don't: direction *and length*. So, vectors are really closer to Polylines than to Points, since they can be described as two points xP And if we run with that comparison, then perhaps a Vector Object could consist of two points - its position (the base of the arrow), and a second point indicating its length and direction (the tip of the arrow), and they could potentially be manipulated independently, with entirely different controls than for Points. Unlike a Polyline, however, it would store Length and Direction rather than the coordinates of the second point. The Direction would be equivalent to Rotation, just with a more appropriate name for the context.

Arguably, vectors also call for a third difference from Points: a scale factor. It's common for vectors to use units other than those used for sizes, but the user may still want to conveniently manipulate/visualise them, and thus may want to set a multiplier so that the handles aren't too distant or too small. For example, if I'm implementing jump pads, I may want to scale the vector preview so it shows the approximate distance it will send the player, which will be in the tens or hundreds of pixels, rather than the impulse that vector represents, which may be in the thousands, way off the map if shown as pixels.

---
I should've been clearer with the "labels" bit - I meant not only the text labels, but the drawn points as well - those are also labels, since they are not themselves the Point, but sit offset from it. That is, I think rotating the points in this way should be something Tiled does automatically (possibly based on zoom level), without storing it in the map file.
Positioning text labels and making them clickable would be a good option as well. Perhaps the automatic positioning would only take effect when clickable text labels are enabled.

---
I realise I'm asking for a lot here - I don't expect you to implement any of this. I can see the usefulness of the feature as you've added it, but I think that as a merged Tiled feature rather than a local change, it would be better to have something more substantial which doesn't conflict with the usual understanding of a Point. Perhaps I should open an issue requesting Vector Objects?

PR #3859, opened 2023-12-19
 Add World scripting features

### LINKED ISSUE #3539: Scripting: Add API for Worlds
At present, the only way for Tiled scripts to work with Worlds is to parse the .world file themselves. This is a hassle, especially for Worlds that use pattern-matching instead of explicitly defined maps. Working with maps in a World would be a lot more convenient if the scripting API provided access to the many functions of World that already exist on the C++ side, including opening/closing worlds, getting the current active world(s?), adding/removing maps, getting a list of maps and their rects, getting the rect for a given map, getting all maps in a given rect, and Signals for worlds loading/unloading/changing.

It looks like on the C++ side, most of the World methods work with map file paths. I think it would be good to preserve this, while *also* providing methods that accept/return TileMaps for convenience.

Example use cases:
- Automatically generate/populate map transition zone Objects based on map adjacency in the World.
- Populate a World automatically based on a system that the current pattern matching can't deal with, such as maps using sequential letters, or based on custom properties on the map.
- Populate map custom properties based on position in the World, e.g. the map's position on the in-game map, to avoid the need to parse the World in-engine.

dogboydog:
Fix #3539 

Add UnloadAllWorlds action, exposed in World > Unload World when two or more worlds are loaded.

Added to scripting: 
```
tiled.worldsChanged : Signal<void>;
tiled.worldLoaded : Signal<string>;
tiled.worldReloaded : Signal<string>;
tiled.worldUnloaded : Signal<string>;  
tiled.worldSaved : Signal<string>;
tiled.worlds : World[]; 

class WorldMapEntry {
  fileName : string;
  rect : rect;
} 

class WorldPattern {
  regExp: RegExp;
  multiplierX: number;
  multiplierY: number;
  offset: point;
  mapSize: size;
}

class World { 
  readonly maps : WorldMapEntry[];
  readonly patterns : WorldPattern[];
  allMaps() : WorldMapEntry[];
  mapsInRect(rect : rect) : WorldMapEntry[];
  containsMap(fileName : string) : boolean;
  setMapRect(fileName: string, rect : rect): void;
  setMapPos(map: TileMap, x: number, y: number): void;
  addMap(fileName: string, rect: rect): void;
  addMap(map: TileMap, x: number, y: number): void;
  removeMap(fileName: string): void;
  removeMap(map: TileMap): void;
  save(): boolean;
}
```
  
 ## TODO 
 
- [x] Update version number for `@since` in new entries in `tiled.d.ts`
- [x] Fix crash on `tiled.worlds`. I think the WorldDocument I'm creating is getting freed when worlds() returns in scriptmodule.cpp, but not sure where to save the references or if I'm right about the problem 
- [x]  adding/removing maps from worlds
- [x] Test and provide examples in below gist of methods on individual worlds returned from `tiled.worlds`

Test script 
https://gist.github.com/dogboydog/42edb0886989467ec3671acdce4e749f#file-worldscripting


src/tiled/editableworld.h
```diff
@@ -0,0 +1,97 @@
...
+    Q_PROPERTY(QString fileName READ fileName CONSTANT)
+    Q_PROPERTY(QRect rect READ rect CONSTANT)
+
+public:
+    ScriptWorldMapEntry(World::MapEntry *mapEntry);
+    QString fileName() const;
+    QRect rect() const;
+
+private:
+    World::MapEntry *mMapEntry;
```
bjorn, line 59: This can't be a pointer, since the pointed to object will be destroyed by the time a script would try to access this.

src/tiled/editableworld.h
```diff
@@ -0,0 +1,97 @@
...
+
+#include <QObject>
+
+namespace Tiled {
+
+/**
+ * Wrapper which allows world structs to be used with the EditableAsset
+ * class.
+ */
+class ScriptWorld : public Object
```
bjorn, line 36: I wonder if not the `World` class itself should derive from `Object` instead.

bjorn (commented):
Some initial comments, but to be continued next week. :-)

src/tiled/editableworld.h
dogboydog: Changed to not be a pointer

src/tiled/editableworld.h
dogboydog: We could do that if you think it's best, I just wasn't sure if you would want me to change the base definition of Worlds to support scripting

bjorn:
Feedback to address:

- [ ] It is not possible to add maps to worlds that also use patterns. This need to be either supported or an error needs to be reported when trying to do this.
- [ ] Add overloads for `addMap` that take a `Tiled::EditableMap*`, with only a position (since this one can calculate the size automatically).

eishiya:
>     * Add overloads for `addMap` that take a `Tiled::EditableMap*`, with only a position (since this one can calculate the size automatically).

If adding maps can be done with a TileMap, then I think removeMap(TileMap) should be available too. That one should be simpler since it just needs to check whether the TileMap has a fileName and use that.

bjorn:
Another small thing: I think there should be a menu separator above the "Unload All Worlds" action.

dogboydog:
I added addMap and removeMap which take EditableMap instances. I'm wondering if we should abstract away any references to mapIndex in any of these world scripting APIs and just have the user specify either an EditableMap or a file name. Thanks for your help with this.

bjorn:
> I'm wondering if we should abstract away any references to `mapIndex` in any of these world scripting APIs and just have the user specify either an `EditableMap` or a file name.

Yeah, I'll have a go at removing all the index-based API. If necessary one can compute the index by iterating `world.maps` anyway. Otherwise for completeness sake we'd have to provide an index, filename and map version of each function...

eishiya:
Is using a QPoint instead of just x, y necessary? I feel like x, y would be more convenient than Qt.point(x, y).

bjorn:
Alright, I think it's done and ready for merge. Would you give it a final round of review and testing, @dogboydog?

dogboydog:
Thanks, I just gave it a try.  I was able to run my example script and add and remove maps, call mapsInRect, containsMap, setMapPos. The signals seem to be firing as expected. It seems good to me.

bjorn:
@dogboydog I've decided to still push a large change and remove most of the world-specific signals, readding some of them by emitting `assetOpened`, `assetAboutToBeSaved` and `assetSaved` also for worlds. I wasn't convinced the existing world-specific signals would be useful enough, and it could be hard to keep API compatibility when we change the way worlds are handled in the future.

PR #3862, opened 2023-12-27
Godot exporter collision fix

ryanpetrie:
The exporter was incorrectly using the collision object's X and Y position as a part of the tile's center. Fixed the calculation for both rectangles and polygons. Tested on a local Windows build.


bjorn:
Nicely spotted and thanks for the fix!

Issue #3412, opened 2022-07-14
Unable to re-dock floating windows


Zer0-One:
Once I float any of the editor panes, I am unable to re-dock them back into the main window.

**To Reproduce**
Click the square icon at the top right of any of the dockable panes to float it. Then try to re-dock the pane.

**Expected behavior**
Window decorations that offer a button to re-dock, or the ability to drag the window back into place.

**Specifications:**
 - OS: Gentoo
 - Tiled Version: 1.8.6
 - Window manager: Sway (wayland, have not tested as an X client)

bjorn:
Hmm, that's likely a problem between Sway and Qt. I doubt there is anything I can do on the Tiled side to fix this issue, unless you're not seeing this issue with other Qt applications using dockable panes.

sami-sweng:
Same issue on Ubuntu 22.04, (tiled installed via snap, so version is "git"...)

eishiya:
Another user reported this issue on the forum, they are using Wayland: https://discourse.mapeditor.org/t/stuck-with-floating-windows-that-cant-be-docked-back/6229

They also don't seem to have a menu bar at all, and have to rely on Search Actions instead.

bkerz:
I have this bug too, I tried to drag the window again back to the right side but it didn't do anything and when I want to reopen those views they are floating again so there's no way to putting them back on the main UI. Ubuntu 22.04, Wayland, Tiled 1.10.2

bkerz:
I just tested this on X11 and it's working as expected, the issue seems to be only on Wayland.

MadPigeon:
> Hmm, that's likely a problem between Sway and Qt. I doubt there is anything I can do on the Tiled side to fix this issue, unless you're not seeing this issue with other Qt applications using dockable panes.

I have the same issue, it happens if I'm installing Tiled from the Snap Store (Ubuntu Software), but the bug doesn't happen if I install the app using `apt-get`. Same system, no reboots

bjorn:
This is Qt issue [QTBUG-87332](https://bugreports.qt.io/browse/QTBUG-87332). It is Closed because for Qt 6.5.4 / 6.6.0 a solution was implemented. I'll update the Qt version for the next Tiled release, but it will only really resolve this issue if your compositor also implements the [proposed xdg-toplevel-drag Wayland protocol](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/204).

PR #3878, opened 2024-01-22
GitHub Actions: Updated to Qt 6.6.1

bjorn:
This might eventually help people on Wayland to have working dock widgets.

Issue #3412


bjorn:
OMG... Of course, Qt 6.5.4 was released for commercial customers only... :S

As an open source Qt user, Tiled will have to jump on Qt 6.6 and help with identifying the regressions.

Issue #3517, opened 2022-11-05
Scripting: Set fileName when using TilesetFormat.read() or MapFormat.read()


eishiya:
A common task in scripted MapFormats is to read an external tileset from file and assign it to the newly created TileMap. There are two ways to open a Tileset from a file:
- `tiled.open()`, which opens the Tileset as a document in Tiled. This is not usually desirable, and takes extra code to close the tab afterwards, and extra work to only close the document if the user didn't already have it open.
- `TilesetFormat.read()`, which doesn't open a tab... but also doesn't set the Tileset's `fileName`, so when that Tileset is added to a map, it becomes embedded.

It's easy to miss the unintentional  proliferation of tileset clones this causes. [Here's an example from the forum of this unintentional embedding](https://discourse.mapeditor.org/t/animated-collection-of-images-different-sets-uses-same-animation-speed-settings/5219/12). Unless there's some specific reason not to, I think `TilesetFormat.read()` and `TilemapFormat.read()` should set the Asset's `fileName`, which should hopefully fix this embedding issue. Why is this not set in the first place?

Of course, sometimes it's desirable to create an embedded tileset on purpose. For this and other purposes, I think `Tileset.clone()` and `TileMap.clone()` would be good additions - the clones should have an empty `fileName`.

If there *is* a reason that TilesetFormat/TileMapFormat.read don't set the fileName, then the API should provide methods to open files, maintain their link to their file, but without opening a document tab for them, e.g. `tiled.openSilently()`, or `tiled.open(path, silent?)`.

bjorn:
> I think `TilesetFormat.read()` and `TilemapFormat.read()` should set the Asset's `fileName`, which should hopefully fix this embedding issue. Why is this not set in the first place?

### Backstory

None of the formats set the file name of the type they return, which traditionally did not even have a file name field. Like an image also does not store the file name from which it was loaded. The file name only comes into play for assets open in the editor, since they have a `Document` instance associated with them, which used to be where the file name was stored.

Eventually, it became a little more complicated, and there was reason to move the file name (and format) into the `Map` and `Tileset` classes. I did this reluctantly, but there were a number of places in the code where I had access to the Map instance but not its respective `MapDocument` instance. Actually looking at the code, this appears to only be due to the need to store a file name for navigating to an error/warning when it is double-clicked. Many such errors are reported from plugins, which don't have access to the `MapDocument` instance.

[...]

PR #3879, opened 2024-01-23
Scripting: Added tiled.load to load assets without opening an editor

bjorn:
Still WIP, has known issues are likely also undiscovered ones.

Intended as a solution for #3517.


justgook:
maybe you can list issues? so some-one could fix / test those?

bjorn:
I've rebased it and gave it a quick test. Appears to work fine, but one noticeable issue is present on exit, it prints:
```
TilesetManager: There are still 3 tilesets loaded at exit!
```
This indicates tileset references are still alive when they shouldn't be. I don't get this message on the `master` branch.

justgook:
just gut filling - that you have forgot to transfer ownership:
```
if (auto document = TilesetDocument::findDocumentForTileset(tileset->sharedFromThis()))
         return new EditableTileset(document);

     editable = new EditableTileset(tileset);
     editable->moveOwnershipToCpp();
```
or something in that area - so you left with tilesets that is not own by anyone

so it is still loaded, and "not closed" properly.

PS. not tested - just from own exp what that could be.
PSS. im not big expert in C++ - just gut filling from 20+ years exp in web dev.

PR #3880, opened 2024-01-24
add containsMap(map:TileMap) scripting method

dogboydog:
In the newly added world scripting feature #3859 , there's a chance that users could  call `containsMap()` with a TileMap and receive a false negative response, as it seems like the TileMap will be coerced to a string.  To avoid this, introduce `containsMap(map:TileMap)` which will take the file name of the given tilemap.

### DIFF
--- a/src/tiled/editableworld.cpp
+++ b/src/tiled/editableworld.cpp
@@ -41,6 +41,21 @@ bool EditableWorld::containsMap(const QString &fileName) const
     return world()->containsMap(fileName);
 }
 
+bool EditableWorld::containsMap(EditableMap *map) const
+{
+    if (!map) {
+        ScriptManager::instance().throwNullArgError(0);
+        return false;
+    }
+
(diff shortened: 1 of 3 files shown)


src/tiled/editableworld.cpp
```diff
@@ -41,6 +41,21 @@ bool EditableWorld::containsMap(const QString &fileName) const
...
 
+bool EditableWorld::containsMap(EditableMap *map) const
+{
+    if (!map) {
+        ScriptManager::instance().throwNullArgError(0);
+        return false;
+    }
+
+    if (map->fileName().isEmpty()) {
+        ScriptManager::instance().throwError(QCoreApplication::translate("Script Errors", "Save this TileMap before checking for its presence in worlds"));
```
bjorn, line 52: Actually, if the map is not saved, couldn't we just return false without throwing an error?

dogboydog: Done

PR #3883, opened 2024-01-31
Replace dead cr31.co.uk doc links with web.archive.org links

TheNullicorn:
The site cr31.co.uk, which is referenced a few times in documentation for Wang tiles, began returning 404 for all pages sometime in the last month.

I've replaced those links with web.archive.org snapshots of them:
| Dead Link | Archive Link |
| ----------------------------------------------- | ---------------------------------------------------------------------------- |
| http://www.cr31.co.uk/stagecast/wang/intro.html | https://web.archive.org/web/20230101id_/cr31.co.uk/stagecast/wang/intro.html |
| http://www.cr31.co.uk/stagecast/wang/2corn.html | https://web.archive.org/web/20230101id_/cr31.co.uk/stagecast/wang/2corn.html |
| http://www.cr31.co.uk/stagecast/wang/2edge.html | https://web.archive.org/web/20230101id_/cr31.co.uk/stagecast/wang/2edge.html |
| http://www.cr31.co.uk/stagecast/wang/blob.html  | https://web.archive.org/web/20230101id_/cr31.co.uk/stagecast/wang/blob.html |

Notes:
  - "20230101" chooses an archive from around appx. Jan 1 2023, when they were still alive
  - "id_" removes the Wayback Machine's page framing, which is a bit distracting


eishiya:
Thank you!

I think it would be better to leave the Wayback Machine frame, so that it's immediately clear the viewer is looking at an archived version rather than the original. It is a little distracting, but not terribly so, and it goes a long way to explain why the page is taking a long time to load, as Wayback Machine pages often do xP

bjorn:
Thanks for noticing and for the fix! Hopefully this site will eventually come back.

PR #3894, opened 2024-02-23
Automapping: Ignore empty outputs per-rule

### LINKED ISSUE #3523: Automapping: Ignore empty outputs per-rule
### Current behaviour
Currently, when an automap rules map has multiple output indices, all rules from that map will choose a random index from all of the available indices and then output whatever is at that index, even if it's completely empty. This is different from how inputs work, where empty `input` layers within a rule are effectively ignored.

### Impact
This behaviour means that if one wants to have multiple rules in a map with differing numbers of random outputs, one has to duplicate the outputs across the different indices, which has the unfortunate side effect of skewing output probabilities. The consequence is that rules that could be simple and grouped together have to either have way more output indices than they should need (to keep the probabilities appropriate), or have to be split into several rule maps.

### Suggested behaviour
I think it would be more intuitive for indices that are completely empty for a given rule to be ignored. In the case that a user *wants* to output nothing at all at random from a particular rule, they can still do so, by placing an Ignore tile on any of that output index's layers within the rule's region. Since this is fairly unusual and the intended way to randomize outputting anything at all is to use `rule_options`, it's not likely that changing this behaviour will affect many people's rules. And where it does, it's a simple (though potentially tedious) fix.

bjorn:
This change introduces a "compileOutputSet" step which is executed as part of collecting the rules. In this step, each rule only receives as possible outputs the output indexes that are non-empty for the region of that rule.

This makes it much easier to add output variation to only some of the rules. Previously, due to the possibility of empty outputs being chosen, having rules with various amounts of variation required setting up separate rule maps.

Closes #3523

### DIFF
--- a/src/libtiled/tilelayer.h
+++ b/src/libtiled/tilelayer.h
@@ -251,16 +251,16 @@ public:
         {
             if (lhs.mChunkPointer == lhs.mChunkEndPointer || rhs.mChunkPointer == rhs.mChunkEndPointer)
                 return lhs.mChunkPointer == rhs.mChunkPointer;
-            else
-                return lhs.mCellPointer == rhs.mCellPointer;
+
+            return lhs.mCellPointer == rhs.mCellPointer;
         }
 
         friend bool operator!=(const iterator& lhs, const iterator& rhs)
         {
             if (lhs.mChunkPointer == lhs.mChunkEndPointer || rhs.mChunkPointer == rhs.mChunkEndPointer)
                 return lhs.mChunkPointer != rhs.mChunkPointer;
-            else
-                return lhs.mCellPointer != rhs.mCellPointer;
+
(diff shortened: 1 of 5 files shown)


eishiya:
It looks like MatchType Ignore is counted as "no tile" for the purposes of determining whether a given output layer contains anything in a region. Ignore should be the main way that users specify a deliberately empty output (Empty is not sufficient, since that *outputs* Empty, which is not the same as "output nothing").

I think only literally empty regions should count as empty. The presence of any tile, including any MatchType tile, should make the layer count as a valid output.

eishiya:
Suggested wording for the documentation of this change, perhaps it could be added to the last paragraph before "Random Output Example":
> If an output index is completely empty for a given rule, it will never be chosen for that rule.

(And don't forget the Changelog too xP)

Issue #3865, opened 2024-01-03
Update godot 4 exporter to use the new tile rotation/flipping system in godot 4.2


Gnumaru:
**Is your feature request related to a problem? Please describe.**
Currently the godot 4 exporter creates 7 tile alternatives for each tile when exportAlternates is set. This was the only way of doing flipping/rotation in godot 4.0 and 4.1, but in godot 4.2 this is no longer needed since flipx/flipy/transpose info is stored in the 3 upper bits of the tile id in the map, in a similar way as tiled does.

**Describe the solution you'd like**
The exporter could be updated to always export the flip and transpose info, deprecating the exportAlternates property and breaking compatibility with godot 4.0 and 4.1. Optionally, instead of breaking compatibility, exportAlternates could become an int property with three available modes: 0: ignore flipping and rotation while exporting; 1: export flipping and rotation using alternate tiles; 2: export flip and rotation using the tiles upper bits.

**Describe alternatives you've considered**
The current exporter works as (visually) expected when using the exportAlternates property, but it creates very large tilesets due to creating tile alternatives which is no longer needed in godot 4.2

**EDIT**
It is not exactly the higher bits. The exact values are here:
https://docs.godotengine.org/en/latest/classes/class_tilesetatlassource.html#constants
```
TRANSFORM_FLIP_H = 4096
TRANSFORM_FLIP_V = 8192
TRANSFORM_TRANSPOSE = 16384
```

bjorn:
> Optionally, instead of breaking compatibility, exportAlternates could become an int property with three available modes: 0: ignore flipping and rotation while exporting; 1: export flipping and rotation using alternate tiles; 2: export flip and rotation using the tiles upper bits.

To keep things simple, possibly we could export using the new method by default, but export using tile alternatives when this option is enabled?

Alternatively, we could introduce an explicit "godotVersion" project property, which can default to the latest version but which can be set to earlier versions to disable the use of the flip bits, for example. Such a setting could help avoid introducing multiple compatibility options.

Gnumaru:
Indeed, an explicit "godotVersion" project property is better since it would be futureproof. Then the transform bits would be used and "exportAlternates" would be totally ignored when "godotVersion" >= 4.2. And the transform bits would be unused and "exportAlternates" would be taken into account when godotVersion is in the  [4.0, 4.2) range.

eishiya:
[...]

PR #3895, opened 2024-02-24
Update Godot 4 exporter to use the new tile rotation/flipping system in Godot 4.2

Skrapion:
Implements #3865


Skrapion:
This was a pretty simple change. The only thing of note is the version number I added to the documentation. Maybe this should be part of 1.11 instead of 1.10.3.

bjorn:
@Skrapion This change looks great and I think it's no problem to merge it in Tiled 1.10.3.

PR #3909, opened 2024-03-15
Automapping: Always apply output sets with empty index

bjorn:
Output sets with an empty index (output_foo, rather than output1_foo, for example) are now considered unconditional outputs and no longer participate in the random output index selection process.

Effectively, when a rule matches that has output sets with both an empty index and output sets with an index, first its unconditional output set will apply and then a randomly selected output set from its output sets with non-empty index.

For compatibility reasons, this behavior does not affect rule maps in "legacy" mode (when the user manually defined the rule regions).


bjorn:
One thing I'm wondering about, is whether it is alright for these rules to partially apply. Currently, when the NoOverlappingOutput option is used, the unconditional and random parts of the rule output can independently decide not to apply. Alternatively, I could first check the option for both and only apply both of neither of them overlaps previous applications.

bjorn:
> The alternative option sounds more intuitive to me.

Yeah, I've changed the behavior. And more importantly, fixed a breakage where rules without indexes would no longer apply at all, heh. Also I hope the docs updates are clear.

eishiya:
Perhaps the docs for updating old rules needing explicit indices could link to the index-adding script I submitted to tiled-extensions?

I think the documentation is fairly clear, Since these changes reflect the way people tended to assume Automapping worked, I worry spelling them out explicitly might actually confuse people 🤣 I think that eventually, perhaps after a couple of minor versions, this information should be incorporated into the main description for output indices, rather than being described as a change, or at least reworded to de-emphasise the fact that it's a recent change.

bjorn:
> Perhaps the docs for updating old rules needing explicit indices could link to the index-adding script I submitted to tiled-extensions?

Right, of course!

PR #3912, opened 2024-03-19
docs: Add reference to 'Add Output Index' script by eishiya

bjorn:


eishiya:
I don't think mentioning me by name is necessary, the script includes a credit. I also think the wording might mislead people into thinking they have to use the script, whereas it's only needed if they want to automate it, e.g. if they have a bunch of maps. Suggested alternate wording:
`You can automate updating your existing rule maps with the "[Add Output Index](https://github.com/mapeditor/tiled-extensions/blob/master/AddOutputIndex.js)" script.`

bjorn:
@eishiya Whoops, read your comment the second after I merged it...

PR #3917, opened 2024-03-22
Made adding "Copy" when duplicating optional

bjorn:
The option is disabled by default. When enabled, the behavior is a little bit nicer when creating many duplicates, since they now get numbered rather than adding "Copy Copy Copy..."


eishiya:
Reminder to mention in the changelog that renaming layers now trims them :]

PR #3924, opened 2024-04-15
Display read-only state of open files using a lock icon

bjorn:
This should be mostly helpful for people using a VCS like Perforce or Subversion, which may require a file to be checked out or locked before it can be edited.

For now the read-only icon is a lock, which is common, but it may be confusing for Subversion users since they actually need to lock the file to make it writable.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,7 @@
 
 * Added --project command-line parameter for use when exporting (#3797)
 * Added group layer names in "Move Object to Layer" menu (#3454)
+* Added lock icon to open tabs for which the file is read-only
 * Made adding "Copy" when duplicating optional and disabled by default (#3917)
 * Layer names are now trimmed when edited in the UI, to avoid accidental whitespace
 * Scripting: Added API for working with worlds (#3539)
--- a/src/tiled/document.cpp
+++ b/src/tiled/document.cpp
@@ -41,9 +41,13 @@ Document::Document(DocumentType type, const QString &fileName,
     : QObject(parent)
     , mType(type)
     , mFileName(fileName)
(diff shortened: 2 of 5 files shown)


eishiya:
The lock is displayed for maps that have not been saved to a file, which doesn't feel appropriate. It makes some sense, the maps do have to be Saved As to write the changes since they can't be Saved normally, but presenting the file as locked/read-only doesn't seem a good fit.
image

I think it's been discussed before that Tiled probably needs to present unsaved files in some different way, e.g. "Unsaved map" instead of "untitled.tmx", which implies a file that doesn't exist.

bjorn:
> The lock is displayed for maps that have not been saved to a file

Thanks for catching that! It actually happens because internally read-only is set to `!file.isWritable()`, which obviously will be `true` for files that do not exist. I'll extend that with a file exists check.

I agree the unsaved map could be better identified, though the use of "untitled" as file name for unsaved files seems to still be quite ubiquitous. What about just "New Map"?

eishiya:
"Untitled" as a **file** name is ubiquitous, yes, but there's no file, and therefore no file name xP "New Map" would work. Even "Untitled" might be fine, the main thing is to get rid of the file extension that implies there's a file xP

eishiya:
Noticed another issue with this: Save As does not update the lock status, only Save does. If you open a read-only file and then Save As to a new file (which will not be locked), the tab keeps the lock icon.

bjorn:
> Noticed another issue with this: Save As does not update the lock status, only Save does.

I don't think Save does (if the filename stays the same, it would just update based on the file watcher). I've addressed the issue with Save As as well now in a55feecc7494d5131a0cec7106d76de5293c3e52, thanks!

Issue #3927, opened 2024-04-15
World Map chunks getting loaded twice.


grumpygamer:
I have notice some odd behavior when editing world maps. A map chuck gets opened twice (2 tabs) and one of them is a old version, it seems these come from a chuck loaded due to it being next to an active chunk and it's showing an old in-memory copy, not the current. I have lost work with this due to the outdated chucking being saved.  Closing Tiled and restarting always fixes it.

 - Mac 14.2.1
 - Tiled Version: 1.10.2

grumpygamer:
By chunk I mean the sub-maps that make up a world in Tiled.

eishiya:
Those are all separate, wholly independent maps, they're just displayed together as part of a World.

However, it *is* possible that a map being opened as part of restoring the session *and* as part of a World could be part of the issue. It shouldn't happen, and normally doesn't, but I suspect there's something weird happening specifically when automatically restoring the last session.

bjorn:
I was able to reproduce this issue today. The problem happens because the implementation of reload-on-change only replaces the map in its own editor tab and not the instances of the same map within the world in other tabs. So after the reload, we have both the old and the new version loaded. When you then click an old version of the map, it will get its own tab alongside the new version. :-/

I will look for ways to make sure no old versions of maps will stay around in a world. This also means they need to be watched for changes even when no editor is currently open for them.

bjorn:
PR #3939 basically fixes this issue, since it should prevent multiple versions of the same map being loaded or even opened side-by-side. However, I'm keeping this issue open until I've also implemented reloading for world maps that aren't currently open for editing, since otherwise you can still end up with an unexpected old version.

Issue #1785, opened 2017-10-17
Detect external changes of tilesets


issy123:
Tiled does detect external changes of the .tmx files but would be nice if it also detects changes on the .tsx files and auto reload.

bjorn:
Hmm, it only does that when the tileset is explicitly opened, rather than just being loaded as a dependency of a map. I agree it should be extended to reload as well in the latter case.

UliAbo:
> Hmm, it only does that when the tileset is explicitly opened, rather than just being loaded as a dependency of a map. I agree it should be extended to reload as well in the latter case.

Yes it would be great to have it auto-reloaded without it being opened.

PR #3939, opened 2024-05-01
Fixed map reload in case map is open as part of a world

bjorn:
This is a re-implementation of the way maps are reloaded, to make sure the reload affects all instances and not just the opened editor tab. Previously, the reload was implemented by closing and reopening that tab, which was only a viable method before the implementation of worlds.

This is a rather risky change since the code was never prepared for the map and its internal structures to be replaced entirely without any specific change events. There's a high probability of leaving a roaming pointer somewhere.

Marked as draft because there are a few open tasks:

* Needs much more testing.
* Would be nice to restore object and layer selection after the reload when possible.

Fixes part of issue #3927 (though not entirely, since that would require watching map files for changes also when they aren't open in their own tab).

### DIFF
--- a/src/tiled/mapitem.cpp
+++ b/src/tiled/mapitem.cpp
@@ -358,6 +358,21 @@ void MapItem::repaintRegion(const QRegion &region, TileLayer *tileLayer)
 void MapItem::documentChanged(const ChangeEvent &change)
 {
     switch (change.type) {
+    case ChangeEvent::DocumentAboutToReload:
+        for (Layer *layer : mMapDocument->map()->layers())
+            deleteLayerItems(layer);
+        break;
+    case ChangeEvent::DocumentReloaded: {
+        // The renderer has been re-created
+        auto lineWidth = Preferences::instance()->objectLineWidth();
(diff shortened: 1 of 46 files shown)


src/tiled/mapitem.cpp
```diff
@@ -357,7 +357,24 @@ void MapItem::repaintRegion(const QRegion &region, TileLayer *tileLayer)
 
 void MapItem::documentChanged(const ChangeEvent &change)
 {
+    // todo: reload hides the darkening of all but currently selected map
```
bjorn, line 360: Alright, that's just the "selected layers highlight" disappearing, because currently after reload no layers are selected.

bjorn:
> (though not entirely, since that would require watching map files for changes also when they aren't open in their own tab)

I've extended the changes in this PR with another change that makes sure any loaded document is watched for changes on disk and reloaded when appropriate (when it doesn't have unsaved changes). This not only applies to maps loaded as part of a world, but also to tilesets that are loaded as part of maps (issue #1785).

PR #3957, opened 2024-06-04
AppImage: Include all platform themes

bjorn:
This might improve native style based on GTK 3, depending on the context.


bjorn:
Hmm, it was worth a try since I don't get the GTK open file dialog on Xfce with the old builds, but the native Qt one which is a little ugly. However, with the GTK3 platform plugin shipped I get a crash on startup:
```
Gtk:ERROR:../../../../gtk/gtkiconhelper.c:494:ensure_surface_for_gicon: assertion failed (error == NULL): Failed to load /org/gtk/libgtk/icons/16x16/status/image-missing.png: Unrecognized image file format (gdk-pixbuf-error-quark, 3)
Bail out! Gtk:ERROR:../../../../gtk/gtkiconhelper.c:494:ensure_surface_for_gicon: assertion failed (error == NULL): Failed to load /org/gtk/libgtk/icons/16x16/status/image-missing.png: Unrecognized image file format (gdk-pixbuf-error-quark, 3)
Aborted (core dumped)
```

PR #3958, opened 2024-06-04
Scripting: Fixed UI issues after setting class values

bjorn:
When setting a class value from a script, the value was not getting converted from a JS object to a QVariantMap. This tripped up code elsewhere, for example making the UI unable to edit the values (though it could still display them, and they also got saved correctly).

This change adds the missing conversion from JS object to QVariantMap to `tiled.propertyValue` and `Object.setProperty`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,7 @@
 * Changed default shortcut for "Save As" to Ctrl+Shift+S and removed shortcut from "Save All" (#3933)
 * Layer names are now trimmed when edited in the UI, to avoid accidental whitespace
 * Scripting: Added API for working with worlds (#3539)
+* Scripting: Added Object.setProperty overload for setting nested values
 * Scripting: Added Tile.image for accessing a tile's image data
 * Scripting: Added Image.copy overload that takes a rectangle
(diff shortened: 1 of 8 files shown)


bjorn:
Even if this PR adds a convenience function `Object.setProperty`, I've passed on adding an equivalent `Object.property` overload that takes a path for now. This functionality doesn't exist on the C++ side either at the moment, so it would be a larger patch, and the usefulness is less obvious since it just makes this sequence a bit less verbose:
```js
// currently supported
let nestedPropertyValue = object.property("foo").value.bar.value.baz;

// vs:
let nestedPropertyValue = object.property(["foo", "bar", "baz"]);
```
In addition, we'd then also need to add path support to `Object.resolvedProperty`.

There was no need for overloading `Object.setColorProperty` or `Object.setFloatProperty`, because class members have their type defined by their class. The `setProperty` overload taking a path will try to convert the values to the type of the class member (though this currently isn't the case of top-level properties).

`Object.setColorProperty` has been deprecated in favor of using the new `tiled.color` to create a color value. Unfortunately I couldn't immediately find a quick way to do the same with `Object.setFloatProperty`.

PR #3960, opened 2024-06-06
Scripting: add displayImage property to Tile to get pre-cropped image

dogboydog:
I saw #3918 is marking these as `@since 1.11` , so I did the same here, but I could change it. 

The idea here was to provide a pre-cropped image (since `Tile.image` returns the full image)  so that you could display it on a dialog. It came up when I was asking if a new type of widget  that lets you select & preview a tile (a TileButton like a color picker button) might be useful.

Let me know what you think 

Example script: 
https://gist.github.com/dogboydog/363a4fec2e7b1c408e9ecc4b93234bde

Example images 
image
image

### DIFF
--- a/docs/scripting-doc/index.d.ts
+++ b/docs/scripting-doc/index.d.ts
@@ -2474,6 +2474,16 @@ declare class Tile extends TiledObject {
   image: Image;
 
   /**
+   * Returns the image of this tile, cropped to its {@link imageRect}
+   * to include only the visible portion of the image.
+   *
+   * This property is read-only.
+   * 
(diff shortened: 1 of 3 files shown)


bjorn:
While this looks like a useful addition, I'm not sure if it really makes sense to add it is almost trivially supported by the existing API as follows:
```js
let rect = tile.imageRect;
let displayImage = tile.image.copy(rect.x, rect.y, rect.width, rect.height);
```
Though I realize the existing behavior of `tile.image` isn't very intuitive, and we should probably at least add the above snippet to its documentation.

Issue #3961, opened 2024-06-06
Make Stamp Brush right-click cut instead of copying when Shift is held


kdx2a:
**Is your feature request related to a problem? Please describe.**
Moving a group of tiles on a tile layer is too convoluted for such a common operation. Unless I miss something, here's the fastest way to do so (with default keybindings):
* [press R] switch to Rectangular Select
* [left click] select area
* [C-x] cut selection
* [b] switch to Stamp Brush
* [left click] paste area

**Describe the solution you'd like**
Allow to cut a rectangular area in Stamp Brush mode by holding Shift and right-clicking (in addition to the current copying behavior).

**Describe alternatives you've considered**
* Add a binding to quickly delete the last area copied with the Stamp Brush.
* Allow moving a selection directly (ie. by holding right click).

eishiya:
Displacing tiles within a layer is a common operation for some users, but is never needed for others, so overall, it's a rarely-needed feature, which is why it's not gotten much attention.

I think your proposed expansion of the current copying behaviour sounds great! Way more elegant than the more complex behaviour I was thinking of. Shift+right-click seems to be free, it currently causes the Brush Stamp to just not do anything.

Big fat +1 to **making Stamp Brush right-click cut instead of copying when Shift is held**.
(For clarity, consider changing your issue title to something that describes your proposed change, as your title makes it sound like you want something more complicated.)

I'd previously scripted a Move Tiles Tool that allows moving tiles in multiple drags and via the arrow keys, and while a more complex style of operation like that can be desirable, in practice, a quicker way to just cut and paste would be way faster, even in scenarios where small, precise movements of large groups of tiles are desired.

kdx2a:
Thank you for your feedback @eishiya! I had time off today, so I looked into the codebase to see how possible it would be and I ended up implementing the feature myself. It wasn't as straightforward as I would have expected, but it works very well now and is very convenient in terms of workflow.

bjorn:
Implemented by #3963.

PR #3963, opened 2024-06-07
Stamp Brush right-click cut when Shift is held

kdx2a:
Implements #3961

Implementing this required a refactor of the state system in StampBrush.  Behavior and State were previously handled in the same property, which resulted both in wrong behavior when using capture and pressing the behavior keys (replicate by holding right click, start holding shift, release right click, the behavior wasn't set to line as it should have). This commit refactors this system by splitting State and Behavior in two enums and two properties, resulting in making the implemented feature possible, fixing behavior bugs, and future-proofing this system for expansion.

### DIFF
--- a/src/tiled/abstracttilefilltool.cpp
+++ b/src/tiled/abstracttilefilltool.cpp
@@ -77,7 +77,7 @@ void AbstractTileFillTool::deactivate(MapScene *scene)
 void AbstractTileFillTool::mousePressed(QGraphicsSceneMouseEvent *event)
 {
     if (event->button() == Qt::RightButton && event->modifiers() == Qt::NoModifier) {
-        mCaptureStampHelper.beginCapture(tilePosition());
+        mCaptureStampHelper.beginCapture(tilePosition(), false);
         return;
     }
 
--- a/src/tiled/abstracttilefilltool.cpp
+++ b/src/tiled/abstracttilefilltool.cpp
@@ -76,8 +76,8 @@ void AbstractTileFillTool::deactivate(MapScene *scene)
 
 void AbstractTileFillTool::mousePressed(QGraphicsSceneMouseEvent *event)
 {
-    if (event->button() == Qt::RightButton && event->modifiers() == Qt::NoModifier) {
-        mCaptureStampHelper.beginCapture(tilePosition(), false);
+    if (event->button() == Qt::RightButton) {
+        mCaptureStampHelper.beginCapture(tilePosition());
         return;
     }
 
@@ -89,7 +89,8 @@ void AbstractTileFillTool::mouseReleased(QGraphicsSceneMouseEvent *event)
     if (event->button() == Qt::RightButton && mCaptureStampHelper.isActive()) {
         clearOverlay();
 
-        TileStamp stamp = mCaptureStampHelper.endCapture(*mapDocument(), tilePosition());
+        const bool cut = event->modifiers() & Qt::ShiftModifier;
+        TileStamp stamp = mCaptureStampHelper.endCapture(*mapDocument(), tilePosition(), cut);
         if (!stamp.isEmpty())
             emit stampChanged(stamp);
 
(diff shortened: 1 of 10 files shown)


bjorn:
Thank you so much for the nice idea and for the contribution! It works great, though I have a few concerns:

> Behavior and State were previously handled in the same property, which resulted both in wrong behavior when using capture and pressing the behavior keys (replicate by holding right click, start holding shift, release right click, the behavior wasn't set to line as it should have).

Indeed it looks like there is less code duplication with the enum split up. The issue was still there though. Due to `setStamp(TileStamp())` in `beginCapture` and the condition in `modifiersChanged` that the stamp isn't empty, it wasn't updating the brush behavior. I've just removed the condition, since it didn't look like it was actually necessary.

I don't really like the code duplication of the erasing code in `CaptureStampHelper::endCapture` and `Eraser::doErase`. I realize it looks rather different but since it should be doing mostly the same thing, I think it could live in a shared place (like adding a `MapDocument::eraseRegion`). I think the `Eraser::doErase` implementation is preferred because it uses only a single undo command. I can look into changing this tomorrow, if you don't beat me to it.

Also, I wonder if we really need the "cut" state to be set when you start capturing? The code would be rather simpler if we'd just take Shift state at the end of the capture.

Finally, there is a `CaptureStampHelper` because there are a bunch of other tools that allow capturing the brush. Now, the Shift modifier only works for the Stamp Brush. Did you consider whether it can behave consistently for the tools derived from `AbstractTileFillTool` as well?

eishiya:
> Also, I wonder if we really need the "cut" state to be set when you start capturing? The code would be rather simpler if we'd just take Shift state at the end of the capture.

I think this would also be more intuitive for the user.

kdx2a:
Thank you for making Tiled!

> I don't really like the code duplication of the erasing code in `CaptureStampHelper::endCapture` and `Eraser::doErase`. I realize it looks rather different but since it should be doing mostly the same thing, I think it could live in a shared place (like adding a `MapDocument::eraseRegion`). I think the `Eraser::doErase` implementation is preferred because it uses only a single undo command. I can look into changing this tomorrow, if you don't beat me to it.

To be completely honest I'm not sure how to make this change properly, I'd need to get more familiar with this part of the codebase. I'll take a glance at it but I'll probably focus on the other bulletpoints tonight.

> Also, I wonder if we really need the "cut" state to be set when you start capturing? The code would be rather simpler if we'd just take Shift state at the end of the capture.

I don't think it's necessary either. I'll make the change, if you decide you want to keep the state afterwise I'll let you reverse the commit.

> Finally, there is a `CaptureStampHelper` because there are a bunch of other tools that allow capturing the brush. Now, the Shift modifier only works for the Stamp Brush. Did you consider whether it can behave consistently for the tools derived from `AbstractTileFillTool` as well?

I didn't consider it, no. My bad. I'll take a look and try to make it behave consistently.

src/tiled/abstracttilefilltool.cpp
```diff
@@ -76,8 +76,8 @@ void AbstractTileFillTool::deactivate(MapScene *scene)
 
 void AbstractTileFillTool::mousePressed(QGraphicsSceneMouseEvent *event)
 {
-    if (event->button() == Qt::RightButton && event->modifiers() == Qt::NoModifier) {
```
bjorn, line 79: Ah, wait a sec, this "no modifier" check is to avoid conflicting with the "select top-most tile layer with a tile at this location" behavior that is implemented in `AbstractTileTool::mousePressed`. So, this should check that the Ctrl modifier isn't held.

src/tiled/stampbrush.cpp
```diff
@@ -143,10 +143,7 @@ void StampBrush::mousePressed(QGraphicsSceneMouseEvent *event)
             }
             return;
         } else if (event->button() == Qt::RightButton) {
-            if (event->modifiers() == Qt::NoModifier)
-                beginCapture(false);
-            else if (event->modifiers() & Qt::ShiftModifier)
-                beginCapture(true);
+            beginCapture();
```
bjorn, line 146: Here also we'll need to check that the Ctrl modifier isn't used, since in that case we want the ctrl+right-click handling in `AbstractTileTool::mousePressed` to kick in.

bjorn:
> I'm not sure if this is intended behavior or not?

As discussed on Discord, it might be better to avoid the needless history entry, but this is the existing behavior so we'll merge this PR as-is.

Thanks again!

PR #3964, opened 2024-06-13
EditorConfig file

kdx2a:
[EditorConfig](https://editorconfig.org) is a [well established](https://github.com/search?q=path%3A**%2F.editorconfig&type=code) file format standard and text editor plugins collection that helps contributor keep a consistent coding style. This pull request shares the one I've been using when working on Tiled. Not much more to add, this is just nice to have.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+# https://editorconfig.org
+root = true
+
+[*.{cpp,h}]
+indent_style = space
+indent_size = 4
+end_of_line = lf
+charset = utf-8
+spelling_language = en-US
+trim_trailing_whitespace = true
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,11 @@
+# https://editorconfig.org
+root = true
+
+[*.{cpp,h}]
+indent_style = space
+indent_size = 4
+end_of_line = lf
```
bjorn, line 7: Let's remove this line because I'm not sure `lf` is used on Windows, it probably depends on your git settings.

bjorn (commented):
Seems like a good idea, thanks!

PR #3969, opened 2024-06-14
docs: Provide snippet to get a tile's image

bjorn:
Since `Tile.image` can refer to a larger image, to get only the part representing the tile one needs to make a copy.

See #3960


eishiya:
If Image.copy had an overload that took a QRect, the snippet could be rather legible one-liner: ` let image = tile.image.copy(tile.imageRect)` hinthintnudgenudge ;D [QImage has such a method](https://doc.qt.io/qt-6/qimage.html#copy), perhaps it could be exposed to the API?

PR #3970, opened 2024-06-14
Allow drawing lines by holding shift while painting

### LINKED ISSUE #3642: Pressing Shift while painting with Stamp Brush should start line mode
When trying to use Shift + Stamp Brush to draw lines, I often end up clicking first and then pressing Shift because that feels more natural to me, but in Tiled this doesn't work - it does not begin line-drawing *and* causes any further cursor movement to do nothing until I let go of the mouse button.

I think it would be more convenient if pressing Shift while already painting would start drawing a line from the cursor's location at that point. If the user releases Shift before the mouse button, then the line should not be drawn, same as the current line-drawing behaviour.

kdx2a:
Allows starting to draw, holding shift to preview a straight line, and releasing to draw it. This is enabled by the addition of a `PaintStartSet` state. Pressing right click during `StartSet` or `PaintStartSet` cancels the operation and returns to Free. This is inspired by the behavior of the pencil in some image editing software. Thinking about the issue #3642 while working on this, I made sure than clicking, holding shift, then releasing shift, switches from `PaintStartSet` to `StartSet` to avoid user confusion.

I invite the reader to test this feature. I think it is good, since it's non intrusive and eases the better flow when mixing many organic and straight lines. But I understand how this might be controversial.

closes #3642

### DIFF
--- a/src/tiled/stampbrush.cpp
+++ b/src/tiled/stampbrush.cpp
@@ -119,6 +119,7 @@ void StampBrush::mousePressed(QGraphicsSceneMouseEvent *event)
     if (brushItem()->isVisible()) {
         if (event->button() == Qt::LeftButton) {
             switch (mBrushState) {
+            case BrushState::PaintStartSet: // shouldn't occur
             case BrushState::StartSet:
                 doPaint();
                 mStampReference = tilePosition();
@@ -134,7 +135,10 @@ void StampBrush::mousePressed(QGraphicsSceneMouseEvent *event)
                 case BrushBehavior::Line:
                 case BrushBehavior::Circle:
                     mStampReference = tilePosition();
-                    mBrushState = BrushState::StartSet;
+                    if (mBrushState == BrushState::Paint)
+                        mBrushState = BrushState::PaintStartSet;
+                    else
+                        mBrushState = BrushState::StartSet;
                     break;
                 }
                 break;
@@ -142,11 +146,15 @@ void StampBrush::mousePressed(QGraphicsSceneMouseEvent *event)
                 break;
             }
             return;
-        } else if (event->button() == Qt::RightButton &&
-                   !(event->modifiers() & Qt::ControlModifier))
-        {
-            beginCapture();
-            return;
+        } else if (event->button() == Qt::RightButton) {
+            if (mBrushState == BrushState::StartSet || mBrushState == BrushState::PaintStartSet) {
+                mBrushState = BrushState::Free;
+                updatePreview();
+                return;
+            } else if (!(event->modifiers() & Qt::ControlModifier)) {
+                beginCapture();
+                return;
+            }
         }
     }
 
@@ -156,11 +164,14 @@ void StampBrush::mousePressed(QGraphicsSceneMouseEvent *event)
 void StampBrush::mouseReleased(QGraphicsSceneMouseEvent *event)
 {
     switch (mBrushState) {
+    case BrushState::PaintStartSet:
     case BrushState::StartSet:
         if (event->button() == Qt::LeftButton) {
             if (mStampReference != tilePosition()) {
                 doPaint();
                 mBrushState = BrushState::Free;
+            } else {
+                mBrushState = BrushState::StartSet;
             }
         }
         break;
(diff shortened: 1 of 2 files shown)


bjorn:
I tested it, seems like a nice improvement and a behavior that one could get used to. 

I don't have any comparison with other painting software though. I tried to compare with GIMP and Krita, but neither appears to have such behavior (or I just couldn't find it):

* GIMP: The Shift modifier always only draws a line from the last painted location to the current mouse on click. There is no handling of "Shift while painting" nor any way to toggle between free drawing and line drawing without releasing the mouse button in between.
* Krita: Drawing lines appears to require using the special line drawing tool, which always requires you to click at the start location and drag to the end location. Shift modifier on Freehand Brush rather provides a quick way to change brush size.

@eishiya Since in #3642 you wrote "If the user releases Shift before the mouse button, then the line should not be drawn", and this PR behaves differently, it would be great if you could check the behavior.

One thing that's a little surprising to me is the behavior when doing the following:

* Hold Shift to start drawing a line.
* Press and hold Left mouse, drag somewhere.
* Release Shift but hold mouse.

This cancels the line drawing and now you're kind of stuck in a useless mode until you also release the mouse button. I had expected the behavior to be the same as when pressing/releasing shift while painting: to draw the line and enter free painting mode. Does that sound reasonable?

eishiya:
I do not like how if I click, then hold shift, and then release shift without releasing the click, the line is drawn anyway. If I release shift, it's because I've changed my mind about drawing a line from that particular location, the last thing I want is for that line to be committed. It's also unnatural for a preview to be committed just because the state of a modifier changed, IMO. I expect to modify the map with a click or a release, never with a modifier change. I see that right-clicking cancels the line, but that's somewhat unexpected since usually Stamp Brush right-click samples tiles.

I think that after Shift is released, the tool should NOT draw the line, exit line mode, and require letting go of LMB to start painting again, similar to Tiled's current behaviour. #3642 just asks for the ability to start line mode if LMB is already held xP

However, I do see the benefits of committing on releasing shift, particularly when blocking in maps (i.e. when precision is less important than speed). Perhaps it could be provided as an option on the Stamp Brush tool, called something like "quick line mode"? When this mode is enabled, releasing shift while LMB is held should commit the line, and the brush should switch to freehand painting mode as Bjorn described, regardless of whether the user clicked before or after pressing Shift.

bjorn:
> So I agree with eishiya's proposal to make it optional. Where should I look in the codebase to see how to add such a setting?

I'm not really comfortable with adding an option for this. It seems the easy way out as soon as there isn't a consensus about the desired behavior, but it puts a burden on both the maintainers and the users forever.

@kdx2a Were you really looking for the implemented behavior or was this something that came out of trying to address #3642?

I think there's also an alternative that still allows mixing many organic and straight lines: 

* When releasing the mouse while painting and holding Shift, it would set the start position of the next line.
* When pressing LMB, it will draw the previewed line, and when you drag from there, it would be freehand painting.

With the above, you can hold Shift, draw straight lines while LMB is released and organic pieces while LMB is pressed. Releasing Shift will always cancel any current line preview.

I think the main tricky part in the above is what to do with the initial "hold Shift and then LMB + drag + release". Currently, this is a quick way to draw a single line segment and it should probably stay that way. That means that organic segments can only be drawn by drag after drawing at least one line.

kdx2a:
> I'm not really comfortable with adding an option for this. It seems the easy way out as soon as there isn't a consensus about the desired behavior, but it puts a burden on both the maintainers and the users forever.

I see, I respect this design philosophy.

> @kdx2a Were you really looking for the implemented behavior or was this something that came out of trying to address #3642?

I was looking for the implemented behavior while keeping in mind #3642. But I warrant it might be better to see this pull request as unrelated, since in the end it doesn't seem to fix #3642 and I think a new behavior in this vein is more beneficial to Tiled as a whole.

> I think there's also an alternative that still allows mixing many organic and straight lines, which might be less controversial:
* When releasing the mouse while painting and holding Shift, it would set the start position of the next line.
* When pressing LMB, it will draw the previewed line, and when you drag from there, it would be freehand painting.

I like this. Here is how I think approaching implementing this behavior:
* Holding Shift and clicking sets a start point.
* Clicking somewhere else draws the line. If Shift is held, repeat, otherwise, start free painting.

PR #3972, opened 2024-06-17
GitHub Actions: Updated macOS and Windows images

bjorn:
The `macos-11` image is now deprecated and doing blackouts.

Switched both to 'latest' to avoid them going stale again in the future.

AppImage remains on `ubuntu-20.04`, since it should generally be built on the oldest still supported version.


bjorn:
Closed this since GitHub UI wasn't showing it was running the CI. It seems to be running after all, but anyway this is not superseded by #3973.

PR #3973, opened 2024-06-17
GitHub Actions: Updated macOS and Windows images

bjorn:
The `macos-11` image is now deprecated and doing blackouts.

Switched both to 'latest' to avoid them going stale again in the future.

AppImage remains on `ubuntu-20.04`, since it should generally be built on the oldest still supported version.

### DIFF
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -184,7 +184,7 @@ jobs:
 
   macos:
     name: macOS (${{ matrix.version_suffix }})
-    runs-on: macos-11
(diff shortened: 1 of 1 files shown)


bjorn:
The builds on `macos-12` have the same issue as `macos-14`, in that the unsigned builds will not run at all, with the OS claiming they are "damaged" (tested on macOS 12 and 14). This is unfortunate since it means there might be no way to run PR builds, but at least the signed builds seem to work so I'm just going with `macos-latest` (14) for now.

PR #3977, opened 2024-06-18
Enabled template resolution for the tmxrasterizer application.

Laskoran:
Enabled template resolution for the tmxrasterizer application.
This then provides that image objects are properly rendered if they stem from a template

### DIFF
--- a/src/tmxrasterizer/main.cpp
+++ b/src/tmxrasterizer/main.cpp
@@ -53,6 +53,8 @@ int main(int argc, char *argv[])
     app.setApplicationVersion(QLatin1String("1.0"));
 
     PluginManager::instance()->loadPlugins();
+
+    // Necessary to enable loading of object templates in XML format
     XmlObjectTemplateFormat xmlObjectTemplateFormat;
     PluginManager::addObject(&xmlObjectTemplateFormat);
 
--- a/src/tmxrasterizer/main.cpp
+++ b/src/tmxrasterizer/main.cpp
@@ -28,6 +28,7 @@
 
 #include "pluginmanager.h"
 #include "tmxrasterizer.h"
+#include "tmxmapformat.h"
 
 #include <QCommandLineParser>
 #include <QDebug>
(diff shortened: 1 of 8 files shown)


src/tmxrasterizer/main.cpp
```diff
@@ -52,6 +53,8 @@ int main(int argc, char *argv[])
     app.setApplicationVersion(QLatin1String("1.0"));
 
     PluginManager::instance()->loadPlugins();
+    XmlObjectTemplateFormat xmlObjectTemplateFormat;
+    PluginManager::addObject(&xmlObjectTemplateFormat);
```
bjorn, line 57: That's a somewhat unexpected fix, especially since it means `tmxrasterizer` now depends on `libtilededitor`, just because of this small interface class.

Alternatively, we could make `Tiled::readObjectTemplate` behave similar to `Tiled::readMap` and `Tiled::readTileset`, such that it falls back on trying to load the file as an XML template file.

Or we could make an actual plugin that implements the file format interfaces for the XML format, so that these formats no longer need special handling (and could even be disabled for those who want to use JSON exclusively).

In any case, thanks for showing where the issue is and a possible fix!

Laskoran: Funnily, had the same thoughts, not liking the new dependency. But while the fix introduces this new dependency, it had the benefit to keep the changes local to the tmxrasterizer sub-project.

But does this not automatically mean that 'XmlObjectTemplateFormat' is currently wrongly located in 'libtilededitor' and should rather be in 'libtiled'?

bjorn: You're right that the `XmlObjectTemplateFormat` could just be moved to the `libtiled` project, along with the related `TmxMapFormat` and `TsxTilesetFormat` classes. From there they could even be automatically added to the `PluginManager`.

Though I kind of liked the idea of putting the XML formats on the same footing as the JSON formats by moving it into a plugin that can be disabled.

bjorn: I've decided to move those classes for now. To introduce a new plugin is a bigger change better reserved for after the 1.11 release. :-)

PR #3978, opened 2024-06-18
CI: Don't fail a build if upload to Sentry failed

bjorn:
The Sentry upload is optional.

Also, increase the log level to possibly find out why the upload appears to have failed with "error: Invalid checksum" last time.


bjorn:
Well, I guess the Sentry upload is optional in general, but not when we're building a release. So the condition should be tweaked somewhat.

PR #3979, opened 2024-06-20
AppVeyor: Updated to Python 3.12

bjorn:


appveyor.yml
```diff
@@ -13,19 +13,19 @@ environment:
...
-      PYTHONHOME: C:\Python38-x64
+      PYTHONHOME: C:\Python312-x64
       MINGW: C:\Qt\Tools\mingw1120_64
       DEFAULT_PROFILE: x86_64-w64-mingw32-gcc-11_2_0
       PUSH_RELEASE: true
       ENABLE_ZSTD: true
       TILED_ITCH_CHANNEL: windows-64bit
     - QTDIR: C:\Qt\5.15\mingw81_32
-      PYTHONHOME: C:\Python38
+      PYTHONHOME: C:\Python312
```
bjorn, line 28: Actually since the 32-bit builds against Qt 5 are meant for compatibility with Windows 7, using Python 3.9+ is a bad idea because it needs at least Windows 10.
```suggestion
      PYTHONHOME: C:\Python38
```

PR #3985, opened 2024-06-22
Update Czech translation

encukou:
Hello!
Here's an update for the Czech translation.

### DIFF
--- a/translations/tiled_cs.ts
+++ b/translations/tiled_cs.ts
@@ -60,7 +60,7 @@
     <message>
         <location line="-33"/>
         <source>Failed to load project &apos;%1&apos;.</source>
-        <translation type="unfinished"></translation>
+        <translation>Nepodařilo se načíst projekt &apos;%1&apos;.</translation>
     </message>
     <message>
         <location line="+44"/>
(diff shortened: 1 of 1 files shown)


translations/tiled_cs.ts
```diff
@@ -7291,92 +7291,92 @@ Prosím vyberte formát ručně.</translation>
...
         <location line="+90"/>
         <source>Godot exporter does not support property of type &apos;%1&apos;</source>
-        <translation type="unfinished"></translation>
+        <translation>Exportér pro Godot nepodporuje vlastnost typu &apos;%1&apos;</translation>
     </message>
     <message>
         <location line="+97"/>
         <source>exportAlternates is deprecated. Godot 4.2 supports native tile rotation. Tileset: %1</source>
-        <translation type="unfinished"></translation>
+        <translation>exportAlternates už není potřeba. Godot 4.2 podporuje otáčení dlaždic přímo. Sada dlaždic: %1</translation>
```
encukou, line 7344: I translated “deprecated” as “not necessary any more”. Should it be something like “will stop working soon” instead?

bjorn: I think that translation is fine. Deprecated does imply it might be removed in the future usually, but it doesn't specify if or when.

bjorn:
Thank you so much for the update, @encukou!

PR #3999, opened 2024-07-08
Qbs: Use pkg-config to depend on zlib

bjorn:
This way we should support building Tiled with zlib in some strange location, like on NixOS.

To support this on the latest version of Qbs, we need to set `Project.qbsModuleProviders` since it no longer falls back on `pkg-config` by default. Since this property was introduced in Qbs 1.21, this becomes the minimum Qbs version.


bjorn:
Since Ubuntu 22.04 apparently ships with Qbs 1.19.2, we'll need wait with this until we no longer support this OS or find a workaround.

PR #4005, opened 2024-07-11
Scripting: Added MapEditor.currentBrushChanged signal

bjorn:
So scripts can respond to brush changes. If they are careful, they can also adjust the brush, though this will trigger another `currentBrushChanged` call.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 ### Unreleased
 
 * Scripting: Added `FileFormat.nameFilter`
+* Scripting: Added `MapEditor.currentBrushChanged` signal
 
 ### Tiled 1.11.0 (27 June 2024)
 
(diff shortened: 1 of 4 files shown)


franciscod:
I've cloned this branch, built tiled and with the new signal my extension works as I wanted!

With this, if there's a `layer` property (set to a layer name) on a tile you select on the tileset, the layer with the matching name gets selected (and all others get deselected).

```
var selectLayerOfTileWhenBrushChanged = function() {
	var asset = tiled.activeAsset
	if (!asset.isTileMap) {
		return
	}
	var brushLayer = tiled.mapEditor.currentBrush.layers[0]
	if (brushLayer.size.height != 1 || brushLayer.size.width != 1) {
		return
	}

	var tile = brushLayer.tileAt(0, 0)
	var tileLayer = tile.property('layer')
	if (!tileLayer) {
		return
	}

	for (var i = 0; i < asset.layerCount; i++) {
		var layer = asset.layers[i]
		var shouldSelect = layer.name == tileLayer
		layer.selected = shouldSelect
	}
}

tiled.mapEditor.currentBrushChanged.connect(selectLayerOfTileWhenBrushChanged)
```

Issue #3997, opened 2024-07-05
Expand/collapse layers API


hnbdr:
When working with maps with a large number of layers, a scripting API for expanding/collapsing GroupLayer and ObjectGroup would be very helpful.

As @bjorn suggested, this can be done by exposing the Layers and Objects views functions in the MapEditor interface.

Oval17:
Hi @hnbdr @bjorn 
I'd like to work on this. I see PR #4009 made progress but has been open since July 2024 with some unresolved issues around segfault protection. 
Would it be appropriate if I work on top of that PR?

bjorn:
@Oval17 Ah, thanks for bringing this PR to my attention again, but I really should provide some feedback to the original author and then we can give them some time to respond. If it turns out to be abandoned you could finish it, but I just had a look and it really seems like it's almost done.

Oval17:
@bjorn Yes sure as I can see you have asked author regarding pr and reviewed the pr as well, will move accordingly.
Thanks

bjorn:
@Oval17 Alright, it's been over a week now with no response from @PalaBeaveR, so I think it's alright if you continue their work. If you pull the branch from their fork and commit on top I can still easily credit you both when merging this.

PalaBeaveR:
Very sorry. I meant to respond to this when i saw it, but lost track of it.
Feel free to take this over as i don't have the time or interest to work on
this right now.

Oval17:
Hey @bjorn @PalaBeaveR 
Yes sure thanks for confirmation, will take over.

PrateekSingh070:
Hi, I’m new to the project and interested in working on this issue.
I’m currently setting up the development environment and would like to try fixing this if it’s available.

sam-coolshrestha:
Hi! I'd like to work on this issue if it's still available.

bjorn:
@PrateekSingh070 @sam-coolshrestha I appreciate your enthusiasm but there is already a PR open for finishing this at #4379 so please look around for another issue to work on.

I realize this is currently the last open issue with explicit "good first issue" label, but any issue is a potentially good first one! You can also play around with Tiled to find things that could be improved. :-)

PR #4009, opened 2024-07-14
Added Expand/Collapse API to GroupLayer

PalaBeaveR:
Related issue: #3997

Added a new scripting field in GroupLayer called expanded that controls the layer's expand state.

From what I'd been able to gather, in order to expand/collapse layers I would need access to LayerView(QTreeView) which controls layer expansion with setExpanded. It wasn't possible since those properties were privated so I added getters to access [LayerDock](https://github.com/PalaBeaveR/tiled/blob/d684b2810cf5b1aeb9d70a9e91bcb98ee935de0d/src/tiled/mapeditor.h#L86), [LayerView](https://github.com/PalaBeaveR/tiled/blob/d684b2810cf5b1aeb9d70a9e91bcb98ee935de0d/src/tiled/layerdock.h#L57), [QAbstractProxyModel](https://github.com/PalaBeaveR/tiled/blob/d684b2810cf5b1aeb9d70a9e91bcb98ee935de0d/src/tiled/layerdock.h#L95).

Right now the expanded state is stored in [GroupLayer](https://github.com/PalaBeaveR/tiled/blob/d684b2810cf5b1aeb9d70a9e91bcb98ee935de0d/src/libtiled/grouplayer.h#L70) and is not synced, however since Qt already stores that state itself, the scripting interface can be made to use that instead.

I also don't know if the same API is appropriate for ObjectGroup(as requested by #3997), but i have that code made in the same way as GroupLayer if it is needed.

I'm not sure if the new definitions in changelayer.h/undocommands.h are needed since i was just copying code from other places in the repo so it would be nice to have that clarified.

### DIFF
--- a/src/libtiled/grouplayer.h
+++ b/src/libtiled/grouplayer.h
@@ -56,6 +56,8 @@ public:
     QList<Layer*>::const_iterator begin() const { return mLayers.begin(); }
     QList<Layer*>::const_iterator end() const { return mLayers.end(); }
 
+    bool isExpanded() const;
+    void setExpanded(bool expanded);
 protected:
     void setMap(Map *map) override;
     GroupLayer *initializeClone(GroupLayer *clone) const;
@@ -64,8 +66,19 @@ private:
     void adoptLayer(Layer &layer);
 
     QList<Layer*> mLayers;
+
+    bool mExpanded = true;
 };
 
+
+inline bool GroupLayer::isExpanded() const
+{
+    return mExpanded;
+}
+
+inline void GroupLayer::setExpanded(bool expanded)  {
+    mExpanded = expanded;
+}
 
 inline int GroupLayer::layerCount() const
 {
--- a/src/tiled/editablegrouplayer.cpp
+++ b/src/tiled/editablegrouplayer.cpp
@@ -20,10 +20,16 @@
 
 #include "editablegrouplayer.h"
 
+#include "changelayer.h"
 #include "addremovelayer.h"
 #include "addremovetileset.h"
 #include "editablemap.h"
 #include "scriptmanager.h"
+#include "documentmanager.h"
+#include "mapeditor.h"
+#include "layerdock.h"
+#include "layermodel.h"
+#include "qabstractproxymodel.h"
 
 #include <QCoreApplication>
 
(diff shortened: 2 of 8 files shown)


bjorn:
I think it would be better to allow the scripting API to interact directly with the UI. There is no need for this to go through an undo command. I also don't think we need to store that state on the GroupLayer object itself.

There already is `QSet<int> expandedGroupLayers` on the `MapDocument`, which actually exists only because it is saved in the session. But if the scripting API just talks to the view, updating that container should happen automatically.

Maybe it could just be some additional functions like `tiled.mapEditor.layersView.setExpanded(layer, true)`, along with `isExpanded`.

PalaBeaveR:
Alright I've removed the state from `GroupLayer` and added `setExpanded/isExpanded` to `tiled.mapEditor.layersView`.
`ObjectGroup` and `tiled.mapEditor.objectsView` got the same treatment.

One thing I noticed is that for now changing group layer expanded state does not reflect in `ObjectsView`. This is expected, but need to figure out how to go about it, because i assume it would be better to not sync them between trees, but need some input on that.

Also I got segfaults when passing the wrong layer to the `setExpanded` method and want to know whether this should be handled or not.

src/libtiled/grouplayer.h
```diff
@@ -55,7 +55,6 @@ class TILEDSHARED_EXPORT GroupLayer : public Layer
     QList<Layer*>::iterator end() { return mLayers.end(); }
     QList<Layer*>::const_iterator begin() const { return mLayers.begin(); }
     QList<Layer*>::const_iterator end() const { return mLayers.end(); }
-
```
bjorn, line 58: Please avoid changing this file at all. I realize these are remnants of earlier changes.

src/tiled/editablegrouplayer.cpp
```diff
@@ -136,6 +139,20 @@ void EditableGroupLayer::addLayer(EditableLayer *editableLayer)
...
+    auto mapEditor = static_cast<MapEditor*>(documentManager->editor(Document::MapDocumentType));
+    return mapEditor->layerDock()->isExpanded(groupLayer());
+}
+
+inline void EditableGroupLayer::setExpanded(bool expanded)
+{
+    auto documentManager = DocumentManager::instance();
+    auto mapEditor = static_cast<MapEditor*>(documentManager->editor(Document::MapDocumentType));
+    mapEditor->layerDock()->setExpanded(groupLayer(), expanded);
+}
```
bjorn, line 154: Please revert your changes to `editablegrouplayer.cpp/h` and `editableobjectgroup.cpp.h` entirely. Since we decided to go with a view-based API we don't need these functions here at all.

src/tiled/layerdock.h
```diff
@@ -54,6 +55,11 @@ class LayerDock : public QDockWidget
      */
     void setMapDocument(MapDocument *mapDocument);
 
+    Q_INVOKABLE bool isExpanded(EditableGroupLayer *layer) const { return isExpanded(layer->groupLayer()); };
+    Q_INVOKABLE void setExpanded(EditableGroupLayer *layer, bool expanded) { setExpanded(layer->groupLayer(), expanded); };
```
bjorn, line 59: These functions need to check at the very least that `layer` is not a `nullptr`. If so, raise a script error (see other script functions for how to do this) and return. I think their implementation should be moved into the cpp file.

src/tiled/objectsdock.h
```diff
@@ -41,6 +43,11 @@ class ObjectsDock : public QDockWidget
 
     void setMapDocument(MapDocument *mapDoc);
 
+    Q_INVOKABLE bool isExpanded(EditableObjectGroup *layer) const { return isExpanded(layer->objectGroup()); };
+    Q_INVOKABLE void setExpanded(EditableObjectGroup *layer, bool expanded) { setExpanded(layer->objectGroup(), expanded); };
```
bjorn, line 47: Same here regarding `nullptr` checks.

bjorn:
I'll close this one now that @Oval17 has taken over at #4379. Thanks for getting this started, @PalaBeaveR!

PR #4034, opened 2024-08-15
Added "Tileset > Edit Tileset Image Parameters..." action

bjorn:
This makes the functionality easier to discover. Previously editing the tileset image parameters was only possible by clicking the value of the Image property in the Properties view, and then the appearing "Edit" button.


bjorn:
@CodiumAI-Agent /review

(just curious what feedback it might have)

bjorn:
Oh well, the agent failed to understand a line was added to a function rather than a "duplicate" being added. For a moment I thought it might have caught the duplication between `TilesetEditor::editTilesetParameters` and `TilesetParametersEdit::buttonClicked`, but no, I think the latter was likely not in the context.

PR #4039, opened 2024-08-21
Qbs: Add option to specify custom Python pkg-config name

mechakotik:
I am Gentoo maintainer of Tiled. In Gentoo, there may be many Python versions installed at the same time for compatibility reasons, and each version has separate name in pkg-config.

For example, for Python 3.12 it would be named python-3.12-embed, so running pkg-config for python-embed will fail. Adding this option so it's possible to specify the correct name from build script.


bjorn:
Sounds reasonable, thanks!

PR #4043, opened 2024-08-28
Refactored World management

### LINKED ISSUE #4042: Segfault when attempting to operate on a world accessed through a directory symlink
**Describe the bug**
Attempting to operate in any meaningful way on a world accessed through a directory symlink will always cause a segfault. It's possible to open the world, but only the active map will be displayed, not the whole world, and if Tiled has been relaunched since the world was created, then switching to the world tool will cause an immediate segfault, as will closing the world. Since it's not possible to close the world through Tiled's interface, external measures are necessary to recover a usable session, e.g. editing the session file by hand, or deleting the offending directory symlink so that Tiled can't load the world on relaunch.

**To Reproduce**
1. Boot a Linux environment with Tiled installed.
2. Open the terminal.
3. If your terminal or shell is configured to start somewhere other than your home directory, run `cd` (or adjust remaining steps for use with your preferred working directory).
4. Run `mkdir foo`. Adjust filename as needed if something already exists at `foo`.
5. Run `ln -s foo bar`. Adjust filename as needed if something already exists at `bar`.
6. Place an image file in `/home/$USER/foo`. (Replace `$USER` with your username.) Any image should do.
7. Launch Tiled.
8. Create a tileset in `/home/$USER/bar` using the image file there. Be sure to create the tileset in `bar`, not in `foo`, even though we know they're actually the same location.
9. Create two maps in `/home/$USER/bar` which use the tileset.
10. Create a new world in `/home/$USER/bar`.
11. Add both maps to the world. Notice that even with the world tool active, each tab now shows only the associated map, not the world.
12. Exit. Save changes to the world when prompted to do so.
13. Relaunch Tiled. If you closed the tileset, maps, and/or world, reopen them. Be sure to open them through `bar`, not `foo`.
14. Switch to the world tool. Tiled should now segfault.
15. Relaunch Tiled. It should restore the session you closed out of.
16. Try to unload the world or close all files open in the session. Tiled should now segfault again.
17. Relaunch Tiled and observe that the uncooperative world is still open.
18. Close Tiled and delete `/home/$USER/bar`.
19. Relaunch Tiled and observe that the frequent crashes have abated.

**Expected behavior**
The behavior I would expect would be identical to that presented in the case that the path does not include a directory symlink: namely, the initial action of adding both maps to the world should cause display of both maps next to each other, rather than only one map, and then, switching to the world tool after relaunching should allow the user to move the maps relative to each other, rather than crashing.

**Media**
Not applicable; issue occurs regardless of the graphics involved.

**Specifications:**
 - OS: EndeavourOS on x86_64
 - Tiled Version: 1.11.0

bjorn:
This way it is more inline with how the other assets are managed. It also avoids silly issues, like a `WorldDocument` pointer not getting looked up successfully due to a file name mismatch.

Fixes #4042

### DIFF
--- a/src/tiled/changeworld.cpp
+++ b/src/tiled/changeworld.cpp
@@ -29,38 +29,50 @@
 
 namespace Tiled {
 
-AddMapCommand::AddMapCommand(const QString &worldName, const QString &mapName, const QRect &rect)
-    : QUndoCommand(QCoreApplication::translate("Undo Commands", "Add Map to World"))
-    , mWorldName(worldName)
+AddRemoveMapCommand::AddRemoveMapCommand(WorldDocument *worldDocument,
+                                         const QString &mapName,
+                                         const QRect &rect,
+                                         QUndoCommand *parent)
+    : QUndoCommand(parent)
+    , mWorldDocument(worldDocument)
     , mMapName(mapName)
     , mRect(rect)
 {
 }
 
-void AddMapCommand::undo()
+void AddRemoveMapCommand::addMap()
 {
-    WorldManager::instance().removeMap(mMapName);
+    auto world = mWorldDocument->world();
+    world->addMap(mMapName, mRect);
+    emit mWorldDocument->worldChanged();
 }
 
-void AddMapCommand::redo()
+void AddRemoveMapCommand::removeMap()
 {
-    WorldManager::instance().addMap(mWorldName, mMapName, mRect);
+    auto world = mWorldDocument->world();
+
+    const int index = world->mapIndex(mMapName);
+    if (index < 0)
+        return;
+
+    world->removeMap(index);
+    emit mWorldDocument->worldChanged();
 }
 
 
-RemoveMapCommand::RemoveMapCommand(const QString &mapName)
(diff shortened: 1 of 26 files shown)


src/tiled/changeworld.cpp
```diff
@@ -29,74 +29,90 @@
...
 {
-    WorldManager::instance().addMap(mWorldName, mMapName, mRect);
+    auto world = mWorldDocument->world();
+
+    const int index = world->mapIndex(mMapName);
+    if (index < 0)
+        return;
+
+    world->removeMap(index);
+    emit mWorldDocument->worldChanged();
```
bjorn, line 59: Each change affecting a world just emits `worldChanged` now, whereas before the `WorldManager::worldsChanged` signal was emitted. Probably we should forward the former to the latter, because existing code relies on getting `worldsChanged` (and often, we need to respond to any world changing, because we do not know beforehand which world a map will then belong to).

bjorn: It appears I already added the connection between `worldChanged` and `worldsChanged`.

PR #4044, opened 2024-08-29
Delay loading of tile stamps until after initializing the session

bjorn:
The tilesets referenced by stored tile stamps may rely on custom types defined in the project and possibly even on custom tileset formats added by extensions. Hence, it is better to wait with loading them until after the project has been loaded and script manager is initialized.

The user will still run into issues when the project isn't loaded on startup, but this at least makes running into issues less likely.

It could help to force reloading of the stamps when switching projects.


bjorn:
> Language not supported

Oh well, that's disappointing.

PR #4046, opened 2024-08-30
docs: Mention DotTiled library in C#/.NET section

dcronqvist:
Hello! Due to the amount of unmaintained and/or relatively old libraries in the .NET ecosystem for Tiled parsers, I've decided to make my own ([DotTiled](https://github.com/dcronqvist/DotTiled)) that I aim to maintain actively :)

Some of the ideas I want to accomplish with this library:
- fast and memory-efficient (see benchmark results for current status)
- easy to use and flexible API
- support for custom types with (in the future) automatic mapping between classes and Tiled custom types
- detailed usage documentation
- abstract renderer with ability to hook in relevant rendering backend

Please let me know if the description I've entered is too positively biased or should occur later in the list of mentioned libraries.

### DIFF
--- a/docs/reference/support-for-tmx-maps.rst
+++ b/docs/reference/support-for-tmx-maps.rst
@@ -68,6 +68,7 @@ C++
 C#/.NET
 ~~~~~~~
 
+-  `DotTiled <https://github.com/dcronqvist/DotTiled>`__: A fast, memory-efficient, and easy to use .NET library for loading Tiled maps and tilesets with support for both TMX and JSON formats.
 -  `TiledCS <https://github.com/TheBoneJarmer/TiledCS>`__: A dotnet library for loading Tiled tilesets and maps (TMX/TSX or JSON).
 -  `MonoGame.Extended <https://github.com/craftworkgames/MonoGame.Extended>`__
    has a Tiled map loader and renderer that works with MonoGame on all
--- a/docs/reference/support-for-tmx-maps.rst
+++ b/docs/reference/support-for-tmx-maps.rst
@@ -69,11 +69,13 @@ C#/.NET
 ~~~~~~~
 
 -  `DotTiled <https://github.com/dcronqvist/DotTiled>`__: A fast, memory-efficient, and easy to use .NET library for loading Tiled maps and tilesets with support for both TMX and JSON formats.
--  `TiledCS <https://github.com/TheBoneJarmer/TiledCS>`__: A dotnet library for loading Tiled tilesets and maps (TMX/TSX or JSON).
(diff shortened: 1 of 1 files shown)


bjorn:
Thanks for providing an updated alternative to the existing Tiled loaders for .NET / C#! I think the description is fine, and it's great that you've done a thorough comparison to the other options.

Since also TiledCSPlus and TiledLib appear to be actively maintained, I guess it would be a good idea to add those to the list as well (and move TiledCS to the list of no longer maintained ones).

Btw, I noticed TiledLib was [recently updated to .NET 8.0](https://github.com/Ragath/TiledLib.Net/releases/tag/v5.0.0), so your comparison table could need a slight update in that regard. :-)

Finally, regarding ease of use, the quick start docs do look a little intimidating and have a lot of duplicated code. I understand the flexibility is desired, but it would be nice if it could be done with less duplication and passing lots of parameters around. For [`rs-tiled`](https://github.com/mapeditor/rs-tiled), we have for example two interfaces, the [`ResourceReader`](https://docs.rs/tiled/latest/tiled/trait.ResourceReader.html#associatedtype.Resource) and the [`ResourceCache`](https://docs.rs/tiled/latest/tiled/trait.ResourceCache.html) and then a [`Loader`](https://docs.rs/tiled/latest/tiled/struct.Loader.html) that uses those interfaces. This way we provide flexibility while also providing convenience based on a `DefaultResourceCache` and a `FilesystemResourceReader`, which are used when the user just calls [`Loader::new()`](https://docs.rs/tiled/latest/tiled/struct.Loader.html#method.new). It might be worth considering how such a setup could be applied to make DotTiled easier to use in the common case (and to add caching, which as far as I understand would need to be manually added, for example to avoid loading the same tileset multiple times).

dcronqvist:
Good idea! I've added TiledLib and TiledCSPlus to the list of libraries, and moved TiledCS to the unmaintained list. For TiledLib and TiledCSPlus, I simply used their corresponding GitHub repository descriptions.

Will update comparison table in README to reflect TiledLib's latest release.

Regarding your ease-of-use comment, I 100% agree that I need some common implementations of the resolving mechanisms, including built-in caching. However, they're further down the todo-list than what I had planned for 0.1.0, so hopefully in some upcoming version I'll have them ready for people to use :) I'll have a closer look at the rs-tiled crate and use it as inspiration!

bjorn:
Thanks for the update! It is now live on https://doc.mapeditor.org/de/latest/reference/support-for-tmx-maps/#c-net, whereas the "stable" docs will update upon the next tagged release.

PR #4048, opened 2024-09-05
docs: Move TiledCSPlus to list of unmaintained C#/.NET libraries

dcronqvist:
I was recently informed by the author and maintainer of [TiledCSPlus](https://github.com/krnlexception/TiledCSPlus) that they will no longer be maintaining the library, which was solidified by the repository being archived.

To reflect this in the Libraries and Frameworks section of the docs, it should be moved to the list of unmaintained (but probably still quite useful) libraries.


bjorn:
Thanks! And it's nice of the author of TiledCSPlus has thought about this and in the end, makes the choice for people looking for a C# loader easier!

Be sure to also update the library status in your own README.md (as well as the .NET Targets for TiledLib). ;-)

PR #4061, opened 2024-09-27
docs: Added image showing an example of transitions to nothing

bjorn:
To make this case more noticeable and explanatory.

Thanks to Game Shovel LLC for the example art!


albertreed:
Just re-iterating, that Game Shovel LLC grants rights to the image under the MIT license.

PR #4073, opened 2024-10-07
Expand Sphinx Dependencies for Documentation Generation

Kureii:
While following the instructions to generate the Tiled documentation, I encountered missing dependencies on my system. To resolve this, I have added the following packages to the setup instructions:

- myst_parser
- sphinx_design
- sphinx_reredirects

These additions ensure that all necessary components are available for successful documentation generation, aligning the setup with the current requirements.

**Changes:**
Updated dependencies list for Sphinx documentation in the setup instructions.
This update should help others avoid missing dependency issues when building the documentation.


bjorn:
Thanks! Actually, I think it would be better to update this documentation to use the `docs/requirements.txt` file that was added at some point. Could you adjust the patch accordingly?

Kureii:
okay, I'll do it.

bjorn:
Thanks! That seems quite verbose, but I guess the note about using virtual environments could be useful (I should probably be using that myself as well, heh).

PR #4074, opened 2024-10-09
Add ignore for other IDEs and add ignore for docs/.venv/

Kureii:
I edit git ignore for other IDEs. This is the first step to add support for cmake build and make the project available to more developers.


Kureii:
Are you okay with my plan?

.gitignore
```diff
@@ -7,8 +7,30 @@ default/
...
 
-# IDE files
+# QT Creator
+*.user
+*.creator.*
+*.autosave
+*.pro.user
+*.qbs.user
 *.qbs.user*
+*.pro.user.*
```
bjorn, line 17: I don't see why we would need to ignore pro files again. You're not going to add back a qmake project, are you?

.gitignore
```diff
@@ -7,8 +7,30 @@ default/
 # Mac stuff
 *.DS_Store
 
-# IDE files
+# QT Creator
+*.user
+*.creator.*
+*.autosave
+*.pro.user
+*.qbs.user
```
bjorn, line 15: This should be covered by the existing `*.qbs.user*` ignore rule, right?

.gitignore
```diff
@@ -7,8 +7,30 @@ default/
 # Mac stuff
 *.DS_Store
 
-# IDE files
+# QT Creator
+*.user
```
bjorn, line 11: If we're going to ignore `*.user`, then a bunch of ignore rules below are not needed.

.gitignore
```diff
@@ -7,8 +7,30 @@ default/
 # Mac stuff
 *.DS_Store
 
-# IDE files
+# QT Creator
```
bjorn, line 10:
```suggestion
# Qt Creator
```

bjorn (commented):
Well, I'm curious what the CMake project would look like.

Note that Qbs can also generate Visual Studio project files (I haven't tested it) and that there is a Qbs extension for VS Code. So I think Qbs is quite available (though not sure about CLion or Xcode).

bjorn:
Oh, and the commits from #4073 came back. Please rebase instead of merging upstream. :-)

bjorn:
Hmm, you do need to resolve the conflicts before doing `git rebase --continue`. Alternatively, if you rebase using `git rebase -i upstream/master`, then an editor will open where you can drop the changes that were already applied upstream, so you don't need to resolve conflicts in them.

PR #4076, opened 2024-10-10
Fixed issue with placing tile objects after switching maps

### LINKED ISSUE #3497: "Insert Tile" doesn't work anymore reliably after switching between Maps?
**To Reproduce**
1. I have two maps opened, both in same Project, share same Tilesets.
2. Go to tab for Map A and select nothing the Map. But select a Tile in the Widget "Tilesets".
3. Go to tab for Map B and click somewhere in the Map.
4. Go to tab for Map A, the former Object is still selected in the Tileset. Now try to place another Object by selecting Tool "Insert Tile". It won't work, like the function doesn't work at all.

**Specifications:**
 - OS: Win10
 - Tiled Version: 1.9.2

### LINKED ISSUE #3681: Can't place Object Tiles when TAB-ing between opened Map files.
**To Reproduce**
1. Open two TMX Maps with these properties: Both files have different names, but have the identical amount and name of ObjectLayers. They also have both one same TSX file connected.
2. Select on both Maps the same ObjectLayer.
3. Open the same Object Tileset in Tileset View on both Maps.
4. Select Tool "Insert Tile" and select one Tile from TSX and place it in the Map. This works.
5. Now only press "CTRL+TAB" for jumping to the other opened TMX Map file.
6. Here, the same Tile is of course already selected (and also the same ObjectLayer is selected). But the MouseCursor preview already shows that you can't place the Tile anymore. You need to select another Tile first and go back to the one you want.

**Media**
grafik

**Specifications:**
 - OS: Win 10
 - Tiled Version: 1.10.0

bjorn:
The "current tile" was not being restored correctly after switching maps, when another type of object had been focused (for example due to switching layers or placing objects).

Now, when there is no "current tile" referenced by the `MapDocument`, it is set based on the tile selected in the current tileset view.

Closes #3497
Closes #3681

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,7 @@
 * Scripting: Added `MapEditor.currentBrushChanged` signal
 * Scripting: Added `tiled.cursor` to create mouse cursor values
 * Fixed saving/loading of custom properties set on worlds (#4025)
+* Fixed issue with placing tile objects after switching maps (#3497)
 * Fixed crash when accessing a world through a symlink (#4042)
 * Fixed error reporting when exporting on the command-line (by Shuhei Nagasawa, #4015)
 * Fixed minimum value of spinbox in Tile Animation Editor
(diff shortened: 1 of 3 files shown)


UliAbo:
@bjorn Thanks :D

bjorn:
I've identified a minor issue with this patch, which is that it switches the "current object" (shown in Properties view) to the selected tile whenever you switch maps. I should try to find a way to suppress that.

PR #4086, opened 2024-10-18
CI: Updated builds to Qt 6.8

bjorn:


bjorn:
Alright, likely the `install-qt.sh` will need some adjustment for Qt 6.8:

    Could not determine a remote URL for qtbase with version 6.8.0

PR #4100, opened 2024-11-13
Scripting: Expose Tileset.transformationFlags

### LINKED ISSUE #3753: Scripting: Expose "Allowed Transformations" properties
I'd like to write a script that modifies a tileset's "Allowed Transformations", but these properties do not appear to be exposed via the scripting API. Since these properties are shared across the whole tileset, I have to change them frequently based on the Terrains I'm using, and I was hoping to use scripting to automate this.

It would be nice to have these properties exposed, perhaps as `Tileset.allowFlipHorizontally`, `Tileset.allowFlipVertically` and `Tileset.allowRotate`.

If #3016 is implemented, then the Tiles' allowed transformations should be exposed similarly. This would allow users to conditionally set these properties en masse.

dogboydog:
Fix #3753 

Examples (scripting console with a Tileset as the active asset)
```
>tiled.activeAsset.transformationFlags |= Tileset.AllowFlipVertically
2
> tiled.activeAsset.transformationFlags
$8 = 2
> if ((tiled.activeAsset.transformationFlags | Tileset.AllowFlipVertically) > 0) { tiled.log('vertical flip enabled'); } 
vertical flip enabled
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,7 @@
 * Scripting: Added `FileFormat.nameFilter`
 * Scripting: Added `MapEditor.currentBrushChanged` signal
 * Scripting: Added `tiled.cursor` to create mouse cursor values
+* Scripting: Added `Tileset.transformationFlags` (#3753)
 * Fixed saving/loading of custom properties set on worlds (#4025)
 * Fixed issue with placing tile objects after switching maps (#3497)
 * Fixed crash when accessing a world through a symlink (#4042)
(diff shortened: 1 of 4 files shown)


bjorn (commented):
Thanks, this change looks mostly fine, I only see some empty lines missing between functions and there is no `@since` on the new property (and needs addition to `CHANGELOG.md`).

Thanks for implementing this!

dogboydog:
What should be used for `@since` ? I'll make the other updates too.

bjorn:
@dogboydog Ah, for other scripting additions I've used `@since 1.11.1` so I'd suggest you do the same. I may still do a 1.11.1 release after all.

dogboydog:
Done, thanks

PR #4104, opened 2024-11-18
tiled-api: Fixed strict mode error

bjorn:
While trying to publish to [jsr.io](https://jsr.io/), it reported:

    error[syntax-error]: 'eval' and 'arguments' cannot be used as a binding identifier in strict mode


docs/scripting-doc/index.d.ts
```diff
@@ -5022,7 +5022,7 @@ declare class Process {
   /**
    * Executes the program at filePath with the given argument list and blocks until the process is finished. If an error occurs (for example, there is no executable file at filePath) and throwOnError is true (the default), then a JavaScript exception will be thrown. Otherwise, -1 will be returned in case of an error. The normal return code is the exit code of the process.
    */
-  exec(filePath: string, arguments: string[], throwOnError?: boolean): number;
+  exec(filePath: string, args: string[] , throwOnError?: boolean): number;
```
bjorn, line 5025:
```suggestion
  exec(filePath: string, args: string[], throwOnError?: boolean): number;
```

PR #4105, opened 2024-11-21
Scripting : Make MapEditor|TilesetEditor.currentWangSet and MapEditor|TilesetEditor.currentWangColorIndex writeable

### LINKED ISSUE #4101: Scripting : Make MapEditor.currentWangSet and MapEditor.currentWangColorIndex writeable
``` 
bjorn: [...] making MapEditor.currentWangSet and MapEditor.currentWangColorIndex writable. They're currently read-only. I don't see an open issue about this, but I think @eishiya might still have a use for that.
eishiya — 2024-11-14 09:44 AM
I do indeed, for my terrain tools that use right-click sample :D
```

dogboydog:
Fix #4101

currentWangSet setter: seems to be working: 
```js
tiled.mapEditor.currentWangSet = tiled.activeAsset.tilesets[0].wangSets[0] 
```
~~Currently crashing,  I'm doing something wrong setting the currentWangColorIndex~~ Edit: working :

```
tiled.mapEditor.currentWangColorIndex  = 2 
```
I didn't see an existing public method to  change the selection so I tried to make one based on reading the rest of the code in wangdock.cpp . Maybe I missed something existing or just have  a logic error in my setter. Visually at least, it seems to change the color selection though. 

Causes crash: 
image 

TODO (@dogboydog ) 

- [x] Update NEWS 
- [x] Update scripting doc

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ default/
 *.user*
 *.creator.*
 *.autosave
+build/ 
 
 # Visual Studio
 *.sln
--- a/src/tiled/mapeditor.cpp
+++ b/src/tiled/mapeditor.cpp
@@ -1083,7 +1083,7 @@ int MapEditor::currentWangColorIndex() const
 
 void MapEditor::setCurrentWangColorIndex(int newIndex)
 {
-    mWangDock->setCurrentWangColor(newIndex);
+    mWangDock->onColorCaptured(newIndex);
 }
 AbstractTool *MapEditor::selectedTool() const
 {
--- a/src/tiled/mapeditor.cpp
+++ b/src/tiled/mapeditor.cpp
@@ -1067,11 +1067,24 @@ EditableWangSet *MapEditor::currentWangSet() const
     return EditableWangSet::get(mWangDock->currentWangSet());
 }
 
+void MapEditor::setCurrentWangSet(EditableWangSet *wangSet)
+{
+    if (!wangSet) {
+        ScriptManager::instance().throwNullArgError(0);
+        return;
+    }
+    mWangDock->setCurrentWangSet(wangSet->wangSet());
+}
+
 int MapEditor::currentWangColorIndex() const
 {
     return mWangDock->currentWangColor();
 }
 
+void MapEditor::setCurrentWangColorIndex(int newIndex)
+{
+    mWangDock->setCurrentWangColor(newIndex);
+}
 AbstractTool *MapEditor::selectedTool() const
 {
     return mSelectedTool;
--- a/src/tiled/mapeditor.cpp
+++ b/src/tiled/mapeditor.cpp
@@ -281,7 +281,7 @@ MapEditor::MapEditor(QObject *parent)
     connect(mWangDock, &WangDock::wangColorChanged,
             mWangBrush, &WangBrush::setColor);
     connect(mWangBrush, &WangBrush::colorCaptured,
-            mWangDock, &WangDock::onColorCaptured);
+            mWangDock, &WangDock::setCurrentWangColor);
 
     connect(mTileStampsDock, &TileStampsDock::setStamp,
             this, &MapEditor::setStamp);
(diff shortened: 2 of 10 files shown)


bjorn:
The error message is from `QSortFilterProxyModel`, and it says the `QModelIndex` you're using is from the wrong model. That's because `mWangColorModel->colorIndex` returns an index in the source model, which still needs to be mapped to the model used by the view (the `mWangColorFilterModel`).

I don't recall now what this sort-filter-model is used for (I couldn't immediately find a reason for it to exist in the current code). The easiest thing to do for now is to call `mWangColorFilterModel->mapFromSource(index)`, as done in a slightly mode convoluted way in `WangDock::editWangColorName`.

Hmm, in fact, you could likely adjust `WangDock::onColorCaptured`, which is basically doing what you're after with `WangDock::setCurrentWangColor`.

Just note that there's an assert in `WangColorModel::colorIndex`, which should not be possible to trigger from script. Potentially we could just turn the assert into a regular condition and return `QModelIndex()` in that case.

dogboydog:
Thanks, I think I implemented what you suggested. It's working without errors now. 
I updated the scripting docs and NEWS.md

.gitignore
```diff
@@ -11,6 +11,7 @@ default/
 *.user*
 *.creator.*
 *.autosave
+build/
```
dogboydog, line 14: Is it okay to add this? When I run Qt Creator locally it seems to create this directory

src/tiled/wangcolormodel.cpp
```diff
@@ -42,9 +42,7 @@ WangColorModel::WangColorModel(TilesetDocument *tilesetDocument,
 
 QModelIndex WangColorModel::colorIndex(int color) const
 {
-    if (mWangSet)
-        Q_ASSERT(color <= mWangSet->colorCount());
-    else
+    if (!mWangSet || color > mWangSet->colorCount())
```
dogboydog, line 45: Should I put the assert back now that the script error is implemented? Or do we just want to make absolutely sure no one can trigger the assert by scripting

src/tiled/mapeditor.cpp
```diff
@@ -1067,11 +1067,26 @@ EditableWangSet *MapEditor::currentWangSet() const
...
+
 int MapEditor::currentWangColorIndex() const
 {
     return mWangDock->currentWangColor();
 }
 
+void MapEditor::setCurrentWangColorIndex(int newIndex)
+{
+    if(newIndex < 0 || newIndex > mWangDock->currentWangSet()->colorCount())
+        ScriptManager::instance().throwError(QCoreApplication::translate("Script Errors", "An invalid wang color index was provided"));
```
dogboydog, line 1087: Should I update tiled_en.ts with this error? 
How do you keep the line numbers in that file up to date ?

src/tiled/mapeditor.cpp
```diff
@@ -1067,11 +1067,26 @@ EditableWangSet *MapEditor::currentWangSet() const
...
+    mWangDock->setCurrentWangSet(wangSet->wangSet());
+}
+
 int MapEditor::currentWangColorIndex() const
 {
     return mWangDock->currentWangColor();
 }
 
+void MapEditor::setCurrentWangColorIndex(int newIndex)
+{
```
dogboydog, line 1085: I noticed Tiled crashes if you call this when no terrain set is loaded / no map is open.

Tried adding this but it doesn't prevent it. Seems to be a crash trying to get the colorCount(), but currentWangSet() isn't null 

```cpp
    if(!mWangDock->currentWangSet())
        ScriptManager::instance().throwError(QCoreApplication::translate("Script Errors", "No wangset is loaded"));
```

.gitignore
bjorn: I have my Qt Creator set up with a different path, but since I guess this is the default, it's alright to add this.

src/tiled/mapeditor.cpp
bjorn: That check is fine, but you're forgetting to use `return;` to exit the function early. The `throwError` call is going to throw an error in the script engine, it's not a C++ exception.

src/tiled/mapeditor.cpp
bjorn: No, I will update the `tiled_en.ts` 1-2 weeks before the 1.12 release. The line numbers are not actually relevant, they just help translators find the source of a string in case it helps to see them in context.

Maybe we can include this patch in Tiled 1.11.1, but then this string just won't be translatable yet. Usually I avoid touching UI strings entirely in patch releases, but I've made exceptions for rather hidden stuff like these script errors.

src/tiled/wangcolormodel.cpp
bjorn: I think it's fine to leave it like this now.

src/tiled/mapeditor.cpp
dogboydog: Thanks, that fixed it.

Issue #4014, opened 2024-07-18
Scripting: Allow setting starting value and tooltip in Dialog add widget methods


eishiya:
A common pattern when writing scripts that use Dialogs is something like
```js
let widget = dialog.addColorButton("My Color");
widget.color = defaultColor;
widget.toolTip = "My explanatory tooltip";
widget.colorChanged.connect(function(state) {/*do stuff*/});

/*the widget variable is never referenced again after this*/
```
This makes code building dialogs rather verbose because one needs multiple statements to set all the properties one wants, and because one has to keep a variable to reference the widget to set all those properties. A few of the widget-creating methods, such as addCheckBox, already take a starting value, making using them a little cleaner to create.

For all of the add[Widget] methods, it would help if they all had optional parameters for the starting value and the tooltip, so one could do e.g.
```js
let widget = dialog.addColorButton("MyColor", defaultColor, "My explanatory tooltip");
widget.colorChanged.connect(function(state) {/*do stuff*/});
```
Or, if the scripter really wants a one-liner and only has one signal they care about, they could even do
```js
dialog.addColorButton("MyColor", defaultColor, "My explanatory tooltip").colorChanged.connect(function(state) {/*do stuff*/});
```

A starting value is probably more often needed than a tooltip, so it should come before the tooltip in the parameter list. These methods will also need to be careful to not interpret `null` as `false` or some other "valid" value, for those cases where the programmer wants to skip setting a starting value but still wants to use the tooltip parameter.

With these additional parameters, one would be able to add fully usable widgets with less code, and would usually only need to keep references to those widgets around if they need to modify the widget elsewhere, e.g. to change whether it's enabled in response to other widgets changing.

On Discord, bjorn and I discussed accepting signal handlers as parameters too, but because some widgets emit multiple signals, doing so would probably still have to be pretty verbose, requiring either separate parameters per signal necessitating passing null values, or using a map to name the desired signal(s), e.g. `{ currentIndexChanged: function() {/*do stuff*/} }`. The latter is still somewhat verbose and would require the add[Widget] method to be aware of which signals are available, a complication avoided by sticking with `Signal.connect()`. We concluded that it's probably best to leave signals out of this, even where there's only one possible signal.

Oval17:
Hi @bjorn @eishiya would like to work on this.
Thanks

bjorn:
@Oval17 You're welcome to implement this, I'll assign it to you. I think @eishiya already described the plan very well. Thanks!

PR #4107, opened 2024-11-28
Scripting: add addRadioButtonGroup to Dialog

### LINKED ISSUE #3478: Scripting: Dialog radio buttons
When building Dialogs, it's sometimes useful to have radio buttons. Their functionality can be achieved script-side with checkboxes, or they can be replaced with combo boxes (dropdowns), so this isn't an important feature, but it would be neat.

Qt Dialogs allow QRadioButtons and QButtonGroups, but this functionality is not currently exposed to the scripting API.
Relevant Qt docs: https://doc.qt.io/qt-5/qradiobutton.html

The challenge with radio buttons is that they derive their usefulness from belonging to a group with other radio buttons, and only one can be selected at a time. Qt considers all QRadioButtons that have the same parent to be part of the same group, but since the Dialog API doesn't give you direct control over the parent widgets and adds columns and rows automatically, this can get confusing. Perhaps rather than exposing QRadioButtons directly, the Dialog API could support only something like RadioButtonGroup that behaves a lot like a QComboBox (value changed signals, a single tooltip for the whole thing, a constructor that takes all the labels at once, etc), but which consists of a QButtonGroup and a bunch of QRadioButtons under the hood?

dogboydog:
Fix #3478 

TODO @dogboydog 
- [x] Update scripting API doc 
- [x] Update News.md
- [x] Fix radio buttons being invisible 
- [x] Provide example of responding to selected option changes
- [x] Provide a way to select one of the radio buttons from the script? 
- [x] Decide  on tooltip API

 Examples: https://github.com/dogboydog/tiled-dialog-scripts Radio*.mjs

https://github.com/user-attachments/assets/175361e9-34ea-4f2e-ac0b-1296fecc9c64

### DIFF
--- a/docs/scripting-doc/index.d.ts
+++ b/docs/scripting-doc/index.d.ts
@@ -789,6 +789,12 @@ declare namespace Qt {
     readonly checkedButton: QRadioButton | undefined;
 
     /**
+     * ID / index into {@link QButtonGroup.buttons} that is currently selected.
+     * If no radio button is selected, -1 will be returned.
+     */
+    readonly checkedIndex: number;
+
+    /**
      * Add multiple radio buttons to this group.
      * @param values - Each entry in the array is the text that will appear on the radio button.
      * @param toolTips Optional list of tooltips, where each entry is a tooltip that corresponds to the radio button in the list of values with the same index.
@@ -802,6 +808,11 @@ declare namespace Qt {
      */
     addItem(value: String, toolTip: string | undefined) : QRadioButton;
    
+    /**
+     * Signal emitted when any radio button in this QButtonGroup is selected or deselected. 
+     * If the button that causes the signal to be emitted is now selected, checked will be true.
+     */
+    readonly idToggled: Signal<[index: number, checked: boolean]>;
   }
 }
 
--- a/docs/scripting-doc/index.d.ts
+++ b/docs/scripting-doc/index.d.ts
@@ -763,11 +763,11 @@ declare namespace Qt {
 
   /**
    * This type is returned when calling {@link QButtonGroup.addItem} or {@link QButtonGroup.addItems}.
-   * 
+   *
    * You can set checked to true for a QRadioButton even if you have also set enabled to false on it,
-   * which could allow you to prevent the script user from selecting a radio button by clicking on it, 
+   * which could allow you to prevent the script user from selecting a radio button by clicking on it,
    * but still change the selection in your script.
-   * 
+   *
    * @since 1.11.1
    */
   class QRadioButton extends QAbstractButton {}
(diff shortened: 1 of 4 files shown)


dogboydog:
Currently I'm trying a few different things to try to get the radio buttons to show up. They seem to be invisible. Tried putting them in a QHBoxLayout and that didn't seem to help . Trying to do something like the example here 
https://doc.qt.io/qt-6.5/qgroupbox.html#flat-prop

image

dogboydog:
Also planning on looking at #4014 in the future, and thinking I should start providing the tooltip optional argument on this widget. Only thing is in this case (and with the combo box) the tooltip would be for the whole button group, which might be less useful than  per radio button. I guess you could embed a description of all the options in your tooltip though.

eishiya:
Would it be possible to have the tooltip method accept an array of tooltips, one for each radio button?

dogboydog:
Examples 

https://github.com/user-attachments/assets/60ed734c-c190-4ac8-8bac-5c1a90062230


https://github.com/user-attachments/assets/64ead08c-bcce-47dd-b4c0-9f75a7c5317d

src/tiled/scriptdialog.cpp
```diff
@@ -267,6 +291,21 @@ ScriptDialog::NewRowMode ScriptDialog::newRowMode() const
...
+                                                     const QStringList &values,
+                                                     const QString &toolTip,
+                                                     const QStringList &buttonToolTips)
+{
+    QGroupBox *groupParent = new QGroupBox(this);
+    QHBoxLayout *hBox = new QHBoxLayout(groupParent);
+    ScriptButtonGroup *buttonGroup = new ScriptButtonGroup(groupParent, hBox);
+    buttonGroup->addItems(values, buttonToolTips);
+    hBox->addStretch(rightColumnStretch);
+    groupParent->setLayout(hBox);
```
bjorn, line 304: Ah, you can remove this line because `new QHBoxLayout(groupParent)` already assigned the layout to that widget.

docs/scripting-doc/index.d.ts
```diff
@@ -5327,6 +5380,23 @@ declare class Dialog extends Qt.QWidget {
...
+   * buttons, where only one radio button can be selected at a time.
+   *
+   * Each radio button is a {@link QAbstractButton}. You can force selection of
+   * a radio button by setting checked = true, and also disable
+   *
+   * @param labelText - If the labelText is non-empty, a label widget will be added to the left of
+   *  the widget.
+   * @param values - Each entry in the array is the text that will appear on the radio button.
+   * @param labelToolTip - Optional tooltip for the label to the left of the radio button group.
+   * @param buttonToolTips - Optional list of tooltips, where each entry is a tooltip that corresponds to the radio button in the list of values with the same index.
```
bjorn, line 5394: To make the online docs a little less verbose, I've removed the elements related to params. So this will not be shown and needs to be moved into the description without using the `@param` directive.

bjorn (changes requested):

docs/scripting-doc/index.d.ts
dogboydog: I think I took care of both of these, let me know

src/tiled/scriptdialog.cpp
```diff
@@ -267,6 +291,20 @@ ScriptDialog::NewRowMode ScriptDialog::newRowMode() const
...
+ScriptButtonGroup *ScriptDialog::addRadioButtonGroup(const QString &labelText,
+                                                     const QStringList &values,
+                                                     const QString &toolTip,
+                                                     const QStringList &buttonToolTips)
+{
+    QGroupBox *groupParent = new QGroupBox(this);
+    QHBoxLayout *hBox = new QHBoxLayout(groupParent);
+    ScriptButtonGroup *buttonGroup = new ScriptButtonGroup(groupParent, hBox);
+    buttonGroup->addItems(values, buttonToolTips);
+    hBox->addStretch(rightColumnStretch);
```
bjorn, line 303: Hmm, this stretch will sit after the items added through this function, and before the items added using `addItems`/`addItem` on the returned `ScriptButtonGroup`, right?

dogboydog: Yeah, I think this is unnecessary. I was probably  just trying to figure out why the buttons weren't showing up and messing with the layout. I removed it

bjorn (approved):

PR #4110, opened 2024-12-02
CI: Updated builds to Qt 6.8.1

bjorn:


bjorn:
Urgh, it's installing aqtinstall 3.1.18 on Ubuntu 20.04 because it ships with Python 3.8 and 3.1.19 requires Python 3.9. Maybe it's anyway time to update to Ubuntu 22.04. We could try it for the Qt 6 build only for now.

PR #4121, opened 2024-12-13
Update tiled_pt.ts

laetemn:
UPDATE - Translation to the Brazilian Portuguese language.

### DIFF
--- a/translations/tiled_pt.ts
+++ b/translations/tiled_pt.ts
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!DOCTYPE TS>
-<TS version="2.1" language="pt_BR" sourcelanguage="en">
+<TS version="2.1" language="pt_BR" sourcelanguage="en_US">
 <context>
     <name>AboutDialog</name>
     <message>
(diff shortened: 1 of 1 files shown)


translations/tiled_pt.ts
```diff
@@ -309,7 +309,7 @@
     <message>
         <location filename="../src/plugins/csv/csvplugin.cpp" line="+166"/>
         <source>CSV files (*.csv)</source>
-        <translation>Ficheiros CSV (*.csv)</translation>
```
bjorn, line 312: For Tiled 1.11,  @SantosSi had changed all these the other way around in cfd589bbcb1c81278de1a481c2c9006ac66b81d1. We should try to avoid going back and forth on this, so maybe you could agree on a preferred translation?

bjorn: Or maybe it's not clear that https://hosted.weblate.org/projects/tiled/translations/pt_PT/ is the "Portuguese (Portugal)" translation whereas https://hosted.weblate.org/projects/tiled/translations/pt/ is actually "Portuguese (Brazil)" (just, due to leaving out the `_BR`, also serves as fallback for strings without a specific translation for Portugal?

laetemn: > For Tiled 1.11, @SantosSi had changed all these the other way around in [cfd589b](https://github.com/mapeditor/tiled/commit/cfd589bbcb1c81278de1a481c2c9006ac66b81d1). We should try to avoid going back and forth on this, so maybe you could agree on a preferred translation?

"Ficheiro" is not a word used in Brazilian Portuguese (Brazil - pt_BR). It is a word used in the Portuguese language of Portugal (pt_PT). In Brazilian Portuguese, the word "Arquivo" ("File" in English) is used.

bjorn:
Thanks for the update! I'll take care of merging this with the changes done on Weblate.

PR #4122, opened 2024-12-13
CI: Use jurplel/install-qt-action instead of aqtinstall directly

bjorn:
The `jurplel/install-qt-action` takes care of setting environment variables and caching.

### DIFF
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -50,13 +50,13 @@ jobs:
           qt_version: 5.15.2
           qt_version_major: 5
           qt_arch: gcc_64
-          qt_install_args: ""
+          qt_modules: ""
           qbs_default_profile: gcc
         - ubuntu_version: 22.04
           qt_version: 6.8.1
           qt_version_major: 6
           qt_arch: linux_gcc_64
-          qt_install_args: "--modules qtimageformats"
+          qt_modules: "qtimageformats"
           qbs_default_profile: x86_64-linux-gnu-gcc-11
 
     env:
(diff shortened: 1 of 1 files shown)


bjorn:
Oh well, failures on all platforms:

## Linux
```
Cannot mix incompatible Qt library (5.12.8) with this library (5.15.2)
```
That's likely the system Qbs using the system Qt (5.12.8) somehow picking up a library or plugin from the installed Qt (5.15.2). Solution could be to install `tools_qtcreator` and use the Qbs shipped with that instead.

## macOS
```
src/plugins/python/pythonplugin.h:27:10: fatal error: 'Python.h' file not found
#include <Python.h>
         ^~~~~~~~~~
```
That's odd, because earlier message reports "WARNING: PYTHONHOME not set. The Python plugin will not be available.", but it seems Python is getting picked up by `pkg-config`, but not entirely. Could likely be resolved by specifying `setup-python: false` for the `jurplel/install-qt-action`.

## Windows
```
INFO    : aqtinstall(aqt) v3.1.20 on Python 3.11.9 [CPython MSC v.1938 64 bit (AMD64)]
WARNING : Failed to download checksum for the file 'Updates.xml'. This may happen on unofficial mirrors.
ERROR   : Failed to locate XML data for Qt version '6.8.1'.
```
That doesn't appear to be an issue related to this action, since it was also observed at https://github.com/mapeditor/tiled/actions/runs/12312710252/job/34365308531. It seems something changed at the Qt download server, or it might be an issue that happens only for some mirrors.

bjorn:
Now Sentry compile started to fail:
```
CMake Error at /opt/hostedtoolcache/cmake/3.19.8/x64/cmake-3.19.8-Linux-x86_64/share/cmake-3.19/Modules/FindCURL.cmake:165 (message):
  CURL: Required feature AsynchDNS is not found
```
So we still need to install `libcurl4-openssl-dev`.

bjorn:
Linux build now fails with:
```
Deploying dependencies for ELF file AppDir/usr/lib/tiled/plugins/libpython.so 
ERROR: Could not find dependency: libpython3.11.so.1.0 
ERROR: Failed to deploy dependencies for existing files
```
Not sure why this started failing, maybe Tiled is linking to a different Python version now. Might be another case where we need to add `setup-python: false`.

bjorn:
Reported about the failing Qt 6.8.1 install on Windows, worked around by using aqtinstall 3.1.19 for now.

Windows build is now failing due to Qbs picking the wrong toolchain. Need to force it to use that one for the Qt 5 build:
```
Profile 'i686-w64-mingw32-gcc-8_1_0' created for 'D:/a/tiled/Qt/Tools/mingw810_32/bin/i686-w64-mingw32-gcc-8.1.0.exe'.
```
(and the appropriate other one for the Qt 6 build)

bjorn:
Hmm, https://github.com/mapeditor/tiled/actions/runs/12333087797 failed so there's something still not right about this... Maybe we can't use `setup-python: false` and just need to explicitly disable the Python plugin for now.

A bit strange that the build succeeded before, which might have to do with using a cache with `setup-python: false`, where the cache had actually been created with that option enabled. It seems that option is not part of the cache key.

PR #4123, opened 2024-12-20
add qaseprite install step to packages.yml

### LINKED ISSUE #4109: feature request: use aseprite file in tilesheet
If we want to use aseprite to make a tilesheet, then we must export it to a format supported by Tiled (each time again)

https://www.aseprite.org/

It would be great if Tiled supports the .aseprite format itself so that we can load it directly in Tiled.
The source code is here 

https://github.com/aseprite/aseprite/tree/main/src

dogboydog:
Fix #4109  

Build and bundle [qaseprite](https://github.com/mapeditor/qaseprite) plugin with Tiled

- [x] Linux (AppImage)
- [x] Linux (Snap) 
- [x] MacOS
- [x] Windows


.github/workflows/packages.yml
```diff
@@ -112,6 +122,13 @@ jobs:
         sudo cmake --install build --prefix /usr --config RelWithDebInfo
         popd
 
+    - name: Install qaseprite plugin
+      working-directory: qaseprite
+      run: |
+        cmake -B build -DCMAKE_BUILD_TYPE=Release
+        cmake --build build --config Release --target qaseprite
```
bjorn, line 129:
```suggestion
        cmake --build build --config Release
```

.github/workflows/packages.yml
```diff
@@ -112,6 +122,13 @@ jobs:
         sudo cmake --install build --prefix /usr --config RelWithDebInfo
         popd
 
+    - name: Install qaseprite plugin
+      working-directory: qaseprite
+      run: |
+        cmake -B build -DCMAKE_BUILD_TYPE=Release
```
bjorn, line 128: The plugin should be build against the system libraries, as done in https://github.com/mapeditor/qaseprite/blob/c388282e746fa41e15b38750b502f7a8e4361cad/.github/workflows/build.yml#L51-L55 (also needs those dependencies installed).

bjorn (commented):
Two small comments. Overall, I think we can wait a bit with merging this until we have tried adding it for other platforms as well.

bjorn:
The macOS build failure is a recent breakage in the workflow, which I'm trying to address in #4124 now.

dogboydog:
image

bjorn:
Hmm, for both versions I get this on Sequoia 15.1:

image

Not sure yet what the issue is, I don't get any additional output when running from a terminal. My first guess would be some missing library, but I couldn't find any. I also tried removing the `Contents/PlugIns/imageformats/libqaseprite.dylib` file, but Tiled still wouldn't launch.

dogboydog:
I have this problem all the time on mac (including with the Tiled builds I tested). I have to do this: 

`xattr -r -d com.apple.quarantine Tiled.app`

bjorn:
> I have this problem all the time on mac (including with the Tiled builds I tested).

Ah, of course, since these builds are from the pull request, they're not signed and notarized. Boy, what a strangely misleading error message... Indeed, after doing that the builds work and the plugin loads!

dogboydog:
I agree the error message is not helpful at all, very aggressive how it tells you to send it to the trash lol... I get the same when I try to test my games on a mac

bjorn:
Hmm, while this PR adds qaseprite to the GitHub Actions workflow, for the Windows releases we actually need it to compile as part of `appveyor.yml`.

bjorn:
Alright, AppVeyor works! Unfortunately the plugin is still not functional in the snap. We can see it sets the environment variable:
```
:: ++ qmake -query QT_INSTALL_PLUGINS
:: + export QT_PLUGIN_PATH=/usr/lib/x86_64-linux-gnu/qt5/plugins
:: + QT_PLUGIN_PATH=/usr/lib/x86_64-linux-gnu/qt5/plugins
:: + craftctl default
:: + cmake /root/parts/qaseprite/src -G 'Unix Makefiles'
```
But later on when installing, it goes to the wrong location:
```
: Install the project...
:: -- Install configuration: "RelWithDebInfo"
:: -- Installing: /root/parts/qaseprite/install/usr/lib/plugins/imageformats/libqaseprite.so
```
This is the same location as where it used to go before, so it appears the `QT_PLUGIN_PATH` environment variable might not be getting through to the CMake configure step.

PR #4136, opened 2025-01-07
AppVeyor: Updated builds to Qt 6.8

bjorn:
Qt 6.7 is no longer available on AppVeyor.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,7 +1,7 @@
 clone_depth: 200
 version: '1.11.{build}'
 image:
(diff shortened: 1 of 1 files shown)


bjorn:
Hmm, it seems there might actually be no Qt available at `C:\Qt\6.8\msvc2019_64`:
```
Detecting Qt installations...
WARNING: Could not find any qmake executables in PATH. Either make sure a qmake executable is present in PATH or set the moduleProviders.Qt.qmakeFilePaths property to point a qmake executable.
```

PR #4140, opened 2025-01-12
Added composition mode to layer

### LINKED ISSUE #3932: Layer Blend Mode Property
Images intended to be rendered with blend modes other than alpha blend don't display accurately in Tiled.

It would be great if layers had a blend mode property, choices would include additive, multiplied, subtractive, and default alpha blend.

Currently, I just a custom property to set the blend mode in my engine, but obviously this doesn't fix it rendering inaccurately in Tiled.

Demo:
gui
side_by_side

bjorn:
TODO:

- [x] Serialization in XML, JSON and Lua formats
- [x] Decide which modes to support
- [x] Consider what should happen when set on group layers
- [x] Scripting API
- [x] Documentation

Closes #3932

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 ### Unreleased
 
+* Added support for SVG 1.2 / CSS blending modes to layers (#3932)
 * YY plugin: Fixed compatibility with GameMaker 2024 (#4132)
 * snap: Fixed crash on startup on Wayland
 * Raised minimum supported Qt version from 5.12 to 5.15
--- a/docs/manual/layers.rst
+++ b/docs/manual/layers.rst
@@ -235,7 +235,7 @@ calculation details.
 Mode         SVG equivalent
 ===========  ===========================================================================
 Normal       `src-over <https://www.w3.org/TR/SVGCompositing/#comp-op-src-over>`__
-Plus         `plus <https://www.w3.org/TR/SVGCompositing/#comp-op-plus>`__
+Add          `plus <https://www.w3.org/TR/SVGCompositing/#comp-op-plus>`__
 Multiply     `multiply <https://www.w3.org/TR/SVGCompositing/#comp-op-multiply>`__
 Screen       `screen <https://www.w3.org/TR/SVGCompositing/#comp-op-screen>`__
 Overlay      `overlay <https://www.w3.org/TR/SVGCompositing/#comp-op-overlay>`__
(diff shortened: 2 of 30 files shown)


eishiya:
Draft of docs, to add to https://github.com/mapeditor/tiled/blob/master/docs/manual/layers.rst after the Tinting Layers section.
```
.. _blend-mode:

Blend Modes
--------------

Tiled supports several common blend modes for Layers, which modify their
appearance by blending them with the layers below in different ways.
The default blend mode is Inherit, which uses the blending mode from
the layer's parent Group. The blend mode inherited by top-level Layers
and Groups is Normal.

Below is the full list of blend modes available in Tiled, along with links
to their equivalents in the SVG Compositing Specification, where you can see
examples and calculation details.

+-------------+--------------------------------------------------------------------------+
| Mode        | SVG equivalent                                                           |
+=============+==========================================================================+
| Inherit     | --                                                                       |
| Normal      | [src-over](https://www.w3.org/TR/SVGCompositing/#comp-op-src-over)       |
| Plus        | [plus](https://www.w3.org/TR/SVGCompositing/#comp-op-plus)               |
| Multiply    | [multiply](https://www.w3.org/TR/SVGCompositing/#comp-op-multiply)       |
| Screen      | [screen](https://www.w3.org/TR/SVGCompositing/#comp-op-screen)           |
| Overlay     | [overlay](https://www.w3.org/TR/SVGCompositing/#comp-op-overlay)         |
| Darken      | [darken](https://www.w3.org/TR/SVGCompositing/#comp-op-darken)           |
| Lighten     | [lighten](https://www.w3.org/TR/SVGCompositing/#comp-op-lighten)         |
| Color Dodge | [color-dodge](https://www.w3.org/TR/SVGCompositing/#comp-op-color-dodge) |
| Color Burn  | [color-burn](https://www.w3.org/TR/SVGCompositing/#comp-op-color-burn)   |
| Hard Light  | [hard-light](https://www.w3.org/TR/SVGCompositing/#comp-op-hard-light)   |
| Soft Light  | [soft-light](https://www.w3.org/TR/SVGCompositing/#comp-op-soft-light)   |
| Difference  | [difference](https://www.w3.org/TR/SVGCompositing/#comp-op-difference)   |
| Exclusion   | [exclusion](https://www.w3.org/TR/SVGCompositing/#comp-op-exclusion)     |
+-------------+--------------------------------------------------------------------------+
.. raw:: html

   <div class="new new-prev">Since Tiled 1.12</div>
```

It would be nice to include a figure that demonstrates blending modes, perhaps a demo with 3 layers where all are Normal, compared to where the bottom layer is Normal and the other two are Screen and Multiply or something. I don't have any art handy for that, and stock art designed for blending modes is uncommon.

I was a little tempted to include a paragraph with some usage examples, but figured the docs aren't the place. But just in case you do want it, here:
```
In most cases, you'll want to use the default Inherit or Normal mode,
which is also the default supported by practically all game engines.
Multiply is often used for lightmaps, while Plus or Screen are common for
light cookies. Note that not all engines support all of these blending modes.
```

PR #4142, opened 2025-01-13
YY plugin: Adjusted to changes in GameMaker file format

### LINKED ISSUE #4132: GameMaker Studio 2 Export Broken With Recent GM Update
Reposting here for maximum visibility

A recent GM update seems to have rendered maps exported from Tiled incompatible. Attempting to load the project in GM will result in an error reading

>     The JSON file reader encountered parsing errors
>     C:\Git\Ancient Keys DX\rooms\rm_area_a\rm_area_a.yy(2,3): Error: A type tag field is required at the start of the JSON record, eg: “$MyTypeName”: …

Adding this field manually will turn the error into

>     C:\Git\Ancient Keys DX\rooms\rm_area_a\rm_area_a.yy(3,3): Error: Field “%Name”: expected.

Adding that field turns it into

>     C:\Git\Ancient Keys DX\rooms\rm_area_a\rm_area_a.yy(4,3): Error: Field “creationCodeFile”: expected.

And so on and so forth.

I’m sure this is known and is being worked on, but I wanted to make absolutely sure. Thank you for your time.

bjorn:
It seems an update to GameMaker in late 2024 has broken compatibility in several ways:

* It now requires a "type tag field" at the start of applicable JSON records.

* The name is expected to follow the "type tag field" as a "%Name" field, in addition to the regular "name" field.

* Except for these special properties, the JSON fields are now required to be ordered alphabetically.

Other minor differences include:

* The "tags" field is left out when there are no tags.
* There is no longer a space included after JSON field names.
* Some new fields were added.

This patch does not address all alphabetical ordering issues. Propably the preferred way will be to collect the fields in a QJsonObject first and rely on automatic sorting of the fields while serializing.

Closes #4132

### DIFF
--- a/src/plugins/yy/jsonwriter.cpp
+++ b/src/plugins/yy/jsonwriter.cpp
@@ -255,7 +255,7 @@ void JsonWriter::writeKey(const char *key)
     prepareNewLine();
     write('"');
     write(key);
-    write(m_minimize ? "\":" : "\": ");
+    write("\":");
 }
 
 void JsonWriter::write(const char *bytes, qint64 length)
--- a/src/plugins/yy/yyplugin.cpp
+++ b/src/plugins/yy/yyplugin.cpp
@@ -28,13 +28,13 @@
 #include "mapobject.h"
 #include "maprenderer.h"
 #include "objectgroup.h"
-#include "savefile.h"
 #include "tile.h"
 #include "tilelayer.h"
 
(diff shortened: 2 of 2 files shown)


bjorn:
Whoops, I forgot I had already opened a pull request for this. Anyway, this one is now superseded by #4143.

PR #4145, opened 2025-01-15
snap: Try to remove Wayland support

bjorn:
Such that the Tiled snap will hopefully start working again on Wayland based on XWayland.

Currently we're seeing this when trying to launch Tiled on Wayland:
```
$ snap run tiled
xkbcommon: ERROR: failed to add default include path /usr/share/X11/xkb
qt.qpa.wayland: failed to create xkb context
QSocketNotifier: Can only be used with threads started with QThread
Segmentation fault (core dumped)
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 ### Unreleased
 
+* snap: Fixed crash on startup on Wayland
 * Raised minimum supported Qt version from 5.12 to 5.15
 
 ### Tiled 1.11.1 (11 Jan 2025)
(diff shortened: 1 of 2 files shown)


bjorn:
Removing Wayland support like this yields:
```
$ snap run tiled
qt.qpa.plugin: Could not find the Qt platform plugin "wayland-egl" in ""
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: eglfs, linuxfb, minimal, minimalegl, offscreen, vnc, xcb.

Aborted (core dumped)
```

PR #4157, opened 2025-01-26
Docs: "Manual" draw order -> "Index Order"

eishiya:
At some point, the Manual object draw order was renamed to Index Order, but the docs were not updated.

This PR updates the Working with Objects and Keyboard Shortcuts manual pages. I'm not sure why the Keyboard Shortcuts page even mentions the draw order, as the shortcuts work to reorder Objects regardless of draw order, the order just doesn't impact rendering in Top Down mode.

I **did not** update the mention of the manual draw order in the scripting docs, as there it is unclear whether "manual" refers to the name of the mode or its function. Perhaps that should be updated too though.

### DIFF
--- a/docs/manual/keyboard-shortcuts.rst
+++ b/docs/manual/keyboard-shortcuts.rst
@@ -104,13 +104,13 @@ When an object layer is selected
 
 -  ``S`` - Activate :ref:`select-objects-tool`
 
-   -  ``PgUp`` - Raise selected objects (with Manual object drawing
+   -  ``PgUp`` - Raise selected objects (with Index Order object drawing
       order)
-   -  ``PgDown`` - Lower selected objects (with Manual object drawing
+   -  ``PgDown`` - Lower selected objects (with Index Order object drawing
       order)
-   -  ``Home`` - Move selected objects to Top (with Manual object
+   -  ``Home`` - Move selected objects to Top (with Index Order object
       drawing order)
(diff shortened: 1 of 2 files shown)


bjorn:
Hmm, this rename happened by accident while implementing the new Properties view. I think we should rather just rename it back to "Manual" before the 1.12 release.

> I'm not sure why the Keyboard Shortcuts page even mentions the draw order, as the shortcuts work to reorder Objects regardless of draw order, the order just doesn't impact rendering in Top Down mode.

Indeed they can change the order, but if Manual object drawing order wasn't mentioned there, then I'm sure people would be confused about not seeing any visible difference. Maybe it just needs a slight rewording.

eishiya:
Ah. I did think about checking whether 1.11 had this wording but got lazy. I think it's better as "Manual". I'll leave this PR open until you change it back in Tiled, feel free to close it when you're ready.

As for the keyboard shortcuts, perhaps remove the parentheticals, but add a note above or below the list that the Object order is visible in the Objects view and that it affects Object draw(ing) order in Manual mode.

The Changing Stacking Order section of the Objects page has a similar issue with implying those shortcuts only work in Manual mode. That whole page also doesn't mention the Objects view, which seems an odd omission. While it's not necessary to work with Objects, it is useful in some scenarios and worth knowing about. Even the neglected Tile Stamps view gets more love from the docs!

PR #4158, opened 2025-01-26
Export plugin for Remixed Dungeon

Mikhael-Danilov:
Adds export to Remixed Dungeon (https://play.google.com/store/apps/details?id=com.nyrds.pixeldungeon.ml, https://github.com/NYRDS/remixed-dungeon) map format.


Mikhael-Danilov:
@bjorn please take a look at this.

src/plugins/rpd/plugin.json
```diff
@@ -0,0 +1 @@
+{ "defaultEnable": true }
```
bjorn, line 1: As far as I understand, this plugin saves files with .json extension? This may conflict with the existing JSON export, so we can't enable this plugin by default.

src/plugins/rpd/qjsonparser/json.cpp
bjorn, line 1: It's a bit unfortunate that we need yet another copy of this JSON writing code. The only real reason this thing still exists is because arrays of numbers are stored incredibly inefficiently by Qt due to putting each value on its own (indented) line. And we're writing huge arrays of numbers.

We can leave this as-is now, but eventually I'd like to move `plugins/yy/jsonwriter.cpp` to `libtiled`, so that it can be used by all plugins. Since this writer can produce a bit nicer output (though currently its formatting is somewhat specific to GameMaker).

src/plugins/rpd/rpdplugin.cpp
bjorn, line 1: Source files are missing a copyright header, probably should look like:
```c++
/*
 * Remixed Dungeon Tiled Plugin
 * Copyright 2025, Mikhael Danilov <...@gmail.com>
 *
 * This file is part of Tiled.
 *
 * This program is free software; you can redistribute it and/or modify it
 * under the terms of the GNU General Public License as published by the Free
 * Software Foundation; either version 2 of the License, or (at your option)
 * any later version.
 *
 * This program is distributed in the hope that it will be useful, but WITHOUT
 * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
 * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
 * more details.
 *
 * You should have received a copy of the GNU General Public License along with
 * this program. If not, see <http://www.gnu.org/licenses/>.
 */
```

src/plugins/rpd/rpdplugin.cpp
```diff
@@ -0,0 +1,319 @@
...
+}
+
+RpdMapFormat::RpdMapFormat(SubFormat subFormat, QObject *parent)
+    : Tiled::WritableMapFormat(parent)
+    , mSubFormat(subFormat)
+{}
+
+QString RpdMapFormat::shortName() const
+{
+    return "RPD";
```
bjorn, line 33: All other formats use lower-case short names, so I'd suggest to do the same here.
```suggestion
    return "rpd";
```

src/plugins/rpd/rpdplugin.cpp
```diff
@@ -0,0 +1,319 @@
...
+            QJsonArray multiexit;
+
+            mapJson.insert("width", layer->map()->width());
+            mapJson.insert("height", layer->map()->height());
+
+            mapJson.insert("map", packMapData(layer->asTileLayer()));
+
+            for (int i = 0; i < layer->map()->width(); ++i) {
+                for (int j = 0; j < layer->map()->height(); ++j) {
+                    int tileId = layer->asTileLayer()->cellAt(i, j).tileId();
```
bjorn, line 110: Plugin will crash if exporting a map with a layer named "logic" that isn't a tile layer. Layer type needs to be checked along with the name above (same problem with other layers below).

src/plugins/rpd/rpdplugin.cpp
```diff
@@ -0,0 +1,319 @@
...
+            mapJson.insert("decoTileVar", packMapData(layer->asTileLayer()));
+            mapJson.insert("customTiles", true);
+
+            if (!insertTilesetFile(*layer, "tiles", mapJson))
+                return false;
+
+            QJsonArray decoDesc;
+            QJsonArray decoName;
+
+            auto tilesets = layer->asTileLayer()->usedTilesets();
```
bjorn, line 160: Since this function is virtual we don't need to get a `TileLayer*` first.
```suggestion
            auto tilesets = layer->usedTilesets();
```

src/plugins/rpd/rpdplugin.cpp
```diff
@@ -0,0 +1,319 @@
...
+            }
+
+            mapJson.insert("decoName", decoName);
+            mapJson.insert("decoDesc", decoDesc);
+        }
+
+        if (layer->name() == "objects") {
+            QMap<QString, QJsonArray> objects;
+
+            for (auto object : layer->asObjectGroup()->objects()) {
```
bjorn, line 180: Better verify `layer->isObjectGroup()` in the check above.

src/plugins/rpd/rpdplugin.h
```diff
@@ -0,0 +1,86 @@
...
+
+
+class RPDSHARED_EXPORT RpdMapFormat : public Tiled::WritableMapFormat
+{
+    Q_OBJECT
+    Q_INTERFACES(Tiled::MapFormat)
+
+public:
+    enum SubFormat {
+        Rpd
```
bjorn, line 37: `mSubFormat` is actually unused so this can be removed.

src/plugins/rpd/rpdplugin.h
```diff
@@ -0,0 +1,86 @@
...
+protected:
+    QString mError;
+    SubFormat mSubFormat;
+
+private:
+    bool insertTilesetFile(Tiled::Layer &layer, const QString &tiles_name, QJsonObject &mapJson);
+};
+
+
+class RPDSHARED_EXPORT RpdTilesetFormat : public Tiled::TilesetFormat
```
bjorn, line 63: You can derive from `Tiled::WritableTilesetFormat` instead so you don't have to implement those dummy `read` and `supportsFile` overrides.

src/plugins/rpd/rpdplugin.h
```diff
@@ -0,0 +1,86 @@
+#ifndef RPDPLUGIN_H
+#define RPDPLUGIN_H
```
bjorn, line 2: I'm no longer using include guards in this project, please just write `#pragma once`.

bjorn (changes requested):
> @bjorn please take a look at this.

Sure, I'm sorry I didn't have a look sooner!

Overall looks like a fine addition. Is the idea to let the community add maps to Remixed Dungeon?

It seems the plugin is lacking some error checking. It should at least not crash when used with unexpected data, but you might also want to add some helpful error messages (check out `yyplugin.cpp` for example, which uses `Tiled::ERROR` and `Tiled::WARNING` to add messages to the "Issues" view).

Don't forget to also edit `dist/win/installer.wxs` to add the `rpd.dll` to the list of plugins included in the installer.

src/plugins/rpd/rpdplugin.cpp
```diff
@@ -0,0 +1,319 @@
...
+{
+    Tiled::SaveFile file(fileName);
+
+    if (!file.open(QIODevice::WriteOnly | QIODevice::Text)) {
+        mError = tr("Could not open file for writing.");
+        return false;
+    }
+
+    Tiled::MapToVariantConverter converter;
+    QVariant variant = converter.toVariant(tileset, QFileInfo(fileName).dir());
```
bjorn, line 276: This looks like just the same thing as `JsonTilesetFormat`. Why do you need it again here?

Mikhael-Danilov:
Thank you for the thorough review and valuable feedback!

Yes, the primary goal of this plugin is to enable the community to create and integrate custom maps for their mods in *Remixed Dungeon*. The game supports a modding system that allows users to create and install mods, which can include assets like maps, sprites, configuration files, and some lua scripts.

I appreciate your suggestions regarding error handling and user feedback.

I’ll work on these changes and provide an update once they’re implemented.

src/plugins/rpd/plugin.json
Mikhael-Danilov: fixed

src/plugins/rpd/rpdplugin.cpp
Mikhael-Danilov: done

src/plugins/rpd/rpdplugin.h
Mikhael-Danilov: done

src/plugins/rpd/rpdplugin.h
Mikhael-Danilov: done

src/plugins/rpd/rpdplugin.h
Mikhael-Danilov: done

src/plugins/rpd/rpdplugin.cpp
Mikhael-Danilov: Well, I just copied and pasted it from JSON plugin template a long time ago (in another fork) and not touched it since. It not need and removed now.

Mikhael-Danilov:
Hi again, @bjorn!

I’ve implemented the suggested changes. Please let me know if there’s anything else that needs to be addressed.

bjorn (approved):
I think it looks fine now, thanks!

PR #4166, opened 2025-02-03
feat: remove image rect restrictions in tilesets

deathbeam:
Previously, image rect could only be set on collection tileset tiles. This change allows setting image rect on any tile type.

This allows making "multi tile" tiles without changing any of the existing data layout or needing to painfully create new collections.

Godot 4 has similar feature in merging tiles.

This is mostly useful for y sorting and multi tile properties.

**Still unfinished, just PRing the first draft for opinions**

The baseline currently works, just need to sort out some rendering quirks and make sure I did not broke anything. And maybe improve the usage by a bit.


eishiya:
This sounds like a partial implementation of the Tile Atlas concept discussed in #2863? Image rects were added as a first step towards this. You may want to see the discussion there about why this should probably be a separate tileset type rather than just extending image rects to "Based on Tileset Image" tilesets.

deathbeam:
Looking through it, yea. Even though I think you are overcomplicating it a bit there with the off-grid tiles like the 16x19 example there as that one also fits in 16x32 so the grid display can always just remain in the regular baseline grid set on the tileset, so dont think new view is even needed.

New tileset type, I guess the benefit would be to just inherently save all the image recs instead of saving just the custom ones? which means image recs would be able to also get removed (which would be ideal for merging tiles, which is the feature im trying to replicate here from godot and I think godot mostly works like that anyway, and I think you are also talking about ending on something like that eventually).

deathbeam:
image

ok switched back to grid view for simplicity and grid spans, now I can delete tiles without breaking grid and selection also works perfectly. Just need logic for adding the tiles back after they are deleted i guess

deathbeam:
I think as far as functionality goes i think i have most of stuff working properly now

Issues:

- atm im kinda bastardizing rearrange tiles mode for splitting existing multi tile tiles and readding removed tiles back. I think mode like this is ideal tho, but maybe it should be called differently when in tile atlas mode. And maybe it should remove tiles on right click, add on left click, not sure
- merge multiple tiles into one is only done manually atm, ideally this would be hold and drag in rearrange tiles mode I think
- after the above is done I think disabling manual input to the image rect would be best
- no way to regenerate all the image recs from scratch, currently they are generated when tileset is opened without any image rects (maybe new button, or maybe not even needed)

deathbeam:
Ok got merging and everything else done, for now in **Rearrange Tiles** mode:

- Tiles can be merged by holding down left click (this creates multi tile tile)
- Tiles can be split by right clicking on multi tile edges
- Tiles can be removed from atlas by right clicking single tiles (so needs to be split first then removed) - current limitation, ideally right click hold would delete multiple, but this is probably not super common use case and the visual seection doesnt work for right click, QT limitation maybe?
- Tiles can be added back to atlas by left clicking empty tiles

deathbeam:
Ok I think this is now properly ready for review, I was doing some testing since yesterday and I think i fixed everything I found and im happy with the implementation. I did not adjusted documentation or stuff like python bindings etc, I think that can be done as finishing touches after this is reviewed and if the implementation even makes sense (I tried my best but I was not familiar with Tiled codebase at all until I started working on this PR).

bjorn:
First off, I really appreciate the effort you're putting into this, @deathbeam! It worked quite well in my tests and it's nice to be able to quick merge and split the tiles. Things I noticed:

* There's an odd issue when toggling the "Rearrange Tiles" option in one tileset and then switching to another tileset, which then shows the toggled button but isn't in the rearrange mode. However, I noticed this is an existing issue which should be fixed in a separate PR.

* There's an issue with adjusting the draw margins which might be related to the maximum tile size not getting updated for atlas tilesetes? I'm not sure why `Tileset::updateTileSize` does an early return in this case. This information is used in `Map::recomputeDrawMargins` to make sure enough of the view is updated. I think there is also an existing issue with this mechanism, but not sure if it's the same.

* If I click and drag a small amount on a tile and then click another tile, it decides to merge the tiles. I think intended behavior is to only merge when I do a single drag covering multiple tiles.

* The dynamic wrapping mode doesn't work for atlas tilesets (which couldn't be combined with the "rearrange tiles" mode, at the very least). It does work for image collection tilesets and I feel it should work for atlas tilesets while not rearranging them.

* I feel like splitting/merging of tiles should be a separate mode from "Rearrange Tiles" mode. There doesn't appear to be a reason the original Rearrange Tiles mode couldn't be supported for atlas tilesets (though probably only in "dynamic wrapping" mode).

Unfortunately, apart from these small issues this approach is not what we (or I) had planned for the atlas tileset. It was meant to cover not just grid-aligned sub-rects but allow for tiles of any size (to resolve #1008). I realize technically you could do this in the current solution with a base tile size of 1x1, but that doesn't seem ideal.

Another issue is the automatic assignment of tile IDs based on the location. This way, it is not possible to move the tile rects in response to a rearranged tileset image while keeping the tile IDs the same. In addition, using a "12+12 bit spatial hash" creates huge gaps in the IDs and ends up placing a bunch of arbitrary limits (not just the 4096x4096 tile coordinates limit but since we have only 28 bits (+4 flags) available for a global tile ID, also limiting the amount of such tilesets in a map to 16). These limits are probably reasonable, but I think if such ID management is desired it's better added as a separate option.

I'm a little undecided how to move forward with this. Even if we fix the above issues and move the ID assignment to a separate option, the tile grid limitation remains and forms the basis of the entire implementation. Yet, the file format does not have this limitation, so the tileset view could be replaced later to enable per-pixel tile rects to be drawn or dragged around. Alternatively we keep this PR open for now until we've found time to support that as well. Let me know your thoughts.

deathbeam:
Ok tried 1x1 grid and not even my PC could handle that, QT not good enough to handle so many entries in grid i guess :d For now enabled editing the rectangle at least and fixed the selection bug.

deathbeam:
So fixed 1, 2, 3, 4 and 7.
Partially fixed 6 by allowing the rectangle to be editable in the properties window and making sure the spans at least try to support it properly (works somewhat well).
For 7. Decided to just remove the new tile ID format compltely and use the old old. Even if I think that the new tile ID format is better for atlas tilesets, maybe that is something that can be done in separate PR to not add extra complexity to this PR when its not absolutely necessary

dogboydog:
I didn't get the chance to try this yet, but... 

> Unfortunately, apart from these small issues this approach is not what we (or I) had planned for the atlas tileset. It was meant to cover not just grid-aligned sub-rects but allow for tiles of any size (to resolve https://github.com/mapeditor/tiled/issues/1008). I realize technically you could do this in the current solution with a base tile size of 1x1, but that doesn't seem ideal.

This is basically my only use case for having an atlas tileset - arbitrary tile size, to allow using a sprite atlas as a tileset where the  individual tiles have been packed to take up the least amount of space rather than a grid. So it'd be too bad if this feature wouldn't allow that setup. The current workaround I use is having an image collection tileset plus the sub rect feature.  Others mention sprite atlases in the linked issue as well, so they'd be similarly  prevented from using the feature

deathbeam:
image

im also trying slightly different approach where the tileset is just 1 cell on the "grid" and it just draws whole tileset and tiles are just selection on the tileset and nothing else. This makes any sized rects easy, and grid somewhat too because i can just auto snap to grid. Downside is that spacing cannot really work with this I think very easily, margin still might tho.

deathbeam:
Ok pushed changes with different way of doing this, which supports moving tiles, drawing any size tiles and adding/deleting in the rearrange tiles mode.

Currently that is all that is supported (animation and map editor view doesnt work with it yet, selection simply doesnt propagate due to how this is implemented and I did not bothered with trying to support dynamic wrap yet either), just want to get some feedback on the UX. But it should support everyones use cases I believe.

deathbeam:
Ok I believe I resolved all issues except dynamic view now:

image

Im not sure if supporting dynamic view for this new approach is even worth it, maybe? But it would also be a lot simpler to just disable it for tileset view :d 

Would appreciate if someone other than me also checked the new UX and if everything works fine outside of dynamic view

deathbeam:
Dynamic wrapping should now work as well

deathbeam:
Added support for corner resizing as well, with left click drag on existing tile near corners

dogboydog:
I tried this out via the Linux snap . Some of the behavior might have changed from the last time OP tested,  when the branch was updated to include the new properties framework, but this is the current state from the last build. 

Some thoughts 
image
image

It looks like you can leave the Source image field blank when creating a tileset, and then I don't see a way after the tileset has been created to add a source. So I think it should probably just be required when creating the tileset. If you do select an image while creating, it looks like it's already supported to then change which image is used for the tileset. 

I'm a little confused by the inclusion of "generate tiles based on grid"  and the tile width and height properties in the context of an atlas tileset, but maybe there is a use case where you would start with a grid and then resize some of the tiles that don't fit into a clean grid.  Either way, if you select this option when creating a tileset,  right now the tiles that are generated cannot be deleted or resized. It seems to just act like a regular grid tileset. 

When "generate tiles based on a grid" is unchecked, the resulting tileset no longer renders an image , so  you seemingly can't make an atlas tileset successfully anymore in this current version 
image

deathbeam:
> I tried this out via the Linux snap . Some of the behavior might have changed from the last time OP tested, when the branch was updated to include the new properties framework, but this is the current state from the last build.

You need to use Rearrange Tiles mode to edit the tiles 
image

Its described in both commit description and in the tileset editor at bottom. Only the tiles that have rects over them are drawn, that is consistent with non-tileset atlas behaviour. In rearrange tiles mode they are always drawn. And yes the generate grid purpose is exactly what you said (consistent with godot does this for example, and use case is where for example you have multi tile buildings but you need single tileset object for for example y sorting, same applies to trees etc and similar stuff). And the generated tiles also are editable in rearrange tiles mode.

deathbeam:
The source not being required atm is def an issue i will look at that

EDIT: Fixed

dogboydog:
Got it, I see that message at the bottom now and I see how to edit tiles.  It's true you made that warning pretty prominent, but it just feels very different than managing other tilesets like "collection of images" tileset.  It's definitely more intuitive than the last time I tried this PR.   I understand the rearrange tile mode is the basis of this PR but it seems to force  this workflow into a kind of weird UX in my opinion, where the add and remove tiles button don't interact with this feature at all, and you can't select a tile to see its properties without exiting the rearrange tile mode.   It definitely achieves making an atlas tileset though, and it's pretty cool to see what you've accomplished. 

I still wonder why we should have a grid size in the tileset image properties that constrains the tiles you can make to only be increments of the grid size. To me, the behavior when you set the grid size to 1x1 pixels is the way I would expect this feature to work, and I wonder why  you would want to lock your tiles to not only be multiples of 8 or 16 pixels for example, but force the edges of each tile onto those multiples as well .

deathbeam:
Yea i agree its a bit different workflow, but this feature needs ability to do the add/edit/delete operations quickly otherwise its kinda painful to use especially if you have a lot of tiles that you need to manually describe like this. It works for collections of images because you are just adding whole tiles 1 by 1, but even that workflow is painfully slow if you want to replicate something like tileset atlas even though it can technically be used as alternative, but the UX there doesnt really work for many operations at once. I can try and spend some time to make the add/remove buttons work but its more pain than its worth doing when its not better UX than going to rearrange tiles mode. 

Also for 1x1 mode, the grid is just default snapping behaviour, when you hold shift it unsnaps, so you can arrange the tiles in whatever fashion you want. But maybe doing the opposite would be better, depends on what is the more common use case for this (e.g do not snap by default and snap when holding shift only). As I said, the main purpose i made this for was multi tile tiles while also supporting arbitrary sizes at same time, and both non-grid tilesets and grid tilesets with multi tile tiles are very valid use cases so this needed to support both, and i needed the first one more often than second one so the default behaviour is simply configured to that use case.

dogboydog:
Thanks for explaining. It's true this approach allows you to more quickly edit multiple tiles at once. Curious if others would agree that not snapping to the grid is more expected as a default behavior

Kiamo2:
When I look at it from Godot's perspective, only the grid-based (snapping) atlas corresponds reasonably well to the Godot atlas.
So I would disagree (from that perspective).

dogboydog:
> When I look at it from Godot's perspective, only the grid-based (snapping) atlas corresponds reasonably well to the Godot atlas. So I would disagree (from that perspective).

Which Godot feature are you referring to? AtlasTexture doesn't snap by default for me, it started out with Snap Mode: None which allows any dimensions without snapping

deathbeam:
> > When I look at it from Godot's perspective, only the grid-based (snapping) atlas corresponds reasonably well to the Godot atlas. So I would disagree (from that perspective).
> 
> Which Godot feature are you referring to? AtlasTexture doesn't snap by default for me, it started out with Snap Mode: None which allows any dimensions without snapping

tileset editor does what i tried to achieve here:

image

It basically does what I do here by default, pre-generates the tiles at first based on grid and then you can create multi tiles based on the grid as well.

dogboydog:
Gotcha, thanks for clarifying . I'm not sure which workflow is more common than the other - I personally hadn't used this multi tile functionality at all and approached this more from a generated sprite atlas perspective (for example, an atlas generated by a program like Aseprite) where there is no consistent size or spacing for any sprite. But I see the use case for what you're showing

Kiamo2:
Yeah a Godot atlas is a different animal compared to an Aseprite atlas.
A multi tile in Godot is defined by it's atlas coordinates (row, column) and width and height in (base) tile counts.

dogboydog:
Both types do exist in Godot though.  Even with YATI I use a "collection of images" tileset as a makeshift sprite atlas from aseprite, and you can also manually make an AtlasTexture which works with the arbitrary size by default.  For repeating tiles used in a tile layer I can see the snapping use case, but for sprites/objects from an atlas the non snapping workflow is useful. I can see either one being the default

Kiamo2:
In a collection of images every image in Godot gets its own tile source (the image) which can have a sub rect.
So that indeed has an equivalent in Godot.
As long as it's only one sub rect.
But assume one image and several sub rects in that same image...??

Kiamo2:
I'm currently testing.
After getting used to these quirks
- must delete tile at starting point first before I can expand a rect
- can only get hold of a tile object in the upper left corner

it seems to work reasonably well.

But in my test map there are some strange artifacts, which I don't know where they come from.
And they are permanent i.e. they survive saving and loading.
In fact they are tiles which are contained in the map but shouldn't.
I've marked them with red arrows.

artifacts

Kiamo2:
Artifacts are caused by those underlying tiles:
asset1

After deleting those everything works fine:
asset2

bjorn:
I've added a rebased version of #4322 to this branch because it was bothering me during testing. I intend to give this change a good look soon.

_Edit: rebased this change on that fix now._

bjorn:
First of all I'd like to apologize for letting this PR sit around for so long. I do have a bunch of other projects taking my time and this PR isn't the easiest one to go through because I'd have approached it rather differently, but I should have made time for this earlier. Anyway, here we are.

I'll have a good look at the code next week, for now I've just been trying out the feature and left these notes.

* There are what appears to be corner handles rendered on hover, but when resizing the tiles I can in fact only grab the edges and resize in one direction. I think either we need to make the corner handles allow resizing in both directions or we should change this hover highlight to draw boxes along the edges instead.

* Regardless of whether grid mode is on by default, there should be a persistent toggle for this because for those that don't want it, holding Shift all the time is not an option. Of course there already is View > Snapping menu but I guess we shouldn't share this setting with map view snapping.

* Just left-clicking a tile should not trigger a "Change Image Rect" undo command.

* Resizing a tile over other tiles is very confusing. Probably I would expect the other tiles to be automatically deleted, but it may not always be what the user wants. At the very least there's a mismatch between the rendering order and the interaction priority, which makes it so that one ends up hovering/dragging a tile that is rendered below other tiles.

* I'm really bothered by the tileset zooming into to the top-left, but of course that is existing behavior. It just becomes way more prominently annoying if you try to zoom in/out while defining the tile rectangles. I believe I've already looked into this at some point though, and it's probably not something that's easy to change about a `QTableView`.

* It seems silly to have a "Edit Tileset" dialog to change the image and edit grid settings for this type of tileset, but it's something we can improve in a follow-up change.

* Instead of "Atlas: On" in Properties view I think we could add "Type: ComboBox" (matching the values in New Tileset Dailog).

* We seriously need to find a new name and icon for the "Rearrange Tiles" mode, being open to splitting off atlas editing mode from the actual rearrange mode.

* We have one atlas texture in the examples (`examples/forest/squirrel.png`) and it just begs for a automatic detection of tile boundaries, either to find all tiles in the tileset or to set determine the tile's exact rectangle by just clicking somewhere on a tile.

deathbeam:
Thanks for testing :) 

>     * There are what appears to be corner handles rendered on hover, but when resizing the tiles I can in fact only grab the edges and resize in one direction. I think either we need to make the corner handles allow resizing in both directions or we should change this hover highlight to draw boxes along the edges instead.

Fixed this in latest commit, added proper support for 8 direction resizing

>     * Regardless of whether grid mode is on by default, there should be a persistent toggle for this because for those that don't want it, holding Shift all the time is not an option. Of course there already is View > Snapping menu but I guess we shouldn't share this setting with map view snapping.

Yea I was thinking about adding it as option somewhere too but was unsure where as it could be a bit confusing exactly because it can be easily confused with actual grid snapping. For now changed the shift held requirement to simple shift toggle that is for now displayed in the status bar only

>     * Just left-clicking a tile should not trigger a "Change Image Rect" undo command.

Fixed in latest commit

>     * Resizing a tile over other tiles is very confusing. Probably I would expect the other tiles to be automatically deleted, but it may not always be what the user wants. At the very least there's a mismatch between the rendering order and the interaction priority, which makes it so that one ends up hovering/dragging a tile that is rendered below other tiles.

Yea there is definitely use case for overlaps I think, originally I tried to delete overlapping tiles but I think someone pointed out that people might want that (for example tree next to house on tileset overlapping on corner tile as we are limited to rectangles only instead of more complex shapes). For the rendering/interaction mismatch, not sure what would be best approach to fix that but I can look at it next week probably.

EDIT: Fixed in latest commit, just reversed the selection order.

>     * It seems silly to have a "Edit Tileset" dialog to change the image and edit grid settings for this type of tileset, but it's something we can improve in a follow-up change.

Yea I tried to go for minimal codebase changes so I tried to avoid touching stuff I absolutely did not had to for now

>     * Instead of "Atlas: On" in Properties view I think we could add "Type: ComboBox" (matching the values in New Tileset Dailog).

Would this mean also storing the tileset type in the tileset data instead of just storing the atlas boolean? And then moving the enum somewhere else I guess as I think atm the type enum is mostly tied to the dialog box only.

>     * We seriously need to find a new name and icon for the "Rearrange Tiles" mode, being open to splitting off atlas editing mode from the actual rearrange mode.

Edit tiles maybe? Not sure what would be good icon tho

>     * We have one atlas texture in the examples (`examples/forest/squirrel.png`) and it just begs for a automatic detection of tile boundaries, either to find all tiles in the tileset or to set determine the tile's exact rectangle by just clicking somewhere on a tile.

Hmm so automatic edge detection based on transparency/background? i guess that could be kinda interesting, but it also would most likely break in the tree/house corner tile overlapping scenario, but for most normal tilesets it can work. I am not very experienced with qt but I can give it a shot.

deathbeam:
image

so for auto edge detection, i got something working with ctrl-click, but for example the chairs at bottom that are directly connected with tiles below, there the edge detection stops working and basically hits rest of the image as im looking for clearly separated space with alpha, so not ideal. And not sure what would be approach in scenario like this (and this scenario is fairly common on tilesets i think)

Issue #3313, opened 2022-03-26
Automap While Drawing doesn't apply all relevant rules


eishiya:
When Automapping While Drawing, only the rules that look at the layer(s) being edited get applied. If I edit Layer A and a rule updates Layer B in response to a change in Layer A, I expect later rules that only look at Layer B to also apply, but they do not.

It's possible to get around this bug by adding an empty `input_Layer A` layer in the rule maps, but I don't think this should be necessary.

I've encountered this issue in both the stable builds and in the new automapping branch.

bjorn:
> When Automapping While Drawing, only the rules that look at the layer(s) being edited get applied. If I edit Layer A and a rule updates Layer B in response to a change in Layer A, I expect later rules that only look at Layer B to also apply, but they do not.

Hmm, yes I would expect the same. I guess it would be alright though, if this only happened for rules from subsequent rule maps, rather than rules from the same rule map? It would have to work like that in any case in the context of concurrent rule matching.

bjorn:
This issue happens because the filtering of the rule maps happens before any AutoMapping has been done:

https://github.com/mapeditor/tiled/blob/d23131cd156b9006e3d7d2a6eb1020c0768dabb0/src/tiled/automappingmanager.cpp#L122-L138

eishiya:
> Hmm, yes I would expect the same. I guess it would be alright though, if this only happened for rules from subsequent rule maps, rather than rules from the same rule map? It would have to work like that in any case in the context of concurrent rule matching.

I think that would be fine, yes. It would be more intuitive if `MatchInOrder` allowed rules to affect later rules within the same file this way, but it would be fine if it doesn't, as long as you make a note about that in the documentation, perhaps as a warning in the bit on `MatchInOrder`.

bjorn:
I looked into fixing this for Tiled 1.8.3, but the changes had too much overlap with the `automapping` branch, so it will be fixed in Tiled 1.9.

> I guess it would be alright though, if this only happened for rules from subsequent rule maps, rather than rules from the same rule map?

While implementing this, I realized that this distinction is not relevant here, because the skipping happens on a per-rule-map basis and not per-rule. If a rule map does not have a given modified layer as input, it can never happen, that some of its rules would end up being relevant indirectly.

[...]

PR #4168, opened 2025-02-03
AutoMapping: Removed the "touched layers" optimization

bjorn:
The "AutoMap While Drawing" had an optimization where it skipped entire rule maps when none of their input layers had been "touched". This was based on the assumption that its rules would not need to be re-evaluated anyway.

Unfortunately, we can't make this assumption in general. The effect of certain rules might be relevant even if none of their input or output layers have been touched.

### DIFF
--- a/src/tiled/automapper.cpp
+++ b/src/tiled/automapper.cpp
@@ -155,14 +155,9 @@ AutoMapper::AutoMapper(std::unique_ptr<Map> rulesMap)
 
 AutoMapper::~AutoMapper() = default;
 
-QString AutoMapper::rulesMapFileName() const
+const QString &AutoMapper::rulesMapFileName() const
 {
     return mRulesMap->fileName;
-}
-
-bool AutoMapper::ruleLayerNameUsed(const QString &ruleLayerName) const
-{
-    return mRuleMapSetup.mInputLayerNames.contains(ruleLayerName);
 }
 
 template<typename Type>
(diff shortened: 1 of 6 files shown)


eishiya:
Had a go with this using my enormous set of Automapping rules. None of the rules were affected negatively by the optimization. Removing the optimization has made Automap While Drawing even with a 1x1 brush noticeably more sluggish. It's still acceptable, but less pleasant than before, when it was practically instant.

I wonder if perhaps the optimization could be brought back, but with a rule map property like "NeverSkip" (previously discussed in https://github.com/mapeditor/tiled/issues/3313) that'll force that rule map to be used even if its layers weren't touched.

bjorn:
> I wonder if perhaps the optimization could be brought back, but with a rule map property like "NeverSkip" (previously discussed in #3313) that'll force that rule map to be used even if its layers weren't touched

With this generic name I'm a bit worried it might be understood as a way to make an exception for file name filters. Although, such an exception isn't needed at that level because it can be made by temporarily resetting the filter. In the end I guess that's a documentation issue (and another reason to enable tooltips on custom properties).

eishiya:
It could be named something else, especially since Automapping properties are automatically suggested and don't need to be typed and can thus be fairly descriptive, the only drawback to long names is that they're only helpful for English-speakers. Tooltips would be good even for basic names in combination with translations.

"NeverSkip" being generic is a potential problem in case other, orthogonal optimisations are added in the future. Might even be cool to have a "SkipOptimizations" property that's a bunch of flags for different skippable optimizations.
Perhaps if we call this property "SkipOptimizations" (or "IgnoreOptimizations") for now, it could be converted to flags later if needed.

PR #4182, opened 2025-03-21
Qbs: Use exportingProduct.sourceDirectory

bjorn:
This fixes the following warning when using Qbs 2.6: "Resolving path properties relative to the exporting product's location is deprecated."

Raises the minimum version of Qbs to 1.18 (dropping support for compiling on Ubuntu 20.04 out of the box, which shipped with Qbs 1.13).

### DIFF
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -263,7 +263,9 @@ jobs:
       run: |
         brew install qbs
         qbs setup-toolchains --detect
-        qbs config defaultProfile xcode
+        qbs setup-qt --detect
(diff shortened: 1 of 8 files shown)


bjorn:
Some issue...
```
ERROR: /Users/runner/work/tiled/tiled/src/qtsingleapplication/qtsingleapplication.qbs:3:1 Error while handling product 'qtsingleapplication':
/opt/homebrew/Cellar/qbs/2.6.0/share/qbs/imports/qbs/base/QtModule.qbs:49:14 cannot read property 'version' of undefined
```

PR #4184, opened 2025-03-27
docs: Fixed documentation for JSON Map Format - Chunk

Charlie-83:
The example was wrong since width 16 and height 16 implies 256 elements in data rather than 16

It also had trailing commas which made it invalid json

I also updated the description of width and height to be the same as for Layer.


bjorn:
Looks good, thanks!

PR #4187, opened 2025-04-02
GitHub Actions: Pin cmake to 3.31

bjorn:
We can't use CMake 4 for now because of qaseprite not compiling due to some of its dependencies setting a CMake version below 3.5.


bjorn:
That fixed the Windows build of qaseprite, so let's have it. The macOS build is broken for another reason (#4182).

PR #4190, opened 2025-04-10
add world properties dialog to World menu

### LINKED ISSUE #4188: Provide a properties editor for worlds
Currently World properties can only be accessed via scripting.

A menu such as  World > World Properties > <loaded world> would open up a properties editor similar to  Map Properties or Tileset Properties. 

One consideration would be how can we get this to work via the action search /  ctrl + shift + p. 

The view should also auto update when the world properties are changed via scripting. 

Perhaps we could also display in a read only fashion any patterns and maps added to the world. 

The current workaround is to edit the custom properties via scripting.

dogboydog:
Fix #4188 
image

Works with Ctrl + Shift + P / action search 
### TODO 
 
  - [x] Prefer the world for the currently active map, if any ( I think I can do this, I just didn't prioritize it yet) 
  - [x] Fix  custom properties actually working . Right now it's not working, possibly because I didn't differentiate the  temporary world being edited the way the project properties does with the "local" project concept?  
  - [x] Update NEWS.md

### DIFF
--- a/src/tiled/mainwindow.cpp
+++ b/src/tiled/mainwindow.cpp
@@ -1971,11 +1971,7 @@ void MainWindow::editWorldProperties()
     if (!world)
         world = WorldManager::instance().worlds().first();
 
-    if (WorldPropertiesDialog(world, this).exec() == QDialog::Accepted) {
-        QString saveError;
-        if (!world->save(world->fileName(), &saveError))
-            QMessageBox::critical(this, tr("Error Saving World"), saveError);
-    }
+    WorldPropertiesDialog(world, this).exec();
 }
 void MainWindow::autoMappingError(bool automatic)
 {
--- a/src/tiled/mainwindow.cpp
+++ b/src/tiled/mainwindow.cpp
@@ -680,8 +680,10 @@ MainWindow::MainWindow(QWidget *parent, Qt::WindowFlags flags)
 
     connect(mUi->actionTilesetProperties, &QAction::triggered,
             this, &MainWindow::editTilesetProperties);
+
     connect(mUi->actionWorldProperties, &QAction::triggered,
             this, &MainWindow::editWorldProperties);
+
     mUi->actionWorldProperties->setEnabled(false);
     connect(&WorldManager::instance(), &WorldManager::worldsChanged, this, [this] {
         mUi->actionWorldProperties->setDisabled(WorldManager::instance().worlds().empty());
--- a/src/tiled/mainwindow.cpp
+++ b/src/tiled/mainwindow.cpp
@@ -74,6 +74,7 @@
 #include "world.h"
 #include "worlddocument.h"
 #include "worldmanager.h"
+#include "worldpropertiesdialog.h"
 #include "zoomable.h"
 
 #include <QActionGroup>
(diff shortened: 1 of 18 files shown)


src/tiled/mainwindow.ui
```diff
@@ -50,7 +50,7 @@
       <string>&amp;Recent Files</string>
      </property>
      <property name="icon">
-      <iconset resource="tiled.qrc">
+      <iconset resource="../../default_Imported_2fa4f2ef8a59764f_Desktop__69e7352954baa862/libtilededitor.877b2b3a/tilededitor.qrc">
```
dogboydog, line 53: I didn't mean to change this...Qt Creator must have automatically done this when I opened mainwindow.ui ?

bjorn: Yeah, this is quite annoying. Qt Creator (or, the embedded Qt Designer) does not appear to handle nicely the fact that the resource files are generated by Qbs. I've just been excluding these changes from my commits...

Maybe I should work it around by removing resource file references from `.ui` files entirely and assigning resources, most commonly icons, from C++ code instead. Though that's going to be quite a few lines in case of the `MainWindow`. A patch for another time, in any case. :-)

dogboydog:
Hey so I was thinking for this PR - maybe showing the list of maps added to the world  should be a follow up , so that we can provide a way for custom properties to be edited for worlds in the UI?  

I have one non-custom property added right now which is the filename for the world,  and editing it kind of works except for these issues: 
1) The file browser that is opened by the QUrl property widget  is expecting an existing file, so the only way to pick a new one  is to manually type in a new path
2) The old  filename for the world isn't renamed 
3) I don't think the undo for the filename change is working 

Before I try to fix this more, I'm wondering if we even want the filename editable here, or if it should just be a readonly label

bjorn:
> Hey so I was thinking for this PR - maybe showing the list of maps added to the world should be a follow up , so that we can provide a way for custom properties to be edited for worlds in the UI without waiting to implement more features?

Yes, that's reasonable. The list of maps can already be edited through the World Tool anyway.

> I have one non-custom property added right now which is the filename for the world, and editing it kind of works except for these issues:

This makes me wonder, why did you add the file name? Is that because we don't have a "Save As" action for worlds? Cause no other asset works this way. You can't just change an asset's file name, you'd have to save it with a different file name instead. We also don't have any "rename" functionality (though it would be nice to have).

I think the file name might be interesting to include as read-only if it is otherwise hard to identify which world the properties are about.

> Edit: Oh, also I tried to set the dialog title to include the World's display name, but it doesn't seem to be working. I thought it would work as long as I did it after the setupUi call

That appears to be because your wrong usage of the `tr` function. The second parameter is the "disambiguation" parameter. You seem to have confused it with `QCoreApplication::translate`, which does take a context first. But for `tr`, the context is set in the `Q_OBJECT` macro that defines the static `tr` function for that object.

dogboydog:
It just displays the custom properties in the dialog body now and has the world filename in the title

src/tiled/mainwindow.cpp
```diff
@@ -678,7 +680,12 @@ MainWindow::MainWindow(QWidget *parent, Qt::WindowFlags flags)
 
     connect(mUi->actionTilesetProperties, &QAction::triggered,
             this, &MainWindow::editTilesetProperties);
-
+    connect(mUi->actionWorldProperties, &QAction::triggered,
+            this, &MainWindow::editWorldProperties);
+    mUi->actionWorldProperties->setDisabled(WorldManager::instance().worlds().empty());
```
bjorn: I'd suggest just doing this, because in the `MainWindow` constructor no worlds should be loaded yet.
```suggestion
    mUi->actionWorldProperties->setEnabled(false);
```

src/tiled/mainwindow.cpp
```diff
@@ -1953,7 +1960,23 @@ void MainWindow::editTilesetProperties()
...
+    QSharedPointer<WorldDocument> world;
+    Document *currentDocument = DocumentManager::instance()->currentDocument();
+    if (currentDocument && currentDocument->type() == Document::MapDocumentType)
+        world = WorldManager::instance().worldForMap(currentDocument->fileName());
+    if (!world)
+        world = WorldManager::instance().worlds().first();
+
+    if (WorldPropertiesDialog(world, this).exec() == QDialog::Accepted) {
+        QString saveError;
+        if (!world->save(world->fileName(), &saveError)) // my hero ...
```
bjorn: What does "// my hero ..." mean?

dogboydog: Sorry just a joke.  "Save the world". I'll remove it 😛

dogboydog:
Feel free to squash

PR #4192, opened 2025-04-13
convert issue templates to issue forms

NetSysFire:
As discussed.

You can preview them by clicking on the file > "View file".

I left them mostly intact but made some smaller changes.

1. Added a config.yml so blank issues remain an option, also added the Forum there as an alternative venue for questions. So when people try to open an issue, they get presented with basically "There is also a forum".
2. Removed the "Media" section for bug reports. You can already drag and drop files into the fitting section. I think that one makes very little sense in general as screenshots feel more at home where they are of relevance instead of separated somewhere. But re-adding it is just another optional `textarea`.
3. As the issue forms support placeholder text and descriptions, I made use of those and slightly adjusted descriptions so they work with that.


bjorn:
Thanks @NetSysFire, I think that's a nice improvement!

PR #4211, opened 2025-06-12
RenameProperty shortcut (F2 as base shortcut)

djad04:
i Added the Rename shortcut for properties , its set to F2 than i registered it so it can be displayed in the keyboard menu and changed  by users .

### DIFF
--- a/src/tiled/propertieswidget.cpp
+++ b/src/tiled/propertieswidget.cpp
@@ -2232,7 +2232,7 @@ PropertiesWidget::PropertiesWidget(QWidget *parent)
     mActionRemoveProperty->setPriority(QAction::LowPriority);
     connect(mActionRemoveProperty, &QAction::triggered,
             this, &PropertiesWidget::removeProperties);
-    this->addAction(mActionRemoveProperty);
+    addAction(mActionRemoveProperty);
 
     mActionRenameProperty = new QAction(this);
     mActionRenameProperty->setEnabled(false);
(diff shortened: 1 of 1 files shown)


src/tiled/propertieswidget.cpp
```diff
@@ -2236,8 +2236,10 @@ PropertiesWidget::PropertiesWidget(QWidget *parent)
     mActionRenameProperty->setEnabled(false);
     mActionRenameProperty->setIcon(mRenameIcon);
     mActionRenameProperty->setPriority(QAction::LowPriority);
+    mActionRenameProperty->setShortcut(Qt::Key_F2);
```
bjorn, line 2241: While this seems to nicely do the trick, using a shortcut on the action unfortunately hooks it up to F2 regardless of the currently focused widget. This means you can select a property, then focus something else (like the Layers view), then press F2 and expect to rename a layer, but you end up renaming the custom property.

I noticed this is even already the case for the existing shortcut to delete a property:
```c++
mActionRemoveProperty->setShortcuts(QKeySequence::Delete)
```
I think we might be able to work this around by calling `setShortcutContext(Qt::WidgetWithChildrenShortcut)` on the action. But I just tested this and I couldn't get it to work, the shortcut just stopped working entirely.

Alternatively we can handle this shortcut in `PropertiesWidget::keyPressEvent`, with the main downside that it's not easy to make this shortcut configurable (but I don't think that's critical for such a standard shortcut).

djad04: i missed that! but setShortcutContext(Qt::WidgetWithChildrenShortcut) should actually work , did you add the action to the propertieswidget ?  when you put the ShortcutContext to Qt::WidgetWithChildrenShortcut  it will look for that specific action in a specified widget and its children so you must add the action to that widget  
`this->addAction(mActionRenameProperty); `

djad04: Confirmed✅ both shortcuts now trigger correctly as expected.

bjorn: > did you add the action to the propertieswidget

Ah, I didn't, I thought the widget being the parent of the action would be enough, but I guess the `QObject` parent doesn't automatically translate to a widget relationship. Thanks for catching that!

bjorn:
@djad04 I finally remembered to come back to this change. It works great now, so thank you for this nice improvement!

PR #4213, opened 2025-06-26
Add tmx++ to libraries and frameworks

mechakotik:


bjorn:
Looks like a nice new library you're writing there, thanks for adding the link to the manual!

I do think this section (probably the entire page) needs a good cleanup at some point. So many linked libraries are now outdated or no longer maintained, that it's hard for users to find the good options. It's an effort I once started at https://www.mapeditor.org/libraries, which is based on https://github.com/mapeditor/tiled/blob/gh-pages/_data/libs.yml.

PR #4225, opened 2025-08-15
Fixed crash on macOS when JS code is JIT-compiled

### LINKED ISSUE #4218: Bizarre crash in JavaScript engine on macOS Monterey
### Describe the bug

I had recently written a [custom extension](https://github.com/utdrwiki/maps/tree/master/extensions) and tried to get two macOS users to run it. Both users are on macOS Monterey and for both of them Tiled unexpectedly crashed when they performed a specific action in my extension.

The crash narrows down to: a JavaScript function takes another function as an argument, runs it at least 4 times, and then Tiled crashes. The reported exception type is `EXC_BAD_ACCESS (SIGKILL (Code Signature Invalid))`. This could have something to do with QtQml's JavaScript engine, or Tiled's packaging of it.

I'm attaching two crash logs, first is from performing my reproduction steps below on macOS 12.7.6, second is when loading my above extension and performing the `File > Publish to wiki` action on macOS 12.7.4 (the specific part that crashes is the `Array.prototype.map` call in includes/auth.mjs:generateCodeChallenge).

[crash1.txt](https://github.com/user-attachments/files/21520567/crash1.txt)
[crash2.txt](https://github.com/user-attachments/files/21520566/crash2.txt)

This cannot be reproduced on Windows or Linux.

### Steps to reproduce

1. Open Console
2. Run `((c)=>{c();c();c();c()})(()=>{})`
3. Tiled crashes

### Expected behavior

Expected behavior of the above function is to not do anything.

### OS version

macOS 12.7.4 and 12.7.6

### Tiled version

1.11.2

bjorn:
The codesigned release did not have the right entitlement to be able to execute JIT-compiled code.

Also, the MAP_JIT memory flag is only set by Qt since 6.8.2, so when Tiled is compiled against earlier Qt versions, we now force the use of the JS interpreter to disable the JIT compiler.

TODO: This change relies on the Qbs codesign module (available since Qbs 1.19), which will likely need CI updates.

Closes #4218

### DIFF
--- a/src/tiledapp/app.entitlements
+++ b/src/tiledapp/app.entitlements
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+  <dict>
+    <key>com.apple.security.cs.allow-jit</key>
+    <true/>
+  </dict>
+</plist>
--- a/src/tiledapp/main.cpp
+++ b/src/tiledapp/main.cpp
@@ -422,6 +422,13 @@ int main(int argc, char *argv[])
     }
 #endif
 
+    // Disable JIT compilation in the Qt Quick V4 engine on macOS, since it
+    // triggers a crash in code-signed releases. This is resolved by adding the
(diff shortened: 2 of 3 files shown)


bjorn:
Closing this PR since we can't test whether the problem is fixed in the signed version because the PR builds are not being signed because this PR was opened from my development fork.

PR #4228, opened 2025-08-19
Qbs: Only link with QtSvg on macOS

bjorn:
This linking was introduced in ce441e18339ec197d010949487bf1d4d5625edaf to make sure `macdeployqt` copies over the SVG icon engine and image format plugins.

Removed the dependency on other systems since it's not needed there and it was causing issues for somebody trying to compile Tiled on OpenSUSE.


bjorn:
I've verified that the AppImage still ships with SVG support after this change, so merging this.

PR #4234, opened 2025-08-23
Allow JS expressions in number inputs

dogboydog:
Requested in Discord chat.

Replaces `QSpinBox` and `QDoubleSpinBox` throughout the app and in script dialogs with a subclass that allows  setting a value from JavaScript expressions.

Needs more testing, but a lot of things work in the current version.


src/tiled/expressionspinbox.cpp
```diff
@@ -0,0 +1,126 @@
...
+    QJSValue result = this->evaluate(text);
+    if (result.isNumber())
+        return result.toNumber();
+
+    return originalValue;
+}
+
+QValidator::State ExpressionSpinBox::validate(QString &text, int &pos) const
+{
+    return QValidator::Acceptable;
```
dogboydog, line 60: I tried various approaches in the validate()  methods, including returning Intermediate or Invalid if there was an error evaluating the text as an expression, but it seems like the only way this PR works is if I always return Acceptable

src/tiled/propertyeditorwidgets.cpp
```diff
@@ -348,7 +348,7 @@ void SpinBox::wheelEvent(QWheelEvent *event)
 
 
 DoubleSpinBox::DoubleSpinBox(QWidget *parent)
-    : QDoubleSpinBox(parent)
+    : ExpressionDoubleSpinBox(parent)
```
dogboydog, line 351: It doesn't seem like the expressions always work for this subclass ... for instance trying to set the Parallax origin on a Tilemap  is not allowing any characters like * +  to be typed.  I tried seeing if removing the textFromValue override  made it work , but it doesn't seem so

dogboydog:
Demo 

https://github.com/user-attachments/assets/7604d5c6-37b4-481d-956d-5e9fc84e087c

Another issue is that as you can see in the video, some of the spinboxes seem to evaluate the expression too aggressively while I'm still typing, so I can only do simple expressions like *2  or /2, because as soon as I've typed in an expression that can be evaluated, the value is immediately replaced. I'm not sure why this only happens for some of the widgets though.  I was looking around to see if there's a way I could  wait until the user has stopped editing for a certain amount of time before evaluating, but nothing in the Qt docs jumped out immediately

src/tiled/propertyeditorwidgets.cpp
bjorn: Hmm, it works for me, but it's replacing the text too fast. So when there is "0.0" and I type "+1", it immediately changes to "1.0" rather than staying at "1.0+1" and allowing me to type "1.0+10" for example.

This is probably due to a feedback loop, where change of text value causes `valueFromText` to be called, which creates undo command with new value, which in turns sets the value of the spin box again. We should be able to avoid that by suppressing the setting of the spinbox value in this case.

eishiya:
If possible, I think evaluation should only happen on commit (e.g. lose focus, Enter pressed, etc) and not while typing at all. If I type in (current number)/16 and then alt+tab to my documentation to check that that's the correct divisor, I don't want to lose what I'd typed. I want it to evaluate when I tell it I'm ready for it to evaluate.

src/tiled/propertyeditorwidgets.cpp
dogboydog: Interesting. I have that issue with some of the inputs but not others, and the parallax origin one seems too fast to use at all. Maybe I need to figure out how to only evaluate the expression when the input is no longer focused. Eishiya commented that maybe on pressing enter would be good, but in the video I posted below,  hitting enter while focused on a spinbox  pressed a button on the dialog instead, so i'm not sure if there is a good way to make it so that it works when hitting enter. Focus loss might be a good compromise? Thanks for taking a look

bjorn:
> If possible, I think evaluation should only happen on commit (e.g. lose focus, Enter pressed, etc) and not while typing at all. If I type in (current number)/16 and then alt+tab to my documentation to check that that's the correct divisor, I don't want to lose what I'd typed.

This is essentially the behavior it has (except in Properties view, where it probably gets messed up while handling the value change), but I'm afraid Alt+Tabbing to another application also counts as focus lost.

dogboydog:
Just tested out. It looks great after all your changes!  All the bugs I noticed in my initial version are gone 😎 

The only potential gotcha I see is when working with inputs that have a suffix. If you type your numbers after the suffix, the suffix is added back to the end of the string, and the expression won't parse anymore since it will have for example 'px'  in the middle of it. I think this is probably people will quickly realize they have to type after the number instead of after the suffix though. 

image

eishiya:
The units staying in the text box have been a long-standing annoyance for me in Tiled, I often end up accidentally trying to type after or mid-unit. I wonder if it's possible to have the unit text disappear when the input is focused for editing, or otherwise appear sort of like a placeholder, not actually part of the value? Tiled already skips you before/after the unit when you click on it so presumably it's something special and not regular text in the value.

dogboydog:
Hmm, I looked a bit at the docs here, and nothing immediately jumps out as a built in feature that would accomplish what you're asking for.  The suffix  and prefix are stored separately than the value for sure, but I don't see anything that lets you hide them while editing.  We could try to strip the suffix out of the middle of the string as a workaround, but that seems messy since it could potentially  conflict with what you're trying to type if you're using a javascript function like Math.floor() or Math.pow() 

https://doc.qt.io/qt-6/qdoublespinbox.html

bjorn:
@dogboydog I guess we could try handling focus events to reset/restore the suffix/prefix in our custom spin boxes?

dogboydog:
It might be enough to just remove the prefix and suffix when the text input is focused if possible. The spinbox should put them back after the value is evaluated automatically . Not sure how we edit the text directly though since mostly you interact with the value() instead

bjorn:
@dogboydog Now the prefix/suffix is temporarily removed while the widget has focus. Please let me know how this behavior feels to you!

dogboydog:
I think that feels really nice! Thanks for adding that

dogboydog:
It's nice the way this feature worked out where it works throughout the app, including in Script Dialogs (just re-tested with the focus changes and still good there)

PR #4235, opened 2025-08-23
feat: add binmap plugin

BCsabaEngine:
Tiled BinMap Export Plugin
==========================

This is a [Tiled](https://mapeditor.org) plugin to dump tilemaps into a brainfuckingly simple binary format.

It ships BinMap as a plugin (which can be compiled into an .so file) as well as an extension (a .js file).

Installation
------------

You'll only need one of these, as they do exactly the same.

### JavaScript Extension

Copy `src/plugins/binmap/binmap-format.js` into `~/.config/tiled/extensions`.

### Tiled Plugin

1. Download the tiled source from [github](https://github.com/mapeditor/tiled).
2. Install all its dependencies (see its README for details).
3. Copy `src/plugins/binmap` directory into the downloaded repository.
4. Edit `src/plugins/plugins.qbs` and add `"binmap",` to the list.
5. In the repository's main directory, run `qbs`.

Documentation
-------------

The file format specification can be found [here](src/plugins/binmap/FORMAT.md). It is the simplest format possible,
just contains a fixed sized header and the uncompressed tile ids for each layer, that's all.

### DIFF
--- a/src/plugins/binmap/binmap-format.js
+++ b/src/plugins/binmap/binmap-format.js
@@ -1,0 +1,63 @@
+/// <reference types="@mapeditor/tiled-api" />
+
+/*
+ * binmap-format.js
+ *
+ * This extension adds the 'binary map format' type to the Export As menu.
+ * This is an alternative to the .so plugin, written in JavaScript.
+ * To install, copy to `~/.config/tiled/extensions`
+ *
+ * Simplest format possible. Consist of a 16 bytes long header and then data.
+ * See FORMAT.md for details.
+ */
+
+tiled.registerMapFormat("binmap", {
+	name: "Binary map files",
+	extension: "map",
+
+	write: (map, fileName) => {
+		let numLayers = 0, gid = 1;
+		/* fix missing firstgid properties on tilesets */
+		for (let i = 0; i < map.tilesets.length; ++i) {
+			map.tilesets[i].firstgid = gid;
+			gid += map.tilesets[i].tileCount;
+		}
+		/* count tile type layers */
+		for (let i = 0; i < map.layerCount; ++i) {
+			const layer = map.layerAt(i);
+			if (layer.isTileLayer) numLayers++;
+		}
+		/* file and buffers */
+		var file = new BinaryFile(fileName, BinaryFile.WriteOnly);
+		let header = new ArrayBuffer(16);
+		let item = new ArrayBuffer(2);
+		let hu16 = new Uint16Array(header);
+		let iu16 = new Uint16Array(item);
+		/* write header */
+		hu16[0] = 0x414D;
+		hu16[1] = 0x0050;
+		hu16[2] = map.width;
+		hu16[3] = map.height;
+		hu16[4] = numLayers;
+		hu16[5] = map.tileWidth;
+		hu16[6] = map.tileHeight;
+		/* looks like the Javascript API can't handle hexagonal maps */
+		hu16[7] = map.Isometric ? 1 : 0;
+		file.write(header);
+		/* write layer tiles */
+		for (let i = 0; i < map.layerCount; ++i) {
+			const layer = map.layerAt(i);
+			if (!layer.isTileLayer) continue;
+
+			for (let y = 0; y < map.height; ++y) {
+				for (let x = 0; x < map.width; ++x) {
+					const tile = layer.tileAt(x, y);
+					/* convert local tile id to global tile id */
+					iu16[0] = tile == null ? 0 : tile.id + (tile.tileset ? tile.tileset.firstgid : 1);
+					file.write(item);
+				}
+			}
+		}
+		file.commit();
+	},
+});
--- a/src/plugins/binmap/binmapplugin.cpp
+++ b/src/plugins/binmap/binmapplugin.cpp
@@ -1,0 +1,133 @@
+/*
+ * BinMap Tiled Plugin
+ * Copyright 2025, bzt <bztsrc@gitlab>
+ *
+ * This file is part of Tiled.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the Free
+ * Software Foundation; either version 2 of the License, or (at your option)
+ * any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along with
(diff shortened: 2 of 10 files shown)


src/plugins/binmap/binmap-format.js
```diff
@@ -0,0 +1,63 @@
...
+		/* write header */
+		hu16[0] = 0x414D;
+		hu16[1] = 0x0050;
+		hu16[2] = map.width;
+		hu16[3] = map.height;
+		hu16[4] = numLayers;
+		hu16[5] = map.tileWidth;
+		hu16[6] = map.tileHeight;
+		/* looks like the Javascript API can't handle hexagonal maps */
+		hu16[7] = map.Isometric ? 1 : 0;
```
bjorn, line 45: `map.Isometric` doesn't exist. What you're looking for is [`map.orientation`](https://www.mapeditor.org/docs/scripting/classes/TileMap.html#orientation), which will be`TileMap.Orthogonal`, `TileMap.Isometric`, `TileMap.Staggered` or `TileMap.Hexagonal`.

src/plugins/binmap/binmap-format.js
```diff
@@ -0,0 +1,63 @@
...
+ * See FORMAT.md for details.
+ */
+
+tiled.registerMapFormat("binmap", {
+	name: "Binary map files",
+	extension: "map",
+
+	write: (map, fileName) => {
+		let numLayers = 0, gid = 1;
+		/* fix missing firstgid properties on tilesets */
```
bjorn, line 20: This is not a "fix". "First global IDs" simply don't exist outside of the file format. It's fine to assign them as dynamic properties like this, but you'll want to reword that comment. :-)

src/plugins/binmap/binmap-format.js
```diff
@@ -0,0 +1,63 @@
...
+		for (let i = 0; i < map.tilesets.length; ++i) {
+			map.tilesets[i].firstgid = gid;
+			gid += map.tilesets[i].tileCount;
```
bjorn, line 23: I'll have to note that using `tileCount` unfortunately only works when there are no gaps in the tile IDs. Image collection tilesets can contain gaps, since when you remove tiles from such a tileset the other tiles don't change their ID.

If you want to support image collection tilesets, you'd currently need to iterate all tiles to find the maximum used tile ID.

src/plugins/binmap/binmapplugin.cpp
```diff
@@ -0,0 +1,133 @@
...
+    header.mapWidth = map->width();
+    header.mapHeight = map->height();
+    header.tileWidth = map->tileWidth();
+    header.tileHeight = map->tileHeight();
+    switch( map->orientation() ) {
+        case Map::Orthogonal: header.orientation = O_ORTHO; break;
+        case Map::Isometric:  header.orientation = O_ISO; break;
+        case Map::Hexagonal:  header.orientation = map->staggerAxis() == Map::StaggerX ? O_HEXH : O_HEXV; break;
+        default:
+            mError = QCoreApplication::translate("File Errors", "Unsupported map format.");
```
bjorn, line 94: Why not support `Map::Staggered` in this file format? Seems like it'd just take one more possible orientation value. At least, since `staggerIndex` is not getting exported.

src/plugins/binmap/binmap-format.js
```diff
@@ -0,0 +1,63 @@
...
+		for (let i = 0; i < map.layerCount; ++i) {
+			const layer = map.layerAt(i);
+			if (!layer.isTileLayer) continue;
+
+			for (let y = 0; y < map.height; ++y) {
+				for (let x = 0; x < map.width; ++x) {
+					const tile = layer.tileAt(x, y);
+					/* convert local tile id to global tile id */
+					iu16[0] = tile == null ? 0 : tile.id + (tile.tileset ? tile.tileset.firstgid : 1);
+					file.write(item);
```
bjorn, line 57: From a performance perspective I'd suggest writing entire layers at once rather than tile-by-tile.

bjorn (commented):
I'm actually a bit reluctant to merge this because "simple binary formats" come in so many flavors, given that "simple" usually means leaving out a bunch of information and different projects will need different subsets of the information available.

So you apparently only care about tile layers and might not care about image collections, isometric staggered, custom properties, the list of tilesets, parallax settings, layer blend modes, etc. But what's the use of shipping this "binmap" plugin, which presents this subset to many others who'll likely find some of the information they need missing?

As such, I do think the `binmap-format.js` can be helpful. For those who need just that, they can copy it over, but for those who need a little more or even less information, they can easily tweak it to their needs. But I'm not sure it makes sense to ship this extension with Tiled. What about using it as an example binary map format in [mapeditor/tiled-extensions](https://github.com/mapeditor/tiled-extensions) and linking to it from the documentation?

BCsabaEngine:
> So you apparently only care about tile layers

Precisely. This format is specifically for those who don't need anything else (99% of beginner projects) and find including an XML parser an overkill and overcomplicated.
This format is deliberately the simplest possible, needs no parsing at all. In return, it stores tile layers only, and nothing else.

> As such, I do think the binmap-format.js can be helpful.

Unfortunately it can't work correctly without firstGid properties. Current workaround isn't perfect as you've noted.

> What you're looking for is map.orientation

I've tried to use that, but didn't work. I'm open to whatever suggestion to query the TileMap format. Although this isn't an issue in the C++ version, that works well.

> I'll have to note that using tileCount unfortunately only works when there are no gaps in the tile IDs.

This is exactly why the C++ implementation is needed because that uses the GidMapper class so it can save global tile ids correctly.

bjorn:
> I've tried to use map.orientation, but didn't work. I'm open to whatever suggestion to query the TileMap format. Although this isn't an issue in the C++ version, that works well.

It should work, so please provide more details so we may find out what was going wrong here. A quick test in the Console view didn't reveal a problem to me:

```
> tiled.activeAsset.orientation
$0 = 1
> TileMap.Orthogonal
$1 = 1
```

> This is exactly why the C++ implementation is needed because that uses the GidMapper class so it can save global tile ids correctly.

While it does highlight a missing property in the scripting API, it can be worked around with minimal code:

```js
maxTileId = tileset.tiles.reduce((maxId, tile) => Math.max(maxId, tile.id), -1);
// instead of tileset.tileCount, use maxTileId + 1
```

Internally tilesets have a `nextTileId` property, which is essentially `maxTileId + 1` and could be directly exposed in the scripting API to make the above a little faster. To alleviate potential performance issues, one could use this as fallback and still rely on `tileCount` when `isCollection` is false.

> This format is specifically for those who don't need anything else (99% of beginner projects) and find including an XML parser an overkill and overcomplicated.
> This format is deliberately the simplest possible, needs no parsing at all. In return, it stores tile layers only, and nothing else.

While I entirely agree the format is simple to write and read, I still think it's better served as example and a starting point rather than a fixed format.

PR #4236, opened 2025-08-24
fix some issues

Ahmed-Almahi:


src/libtiled/compression.cpp
```diff
@@ -220,12 +227,11 @@ QByteArray Tiled::compress(const QByteArray &data,
 #ifdef TILED_ZSTD_SUPPORT
     } else if (method == Zstandard) {
         if (compressionLevel == -1)
-            compressionLevel = 6;
+            compressionLevel = 3; // ZSTD default
```
bjorn, line 230:
```suggestion
            compressionLevel = ZSTD_CLEVEL_DEFAULT;
```
Maybe we could use this define? Not sure since when it is available.

src/libtiled/compression.cpp
```diff
@@ -220,12 +227,11 @@ QByteArray Tiled::compress(const QByteArray &data,
...
         else
-            compressionLevel = qBound(1, compressionLevel, 22);
-
-        size_t const cBuffSize = ZSTD_compressBound(data.size());
+            compressionLevel = qBound(1, compressionLevel, 22); // ZSTD range
 
+        const size_t cBuffSize = ZSTD_compressBound(data.size());
```
bjorn, line 234: I'd suggest to leave this line alone, since this is the same thing. Or change the lines above and below as well, for consistency.
```suggestion
        size_t const cBuffSize = ZSTD_compressBound(data.size());
```

src/libtiled/compression.cpp
```diff
@@ -237,7 +243,6 @@ QByteArray Tiled::compress(const QByteArray &data,
 
         out.resize(cSize);
         return out;
-#endif
```
bjorn, line 240: You can't remove this line.

src/libtiled/compression.cpp
```diff
@@ -144,11 +146,16 @@ QByteArray Tiled::decompress(const QByteArray &data,
         return out;
 #ifdef TILED_ZSTD_SUPPORT
     } else if (method == Zstandard) {
-        size_t const dSize = ZSTD_decompress(out.data(), out.size(), data.constData(), data.size());
+        unsigned long const dBuffSize = expectedSize > 0 ? expectedSize : 10 * data.size();
```
bjorn, line 149: How do you run into the case of `expectedSize <= 0`? I don't think we have any such code path.

We do have a problematic code path though, which is when somebody tries to decompress zstd data through the scripting API. But in this case `expectedSize` matches `data.size()`.

bjorn (changes requested):

.vscode/settings.json
bjorn, line 1: Please don't add this file. It appears to be user-specific.

joereynolds: I think you're wasting your time reviewing this. Looks like this user has opened many PRs across many codebases to do these fixes. I'm assuming via AI to get internet points ¯\_(ツ)_/¯

bjorn: @joereynolds Ah, alright, thanks for having a look! It looked odd to me already, but it didn't look like AI to me especially with that PR title.

Ahmed-Almahi: @joereynolds I'm just trying to learn, but thanks for your support.

bjorn: @Ahmed-Almahi It's fine if you contribute to Tiled while learning, and I'm certainly available for explaining stuff, but it helps if you provide some context for your changes.

PR #4238, opened 2025-08-29
GitHub Actions: Update Qt to 6.9.2

bjorn:


bjorn:
Tested AppImage, Windows installer and macOS app. Seems to work fine!

Issue #3577, opened 2023-02-17
Ability to Search for and Group Tilesets


benjamin-kirkbride:
**Is your feature request related to a problem? Please describe.**
I have a ton of tilesets (literally hundreds, don't ask) and would like to be able to search through them and also categorize them.

**Describe the solution you'd like**
Any combination of:
- a search
- a tagging system
- folders
- ordering

**Describe alternatives you've considered**
excruciating scrolling

eishiya:
I thought there was an issue for this already, but apparently not!
Related feature requests: #873, #2901

For now, there is the Projects panel and the Search In Project (default shortcut: Ctrl+P), which can be used bring up the relevant tilesets by their location or name. However, while this is useful for opening a tileset, it won't help if the problem is that you already have too many tilesets open to begin with.

I think a related feature request would be the option to hide Tilesets panel tabs for tilesets that are not part of the current map.

benjamin-kirkbride:
Yeah certainly a lot of discussion to be had about exactly how this should/could be implemented.

Is anyone familiar with the JS or Python Plugins able to speak to whether this could be added via that?

benjamin-kirkbride:
Looking through the docs, it seems that it's only possible to create temporary dialogs/modals, and that it is not possible to create new views.

Is this correct?

https://www.mapeditor.org/docs/scripting/

eishiya:
That is correct.

It's possible to make a Dialog that lets you search currently open tilesets based on custom properties and bring up a single tileset as the currently active tileset in the Tilesets panel though, which should cover most needs. If multiple matches are found, the dialog can display a list for you to select one from.

benjamin-kirkbride:
I see, but you can't put it into the UI like the Layers Panel, for instance. Perhaps a new issue should be made for that haha

PR #4239, opened 2025-08-29
 add tileset name filter in tileset dock

dogboydog:
https://github.com/user-attachments/assets/fceb606e-53a4-4df2-87d2-c870653e144d

Related: #3577

Not sure if this PR is enough to close that issue though, since various options are described in the issue. This works great for my use case personally though.

### DIFF
--- a/src/tiled/tilesetdock.cpp
+++ b/src/tiled/tilesetdock.cpp
@@ -75,7 +75,7 @@
 using namespace Tiled;
 
 namespace preferences {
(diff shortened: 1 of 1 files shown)


dogboydog:
image
Added a toggle for the visibility of this filter for folks who prefer having more vertical space

dogboydog:
Thanks for the cleanup! Still working well for me. I think we're good to go

bjorn (approved):
Nice addition, and interesting that it was so easy with the `TilesetDocumentsFilterModel` already in place! :-)

PR #4241, opened 2025-09-08
Added additional description to Stagger Axis property

bjorn:
A staggered X axis leads to flat-top hexes whereas a staggered Y axis leads to pointy-top hexes.

image

This setting also applies to maps using Isometric (Staggered) orientation, in which case this distinction does not exist, so I'm not entirely sure whether this is a good idea. Should we have the property only use the additional labels when the map is using Hexagonal orientation?

### DIFF
--- a/src/tiled/propertieswidget.cpp
+++ b/src/tiled/propertieswidget.cpp
@@ -99,8 +99,8 @@ template<> EnumData enumData<Map::Orientation>()
 template<> EnumData enumData<Map::StaggerAxis>()
 {
     return {{
-        QCoreApplication::translate("StaggerAxis", "X"),
-        QCoreApplication::translate("StaggerAxis", "Y")
+        QCoreApplication::translate("StaggerAxis", "X (Flat-top)"),
+        QCoreApplication::translate("StaggerAxis", "Y (Pointy-top)")
     }};
 }
 


eishiya:
I think if possible, the extra description should only apply to hexagonal maps, yes. It would be confusing on staggered isometric maps, since most people working on those aren't going to be consciously thinking about how they're functionally similar to staggered hexagonal maps xP

Good idea adding this to the GUI rather than the documentation like I'd suggested.

Edit, though maybe this should be an issue: It would be nice if the stagger axis were selectable in the New Map dialog when a staggered orientation is selected. As it is, anyone working with flat-topped hexagons has to create their new map and then go into the properties to change the axis every time, TIled doesn't save that they want stagger axis X since it's not in the dialog.

A fuller section in the documentation also sounds good. Where would it go, though? The Introduction would be bogged down by it. Maybe it's time Maps got their own manual page :] I always miss the Introduction when looking for info on Maps because "Introduction" just doesn't sound like "[how to create/work with] Maps". Additional topics for such a page that would be too much for the Introduction: replacing tilesets, hex side length and how it relates to tile size (perhaps in the section on orientations), how to access the map properties 🙃, background color, layer data formats, chunk size... Some of these things are only explained in the format specs, which is not a place regular users should have to look. Even if they're not imortant to most users, they're still visible options in the editor, and that means some users will want more info on them than you can fit in a tooltip (and the tooltips currently have very little, if any, info).

EJlol:
While a text description is fine, and easiest to implement, making things visual with images would be even better. Not sure how much extra time it would cost though..

eishiya:
> While a text description is fine, and easiest to implement, making things visual with images would be even better. Not sure how much extra time it would cost though..

The visual would be provided by the tile grid in the map, which changes immediately, which I think it clearer than any icon would be, especially since the tiles can also be there for context.
However, if this is added to the New Map dialog, there'll be no such view there. I think a preview of the map or grid has been requested before though ;D

bjorn:
> I think if possible, the extra description should only apply to hexagonal maps, yes. It would be confusing on staggered isometric maps, since most people working on those aren't going to be consciously thinking about how they're functionally similar to staggered hexagonal maps xP

Alright, now it changes depending on the map orientation. :-)

> Edit, though maybe this should be an issue: It would be nice if the stagger axis were selectable in the New Map dialog when a staggered orientation is selected.

I'll look into adding this as part of this PR while I'm at it.

> A fuller section in the documentation also sounds good. Where would it go, though? The Introduction would be bogged down by it. Maybe it's time Maps got their own manual page

Yes, I think giving maps their own page is a good idea, and thanks for considering which other topics could be covered there!

> While a text description is fine, and easiest to implement, making things visual with images would be even better. Not sure how much extra time it would cost though..

Yeah, actually having buttons with clear pictures would be great for this, but sure it'll be a little more effort. And as @eishiya pointed out, I'm not sure it's worth it with the map view already providing direct feedback. If we keep using a "New Map" dialog, it would certainly be nice to add a preview or a visual choice though. Similar situation exists with the "New Tileset" dialog, which could use a preview marking the tiles on the tileset image based on the parameters.

bjorn:
> I'll look into adding this as part of this PR while I'm at it.

I decided not to change the New Map dialog for now, can be done in a later change.

> Yes, I think giving maps their own page is a good idea, and thanks for considering which other topics could be covered there!

There is now a dedicated page about maps at https://doc.mapeditor.org/en/latest/manual/maps/, added as part of #4313. Though since I was focusing on mentioning "Oblique" somewhere, I forgot to cover replacing tilesets, hex side length and how to access the map properties, background color, the various layer data formats and chunk size. So that will need a follow-up...

PR #4244, opened 2025-09-10
Qbs: Restrict debug DLL handling to MSVC toolchain on Windows

### LINKED ISSUE #4233: Build error: qschannelbackendd.dll' does not exist
### Describe the bug

When building with Qt 6.9 in Qt Creator, I receive this error, which can be ignored manually to proceed with building the app

 C:\Users\chris\tiled\dist\distribute.qbs:190: error: File 'C:/Qt/6.9.0/mingw_64/plugins/tls/qschannelbackendd.dll' does not exist.

@bjorn :
>Ah yes, this needs fixing in distribute.qbs. Qt debug builds used to have a d postfix added to the plugin names, but they stopped doing this at some point. Would be good to determine since which Qt version that happens.

Just logging this to track a fix for it at some point 

### Steps to reproduce

Build with Qt 6.9 (potentially older versions too) 

### Expected behavior

The build would complete without error 

### OS version

Windows 10

### Tiled version

master

bjorn:
Avoids warnings when using MinGW, since the Qt MinGW builds are no longer shipping debug DLLs since Qt 5.14.

Closes #4233

### DIFF
--- a/dist/distribute.qbs
+++ b/dist/distribute.qbs
@@ -105,7 +105,7 @@ Product {
 
     property var pluginFiles: {
         if (qbs.targetOS.contains("windows")) {
-            if (qbs.debugInformation)
+            if (qbs.debugInformation && qbs.toolchain.contains("msvc"))
                 return ["*d.dll"];
             else
                 return ["*.dll"];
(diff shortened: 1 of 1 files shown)


dogboydog:
I can confirm I'm no longer asked to ignore any errors with MinGW

PR #4249, opened 2025-09-12
Added tab bar to Properties view

### LINKED ISSUE #3101: Tileset and Map Properties could be easier to access

### LINKED ISSUE #3710: Properties window switch mode button
I'm Pixel Art Girl and I come from Pixel Land, where all people have autism!
I was send to earth to make Tiled more user friendly

There are 3 properties modes possible for the properties window: layer, map and tileset properties

- layer and map properties can only be switched by going to layer and map menu
- tileset properties appear automatically when clicking in the tileset

So if we want to see layer or map properties again, we need to use the menus endlessly. And when clicking in the tileset, the view automatically changes to tileset properties

It could be very handy if a button in the properties window is placed to switch between layer/ map/tileset properties. No need to go to the menu. Also in this way the current properties view doesn't change when we click inside the tileset

bjorn:
This change adds a tab bar that provides quick access to the Map and Tileset properties as well as a way to switch back to editing the properties of other selections like the current layer or last selected object:

image

Closes #3101
Closes #3710

### DIFF
--- a/src/tiled/propertiesdock.cpp
+++ b/src/tiled/propertiesdock.cpp
@@ -297,8 +297,10 @@ void PropertiesDock::tabChanged(int index)
         break;
     case WangColorTab:
         if (auto tilesetDocument = qobject_cast<TilesetDocument *>(mDocument)) {
+            qDebug() << "Switching to WangColor tab" << tilesetDocument;
             if (auto wangSet = tilesetDocument->selectedWangSet()) {
                 const int color = tilesetDocument->selectedWangColor();
+                qDebug() << "  with WangSet" << wangSet->name() << "and color index" << color;
                 if (color > 0)
                     tilesetDocument->setCurrentObject(wangSet->colorAt(color).data());
             }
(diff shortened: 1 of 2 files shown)


bjorn:
I've been trying to finish this feature but there's an open issue that I'm not sure how to resolve. While editing a map it is possible to click a Tile, Terrain or Terrain Set, and in this case the Properties view will display a read-only view on its properties. I kept this because it is sometimes useful to be able to see this information, but what should we do with the tab bar in this case?

> image
>
> Example: I clicked the "Green" terrain and its properties are displayed, but the tab bar is displaying "Map" and "Layer" because we're in the context of editing a map.

I'm kinda thinking we need to keep displaying the tabs related to the map, but add a "temporary" tab for such contextual objects. Maybe it would be good to make those visually distinct, for example by putting their name in italic. This does mean that sometimes up to 4 tabs will display, which can be a bit much for people using a narrow Properties view.

Switching away from such a temporary tab would remove the tab.

Edit: Unfortunately, it doesn't look like [QTabBar](https://doc.qt.io/qt-6/qtabbar.html) provides any way to change the font on a single tab. :-/

Issue #4201, opened 2025-05-02
Bugs / Inconsistencies with Fullscreen, Keybindings, Status Bar, Vertical Sorting.


UliAbo:
Hi everyone, 

I got lots of bugs and inconsistencies in Tiled noticed.

**Specifications:**
 - OS: Win11 Pro
 - Tiled Version: Version 1.11.2

### **1. Bug with Fullscreen mode:**
- In Fullscreen mode (via TAB), no rotation is possible of Objects or flipping, or other transformations - when using keybindings.

### **2. Bug with vertical position:**
- Click on Object in Map and then you can use PageUp and PageDown perfectly as intended. Also Home / End works for shifting Object vertically in its list.
- Click on the same Objects, but via Widget "Objects", then only PageUp / PageDown work the same way (or sometimes not?). But Home / End don't work, here the cursor jumps to first and last entry. Or vice versa, I couldn't reproduce it every time - seems to be also about selecting Objects or where the mouse cursor is?

### **3. Missing feature in Tile Collision Editor:**
- Tile Collision Editor: I can't use "DEL" as a keyboard shortcut to delete a polygon when having a polygon selected. But this shortcut works elsewhere, e.g. on Tiles. Should be more consistent.

### **4. Inconsistencies with Tools / Hotkeys:**

**4a. Rectangular Select Tool:** 
- Here I miss the SHIFT effect I have with the Shape Fill Tool (helping 1:1 square sized form)

**4b. Stamp Brush:**
- When I hold down right mouse button to span a rectangle for picking Tiles, I would expect in the Status bar the size info. 
![Image](https://github.com/user-attachments/assets/63b59b2c-4099-42b7-b407-dd5be72482d0)

**4c. Terrain Brush:** 
- Here I miss the Status bar info about the size / length I'm holding down before confirming.

**4d. Eraser:** 
- When holding down SHIFT, I would expect the same behavior as I have with using the Stamp Brush. With the Stamp Brush, I can hold down SHIFT after clicking the first time and it suggests a line I can confirm to paint. Additionally, I would like to also have the Status bar info here about size / length.

**Overall, all tools should work the same way regarding holding down SHIFT / CTRL and also give the same status bar info about rectangle / line size / length.** 

Cheers,
Abo

bjorn:
> 1. Bug with Fullscreen mode

[...]

PR #4258, opened 2025-09-23
Expand the status bar info for the Terrain Brush

bjorn:
Showing "Drawing line (length: \<length>)" when drawing lines (holding Shift).

See issue #4201


bjorn:
I think it would make sense to also add something to the status bar when using the Alt modifier for rotational symmetry, just not entirely sure what to call this.

eishiya:
Perhaps just "[cursor coordinate] and [mirrored coordinate]"? That avoids any extraneous text and makes it clear that it just works by painting the terrain at two locations (doesn't mirror the tiles).

bjorn:
> Perhaps just "[cursor coordinate] and [mirrored coordinate]"?

Sounds fine so I've just gone with that, thanks for the suggestion!

image

PR #4271, opened 2025-10-15
CI: Added 'brew update' to install latest Qbs

bjorn:
Without this I'm still getting Qbs 3.0.3 but we need a fix in Qbs 3.1.

### DIFF
--- a/.github/workflows/macos-latest.yml
+++ b/.github/workflows/macos-latest.yml
@@ -26,7 +26,8 @@ jobs:
 
     - name: Install dependencies
       run: |
-        brew install qbs
+        brew update
+        brew install qbs qttools
 
     - name: Setup Qbs
       run: |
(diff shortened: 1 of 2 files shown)


bjorn:
Seems that using Qbs 3.1 fixes the "Build Packages" workflow, however there is also a new problem when compiling Tiled against Qt installed from Homebrew:

```
ERROR: The process '/opt/homebrew/bin/lrelease' could not be started: Child process set up failed: execve: No such file or directory. The full command line invocation was: /opt/homebrew/bin/lrelease -silent -qm /Users/runner/work/tiled/tiled/release/translations.31c20e73/tiled_ar_DZ.qm /Users/runner/work/tiled/tiled/translations/tiled_ar_DZ.ts
```

Not sure yet why `lrelease` isn't found anymore. That still worked in the [last successful run](https://github.com/mapeditor/tiled/actions/runs/17391599080/job/49366193409) of this workflow when Qt 6.9.1 was used (now it's Qt 6.9.3).

bjorn:
Alright, it's because `lrelease` is part of `qttools`, which isn't installed as dependency of `qbs`.

PR #4272, opened 2025-10-15
CI: Limit push builds to 'master' and tags

bjorn:
Most other branches are only created to open a pull request, and then we only want the pull request related builds to trigger.


bjorn:
Oh boy, since this didn't trigger the `push` event, it apparently also didn't report the error in my workflow file... (using both `branches` and `branches-ignore` is not allowed)

PR #4286, opened 2025-11-12
Created RunClient button in Build dropdown button

WanSamuel:


bjorn:
6 people co-authored this patch? That must have been fun! :-)

You seem to be building a specialized Tiled version for your engine, which is great. But I guess this PR may have been opened by accident because it makes no sense to add this action to Tiled for everybody.

Btw, for this particular integration, consider using [Custom Commands](https://doc.mapeditor.org/en/stable/manual/using-commands/) instead. Tiled provides this feature to enable you to set up a shortcut for executing your game client.

PR #4288, opened 2025-11-24
Collision capsule support

5gTestSubject:
Adds support for capsule shapes constructed similarly to rectangles and ellipses.

Lots of copy/paste work of ellipse support with a little custom work on drawing.

## Object Layer
Screenshot_20251124_031030
Screenshot_20251124_031435-1

A .tmx snippet of an ellipse and a capsule:
```xml
<object id="1" x="159" y="71" width="152.615" height="70.5215" rotation="325.864">
 <ellipse/>
</object>
<object id="3" x="383" y="24" width="146" height="70" rotation="18.187">
 <capsule/>
</object>
```

## Tile Collisions
Left: ellipse. Right: two capsules.
They happen to be in an isometric project.
Screenshot_20251124_031632

A .tsx snippet of an ellipse and a capsule:
```xml
<object id="1" x="-643.5" y="-442.5" width="110" height="68">
 <ellipse/>
</object>
<object id="2" x="-537" y="-511" width="149.5" height="52.5">
 <capsule/>
</object>
```

## Needs Work
- Icons: Toolbars are currently using the ellipse icons, though they have the correct tooltips. I don't know how these were created and so can't really make one to match.
- Translations: idk how y'all do that.
- Shortcut key: (C was already taken for ellipses :sweat_smile:)
- There's an ellipse edge-case that may need to be adapted to capsules. Unclear.
- Idk the guidelines on who puts what name in a copyright header.

### DIFF
--- a/docs/manual/objects.rst
+++ b/docs/manual/objects.rst
@@ -89,7 +89,7 @@ collision shape needs to represent a circle or ellipse.
 Insert Capsule
 ~~~~~~~~~~~~~~
 
-No default shortcut |insert-capsule|
+Shortcut: ``Shift+C`` |insert-capsule|
 
 Capsules work the same way as `rectangles <#insert-rectangle>`__, except
 that they are rendered as a capsule. Useful for when your area or
--- a/src/tiled/createcapsuleobjecttool.cpp
+++ b/src/tiled/createcapsuleobjecttool.cpp
@@ -32,7 +32,7 @@ CreateCapsuleObjectTool::CreateCapsuleObjectTool(QObject *parent)
     QIcon icon(QLatin1String(":images/24/insert-capsule.png"));
     icon.addFile(QLatin1String(":images/48/insert-capsule.png"));
     setIcon(icon);
-    //setShortcut(Qt::Key_C);
+    setShortcut(Qt::SHIFT + Qt::Key_C);
     Utils::setThemeIcon(this, "insert-capsule");
     languageChangedImpl();
 }


5gTestSubject:
Oh! Almost forgot. Discussion on feature: https://github.com/mapeditor/tiled/issues/2153

bjorn:
That's great, thanks for making this addition! And indeed this really involved mostly minor adjustments to many places in the code and Qt can already draw these shapes so no problem there either.

I think the only remaining thing we need is a fitting icon. I'll try to create one in the style of the other ones.

I've added your name in copyright line for the new files. I think the changes in other files are small enough that it makes little sense to add your name everywhere. I hope that's fine.

Translations files will be updated 1-2 weeks before I make the next release.

I'm also not sure what would make a good default shortcut...

For the edge case, I've opted to also use it there since a 0-width/height capsule is also just a line. We can test later whether this is still an issue today.

5gTestSubject:
Maybe alt+C would work? There doesn't seem to be an existing expectation of what alt is meant to do and it does make a little sense with how similar ellipses and capsules are.

bjorn:
Unfortunately we already put <kbd>Alt+C</kbd> on a hidden "copy hovered coordinates to clipboard" action. :-)

bjorn:
Alright, icons are in, docs are updated, NEWS is updated! I think we're ready to merge this, @5gTestSubject?

5gTestSubject:
@bjorn looks good from my end. Thanks for accepting this! Game devs are gonna be happy.

docs/manual/objects.rst
```diff
@@ -72,18 +72,35 @@ a point object.
...
+
+.. _insert-capsule-tool:
+
+Insert Capsule
+~~~~~~~~~~~~~~
+
+No default shortcut |insert-capsule|
+
+Capsules work the same way as `rectangles <#insert-rectangle>`__, except
+that they are rendered as an capsule. Useful for when your area or
```
joereynolds: Tiniest nitpick but this should be "a capsule"

docs/reference/json-map-format.rst
```diff
@@ -233,6 +233,7 @@ Object
     :header: Field, Type, Description
     :widths: 1, 1, 4
 
+    capsule,          bool,             "Used to mark an object as an capsule"
```
joereynolds: Same here "a capsule"

docs/manual/objects.rst
bjorn: Thanks! That's what I get from simply copy/pasting and replacing "ellipse". :P

bjorn:
@5gTestSubject Thanks again for this contribution! I've opted to put it on Shift+C in the end, which should be still available.

PR #4289, opened 2025-11-25
Change Clipboard Mime-Type to Support Text

indianajson:
Proposal to update Clipboard mime-type to `text/plain` to allow for interoperability of copy/paste functions with external programs. Please see [Issue 4290](https://github.com/mapeditor/tiled/issues/4290) for details.

### DIFF
--- a/src/tiled/clipboardmanager.cpp
+++ b/src/tiled/clipboardmanager.cpp
@@ -33,6 +33,7 @@
 #include "tile.h"
 #include "tilelayer.h"
 #include "tmxmapformat.h"
+#include "preferences.h"
 
 #include <QApplication>
 #include <QClipboard>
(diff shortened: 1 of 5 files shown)


bjorn:
There's a few downsides to this approach:

* The Paste actions will get enabled when there is any text in the clipboard (though they silently fail when triggered).

* It breaks compatibility with older versions of Tiled (though it'd be rather unlikely for somebody to be copy/pasting stuff between two different versions of Tiled).

* If people copy any TMX file to their clipboard, trying to paste it will yield a rather broken thing because Tiled will not be able to resolve any of the tileset or image references (this may the most serious issue). To make this work, Tiled copies TMX in a special mode that forces all external references to use absolute paths.

So first I'd like to know whether it's really not possible to just use the `text/tmx` mime-type.

bjorn:
@indianajson I'm sorry for not getting around to looking at this again earlier. Now it's too late for Tiled 1.12 which went into string freeze last week, but I'll be sure to look into merging this soon after. I intend to not wait too long for the next release!

PR #4293, opened 2025-12-02
UI button

jackireland12:


dogboydog:
Was this PR opened by mistake? It seems to delete some Tiled code and there's no description of what you are contributing

bjorn:
Definitely a mistake. There are some custom additions visible in the patch, but nothing that appears to be meant to be submitted upstream, so I'll close this.

PR #4296, opened 2025-12-05
Updated karchive to 6.21.0

### LINKED ISSUE #4294: Patch for karchive to fix failed compilation
### Describe the bug

[static_cast_mode.patch](https://github.com/user-attachments/files/23890890/static_cast_mode.patch)

When compiling this on Gentoo I've been running into this compilation issue:

```
/usr/include/qt6/QtCore/qstring.h:364:5: note: template argument deduction/substitution failed:
/usr/include/qt6/QtCore/qstring.h: In substitution of ‘template<class ... Args> typename std::enable_if<((sizeof... (Args) >= 2) && conjunction_v<std::conjunction<std::negation<QtPrivate::treat_as_integral_arg<typename std::remove_cv<Args>::type> >, std::is_convertible<Args, QAnyStringView> >...>), QString>::type QString::arg(Args&& ...) const [with Args = {QFlags<QIODeviceBase::OpenModeFlag>&}]’:
/var/tmp/portage/dev-games/tiled-1.11.2_p20250614/work/tiled-f0191439b7f6b18cd727ee051e976756889b4d11/src/karchive/src/karchive.cpp:203:53:   required from here
  203 |         setErrorString(tr("Unsupported mode %1").arg(d->mode));
      |                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~
/usr/include/qt6/QtCore/qstring.h:364:5: error: no type named ‘type’ in ‘struct std::enable_if<false, QString>’
  364 |     arg(Args &&...args) const
      |     ^~~
```

I'm not super familiar with QT, but I fixed it by calling `static_cast<int>()` on uses of mode, maybe it's an enum and needs an int?

My system has: QTCore 5.15.18
Tiled version was: 1.11.2_p20250614

I'll attach the patch I made to allow it to compile on my system

Patch attached, hope this helps somehow.

### Steps to reproduce

emerge it on gentoo.

### Expected behavior

no compile errors.

### OS version

Gentoo Linux

### Tiled version

1.11.2_p20250614

bjorn:
Based on KArchive commit 9db52cdf92c50349b958bdd7e1f7651b13234917.

This update does mean the RpMap plugin is now only enabled for builds against Qt >= 6.8, which is the minimum supported Qt version still supported by KArchive.

Closes #4294


bjorn:
The related issue was fixed with a rather smaller patch at #4297, but I'll leave this open until we've raised the minimum supported Qt version of Tiled as well.

PR #4301, opened 2025-12-11
tscnplugin: Export map custom properties as TileMap metadata

ximop:
This PR updates the Godot 4 exporter to write the map custom properties (if any) as TileMap node metadata when writing the .tscn file.

It loops over all the elements of `map->properties()` and if the property name can be sanitized as a Godot identifier and the type is supported by the exporter it write a new line to the .tscn:

```
metadata/$name = $value
```

Example:

Exporting a map with following Tile custom properties:

Screenshot from 2025-12-11 14-09-54

Will write to the .tscn files the following lines:

```
metadata/myboolprop = true
metadata/mycolorprop = Color(0.827451, 0.211765, 0.509804, 1)
metadata/myproperty = "my value"
```
That will show up in Godot editor as:

Screenshot from 2025-12-11 14-09-29

### DIFF
--- a/src/plugins/tscn/tscnplugin.cpp
+++ b/src/plugins/tscn/tscnplugin.cpp
@@ -855,7 +855,7 @@ static void writeTileset(const Map *map, QFileDevice *device, bool isExternal, A
     device->write("\n");
 }
 
-static void writeTileMapMetadata(const Properties &mapProperties, AssetInfo &assetInfo, QFileDevice *device)
+static void writeMetadata(const Properties &mapProperties, AssetInfo &assetInfo, QFileDevice *device)
 {
     QSet<QString> sanitizedNames;
     sanitizedNames.reserve(mapProperties.size());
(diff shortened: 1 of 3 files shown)


ximop:
The check fails seem unrelated to my PR. Network issues maybe?

src/plugins/tscn/tscnplugin.cpp
```diff
@@ -935,6 +975,12 @@ bool TscnPlugin::write(const Map *map, const QString &fileName, Options options)
 
         device->write("format = 2\n");
 
+        // Write map custom properties as node metadata
+        Properties mapProperties = map->properties();
+        if (!mapProperties.isEmpty()) {
+            writeTileMapMetadata(mapProperties, assetInfo, device);
+        }
```
bjorn:
```suggestion
        writeTileMapMetadata(map->properties(), assetInfo, device);
```
Simplification because `writeTileMapMetadata` will not write anything anyway when there are no properties.

src/plugins/tscn/tscnplugin.cpp
```diff
@@ -854,6 +855,45 @@ static void writeTileset(const Map *map, QFileDevice *device, bool isExternal, A
     device->write("\n");
 }
 
+static void writeTileMapMetadata(const Properties &mapProperties, AssetInfo &assetInfo, QFileDevice *device)
```
bjorn: Is this metadata specific to tile maps? Previously we only supported custom properties on tiles, mapping it to custom data layers, but maybe we can write out properties as metadata for other objects too, like tilesets, layers or objects?

Then I think we could rename this to simply `writeMetadata` and call it in a few more places.

ximop: Any node in Godot can have metadata so we can potentially add it in some other places. I can do it but I'll need some time to investigate the implications since there are many properties that are handled in special ways by the exporter and I could break something.

Meanwhile, I've removed the word "map" from the error messages in `writeMetadata` for when the function is used in other places.

ximop:
Thanks for the review @bjorn. I've addressed your feedback. Please take another look.

src/plugins/tscn/tscnplugin.cpp
Taffer: This is relevant to #4420 depending on which gets merged first. :+1:

PR #4308, opened 2025-12-31
Add .tide support to tbin plugin

spacechase0:
This PR add `.tide` support to the tbin plugin. When I made the plugin in the past, it was for modders - what we got from extracting the game's `.xnb` files was the `.tbin` file format. However, on the developer side, it actually uses `.tide` (which is another xTile format). The xTile map editor (also called tIDE) is still more tedious to use than Tiled though, hence wanting to add this.

I opted to make this part of the tbin plugin instead of a new one since they're basically just different encodings of the same format - `.tbin` is binary, while `.tide` is XML. It also let me reuse a bunch of code between the two without duplication. I considered changing the plugin name to just `xtile` since they are both xTile formats, but I wasn't sure if that would mess anything up (or be confusing if people upgrade and the tbin plugin was gone).

All of the same limitations exist for `.tide` that also exist for `.tbin`, such as no rotated/flipped tiles.

---

There also some additional tweaks I made, which apply to `.tbin` as well:
* Layer visibility is now properly transferred to/from Tiled when de/serializing.
* When de/serializing, tile properties (ie. the `TileData` objects) now attempt to merge/split into larger/smaller object(s) where possible and non-destructive, to make changing them on larger areas than a single tile more useful. (See followup comment for example of before and after, as well as some details on how it functions with overlaps.)
    * I wanted to make an actual setting for this because I feel like there could be cases where it would be undesired, but I couldn't figure out how. I figure more often than not it *would* be desired, though. (It certainly is for SDV.)
    * The algorithm for this is kinda naive and so probably not perfect, but should never be destructive, and is much nicer than having it separate for every single relevant tile.
* This isn't user facing, but "Fake SFML" has been removed, since I had only left it there before from being lazy. However, that `tbin/` folder still isn't using Qt in itself (because I'm still somewhat lazy), and so there's still a `tbin::Vector2i` (previously `sf::Vector2i`). 

---

## Test cases and results

Tested on multiple maps that Stardew Valley uses already, which I believe covers every part of this PR:
* 3 small maps (the farm house interior for new saves, the inside of Linus's tent, and a Joja-related map introduced in SDV 1.4 (being vague because it's kinda a spoiler if you haven't played the game or reached it before).
* 3 medium maps (farm house after two upgrades and while married, a map containing every "spouse room" for the farm house when married, and the interior of Pierre's shop)
* 4 large maps (the default layout for the farm, Cindersap Forest, the beach, and the town)

After loading each of the above maps and resaving it, the following changes were observed in the diff: (`[TBIN]` means this probably already applied to `.tbin` as well, but that's harder to diff)
* Minor whitespace changes in the XML, such as `<SomeTag />` now being `<SomeTag/>`
* The case of the encoding string in the XML header was changed (`"utf-8"` -> `"UTF-8"`)
* [TBIN] Properties on something (a map, tilesheet, tilesheet tile, layer, or specific tile) were sometimes serialized in a different order (I assume due to `std::map` and `QMap` behaving differently than C# `Dictionary`)
* [TBIN, maybe] One map had a layer with no actual tiles on it (intentionally). Originally, for some reason there was a "tilesheet change" marker part way through a row, but since there were no tiles on that row it didn't actually do anything. After resaving, that pointless "tilesheet change" marker disappeared. (See followup comment for example diff.)
* [TBIN] Maps which had a tilesheet reference that had an incorrect size listed (because the tilesheet had been extended since the last time that map was saved, I believe) had the correct size after being resaved.
* [TBIN] Maps lose their `"Id"` field, since Tiled has no place that corresponds for it to be put in. Placing it as a map property could conflict with an actual `"Id"` property (which SDV doesn't have, but maybe other games would)

This also updates the license from my old name to my new one, as discussed in discord. (That was going to be a separate PR, but then I realized it'd be better to put the `.tide` stuff in the same plugin.) The `.tide` serialization code I put a copyright for ConcernedApe LLC instead of myself since I did it on company time, and the other file changes are mostly minor things besides moving stuff around.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -24,6 +24,7 @@
 * Scripting: Added TileMap.chunkSize and TileMap.compressionLevel properties
 * AutoMapping: Don't match rules based on empty input indexes
 * AutoMapping: Optimized reloading of rule maps and load rule maps on-demand
+* tBIN plugin: Added support for the tIDE XML format (by Casey Warrington, #4308)
 * Windows: Fixed issue with opening Tile Animation Editor (#4223)
 * macOS: Fixed crash when JS code is JIT-compiled (#4218)
 * Workaround tileset view layout regression in Qt 6.9


spacechase0:
Example of the disappearing "tilesheet change" marker that was mentioned: 
mintty_IHqmgo8im7

A particularly obvious example of tile data merging (before), overall view and two specific tile data objects as examples:
tiled_zuzLCnWyJE
tiled_UiohI2T84I

The same example of tile data merging but after, overall view and the same two tiles as before:
tiled_gngOWjH27J
tiled_WP6KemyuIA

There some even more egregious examples of the tile data merging, but they're all just the larger versions of the same map.

---

Some information on the tile data merging, and behavior for "incomplete merges" (ie, that example from above).

The entire large bottom area has just "FloorID: Bedroom", while the small tile at the entry way has that as well as "NoFurniture: T". With this merge implementation, the entry way TileData has only "NoFurniture: T", and 'inherits' the "FloorID: Bedroom" from the larger one.
* If they both have the same tile property, the tile data specifically for that tile (ie. a tile size of 1x1) will be preferred.
* If there are multiple larger tile datas over the same spot, I believe the one that comes first in the object list will take precedence. Generally you shouldn't do this though, since it will result in data loss once you save and reload.
* If are multiple 1x1 tile datas for that spot, I believe the one that comes last would take precedence. Generally you shouldn't do this though, since it will result in data loss once you save and reload.

Merging will only happen if the "larger version" is strictly a subset of the original tile. In theory it might be possible to make it slightly more flexible, but it'd complicate the code a fair bit (because of the overlap stuff mentioned previously).

Here's an example of what happens if that single bottom left tile data is given a different "FloorID", and then you saved and reloaded. Less is merged as a result.
tiled_5xxs09GhW0
tiled_2u3COJlANc

src/plugins/tbin/tidemapformat.cpp
```diff
@@ -0,0 +1,566 @@
...
+                            writeProps( tile.props );
+                            xml.writeEndElement();
+                        }
+                    };
+                    xml.writeStartElement( "TileArray" );
+                    std::string lastTilesheet;
+                    for ( int iy = 0; iy < layer.layerSize.y; ++iy )
+                    {
+                        if ( iy == 1 )
+                            iy = iy;
```
bjorn: Self assignment likely not intended?

src/plugins/tbin/tidemapformat.cpp
```diff
@@ -0,0 +1,566 @@
...
+                                }
+
+                                xml.writeEmptyElement( "Null" );
+                                xml.writeAttribute( "Count", QString::number( nullCount ) );
+                            }
+                            else if ( tile.staticData.tileIndex != -1 )
+                            {
+                                if ( tile.staticData.tileIndex == 175 )
+                                {
+                                    ix = ix;
```
bjorn: Self assignment likely not intended?

src/plugins/tbin/tidemapformat.cpp
```diff
@@ -0,0 +1,566 @@
...
+
+                    xml.writeStartElement( "Description" );
+                    xml.writeCDATA( QString(layer.desc.c_str()) );
+                    xml.writeEndElement();
+
+                    xml.writeEmptyElement( "Dimensions" );
+                    xml.writeAttribute( "LayerSize", QStringLiteral( "%1 x %2" ).arg( layer.layerSize.x ).arg( layer.layerSize.y ) );
+                    xml.writeAttribute( "TileSize",  QStringLiteral( "%1 x %2" ).arg( layer.tileSize.x  ).arg( layer.tileSize.y  ) );
+
+                    auto writeStaticTile = [&xml, &writeProps](const tbin::Tile& tile, std::string& lastTilesheet)
```
bjorn: This lambda doesn't use its `lastTilesheet` parameter.

src/plugins/tbin/tidemapformat.cpp
spacechase0: Yeah I usually do stuff like that when I need a conditional breakpoint (out of habit, since they seem really slow for some reason, at least in .net).

src/plugins/tbin/tidemapformat.cpp
spacechase0: Yeah I usually do stuff like that when I need a conditional breakpoint (out of habit, since they seem really slow for some reason, at least in .net).

src/plugins/tbin/tidemapformat.cpp
spacechase0: Ah, it looks like I moved it while getting the output to match xTile's as closely as I could.

Issue #2974, opened 2021-01-14
Tiled fails to render alpha component of tintcolor correctly for some tiles


onpon4:
I have an example in my game engine which uses a Tiled map to design a level. As I made some changes to test the addition of tintcolor support to the engine, I noticed something odd: I had use the tint color to adjust alpha transparency simply because the images used were black (and thus wouldn't be affected by the color component of tint color, but it seems Tiled isn't actually correctly applying the blending correctly to alpha transparency in all cases. At first I thought this might be #2930, so I changed the color of the tiles from black to gray and applied a color tint, confirming that this is a different issue.

Here is the Tiled JSON file (as the JSON format is what I use) and the tileset file and tileset image it uses:

[bad_tiled.zip](https://github.com/mapeditor/tiled/files/5816015/bad_tiled.zip)

Here is a screenshot demonstrating what it looks like in the test case for the engine, and what it should look like:

![Screenshot from 2021-01-14 12-28-30](https://user-images.githubusercontent.com/2217995/104626948-6ec42a00-5664-11eb-8456-a57315bfbf79.png)

But in Tiled 1.4.3, it looks like this:

![Screenshot from 2021-01-14 12-32-14](https://user-images.githubusercontent.com/2217995/104627106-9adfab00-5664-11eb-8b26-7d5a3f2c6334.png)

The image exported by Tiled is quite different, and this may perhaps give some insight into the cause of the problem:

![level](https://user-images.githubusercontent.com/2217995/104627325-e3976400-5664-11eb-9ba7-19c7507ef5d9.png)

I'm not sure how easy this is to see in a Web browser, but if you open it in an image viewer that shows transparent sections as a checkerboard pattern, you can clearly see that the parts with the gray square tiles are seemingly erasing the cyan background color.

Tested via the 1.4.3 AppImage on Debian 10 "Buster".

Kiamo2:
Strangely his example in 'bad_tiled.zip' now works.

But in other cases it still happens that reducing alpha of the tint color produces wrong results.
The color is way to dark and with alpha = 0 it results in a solid black.
It seems that a black instead of e.g. a transparent background color is assumed.
And it seems that this happes only if the tile image does not show the transparent background somewhere i.e. the background is completely concealed.
If a tranparent background in the tile image somewhere shines through the tinting appears correct.

benryhenson:
[...]

PR #4310, opened 2026-01-02
Fixes issue #2974

Kiamo2:
If pngs used for tilesets don't have an alpha channel tinting a tile layer with an alpha below 255 produces wrong results.
This is a small demo for that issue [issue-demo.zip](https://github.com/user-attachments/files/24408582/issue-demo.zip)

That's how it looks currently
before

Result with my proposed fix:
after_fix

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -20,6 +20,7 @@
 * Added export plugin for Remixed Dungeon (by Mikhael Danilov, #4158)
 * Added "World > World Properties" menu action (with dogboydog, #4190)
 * Added Delete shortcut to Remove Tiles action by default and avoid ambiguity (#4201)
+* Fixed alpha component of tint color not applying correctly to opaque images (by Roland Helmerichs, #4310)
 * Scripting: Added API for custom property types (with dogboydog, #3971)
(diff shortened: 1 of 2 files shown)


Kiamo2:
No idea why several checks failed esp. that Windows (64 bit, Qt 6) which I built locally without problems.

bjorn:
> No idea why several checks failed esp. that Windows (64 bit, Qt 6) which I built locally without problems.

Well, the Qt 5 builds are failing because [`QImage::convertedTo`](https://doc.qt.io/qt-6/qimage.html#convertedTo-1) only exists since Qt 6.0:
```
error: ‘class QImage’ has no member named ‘convertedTo’; did you mean ‘convertTo’?
```
Anyway it'll be quite trivial to change this to use `convertTo` instead (and maybe we should use `#ifdef` because I plan to drop Qt 5 support after the 1.12 release).

bjorn:
Well, didn't use `#ifdef` in the end because I preferred the shorter line anyway.

Thanks for the patch, @Kiamo2! It makes sense to force an alpha channel on the result image when we want it to fade out. Though it would still be preferred to use the layer opacity since that doesn't require copying the tile image, heh.

PR #4311, opened 2026-01-10
Adds 'Move Tiles' tool for moving and duplicating tile selections

Syynth:
While looking to see if there was any prior work on this, I saw #1647 and did my best to incorporate the feedback there. 

  - Shortcut: V (standard move tool shortcut)
  - Drag to move selected tiles
  - Alt + Drag to duplicate instead of move
  - Escape or right-click to cancel mid-drag
  - Undo restores both tiles and selection position

The tool uses a floating selection visualization with a marching ants animation during the drag operation. The move is committed as a single command to the undo stack when released.

https://github.com/user-attachments/assets/489611a6-173f-4d02-8824-fe4b2aeade40

Files added:
  - tilemovetool.h/cpp - Tool implementation
  - floatingtileselectionitem.h/cpp - Drag visualization
  - movetiles.h/cpp - Undo command
Files changed:
  - editing-tile-layers.rst - Add to manual
  - libtilededitor.qbs - Update build system
  - mapeditor.cpp - Register tool
  - undocommands.h - Register command

---

Personally I'd kind of like the move tool to be able to create/update the selection without switching to a different tool, but this felt most closely aligned with how the other tools already worked.

If you have any feedback about how it should work, please let me know and I'd be happy to make adjustments; this was just my first pass. If you'd prefer not to add it, I understand but this was to scratch my own itch first and foremost 😁


src/tiled/floatingtileselectionitem.h
```diff
@@ -0,0 +1,69 @@
...
+
+#include <QGraphicsObject>
+#include <QRegion>
+
+namespace Tiled {
+
+class MapDocument;
+class MapRenderer;
+
+class FloatingTileSelectionItem : public QGraphicsObject
```
bjorn: Why did you opt for implementing `FloatingTileSelectionItem` from scratch, rather than adding an "animated outline" option to `BrushItem`? It would probably make sense to share some code here.

src/tiled/floatingtileselectionitem.cpp
```diff
@@ -0,0 +1,185 @@
...
+        painter->setRenderHint(QPainter::Antialiasing);
+        painter->setPen(pen);
+        painter->drawLines(lines, 4);
+
+        pen.setColor(Qt::black);
+        pen.setCapStyle(Qt::FlatCap);
+        pen.setDashPattern({dashLength, dashLength});
+        pen.setDashOffset(mDashOffset);
+        painter->setPen(pen);
+        painter->drawLines(lines, 4);
```
bjorn: When the region is not a rectangle, this approach causes undesired horizontal marching ant lines going through the region. Could we avoid rendering these lines by taking the `QPainterPath` approach done in `OrthogonalRenderer::drawTileSelection`? I realize it will change the direction of the marching ants, so we may have to decide what looks less strange.

Alternatively we could of course try to find a way to adjust those lines to only the necessary parts, which will also be tricky.

bjorn (commented):
@Syynth First of all, thanks for taking up this feature again! And I'm sorry for not getting around to it for a while.

I've tried out the feature and I quite like it. Probably it's indeed a good idea to avoid the persistent floating selection, which was causing most of the problems in #1647.

But why is this now a separate tool rather than being part of the `AbstractTileSelectionTool`?

src/tiled/floatingtileselectionitem.h
Syynth: So the root cause of both of the things you pointed mostly boiled down to me being hesitant to touch too many pieces at once while exploring whether I could get something that did what I wanted. 

However! I was able to get this working by changing `BrushItem` to extend `QGraphicsObject` instead of `QGraphicsItem` to get the animated outline working.

src/tiled/floatingtileselectionitem.cpp
Syynth: I was able to get this working eventually, by calling `QPainterPath::simplified()` which seemed to do the right thing.

The only change I was a little uncomfortable making but convinced myself it was probably okay was making `IsometricRenderer::tileRectToScreenPolygon` in `libtiled` public instead of private. My logic was that it matched the similar thing in the hexagon renderer (`tileToScreenPolygon`) which was already public.

Syynth:
@bjorn Thanks for the feedback! I gave reworking what I had into the `AbstractTileSelectionTool` and it turned out pretty well, I thought. 

I mostly just had to check in each selection tool click handler whether we were trying to move an existing selection or do whatever the tool normally did. 

Let me know if you see any other issues, now that I've reworked this I think it's a really nice addition, I'd be happy to get it finished if you're interested in pulling it in!

PR #4313, opened 2026-02-03
Add Oblique map orientation

### LINKED ISSUE #2917: Oblique tiles / map
Can support for oblique projection style be added? I have searched everywhere but seems that the main styles supported are isometric/hexagon styles etc. This site has a good description & examples from old games - http://www.significant-bits.com/a-laymans-guide-to-projection-in-videogames/

Here are a few tiles that should help!

character
gravel
hole
ladder
planting
wall
wall_a
wall_v

bjorn:
Introduce the "oblique" orientation with per-axis skew (skewx/skewy) attributes.

Updated DTD, XSD, JSON and TMX formats, added ObliqueRenderer implementation, and integrated the new orientation option and related properties in the UI and undo system.

Closes #2917

### DIFF
--- a/docs/manual/automapping.md
+++ b/docs/manual/automapping.md
@@ -230,7 +230,7 @@ IgnoreHexRotate120
 : This boolean layer property can be added to `input` and `inputnot` layers to also match 120-degree rotated tiles on hexagonal maps. However, note that Automapping currently does not really work for hexagonal maps since it does not take into account the staggered axis.
 
 (outputProbability)=
-Probability {bdg-primary}`New in Tiled 1.10`
+Probability {bdg-secondary-line}`Since Tiled 1.10`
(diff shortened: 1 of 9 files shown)


bjorn:
Still has some bugs with certain objects types, Lua export missing, and further testing needed.

bjorn:
Issue to fix: The "Tile Render Order" property should not be disabled in the Properties view, when the map orientation is Oblique.

HeadClot:
Hey @bjorn - I saw a post from you on discord say:

"It basically obsoletes the isometric renderer...
Well, except for the tile rendering order, since only the isometric renderer iterates the tiles diagonally." - [Link](https://discord.com/channels/524610627545595904/524610627545595906/1468712676309074084)

Will this be a breaking change for exporters when it is merged?

bjorn:
@HeadClot Yeah, actually, since I added an "except" line, I basically realized that the Oblique renderer does not in fact obsolete the Isometric renderer. It does essentially obsolete Orthogonal because an Oblique map with Skew X/Y set to 0 will behave basically the same as an Orthogonal map, but I'll leave it as-is for now for clarity.

In terms of this being a breaking change, it only affects people who will use this feature, in which case the importer would need to:

* Recognize the `oblique` value for the `orientation` property of the `map` element.
* Read the `skewx` and `skewy` attributes of the `map` element.
* Apply the skew while rendering the map (if it implements rendering).

PR #4315, opened 2026-02-05
Add ROTATED_HEXAGONAL_120_FLAG in libtiled-java

exoego:
ROTATED_HEXAGONAL_120_FLAG from https://github.com/mapeditor/tiled/blob/d94c6a3fe13c0dee2c4ac1ff68603f925e3c7a1b/src/plugins/csv/csvplugin.cpp#L41
is missing in Java


bjorn:
Thanks for the update!

PR #4317, opened 2026-02-05
docs: add Maven Central badge to increase libtiled-java visibility

exoego:
So libtiled-java users can find latest version (and past versions) easily.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,6 @@ be used in any map. Tilesets can be modified at any time.
 [![Snap](https://snapcraft.io/tiled/badge.svg)](https://snapcraft.io/tiled)
 [![Translation status](https://hosted.weblate.org/widgets/tiled/-/shields-badge.svg)](https://hosted.weblate.org/engage/tiled/?utm_source=widget)
 [![Open Source Helpers](https://www.codetriage.com/mapeditor/tiled/badges/users.svg)](https://www.codetriage.com/mapeditor/tiled)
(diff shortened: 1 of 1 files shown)


dogboydog:
Just adds a badge, I think it looks good. @bjorn  ?

dogboydog (approved):

bjorn:
While it's a nice idea in principle, I'm hesitant about this because libtiled-java has not been properly maintained in a very long time, with the last real updates (apart from some optimizations and dependency version updates) being almost 6 years ago. So I'd prefer to add this badge only to `util/java/libtiled-java/README.md` for now.

README.md
```diff
@@ -17,6 +17,7 @@ be used in any map. Tilesets can be modified at any time.
 [![Snap](https://snapcraft.io/tiled/badge.svg)](https://snapcraft.io/tiled)
 [![Translation status](https://hosted.weblate.org/widgets/tiled/-/shields-badge.svg)](https://hosted.weblate.org/engage/tiled/?utm_source=widget)
 [![Open Source Helpers](https://www.codetriage.com/mapeditor/tiled/badges/users.svg)](https://www.codetriage.com/mapeditor/tiled)
+[![Maven Central Version](https://img.shields.io/maven-central/v/org.mapeditor/libtiled?label=libtiled-java)](https://central.sonatype.com/artifact/org.mapeditor/libtiled)
```
exoego:
```suggestion
```

bjorn:
Thanks @exoego! I see you've also started to modernize libtiled-java in #4325. That's great!

I do think that we should probably move this library into its own repository, like all other such libraries really. But if we do this we'll make sure to do it in a way that keeps its development history.

PR #4318, opened 2026-02-06
file state tracking with visual indicators for deleted and recreated files

### LINKED ISSUE #4130: Make "file not found" behavior more source control friendly
Every time an open file is deleted or removed, Tiled pops up a dialog that requires user interaction to resolve, like this:
image

This gets very frustrating very quickly when you have a lot of maps open and you're working with source control. I use git, and I'm often moving between branches or commits, stashing and un-stashing changes, etc., and there's a constant cacophony of error dialog sounds from Tiled and an ever increasing stack of dialogs like the one above I have to press "OK" on before I can resume work after I'm done.

I love the way Visual Studio Code deals with this situation. The open file gets a red color and a strikethrough on the open file tab, but otherwise nothing happens:
image
and if you bring the file back, it reloads it and goes back to showing it as a normal, open tab:
image

If you happen to be editing the file and it is removed and restored *with different changes* since you started editing it, you will be notified of this only when attempting to save your new changes:
image
and again, it's an important detail that this only shows up when the file being saved is *different* on disk than when the edit started. So even if you made a bunch of edits to the file behind the scenes, it'll still save just fine if the contents of it is back to what VSCode expects by the time you save, or if you deleted it.

Something that does work is closing all the maps before doing file operations as the editor only complains about rug pulls on open files. But it's easy to forget and a less friendly option when making on-going changes to multiple files that you like keeping open.

djad04:
File lifecycle tracking system that detects on-disk deletions, modifications, and restorations, provides clear visual feedback through tab styling, tooltips, context menus, and warning banners, and allows recovery of deleted files with the option to override disk changes using the in-memory version or reload the disk version

###  File Deletion Detection
When a file is detected as missing from disk :

#### Tab turns red with strikethrough text
Screenshot 2026-02-05 155536

#### Tooltip shows "deleted" suffix
- When hovering over the tab, the tooltip displays the filename with "deleted" appended

#### Right-click context menu adds "Restore" and "Save As" options
Screenshot 2026-02-06 135054

#### File deleted warning banner
Screenshot 2026-02-05 155536

---

###  File Recreated Detection
When a previously deleted file is recreated on disk, :

#### Modification Check
- When a deleted file reappears on disk we compare modification timestamps:
- **If the timestamp is identical**: the file is treated as unchanged; the tab returns to its normal state and the warning banner is hidden.  
- **If the timestamp differs**: a recreated state is triggered, the tab changes to yellow, and a warning banner is displayed.

#### Warning banner updates to "FileRecreated" state
- action buttons: "Reload" and "Overwrite"
- Helps user decide whether to reload the new file or overwrite it with current in-memory version
Screenshot 2026-02-05 161004

---

###  Save Handling 
When attempting to save a document whose file has been recreated on disk with different content:
a conflict popup appears 
Screenshot 2026-02-05 193913

---


src/tiled/documentmanager.cpp
```diff
@@ -55,10 +55,15 @@
 #include <QDialogButtonBox>
 #include <QFileDialog>
 #include <QFileInfo>
+#include <QFile>
+#include <QTemporaryFile>
+#include <QCryptographicHash>
+#include <QDir>
```
Oval17: These seems to planned for something but none of them are used anywhere!

src/tiled/tabbar.h
```diff
@@ -33,15 +34,26 @@ class TabBar : public QTabBar
...
+    bool isTabRecreated(int index) const;
 
 protected:
     void mousePressEvent(QMouseEvent *event) override;
     void mouseReleaseEvent(QMouseEvent *event) override;
 
     void wheelEvent(QWheelEvent *event) override;
+    void paintEvent(QPaintEvent *event) override;
+
+protected:
```
Oval17: Duplicate protected here, they can be merged we dont need two protected.

src/tiled/documentmanager.cpp
```diff
@@ -1189,9 +1368,90 @@ void DocumentManager::hideChangedWarning()
...
+    if (isDocumentModified(document))
+        return false;
+
+    QFileInfo fileInfo(fileName);
+    if (!fileInfo.exists())
+        return false;
+
+    return timestampMatches(fileInfo, document);
+}
+
```
Oval17, line 1440: There are unnecessary multiple blank lines here.

src/tiled/documentmanager.cpp
```diff
@@ -1189,9 +1368,90 @@ void DocumentManager::hideChangedWarning()
...
 
     document->setChangedOnDisk(false);
+    mRecreatedDocuments.remove(document);
     mFileChangedWarning->setVisible(false);
+
+    
+    const int index = findDocument(document);
+    if (index != -1)
+     mTabBar->setTabRecreated(index, false);
+
```
Oval17: Here from line 1373 to 1378 , we have bad indentation and few unnecessary blank lines.

src/tiled/documentmanager.cpp
```diff
@@ -1189,9 +1368,90 @@ void DocumentManager::hideChangedWarning()
...
+    return mDeletedDocuments.contains(document);
 }
 
+void DocumentManager::checkForRestoredFiles()
+{
+    // Check all deleted documents to see if their files have been restored
+    QSet<Document*> restoredDocuments;
+
+    for (Document* document : std::as_const(mDeletedDocuments)) {
+        const QString fileName = document->canonicalFilePath();
```
Oval17: I think canonicalFilepath() resolves symlink via the filesystem and always returns an empty string if file doesnt exist. that means our check below `!fileName.isEmpty()` is always false.
Is using document->filename() here better alternative?

src/tiled/documentmanager.cpp
```diff
@@ -1189,9 +1368,90 @@ void DocumentManager::hideChangedWarning()
...
+        fileChanged(fileName);
+    }
+}
+
+
+
+bool DocumentManager::timestampMatches(const QFileInfo &fileInfo, Document *document) const
+{
+    return (fileInfo.lastModified() == document->lastSaved() ||
+            qAbs(fileInfo.lastModified().toMSecsSinceEpoch() - document->lastSaved().toMSecsSinceEpoch()) < 2000);
```
Oval17, line 1423: Why we have 2 second timestamp window here, is there any particular reason for this?

src/tiled/documentmanager.cpp
```diff
@@ -1189,9 +1368,90 @@ void DocumentManager::hideChangedWarning()
...
+}
+
+bool DocumentManager::isRestoredFileIdentical(Document *document, const QString &fileName) const
```
Oval17: Just a suggestion as here we are only comparing timestamps shouldn't we be using something like 
`isRestoredFileUnchanged` or anything on those lines.

Oval17 (commented):
Reviewed this pr, there are three particular test cases which I tested 
1-> delete
2-> restore same file which was deleted
3-> restore different file which is deleted 
and everything seems to be working as expected visually as well
Thanks

Oval17:
Hi @bjorn @djad04 
I have done the review for this PR though I am bit new to reviewing.
Kindly see once ,  I have tested as well this branch and everything seems to be working as expected.

src/tiled/documentmanager.cpp
djad04: yah i forgot those !! those are leftovers from an earlier approach I had because originally I had a 3 steps verification: timestamp → size → hash. The idea was that if the timestamp didn't match, we wouldn't immediately conclude the file had changed we'd still run a hash to check if the file was actually identical to the in-memory version anyway. But I ran into a problem: to do that, the in-memory version of the file had to be serialized into a temp file first before hashing, which made the whole thing heavy and unnecessary since the scenario it was guarding against (someone modifying the file externally, then reverting it to match the exact in-memory version byte-for-byte, then putting it back ) is extremely unlikely in practice, and even if it happened it wouldn't really be a problem since the user would just pick one of the two versions anyway, And I also have a possible update to this feature in mind that would make that check more unnecessary anyway. I think a good addition would be a diff button that opens the two maps side by side so the user can clearly see the differences and then choose whether to override the file or keep their current version.

src/tiled/documentmanager.cpp
djad04: This tolerance is for filesystem timestamp precision differences. During testing I noticed that exact matches sometimes failed, and adding a small offset (around 100 ms) fixed it. After looking into filesystem timestamp resolutions, I found that some filesystems have limited precision, with FAT32 having the worst case at a 2-second resolution. so its there for safety

src/tiled/documentmanager.cpp
djad04: yes you are definitely right document->fileName() is a better choice here.

src/tiled/documentmanager.cpp
djad04: definitely the word ‘identical’ was used originally because the function was meant to verify if the two files were identical. Now that it only checks timestamps I agree a better name would be isRestoredFileUnchanged, and we can maybe merge the timestampMatches function code directly into this function

src/tiled/documentmanager.h
```diff
@@ -248,6 +256,9 @@ public slots:
     QUndoGroup *mUndoGroup;
     FileSystemWatcher *mFileSystemWatcher;
     QHash<QString, Document*> mDocumentByFileName;
+    QSet<Document*> mDeletedDocuments;
+    QSet<Document*> mRecreatedDocuments;
+    QSet<Document*> mRestoringDocuments;
```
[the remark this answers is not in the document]
Oval17: This looks safe!

src/tiled/tabbar.h
```diff
@@ -33,15 +34,25 @@ class TabBar : public QTabBar
...
     void wheelEvent(QWheelEvent *event) override;
+    void paintEvent(QPaintEvent *event) override;
+
+    void tabInserted(int index) override;
+    void tabRemoved(int index) override;
 
 private:
     int mPressedIndex = -1;
+    QSet<int> mDeletedTabs;
+    QSet<int> mRecreatedTabs;
```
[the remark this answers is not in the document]
Oval17: This as well handled!

djad04:
Hey @bjorn , I’ve resolved the conflicts. The only failing check is CodeQL, which shows an authentication error that seems unrelated to my changes.

PR #4320, opened 2026-02-11
Moved 'Erase Terrain' mode into Terrains list

bjorn:
Instead of being a separate button, the erasing mode is now part of the list of terrains. This may be more intuitive.

> Old vs. new:
> 
> image

Came up recently again at https://discourse.mapeditor.org/t/can-i-easily-copy-a-terrain-set-from-one-tmx-to-another/7733/4

### DIFF
--- a/src/tiled/wangdock.cpp
+++ b/src/tiled/wangdock.cpp
@@ -429,11 +429,10 @@ void WangDock::refreshCurrentWangColor()
 void WangDock::wangColorIndexPressed(const QModelIndex &index)
 {
     const int color = index.data(WangColorModel::WangColorIndexRole).toInt();
-    if (!color)
-        return;
-
-    WangColor *currentWangColor = mCurrentWangSet->colorAt(color).data();
-    mDocument->setCurrentObject(currentWangColor, mWangColorModel->tilesetDocument());
+    if (color) {
+        WangColor *currentWangColor = mCurrentWangSet->colorAt(color).data();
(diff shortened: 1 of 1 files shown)


bjorn:
Issues to look into:

* On startup the "erase whole tile" mode is selected. But since Terrains view is the default and not Patterns view, erasing should default to per-index.
* I'm not sure in general if it's helpful to select the erasing mode by default. Previously, the first terrain would be selected.
* Should this entry be called "No Terrain" instead?
* (existing issue) When switching between the Terrains and Patterns views, the editing mode doesn't update.

eishiya:
"Erase Terrain" would ease the transition for old users, but "No Terrain" would probably be better for understanding how the empty terrain really works in the long run. So, I vote "No Terrain" or "Empty Terrain".

I also think this special terrain's name should be styled a little differently, e.g. italics (though that's not good for CJK languages) and/or square brackets. In the forum thread you also mentioned giving it a different icon. I wonder if perhaps the icon should be *both* a dotted outline, and an eraser mini-icon in a corner, to aid those who are expecting an eraser tool rather than a special terrain for this task.

I think the first non-empty terrain should be selected by default, since usually people are looking to start adding terrains, and erasing usually comes after adding in an editing session.

Somewhat related issue: Does the patterns view still only have patterns with non-empty in it? (That's how it was last I checked, don't have time right now to see if it's changed.) If erasing terrains becomes more clearly "just" a terrain, then the lack of patterns with empty becomes even weirder.

PR #4327, opened 2026-02-15
perf(java) optimize image scaling

### LINKED ISSUE #369: libtiled-java should use BufferedImage
libtiled-java use Image instead of BufferedImage because of some old java trick like:

``` java
                     // Deriving a scaled instance, even if it has the same
                        // size, somehow makes drawing of the tiles a lot
                        // faster on various systems (seen on Linux, Windows
                        // and MacOS X).
                        img = img.getScaledInstance(
                                img.getWidth(null), img.getHeight(null),
                                Image.SCALE_FAST);
```

It should not be a performance improvement today: http://today.java.net/pub/a/today/2007/04/03/perils-of-image-getscaledinstance.html

Java image manipulation is easier and faster with BufferedImage, maybe it should be used as preferred image type?

exoego:
Closes #369

Drawing image `sewer_tileset.png` with `MapObject.getImage`:
Case | Before | After | Perf gain
---|--|--|---
50% scaling | 0.819 ms | 0.064 ms | x13
same size | 0.739 ms | 0.021 ms | x35

Note) The original `ImageHelper.bytesToImage()` used `Toolkit.createImage()`, which decodes images asynchronously.
So `getWidth(null)` and `getHeight(null)` can return `-1` before loading completes.
It will cause `IllegalArgumentException` when constructing the `BufferedImage`.
I think the `same-size getScaledInstance(SCALE_FAST)` workaround did not fix this race condition.
So I replaced it with `ImageIO.read(new ByteArrayInputStream(imageData))`, which decodes synchronously and returns a `BufferedImage` directly.
It ensures dimensions are always valid.

### DIFF
--- a/util/java/libtiled-java/src/test/java/org/mapeditor/util/ImageHelperTest.java
+++ b/util/java/libtiled-java/src/test/java/org/mapeditor/util/ImageHelperTest.java
@@ -72,7 +72,7 @@ public class ImageHelperTest {
 
         byte[] pngData = ImageHelper.imageToPNG(img);
         assertNotNull(pngData);
-        assertEquals(pngData.length, 24559);
+        assertTrue(pngData.length > 20000);
 
         BufferedImage roundTripped = ImageHelper.bytesToImage(pngData);
         assertNotNull(roundTripped);


util/java/libtiled-java/src/test/java/org/mapeditor/core/MapObjectTest.java
```diff
@@ -0,0 +1,100 @@
+/*-
+ * #%L
+ * libtiled
+ * %%
+ * Copyright (C) 2004 - 2026 Thorbjørn Lindeijer <thorbjorn@lindeijer.nl>
```
bjorn, line 5: These files seem to be entirely written by you, so I would suggest you put your name here.

bjorn (commented):
Looks like a great improvement!

I wonder if these days we even want to be creating scaled instances of the tiles. It was a way to keep viewport updates fast in the Java version of Tiled even with the viewport scaled, because the scaled versions of tiles were cached. I guess it may still be a noticeable optimization when using software rendering.

util/java/libtiled-java/src/test/java/org/mapeditor/core/MapObjectTest.java
exoego: Thanks, but I don't mind my copyright things here.
I'm fine to present the copyright of these new files to this great project and the original creator.

bjorn: Alright, thanks for the kind words! And of course git blame will still tell us who wrote what in the end. :-)

PR #4330, opened 2026-02-16
Scripting: Added MapEditor.selectedTool and MapEditor.tool

bjorn:
Enables scripts to know which tool is selected, to select a different tool and potentially to modify or interact with the available tools.


eishiya:
It would be nice if there was consistency regarding "id" vs "shortName" for the tool identifier string xP

Issue #4295, opened 2025-12-04
Spacebar doesn't work on first drag after clicking on other elements


RogerLevy:
### Describe the bug

Buggy spacebar detection leads to constant editing mistakes.

### Steps to reproduce

1. Open/Create a map with several layers
2. Edit a layer, pan around with spacebar
3. Select a different layer
4. Hold spacebar and try to pan the tilemap.  It will use the current tool instead of panning.  Try again and it works the second time.  

### Expected behavior

Tiled should be detecting the spacebar at a global level, so that the first time you click in the editing area while holdign it, the view will pan as expected.

### OS version

Windows 10

### Tiled version

1.11.2

Oval17:
Hi @RogerLevy @bjorn  , I am able to replicate this issue.

Root cause Analysis led me to this 

`SpaceBarEventFilter is installed only on MainWindow. Key events go to the focused widget, not MainWindow. When the Layers dock (a QTreeView) has focus, it consumes the Space key itself (to expand/collapse nodes). The event never propagates to MainWindow → mSpacePressed stays false → first click uses the tool instead of panning. After that click, MapView gets focus — it explicitly ignores Space (e->ignore()) so the event bubbles up to MainWindow and works from then on.`

Fix: Install the filter on qApp (application-level). Qt routes all application-level filters before any widget processes an event, so it catches Space regardless of which widget has focus. Add a guard to skip text-input widgets (QLineEdit etc.) so typing spaces doesn't accidentally enable pan mode.
Have a raised a pr as well for same...

Oval17:
Hi @bjorn , 
This issue can be assigned to me , I have already raised a MR for this, 
Kindly have look as well on that.
Thanks

PR #4338, opened 2026-02-20
Fix spacebar panning not working after focus leaves the map view

Oval17:
Resolves : #4295 
Fixes: spacebar pan fails on first click after focus leaves the map view

SpaceBarEventFilter was installed on MainWindow, so it only saw Space key events that propagated up to the main window. Widgets like the Layers dock tree view consume Space themselves (to expand/collapse nodes), so the filter never saw the press and mSpacePressed stayed false. The next click in the map then activated the current tool instead of panning.

Fix: install the filter on qApp so it receives all key events in the application regardless of focus. Added a guard to skip events sent to text-input widgets (WA_InputMethodEnabled) so typing spaces in QLineEdit/QPlainTextEdit fields doesn't accidentally enable pan mode.

Kindly review and have a look @bjorn
Thanks

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -23,6 +23,7 @@
 * Added "World > World Properties" menu action (with dogboydog, #4190)
 * Added Delete shortcut to Remove Tiles action by default and avoid ambiguity (#4201)
 * Fixed alpha component of tint color not applying correctly to opaque images (by Roland Helmerichs, #4310)
+* Fixed panning with space bar not always working on first click (with Oval, #4338)
 * Scripting: Added API for custom property types (with dogboydog, #3971)
 * Scripting: Added TileMap.chunkSize and TileMap.compressionLevel properties
 * AutoMapping: Don't match rules based on empty input indexes


src/tiled/pannableviewhelper.cpp
```diff
@@ -53,16 +53,24 @@ class SpaceBarEventFilter : public QObject
...
         case QEvent::KeyPress:
         case QEvent::KeyRelease: {
             auto keyEvent = static_cast<QKeyEvent*>(event);
             if (keyEvent->key() == Qt::Key_Space && !keyEvent->isAutoRepeat()) {
+                // Don't intercept Space typed into text-input widgets such as
+                // QLineEdit or QPlainTextEdit (identified by WA_InputMethodEnabled).
+                if (auto *widget = qobject_cast<QWidget*>(watched)) {
+                    if (widget->testAttribute(Qt::WA_InputMethodEnabled))
+                        break;
+                }
```
bjorn: I don't think we need this exception. In fact it won't have any effect because the event goes to the window first, before being sent to the input widget.

bjorn:
@Oval17 While investigating this fix I had already come up with an alternative approach which I have just pushed. Could you review and test my change now and let me know if everything seems fine? :-)

Oval17:
Thanks for looking into this more deeply! I had gone through the Qt event handling docs and explored a few approaches before settling on the `qApp` filter and it worked.

But damnn Your windowHandle() approach is much cleaner and simpler. Intercepting at the native window level before dispatch to any widget is the right place to do it, and the WinIdChange handling for the deferred registration is a neat solution. 

Thanks for this improvement! 🫡

bjorn:
Alright, merged after I've verified that this also works in Qt 5. :-)

PR #4339, opened 2026-02-20
Fix TransformMapObjects merge behavior for mixed property updates

### LINKED ISSUE #4332: Resize operation causes many undo steps when it affects position
### Describe the bug

This is a regression. I'm not yet sure since which Tiled version, but it affects 1.11.2 and master.

When resizing an object in a direction that affects its position, the undo commands are not merged, resulting in many undo steps to undo the resizing operation.

### Steps to reproduce

Resize an object from a handle that also will change the position of the object.

### Expected behavior

I expect to be able to undo the resize in one step.

### OS version

Linux

### Tiled version

1.11.2 and master.

kanishka0411:
Fixes #4332 
Fixed ```TransformMapObjects::mergeWith``` in ```[transformmapobjects.cpp]``` to merge undo commands only when they belong to the same dominant gesture type (move, resize/shape, or rotate), while still accumulating changed properties after a valid merge.


src/tiled/transformmapobjects.cpp
```diff
@@ -99,12 +99,26 @@ void TransformMapObjects::redo()
...
+            return 0;
+        if (p & (MapObject::SizeProperty | MapObject::ShapeProperty))
+            return 1;
+        if (p & MapObject::PositionProperty)
+            return 2;
+        return 3;
+    };
+
+    if (dominantType(mChangedProperties) != dominantType(o->mChangedProperties))
+        return false;
```
bjorn, line 113: I think this is pretty clever because it still treats rotating, resizing and moving as separate operations, regardless of whether they also include position updates.

But I think we could simplify it further by saying we need the changed properties to be the same regardless of the `PositionProperty` state:

```suggestion
    auto nonPositionChanges = [](MapObject::ChangedProperties p) {
        return p & ~MapObject::PositionProperty;
    };

    if (nonPositionChanges(mChangedProperties) != nonPositionChanges(o->mChangedProperties))
        return false;
```

This should work because:
* We don't need size/shape changes as part of rotation to be merged (it also should never happen).
* We will always have either size, shape or size+shape flags set in a single operation, depending on whether we're resizing no polygons, only polygons or polygons together with other objects.

bjorn (commented):
Please consider my suggestion for simplifying this somewhat. You can apply it when you agree or provide feedback.

While testing this change I realized a related issue: in some cases I think we're merging too much. When I drag and release a resize handle, I expect that to become a single undo operation. But currently it might merge with the last change if I change the same object(s) with the same kind of operation.

I think the above is something that needs to be addressed in the tools, since they know when the mouse is pressed/released. I think it would involve:

* Adding a `mergeable` bool to the `TransformMapObjects` command.
* Setting this bool from `EditPolygonTool` and `ObjectSelectionTool` and checking it in `mergeWith`.

There is precedence for this approach in undo commands like `PaintTileLayer`.

Is this something you'd like to address, either as part of this change or a future PR?

kanishka0411:
Thanks for the review. I agree on the nonPositionChanges approach it’s much simpler and cleaner, so I'll get that updated.I can also fix the over-merging issue with the drag gestures while I'm at it. I'll just add a mergeable flag to ```TransformMapObjects``` and trigger it from ```EditPolygonTool ```and ```ObjectSelectionTool ```(just like we do in PaintTileLayer).

kanishka0411:
Updated the changes-Please take a look when you get a chance

bjorn (approved):
Looks good, thanks!

I just had to change the default to `true` to preserve the existing merging behavior when changing size/position/rotation in the Properties view. :-)

PR #4342, opened 2026-02-20
Add optional label support for maps in worlds

### LINKED ISSUE #4298: Add optional labels to maps in worlds

When I'm creating levels for a game, it is usually done linearly. Sometimes though, I might find that level 3 works better as level 2.

I then reorder them in the world file so I can visually see this but this isn't clear anywhere else.

This reordering works however I can't visually tell anymore than level 3 is now level 2 and it's something I have to remember myself.

If there was kind of optional label above each, it would reduce the cognitive load.
Something like this:

Image

My suggestion is to add  `name` field to the map. This would be consistent with how objects currently display their names while in the map

_No response_

Oval17:
Maps in a world file can now have an optional label. The label is stored in the world JSON file under the "label" key and displayed above the map in the world view. It can be set via right-click context menu ("Set Label...") with full undo/redo support.

Resolves #4298.

Previously We didnt have any add label option 
Now we can add, edit and remove labels 

Screenshot 2026-02-20 at 11 52 49 PM
Screenshot 2026-02-20 at 11 53 12 PM

### DIFF
--- a/src/tiled/mapitem.cpp
+++ b/src/tiled/mapitem.cpp
@@ -236,9 +236,8 @@ MapItem::MapItem(const MapDocumentPtr &mapDocument, DisplayMode displayMode,
     connect(prefs, &Preferences::highlightCurrentLayerChanged, this, &MapItem::updateSelectedLayersHighlight);
     connect(prefs, &Preferences::propertyTypesChanged, this, &MapItem::syncAllObjectItems);
     connect(prefs, &Preferences::backgroundFadeColorChanged, this, [this] (QColor color) { mDarkRectangle->setBrush(color); });
-    connect(prefs, &Preferences::showMapLabelsChanged, this, &MapItem::updateLabel);
-
     connect(mapDocument.data(), &Document::changed, this, &MapItem::documentChanged);
+    connect(mapDocument.data(), &Document::fileNameChanged, this, &MapItem::updateLabel);
     connect(mapDocument.data(), &MapDocument::mapResized, this, &MapItem::mapChanged);
     connect(mapDocument.data(), &MapDocument::regionChanged, this, &MapItem::repaintRegion);
     connect(mapDocument.data(), &MapDocument::tileLayerChanged, this, &MapItem::tileLayerChanged);
@@ -275,17 +274,6 @@ MapItem::MapItem(const MapDocumentPtr &mapDocument, DisplayMode displayMode,
     // Label shown above each map in the world view
     mLabelItem = new MapLabelItem(this);
     mLabelItem->setZValue(10001);
-
-    // Update label whenever any loaded world changes
-    for (auto &worldDoc : WorldManager::instance().worlds())
-        connect(worldDoc.data(), &WorldDocument::worldChanged, this, &MapItem::updateLabel);
-    connect(&WorldManager::instance(), &WorldManager::worldLoaded,
-            this, [this](WorldDocument *worldDoc) {
-        connect(worldDoc, &WorldDocument::worldChanged, this, &MapItem::updateLabel);
-        updateLabel();
-    });
-
-    updateLabel();
 
     if (displayMode == ReadOnly) {
         setDisplayMode(displayMode);
@@ -957,24 +945,19 @@ void MapItem::updateLabel()
     if (!mLabelItem)
         return;
 
-    const QString &mapFileName = mMapDocument->fileName();
-    auto worldDoc = WorldManager::instance().worldForMap(mapFileName);
-
-    // Only show labels for maps that are part of a world, and when enabled
-    if (!worldDoc || !Preferences::instance()->showMapLabels()) {
-        mLabelItem->setVisible(false);
-        return;
-    }
-
     // Derive the label from the map's file name (without extension)
-    const QString label = QFileInfo(mapFileName).baseName();
-
+    const QString label = QFileInfo(mMapDocument->fileName()).baseName();
     mLabelItem->setText(label);
 
-    // Position centered at the top of the map boundary
-    const QRect border = mapDocument()->renderer()->mapBoundingRect();
+    // Position centered at the top of the map tile boundary
+    const QRectF border = mBorderRectangle->rect();
     mLabelItem->setPos(QPointF(border.center().x(), border.top()));
-    mLabelItem->setVisible(true);
+}
+
+void MapItem::setLabelVisible(bool visible)
+{
+    if (mLabelItem)
+        mLabelItem->setVisible(visible);
 }
 
 void MapItem::updateBoundingRect()
--- a/src/tiled/mapitem.h
+++ b/src/tiled/mapitem.h
@@ -76,6 +76,7 @@ public:
     void setShowTileCollisionShapes(bool enabled);
 
     void updateLayerPositions();
+    void setLabelVisible(bool visible);
 
     // QGraphicsItem
     QRectF boundingRect() const override;
--- a/src/tiled/mapscene.cpp
+++ b/src/tiled/mapscene.cpp
@@ -28,6 +28,7 @@
 #include "containerhelpers.h"
 #include "debugdrawitem.h"
 #include "documentmanager.h"
+#include "preferences.h"
 #include "map.h"
 #include "mapobject.h"
 #include "maprenderer.h"
@@ -69,6 +70,9 @@ MapScene::MapScene(QObject *parent)
 
     WorldManager &worldManager = WorldManager::instance();
     connect(&worldManager, &WorldManager::worldsChanged, this, &MapScene::refreshScene);
+
+    Preferences *prefs = Preferences::instance();
+    connect(prefs, &Preferences::showMapLabelsChanged, this, &MapScene::updateMapLabels);
 
     // Install an event filter so that we can get key events on behalf of the
     // active tool without having to have the current focus.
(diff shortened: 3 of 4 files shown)


joereynolds:
Looks good but I have a few questions/points

- What does it look like with a more populated map? The text sits inside the map currently so if the map was full, the label would be hard to read or perhaps completely illegible
- I have 100+ maps and counting for [my game](https://store.steampowered.com/app/3122220/Mr_Figs). This would mean that I'd have to go through each of these and manually set a label. In my head the label was automatically inferred from the filename and then you'd have a "name" or "label" or perhaps "world label" map property on the map properties section which you could use as an override. Thoughts?

This is looking good and is a good start but I'm unsure of the UX and I don't really have any particular weight in that coversation. I think Bjorn would be a better answer on that direction

bjorn:
Thanks @joereynolds for your feedback! I agree entirely. Even if it's nice that the user can pick their own names, these labels should derive from the map's file name by default. And I think for readability they'll need a background color.

I'm not sure the `QGraphicsSimpleTextItem` will have allow enough customization. Alternatively, have a look at the `MapObjectLabel` implementation, because I think these look generally quite nice. You can derive a `MapLabel` item from those and adjust their positioning code. We can also consider moving these to their own file and have a common `LabelItem` superclass for them to share some code.

Btw, I think the labels should not be stored in the world. I think they belong in each map, like the names of tilesets.

Oval17:
Thanks @joereynolds @bjorn for all the inputs...
1 - > Regarding populated maps, max I tested was with 5 will test and show the result here regarding how does it actually look with 50-60 or something like that.
2 -> Got the point regarding setting label for all of them (first of all I didn't thought about 100+ maps 😅) so obv its very tedious to mark all of them, rest as @bjorn have suggested few things will implement them once and share test results here.
Thanks again!!

Oval17:
Hi @joereynolds @bjorn 
So after testing all the options this sort of seemed middle ground..
How it looks like when a lot we have 50-100+ maps so previously transparent was not visible 
this is how it look now 
Screenshot 2026-02-24 at 9 58 26 PM

Now addressing all the inputs one by one : 
1-> Labels should derive from filename by default : done
2-> Labels need a background color for readability(from observation) : done
3 -> QGraphicsSimpleTextItem shoulnt be usd & Labels should NOT be stored in the world : done

now following I didnt do  : 
1 -> Move label items to their own file as MapLabelItem still lives inline in mapitem.cpp
2 -> Allowing custom label override (via map property)
3 -> setting custom label 

see this now and do give inputs !!

Oval17:
Imp thing so from UX pov this seemed the only best way to have option labels
to have toggle from view 
Screenshot 2026-02-24 at 10 24 56 PM
when selected it shows label for all the maps and can be toggled off

@bjorn @joereynolds

joereynolds:
Looks way better, nice work!

I wouldn't ever expect the maps to overlap like that so doubt it's something to worry about.
From my perspective, I'm happy (it was my request after all heh)

I do have _one_ question. How does it handle maps in other directories?

For example if we have the following

```
dir_1/map1.tmx
dir_2/map1.tmx
```

Will they both be rendered as "map1"?

Not a huge issue and to me it feels like something we can fix after the fact but thought I'd ask.

Good job!

Oval17:
Yes, both would render as "map1" since we use `QFileInfo::baseName()` which only returns the filename without the extension, ignoring the path.
This may be handled to show the relative path from the world file when a name collision exists, e.g. "dir_1/map1" vs "dir_2/map1". Or always show one level of parent directory for disambiguation. 
discussion on this will be required...

joereynolds:
Fair enough, it's not a huge deal and is definitely an edge case.
Personally I'd just leave it but if you wanted to bulletproof it I'd make it so that if the label name ends up being ambiguous (i.e. multiple labels have the same name) then you go up each directory until it's no longer ambiguous and use that as the label

bjorn:
I think that's really a separate feature. We also have the same problem on the open document tabs. So I think we can polish this up and consider the disambiguation of such cases in a future improvement.

The labels look great now, I'll try to get around to reviewing this tomorrow, but there's a lot of active threads now so please be patient. :)

src/tiled/mapitem.cpp
```diff
@@ -188,6 +272,21 @@ MapItem::MapItem(const MapDocumentPtr &mapDocument, DisplayMode displayMode,
 
     mBorderRectangle->setZValue(10000 - 3);
 
+    // Label shown above each map in the world view
+    mLabelItem = new MapLabelItem(this);
+    mLabelItem->setZValue(10001);
+
+    // Update label whenever any loaded world changes
```
bjorn: This seems to me a little inefficient. Like this, each `MapItem` will look up the world again just to determine whether the label should be visible, and that on any world change.

This needs to be moved to the `MapScene`, which already determined the relevant world in `MapScene::refreshScene` and as such could tell its `MapItem` instances whether they should show their name label.

src/tiled/mapitem.cpp
```diff
@@ -153,6 +236,7 @@ MapItem::MapItem(const MapDocumentPtr &mapDocument, DisplayMode displayMode,
     connect(prefs, &Preferences::highlightCurrentLayerChanged, this, &MapItem::updateSelectedLayersHighlight);
     connect(prefs, &Preferences::propertyTypesChanged, this, &MapItem::syncAllObjectItems);
     connect(prefs, &Preferences::backgroundFadeColorChanged, this, [this] (QColor color) { mDarkRectangle->setBrush(color); });
+    connect(prefs, &Preferences::showMapLabelsChanged, this, &MapItem::updateLabel);
```
bjorn: In light of my other comment about connecting `MapItem` to `worldChanged`, I think this connection should probably also be moved to `MapScene`. It might be good to split a `setLabelVisible` function out of `updateLabel`.

src/tiled/mapitem.cpp
```diff
@@ -853,6 +952,31 @@ void MapItem::deleteLayerItems(Layer *layer)
...
+    auto worldDoc = WorldManager::instance().worldForMap(mapFileName);
+
+    // Only show labels for maps that are part of a world, and when enabled
+    if (!worldDoc || !Preferences::instance()->showMapLabels()) {
+        mLabelItem->setVisible(false);
+        return;
+    }
+
+    // Derive the label from the map's file name (without extension)
+    const QString label = QFileInfo(mapFileName).baseName();
```
bjorn: I think we need a connection to `Document::fileNameChanged` (emitted by `mMapDocument`) to update the text on the label.

src/tiled/mapitem.cpp
```diff
@@ -853,6 +952,31 @@ void MapItem::deleteLayerItems(Layer *layer)
...
+
+    mLabelItem->setText(label);
+
+    // Position centered at the top of the map boundary
+    const QRect border = mapDocument()->renderer()->mapBoundingRect();
```
bjorn: Could we just use `mBoundingRect` here instead of calculating it again? I realize it's been through `adjustBoundingRectForOffsetsAndImageLayers`, but not sure if this is good or bad.

Oval17: `updateLabel()` now uses `mBorderRectangle->rect()` instead of recalling `renderer()>mapBoundingRect()` the border rectangle is already set to the unadjusted map boundary, which is the right reference for label positioning

src/tiled/mapitem.cpp
Oval17: Added connect(mapDocument, & `Document::fileNameChanged`, this, & `MapItem::updateLabel`) so label text updates when a map is renamed

src/tiled/mapitem.cpp
Oval17: Have include above

src/tiled/mapitem.cpp
Oval17: - Removed `worldChanged` connections from `MapItem` constructor entirely
- Removed `showMapLabelsChanged` connection from `MapItem`
- Added `setLabelVisible(bool)` as a public method on `MapItem`
- Added `updateMapLabels()` to MapScene it does the single world lookup for the whole scene and calls `setLabelVisible` on each MapItem
- `updateMapLabels()` is called from `refreshScene()` and connected to showMapLabelsChanged

Oval17:
Hi, @bjorn 
Have addressed comments, 
let me explore a bit on part where of to see labels when hovering over map.
Thankss!

src/tiled/mapitem.h
```diff
@@ -133,10 +135,12 @@ class MapItem : public QGraphicsObject
 
     void updateBoundingRect();
     void updateSelectedLayersHighlight();
+    void updateLabel();
 
     MapDocumentPtr mMapDocument;
     QGraphicsRectItem *mDarkRectangle;
     QGraphicsRectItem *mBorderRectangle;
+    MapLabelItem *mLabelItem = nullptr;
```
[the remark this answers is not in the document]
bjorn: Wrong, the `MapLabelItem` will get cleaned up by its parent `MapItem` instance.

src/tiled/mapitem.cpp
```diff
@@ -853,6 +940,26 @@ void MapItem::deleteLayerItems(Layer *layer)
...
+        return;
+
+    // Derive the label from the map's file name (without extension)
+    const QString label = QFileInfo(mMapDocument->fileName()).baseName();
+    mLabelItem->setText(label);
+
+    // Position centered at the top of the map tile boundary
+    const QRectF border = mBorderRectangle->rect();
+    mLabelItem->setPos(QPointF(border.center().x(), border.top()));
+}
```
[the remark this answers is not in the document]
bjorn: Wrong, the ability to specify a map name in the world file was removed intentionally.

Issue #4222, opened 2025-08-11
Scripting: Access user's tile size preferences


eishiya:
When creating Tilesets and TileMaps, Tiled saves the tile sizes last used by the user. It would be nice if scripting had access to these values, so that scripted tileset/tilemap creation could offer these as defaults like the built-in dialogs do. Read-only would be fine, editable even better.

This would allow scripts to be consistent with the user's preferences elsewhere, and avoid the need to redundantly set and store these preferences, and making them writable would allow scripted dialogs to be fully interchangeable with the built-in ones without the UX inconsistencies that would come from only the default dialogs being able to write these preferences.

Perhaps these could be exposed alongside various other preferences (last used map orientation, export settings, snapping, etc) in a `tiled.preferences`, accessible as e.g. `tiled.preferences.mapTileWidth`.

I'd like to use this for my RPG Maker expander and Tile Atlas scripts. The former currently has its own default tile size and saves the last used values in the user's Project, and I'd prefer to avoid repeating that approach for the latter.

kanishka0411:
Hi @bjorn, I’d like to work on this issue

bjorn:
@kanishka0411 You're welcome to have a look into this, thanks!

Note that many preferences could be exposed to scripting by defining them as `Q_PROPERTY` on the `Preferences` class and exposing that as `tiled.preferences` as suggested by @eishiya. However, that currently does not include tile size preferences (because it generally only includes preferences for which parts of Tiled need to know immediately when their value changes, or things which don't have a clear other place in the code).

Because the last used tile size is only used in the New Map dialog, you'll find those values defined in `src/tiled/newmapdialog.cpp`. There it also becomes clear that these are stored in the _session_ rather than the preferences (because one user may use different values for different projects).

I think in the case of the session, we can still expose it to scripts like `tiled.session`, which would return a wrapping class derived from `QObject` that allows getting/setting session variables by their string key as well as providing access to the file states (see `src/tiled/session.h`).

I'll assign the issue to you. If you have any questions it's best to get into contact on Discord but you can ask questions related to this issue here of course.

kanishka0411:
Thanks! I’ll implement tiled.session for scripting with key-based get/set access, including the New Map last-used tile size values and session file states. I’ll ask on Discord if API details come up, then open a draft PR.

kanishka0411:
Hi @bjorn , I’ve opened a PR for this issue.Please have a look:-)

PR #4343, opened 2026-02-21
feat: implement scripting API access to user's tile size preferences

abhinavs1920:
Resolves Issue #4222 

This PR exposes user tile size preferences to the scripting API, allowing scripts to read and modify the default tile dimensions used in the New Map and New Tileset dialogs.

Scripts can now access these preferences through tiled.preferences:

mapTileWidth / mapTileHeight - default tile size for new maps
tilesetTileWidth / tilesetTileHeight - default tile size for new tilesets
The implementation adds a new ScriptPreferences class that wraps the existing session options, making them accessible to the scripting engine. Session options that were previously scoped locally in dialog files have been moved to a shared header (sessionoptions.h) so they can be accessed from the scripting layer.

This is useful for scripts that want to create maps or tilesets programmatically using the user's preferred tile dimensions, or for scripts that need to temporarily adjust these defaults.

I am also adding the screenshot to show successful working of the feature.

image


abhinavs1920:
Hey @bjorn, I’ve added the PR. I just realized someone else has been assigned to it, if it’s not mergeable, I’d really appreciate any feedback. Thanks!

bjorn:
> I just realized someone else has been assigned to it, if it’s not mergeable, I’d really appreciate any feedback. Thanks!

Yeah, I tried to assign issues to avoid duplicate work like this, though on the other hand I wouldn't ask people to wait for my OK to start working on anything. But in this case, you missed my guidance towards the solution and have implemented it directly as suggested in the original issue. This has some downsides:

* It means only some of the session data is available in scripts.
* It creates confusion because we also have actual "preferences", which are global session-independent user settings.

As such I think you'll understand that I'll close this version in favor of #4345.

If you want, you could make a new PR that exposes the actual `Preferences` class as `tiled.preferences` in the scripting API, adding the necessary `Q_PROPERTY` definitions and scripting API documentation.

abhinavs1920:
Got it, that makes sense. I’ll rework this following the suggested approach and open a new PR. Thanks for the feedback!

PR #4344, opened 2026-02-21
macOS: Add Cmd+Shift+[ and Cmd+Shift+] shortcuts to switch tabs

### LINKED ISSUE #4038: Configurable hotkey to switch tabs

I'm trying to switch tabs on MacOS with system hotkeys, `Cmd-Shift-[` and `Cmd-Shift-]`, and I cannot.

I cannot find the function is custom keybindings.

I'd like either:
* Make the function hotkey customizable;
* Or, if it's hardcoded, I'd love it to accept the system hotkeys that switch tabs too.

* `Opt-Letf/Right` works, thought, but it took a while to discover and adjust.
* Ctrl-Tab doesn't switch them, and Cmd-Tab is doing a system global window switch.

Oval17:
These are the standard macOS system shortcuts for switching tabs (used in Safari, Terminal, VS Code, etc.) and are now supported in addition to the existing Alt+Left/Right and Ctrl+Tab shortcuts.

Fixes #4038.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -28,6 +28,7 @@
 * AutoMapping: Don't match rules based on empty input indexes
 * AutoMapping: Optimized reloading of rule maps and load rule maps on-demand
 * Windows: Fixed issue with opening Tile Animation Editor (#4223)
+* macOS: Add Cmd+Shift+[ and Cmd+Shift+] shortcuts to switch tabs (by Oval, #4344)
 * macOS: Fixed crash when JS code is JIT-compiled (#4218)
 * Workaround tileset view layout regression in Qt 6.9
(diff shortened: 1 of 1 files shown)


bjorn:
Since in #4038 it is mentioned that the <kbd>Ctrl+Tab</kbd> and <kbd>Ctrl+Shift+Tab</kbd> shortcuts do not work on macOS (cause they're used by the system), I would suggest we use the `#ifdef` to set these macOS shortcuts on `switchToLeftDocument1` and `switchToRightDocument1` instead of introducing a third shortcut on these actions. Thanks!

Oval17:
Done @bjorn !!

bjorn:
Thanks @Oval17!

PR #4345, opened 2026-02-21
Expose session API to scripting as tiled.session

### LINKED ISSUE #4222: Scripting: Access user's tile size preferences

kanishka0411:
Fixes #4222 

Implemented tiled.session for scripting so scripts can read/write session values by key and access per-file state.
This lets scripted map/tileset creation use the same remembered values as Tiled dialogs, and includes safe behavior in --evaluate mode when no session is active.

### DIFF
--- a/docs/scripting-doc/index.d.ts
+++ b/docs/scripting-doc/index.d.ts
@@ -4632,6 +4632,53 @@ declare namespace Tiled {
   const Zstandard: CompressionMethod;
 }
 
+
+/**
+ * Provides access to session-specific settings. The session stores
+ * per-project preferences like last-used tile sizes, open files, and
+ * file states.
+ *
+ * Values are accessed by their string key (e.g. `"map.tileWidth"`).
+ *
+ * @since 1.12
+ */
+interface Session {
+  /**
+   * Path to the current session file.
+   */
+  readonly fileName: string;
+
+  /**
+   * Returns the session value for the given key, or `defaultValue` if
+   * the key is not set.
+   */
+  get(key: string, defaultValue?: any): any;
+
+  /**
+   * Sets the session value for the given key.
+   */
+  set(key: string, value: any): void;
+
+  /**
+   * Returns whether the given key is present in the session.
+   */
+  isSet(key: string): boolean;
+
+  /**
+   * Returns the per-file state for the given file name.
+   */
+  fileState(fileName: string): { [key: string]: any };
+
+  /**
+   * Sets the per-file state for the given file name.
+   */
+  setFileState(fileName: string, fileState: { [key: string]: any }): void;
+
+  /**
+   * Sets a single value in the per-file state for the given file name.
+   */
+  setFileStateValue(fileName: string, name: string, value: any): void;
+}
 /**
  * The `tiled` module is the main entry point and provides properties,
  * functions and signals which are documented below.
@@ -4766,6 +4813,18 @@ declare namespace tiled {
    * Access the editor used when editing tilesets.
    */
   export const tilesetEditor: TilesetEditor;
+
+  /**
+   * Provides access to the current session, allowing scripts to read and
+   * write session-specific settings like last-used tile size.
+   *
+   * In command-line mode (`--evaluate`), no session may be available. In
+   * that case, methods return safe defaults (empty strings, default
+   * values, etc.).
+   *
+   * @since 1.12
+   */
+  export const session: Session;
 
   /**
    * This function can be used to trigger any registered action. This
--- a/docs/scripting-doc/index.d.ts
+++ b/docs/scripting-doc/index.d.ts
@@ -415,7 +415,7 @@ declare namespace Qt {
    * Can be created with {@link tiled.cursor} and assigned to {@link
    * Tool.cursor}.
    */
-  class QCursor {}
+  class QCursor { }
 
   /**
    * The base type from which all Qt widgets derive.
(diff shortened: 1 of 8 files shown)


docs/scripting-doc/index.d.ts
```diff
@@ -4632,6 +4632,53 @@ declare namespace Tiled {
...
+  /**
+   * Path to the current session file.
+   */
+  readonly fileName: string;
+
+  /**
+   * Returns the session value for the given key, or `defaultValue` if
+   * the key is not set.
+   */
+  get(key: string, defaultValue?: any): any;
```
bjorn, line 4695: I think it's a bit unfortunate that we couldn't use the default value defined in the `SessionOption` instance, but this would require those instances to register their default values somewhere so I think it's fine to leave it like this.

docs/scripting-doc/index.d.ts
```diff
@@ -4632,6 +4632,53 @@ declare namespace Tiled {
...
+   */
+  isSet(key: string): boolean;
+
+  /**
+   * Returns the per-file state for the given file name.
+   */
+  fileState(fileName: string): { [key: string]: any };
+
+  /**
+   * Sets the per-file state for the given file name.
```
bjorn, line 4713: I think the docs for this function could use a warning that it would replace the entire state, with a reference to `setFileStateValue` to set only a single value in the state.

src/tiled/scriptmodule.h
```diff
@@ -38,6 +38,7 @@ class Document;
 class EditableAsset;
 class MapEditor;
 class ScriptImage;
+class ScriptSession;
 class ScriptMapFormatWrapper;
```
bjorn: Let's keep it alphabetical. :-)
```suggestion
class ScriptMapFormatWrapper;
class ScriptSession;
```

src/tiled/scriptmodule.h
```diff
@@ -75,6 +76,8 @@ class ScriptModule : public QObject
     Q_PROPERTY(Tiled::TilesetEditor *tilesetEditor READ tilesetEditor)
     Q_PROPERTY(QList<QObject*> worlds READ worlds)
 
+    Q_PROPERTY(Tiled::ScriptSession *session READ session)
```
bjorn: Please group this property together with the `project` property, since it is somewhat related.

src/tiled/scriptsession.cpp
bjorn, line 1: Please add the copyright header to new files. Can be based on the `.LICENSE-HEADER` file, filling in the file name, year and your name or nickname.

src/tiled/scriptsession.h
```diff
@@ -0,0 +1,40 @@
...
+
+    Q_INVOKABLE QVariantMap fileState(const QString &fileName) const;
+    Q_INVOKABLE void setFileState(const QString &fileName,
+                                  const QVariantMap &fileState);
+    Q_INVOKABLE void setFileStateValue(const QString &fileName,
+                                       const QString &name,
+                                       const QVariant &value);
+};
+
+}
```
bjorn:
```suggestion
} // namespace Tiled
```

src/tiled/scriptsession.h
```diff
@@ -0,0 +1,40 @@
...
+
+    Q_PROPERTY(QString fileName READ fileName)
+
+public:
+    explicit ScriptSession(QObject *parent = nullptr);
+
+    QString fileName() const;
+
+    Q_INVOKABLE QVariant get(const QString &key,
+                              const QVariant &defaultValue = QVariant()) const;
```
bjorn:
```suggestion
    Q_INVOKABLE QVariant get(const QString &key,
                             const QVariant &defaultValue = QVariant()) const;
```

src/tiled/scriptsession.cpp
```diff
@@ -0,0 +1,73 @@
...
+    if (!Session::hasCurrent())
+        return defaultValue;
+
+    const QByteArray latin1Key = key.toLatin1();
+    auto &session = Session::current();
+
+    if (!session.isSet(latin1Key.constData()))
+        return defaultValue;
+
+    return session.get<QVariant>(latin1Key.constData());
```
bjorn: Could we pass the `defaultValue` into the `get` function instead of handling the default explicitly here?

bjorn (changes requested):
I think this is pretty good! I only left some minor feedback I'd ask you to address. Also, please add an entry to the `NEWS.md` file along with the other scripting related additions, like:
```md
* Scripting: Added tiled.session to read and write session properties (by Kanishka, #4345)
```

It would also be good to document what these session properties are, since they are otherwise only visible by looking through the source code. Maybe we can include them as a Markdown table in the JS API docs for the `Session` interface, showing the string and type for each?

kanishka0411:
Thanks for the review, I’ll make all the requested changes and update the PR shortly:-)

kanishka0411:
I've addressed all the requested changes. Please have a look!

bjorn (approved):
That looks great, thanks for this addition!

PR #4346, opened 2026-02-21
Project Dock: add collapse all and only expand to current actions

rhythmcache:
This PR adds Collapse all and only exapand to current actions buttons to the project dock.

I picked up this feature request  ( https://github.com/mapeditor/tiled/issues/4306 ) and implemented both of the suggested options. 
There are now two buttons in the project dock toolbar a collapse all button that instantly collapses every expanded folder, and an Only Expand to current toggle that when enabled automatically collapses everything except the path leading to whichever file is currently active in the editor.

when the toggle is turned on the previous tree state gets snapshotted so that turning it back off restores exactly what we had before. 

i left the keyboard shortcut for collapse all unassigned since i wasn't sure. happy to add one if preferred. The icons are using existing resources as placeholders

image

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 ### Unreleased
 
-* Added 'Collapse All' action and 'Only Expand to Current' mode to Project view (with rhythmcache, #4306)
+* Added 'Collapse All' action and 'Only Expand to Current' mode to Project view (with rhythmcache, #4346)
 
 ### Tiled 1.12.0 (13 March 2026)
 


src/tiled/projectdock.cpp
```diff
@@ -84,6 +89,8 @@ class ProjectView final : public QTreeView
     ProjectModel *mProjectModel;
     ProjectProxyModel *mProxyModel;
     QSet<QString> mExpandedPaths;
+    QSet<QString> mSavedExpandedPaths;
+    bool mHasSavedExpandedPaths = false;
```
bjorn: Seems like a good opportunity to use `std::optional<QSet<QString>>`, then we can also use its `value_or` function in `ProjectView::savedExpandedPaths`.

src/tiled/projectdock.cpp
```diff
@@ -92,21 +99,65 @@ class ProjectView final : public QTreeView
 ProjectDock::ProjectDock(QWidget *parent)
     : QDockWidget(parent)
     , mProjectView(new ProjectView)
+    , mCollapseAllAction(nullptr)
+    , mExpandToCurrentAction(nullptr)
```
bjorn: Please do these `nullptr` initializations at the declaration.

src/tiled/projectdock.cpp
```diff
@@ -92,21 +99,65 @@ class ProjectView final : public QTreeView
...
+            mProjectView->expandToPath(doc->fileName());
+    });
+
     auto widget = new QWidget(this);
     auto layout = new QVBoxLayout(widget);
     layout->setContentsMargins(0, 0, 0, 0);
     layout->setSpacing(0);
 
+    layout->addWidget(toolBar);
     layout->addWidget(mProjectView);
```
bjorn, line 151: In all other views I have put the small tool bar at the bottom of the view. I realize it might not be the best usability, but I'd still suggest that for consistency we also put this one at the bottom.

src/tiled/projectdock.cpp
```diff
@@ -182,6 +233,14 @@ void ProjectDock::selectFile(const QString &filePath)
 void ProjectDock::retranslateUi()
 {
     setWindowTitle(tr("Project"));
+    if (mCollapseAllAction) {
```
bjorn: No need to check whether these actions exist because you're always creating them.

bjorn:
I've pushed a simplification and added some comments. Maybe the change can be simplified further. For example, I don't think we need `mSavedExpandedPaths` at all if we would alternatively just not update `mExpandedPaths` when the `mExpandToCurrentAction` is checked?

rhythmcache:
I've applied the suggested changes and updated the implementation accordingly.

bjorn:
A few more comments:

* Please add the "Collapse All" action also to the context menu (when not clicking any particular item).
* Could you have a look at integrating the following icons for these actions instead of the random existing ones?
    * Collapse All: https://lucide.dev/icons/chevrons-down-up
    * Expand to Active File:  https://lucide.dev/icons/focus (seems appropriate to me)

rhythmcache:
image

added the Collapse All action to the empty area context menu and updated the toolbar icons

bjorn:
In an earlier comment I wrote:

> I don't think we need `mSavedExpandedPaths` at all if we would alternatively just not update `mExpandedPaths` when the `mExpandToCurrentAction` is checked?

@rhythmcache I see in the latest version you've still got both sets. Did you miss this or is there a reason to keep both?

I think the icons look great! Of course the downside of using black icons is that they're barely visible when changing the UI color to a dark one. But this is a problem with the symbolic SVGs in general that will need a generic solution.

rhythmcache:
I missed that there wasn’t really a reason to keep both. I've simplified it now so we just stop updating mExpandedPaths while Only Expand to Current is active.

PR #4347, opened 2026-02-21
Fix Erase Terrain button not activating Terrain brush

### LINKED ISSUE #4177: erase terrain does not change the tool
**Describe the bug**
erase terrain does not switch the tool to the terrain tool

**To Reproduce**
1. select stamp tool
2. click on erase terrain
3. the tool is not changed to the terrain tool

**Expected behavior**
clicking erase terrain should switch the current tool to the terrain tool

**Specifications:**
 - OS: arch linux
 - Tiled Version: Tiled 1.11.2

JustJ01:
Closes #4177 

### Description:
A bug was reported where clicking Erase Terrain in the Terrain Sets dock cleared the current terrain selection but did not switch the active tool to the Terrain Brush. This left users painting with whatever tool was previously selected (Stamp Brush) and broke the expected workflow.

### Before:
In the video , you can see when `Erase Terrain` button was pressed , the current tool remained `Stamp Brush` instead of swtiching to `Terrain Brush`.

https://github.com/user-attachments/assets/cd797b69-4797-44d6-95ca-d8c480bbf0f5

### After:
After the change, clicking `Erase Terrain` always activates the `Terrain Brush` , restoring the intended behaviour.

https://github.com/user-attachments/assets/377e8190-ccd5-491c-a4bc-baeefc9a7726

### DIFF
--- a/src/tiled/wangdock.cpp
+++ b/src/tiled/wangdock.cpp
@@ -429,11 +429,10 @@ void WangDock::refreshCurrentWangColor()
 void WangDock::wangColorIndexPressed(const QModelIndex &index)
 {
     const int color = index.data(WangColorModel::WangColorIndexRole).toInt();
-    if (!color)
-        return;
-
-    WangColor *currentWangColor = mCurrentWangSet->colorAt(color).data();
-    mDocument->setCurrentObject(currentWangColor, mWangColorModel->tilesetDocument());
+    if (color) {
+        WangColor *currentWangColor = mCurrentWangSet->colorAt(color).data();
+        mDocument->setCurrentObject(currentWangColor, mWangColorModel->tilesetDocument());
+    }
 
     emit selectWangBrush();
 }


bjorn:
I've just removed that button in #4320 (not merged yet, but it's mostly just waiting for a slight name change). However, I noticed this issue is not fixed yet. Maybe you could have a look at rebasing your change onto the `erase-terrain-selection` branch?

JustJ01:
Thanks @bjorn for the review, I will look at the codebase of that branch

JustJ01:
Hi @bjorn , I have made some changes onto erase-terrain-selection, also now the bug is fixed too

https://github.com/user-attachments/assets/972400d9-bc88-4c5a-8480-87c5ec5f7e87

bjorn:
@JustJ01 Thanks, that looks like the right fix! I've cleaned up the whitespace changes and merged it into the `erase-terrain-selection` branch, which I'll look into finishing up soon.

PR #4348, opened 2026-02-21
Add "Go to Tile" feature (Closes #1367)

### LINKED ISSUE #1367: Jump to location by tile coordinates

PoonamMehan:
### Feature: Go to tile coordinates

Closes #1367

It adds the ability to jump to a particular tile location on the map by entering tile coordinates. This is accessible via the Map menu: Map -> Go to Tile... menu item or the Ctrl+Shift+G keyboard shortcut.

### How to access:
- **Menu:** Map → Go to Tile... 
- **Shortcut:** Ctrl + Shift + G

### Summary 

Implemented TileLocatorSource class. It parses the input, validates the coordinates, ensures the coordinates exist inside the current map bounds, converts tile coordinates to pixel positions, and ensures the map exists before applying all the changes.

https://github.com/user-attachments/assets/f17fe3d8-08c0-4471-afcf-a56b141fe290

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -14,6 +14,7 @@
 * Added status info for various Stamp Brush, Terrain Brush and Eraser modes (#3092, #4201)
 * Added Escape to clear tile selection when any tile related tool is selected (#4243)
 * Added Escape to cancel tile selection and shape drawing operations
+* Added a "Go to Tile" action to jump to specific coordinates (by PoonamMehan, #4348)
 * Made the shortcut for current tool switch to previous tool (#4280)
 * Allow canceling Select Same Tile, Magic Wand and Bucket Fill operations with right-click and Escape
 * Allow dragging over multiple tiles with Select Same Tile, Magic Wand and Bucket Fill tools (#4276)


src/tiled/locatorwidget.cpp
```diff
@@ -398,6 +402,90 @@ void FileLocatorSource::setFilterWords(const QStringList &words)
...
+}
+
+QAbstractItemDelegate *TileLocatorSource::delegate() const
+{
+    return mDelegate;
+}
+
+QString TileLocatorSource::placeholderText() const
+{
+    return QCoreApplication::translate("Tiled::LocatorWidget", "x, y");
```
bjorn: Let's make it a little more informative.
```suggestion
    return QCoreApplication::translate("Tiled::LocatorWidget", tr("Go to tile: x, y"));
```

src/tiled/locatorwidget.cpp
```diff
@@ -398,6 +402,90 @@ void FileLocatorSource::setFilterWords(const QStringList &words)
...
+
+void TileLocatorSource::setFilterWords(const QStringList &words)
+{
+    bool validX = false;
+    bool validY = false;
+    bool valid = false;
+    int x = 0;
+    int y = 0;
+
+    QString input = words.join(QLatin1Char(' '));
```
bjorn: If we're going to remove spaces, it does not make much sense to join with a space:
```suggestion
    QString input = words.join();
```

src/tiled/locatorwidget.cpp
```diff
@@ -398,6 +402,90 @@ void FileLocatorSource::setFilterWords(const QStringList &words)
...
+
+    if (!valid && words.size() == 2) {
+        x = words.at(0).toInt(&validX);
+        y = words.at(1).toInt(&validY);
+        valid = validX && validY;
+    }
+
+    beginResetModel();
+    mHasValidCoord = valid;
+    if (valid) {
```
bjorn: Let's just remove the `valid` local variable, set `mHasValidCoord = validX && validY` and check that member in the above condition.

src/tiled/locatorwidget.cpp
```diff
@@ -398,6 +402,90 @@ void FileLocatorSource::setFilterWords(const QStringList &words)
...
+    const int x = qBound(0, mTileX, map->width() - 1);
+    const int y = qBound(0, mTileY, map->height() - 1);
+
+    auto renderer = mapDoc->renderer();
+    QPointF pixelPos = renderer->tileToPixelCoords(x + 0.5, y + 0.5);
+
+    if (auto mapView = dm->currentMapView())
+        mapView->forceCenterOn(pixelPos);
+
+    mapDoc->setSelectedArea(QRegion(QRect(x, y, 1, 1)));
```
bjorn: This works nice as a visualization, but it entirely breaks editing because edits are limited by an active selection. It also creates an undo command we don't want since selection changes are part of the undo stack.

So we can't do this, but it is nice to draw attention to the location. Maybe you could create a temporary `QGraphicsItem` (implement a subclass based on `TileSelectionItem`) which blinks a few times?

bjorn (changes requested):
This is a pretty nice and concise implementation, thanks! I did have a few suggestions, most important we'll need to find another way to highlight the location on the map since we don't want to limit edit operations or affect the undo stack.

I think you could add a graphics item for this to the map view which automatically removes itself again after say one second, and it can blink a few times to make it clear it is a temporary highlight.

src/tiled/locatorwidget.cpp
PoonamMehan: @bjorn I removed setSelectedArea() so "Go to tile" no longer affects selection or undo, and added a TileHighlightItem class: a temporary QGraphicsObject that draws a single-tile highlight with MapRenderer::drawTileSelection() and is positioned using the current layer’s offset, blinks three times via an opacity animation, then removes itself so the target tile is shown without changing selection or undo. Please have a look when you have a moment. Thank you very much!

bjorn: The highlight is nicely done, thanks!

bjorn:
Thanks @PoonamMehan! Nice job closing a 10 year old issue. :-)

PR #4349, opened 2026-02-21
Add Erase Empty Tiles mode to Stamp Brush

rhythmcache:
Ref https://github.com/mapeditor/tiled/issues/3643#issuecomment-3938748431

This PR adds an optional Erase Empty Tiles mode to the Stamp Brush.

### DIFF
--- a/src/tiled/abstracttilefilltool.cpp
+++ b/src/tiled/abstracttilefilltool.cpp
@@ -47,6 +47,8 @@ AbstractTileFillTool::AbstractTileFillTool(Id id,
 
     connect(mStampActions->random(), &QAction::toggled, this, &AbstractTileFillTool::randomChanged);
     connect(mStampActions->wangFill(), &QAction::toggled, this, &AbstractTileFillTool::wangFillChanged);
+    connect(mStampActions->eraseMode(), &QAction::toggled, this, &AbstractTileFillTool::setEraseMode);
+    connect(this, &AbstractTileFillTool::eraseModeChanged, mStampActions->eraseMode(), &QAction::setChecked);
 
     connect(mStampActions->flipHorizontal(), &QAction::triggered, this,
             [this] { emit stampChanged(mStamp.flipped(FlipHorizontally)); });
@@ -118,7 +120,8 @@ void AbstractTileFillTool::populateToolBar(QToolBar *toolBar)
 {
     mStampActions->populateToolBar(toolBar,
                                    mFillMethod == RandomFill,
-                                   mFillMethod == WangFill);
+                                   mFillMethod == WangFill,
+                                   mEraseMode);
 }
 
 void AbstractTileFillTool::setFillMethod(FillMethod fillMethod)
@@ -146,6 +149,21 @@ void AbstractTileFillTool::setWangSet(WangSet *wangSet)
     mWangSet = wangSet;
 
     invalidateRandomAndMissingCache();
+}
+
+void AbstractTileFillTool::setEraseMode(bool value)
+{
+    if (mEraseMode == value)
+        return;
+
+    mEraseMode = value;
+    emit eraseModeChanged(value);
+
+    // Don't need to recalculate fill region if there was no preview
+    if (!mPreviewMap)
+        return;
+
+    tilePositionChanged(tilePosition());
 }
 
 void AbstractTileFillTool::mapDocumentChanged(MapDocument *oldDocument,
@@ -227,6 +245,17 @@ void AbstractTileFillTool::updatePreview(const QRegion &fillRegion)
     }
     }
 
+    if (mEraseMode) {
+        if (preview->layerCount() == 0) {
+            for (Layer *layer : targetLayers()) {
+                if (TileLayer *tileLayer = layer->asTileLayer()) {
+                    preview->addLayer(new TileLayer(tileLayer->name(), mFillBounds.topLeft(), mFillBounds.size()));
+                }
+            }
+        }
+        erasePreview(*preview, fillRegion);
+    }
+
     preview->addTilesets(preview->usedTilesets());
 
     brushItem()->setMap(preview);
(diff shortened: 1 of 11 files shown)


bjorn:
I'm wondering whether you tested this change because this doesn't seem to work as intended at all. Some noticeable problems:

* When hovering the map with the "Erase Empty Tiles" toggle enabled, the brush preview's highlight is stuck in the top-left and clipping the preview tiles. I guess it almost works, but its position is off.

* When clicking to paint, a 16x16 section of the map is erased. This seems to be due to the way the `Checked` flag is set on empty tiles, since you're iterating all non-empty chunks of the layer.\
  \
  image

* The "Erase Empty Tiles" mode is also visible for the Bucket Fill and Shape Fill tools, but its state is not persistent and is not taken into account. I think these tools will also need to support it, since they may also paint stamps with empty cells.

I think the implementation will need to look a little different. Note that we can't decide which empty tile to erase by the time we need to draw a stamp. The stamp will need to already have the empty tiles that are supposed to be erased marked with the `Checked` flag at the time the stamp is created. As such, the toggle would rather control how `StampBrush::drawPreviewLayer` should erase or ignore tiles with a `Checked` flag in the stamp.

But actually the way you just used the Eraser icon for this mode makes me think we could also simplify this a little. For the use-case of erasing lines, or erasing in general, it might be actually easier if this mode would simply erase instead of paint. Erase any tile that would have been painted by the stamp, that is. Maybe you could change this PR to work that way?

rhythmcache:
I've reworked the implementation to follow the simpler approach you described.

Instead of marking empty cells as Checked, erase mode now just erases the same region the stamp would normally paint.

bjorn:
Hmm, I think it would still be better if the erasing was reflected in the preview, so this erasing mode should be implemented in `drawPreviewLayer` and not `doPaint`. I just meant that such a mode should be easier, because a mode that only erases empty tiles would need such tiles to be explicitly marked as empty-but-part-of-the-brush when it is created.

I think as part of `drawPreviewLayer`, it could be implemented as a kind of post-processing operation, where all tiles to be drawn are replaced with empty tiles (maybe needing "checked" flags), such that the preview renders empty to also give the user a hint that erase mode is activated. 

Also, you did not address my third point. Either the Bucket Fill and Shape Fill tools should also support this Erase mode, or they should not display that toggle. I think they could just support it as well. Finally, the mode is still called "Erase Empty Tiles" but maybe should not just be called "Erase Mode".

@eishiya I'd really appreciate your feedback on this direction. This "Erase" mode would allow erasing lines, circles, shapes and fill-erase. It's not yet the "erase empty tiles" mode, but I think that could be added as a third option later so we'd have a "Drawing Mode: Normal, Replace, Erase" state that could be supported by all regular drawing tools.

eishiya:
#### Whole-stamp erase
Being able to turn any stamp or fill shape into an eraser sounds great to me! Not the feature I and others asked for, but a useful one too. I'd love to see the bucket and shape fills support that too, not just the brush.

Incidentally, this is similar to how Krita works - it has a global eraser mode which causes its pixel-painting tools to erase within their region instead of painting. I think a separate Eraser tool should still be kept though, for the same reasons I think Krita should add a separate Eraser tool - it can keep its own settings independent of the brush (especially useful if Eraser gains the ability to have its own size and circle/rectangle shapes) so that the user doesn't have to change their stamp to go between painting and erasing, and switching to a separate tool to a separate task can be more intuitive than toggling modes, especially since many people are already used to Erasers being separate tools.

#### Brushes with erase-tiles
I don't think we need three brush modes as such. Rather, I think *at sample-time*, it should be determined whether empty tiles in the selection should be recorded (and thus always erase) or ignored (and behave as brushes do currently). This way there's no need to think about modes when using saved stamps, and the Stamp Brush tool itself (and the Fill tools) can have a single eraser mode toggle which could have nothing to do with whether the current stamp contains any "erase" tiles.

The challenge is figuring out where to place this toggle. It's irrelevant for brushes made by selecting from tilesets (those contain no empty tiles) or from an existing brush (those have already had their region, including any erase-tiles, determined), so it's only relevant for brushes being sampled from a selection in the map, where there may not only be a mix of relevant and irrelevant empty tiles, but also multiple layers that may be a mix of relevant-empty, irrelevant-empty, and non-empty in the same cell. It could perhaps be a selection tool toggle, but selections aren't always made for copying/cutting, many are just there for deleting and filling. Perhaps the Edit menu could have a Copy/Cut mode sub-menu? But then the user has to access that menu to see which mode is currently active. Maybe it could be there and also in a dedicated area in the status bar?

Even with only a single layer, IME brushes that need to erase are nearly always complex in shape, while rectangular selections are often the easiest way to grab a region to make a brush. So an initial selection will nearly always contain empty tiles that need to be recorded as empty, and some that need to be ignored. The user can refine this by deselecting irrelevant tiles, but perhaps a "smart" mode can be added that ignores selected empty tiles that are are contiguous with the selection bounds? That's a common deselection task I'd need to do for my brushes, at least. Maybe it's not actually that common and doesn't warrant automation (or maybe is better as a script).

With multiple layers, it's a right mess, and I suspect there would need to be at least three modes for treating empty tiles: ignore all empty cells, ignore cells that are empty on all selected layers, include (=turn into erase-tiles) all empty cells. For scenarios where even that's not enough, there is scripting.

---

I hope an essay that goes off on a tangent is what you meant when you asked for my feedback 🙃

bjorn:
> I hope an essay that goes off on a tangent is what you meant when you asked for my feedback 🙃

It's not what I was looking for, but I still really appreciate your thoughts on this. To keep this PR actionable I'd suggest we focus on the "Whole-stamp erase" part.

@rhythmcache Could you address my earlier feedback, in short:

* Move the implementation from `doPaint` to `drawPreviewLayer`, such that the preview looks different in erasing mode.
* Support this in the Bucket Fill and Shape Fill tools as well (the toggle is already displayed there as well, just not synchronized/used).
* Rename "Erase Empty Tiles" to just "Erase Mode".

rhythmcache:
Hi @bjorn ,

I've addressed all your feedback =>

The erase logic is now implemented in drawPreviewLayer as a post-processing step. After the preview is fully built, erasePreview() replaces all painted cells with checked empty tiles, so the preview correctly reflects what will be erased before the user commits. The actual erase in doPaint uses eraseTileLayers with the stamp's modified region, so preview and paint behavior are consistent.
Bucket Fill and Shape Fill tools now support Erase Mode via AbstractTileFillTool. The toggle state is also synchronized across all three tools via MapEditor, so switching between tools keeps the mode active.
Renamed "Erase Empty Tiles" to "Erase Mode" everywhere action text, action ID, member variables, signals and slots

PR #4350, opened 2026-02-21
Add viewport support

### LINKED ISSUE #4237: Option to show the camera bounds of the in-game view

I'm making a little GBA game with tons of parallax. So I use the parallax feature of the Tiled editor to represent that. But I've noticed that the parallax appears far more extreme in the editor, because my view is likely zoomed out significantly, which scales the parallax scrolling.

So I thought it'd be nice to be able to see exactly what's going to appear in-game while editing, in the form of a little box in the center of the screen that has the same size as the in-game view.

This will likely benefit level designs that don't use parallax, as well, since level designers can more clearly see what's on-screen at any given moment.

I suggest adding an option in, say, the map properties to specify the size of the in-game view (such as 240×160 pixels in my case). In addition, there should be a toggle in the "view" menu that simply draws a rectangle of that size in the middle of the viewport.

I've also come up with a simple workaround that I'm currently using, in which I have an object layer with a parallax factor of 0 (so it's fixed to the viewport) with a rectangle object in the center. This works pretty well, and the only real downsides are that it's a bit hacky and will require that I inform the engine that this isn't an actual object.

I'm honestly fine with it, so this proposal is really more about making something more standardized and less hacky than filling some gap in the editor's feature set.

Image

abhinavs1920:
### Add viewport bounds visualization

Closes #4237

Adds a viewport overlay feature that displays the in-game camera bounds as a rectangle in the editor. This helps level designers see exactly what will be visible on screen, especially useful for games with parallax scrolling where the editor view can be misleading.

* Added "Viewport Size" property to Map Properties (width × height in pixels)
* Added "Show Viewport" toggle in the View menu
* Viewport is displayed as a red outline rectangle centered in the view
* Viewport size is saved with the map file

### Usage

1. Set the viewport size in Map Properties (e.g., 240×160 for GBA)
2. Enable View → Show Viewport
3.  A red rectangle shows your in-game camera bounds

The viewport stays centered as you pan around, similar to a parallax factor 0 layer.
Benefits over the workaround

The issue mentioned using an object layer with parallax factor 0 as a workaround. This implementation is better because:

* No game objects that need to be filtered out by the engine
* Clean outline with no tinted fill
* Easy to toggle on/off
* Properly integrated with the editor's preference system
* Viewport size is stored as map metadata, not as a game object

### Screenshots

Screenshot from 2026-02-21 23-23-32

Screenshot from 2026-02-21 23-23-47

Feedback welcome!

### DIFF
--- a/src/tiled/mapitem.cpp
+++ b/src/tiled/mapitem.cpp
@@ -409,6 +409,7 @@ void MapItem::documentChanged(const ChangeEvent &change)
         case Map::LayerDataFormatProperty:
         case Map::CompressionLevelProperty:
         case Map::ChunkSizeProperty:
+        case Map::ViewportSizeProperty:
             break;
         }
         break;


bjorn:
@abhinavs1920 Thanks for working on this issue, this will be very useful!

Unfortunately, the current implementation has a problem with scrolling, because you're painting it as part of `MapView::paintEvent`. It's not noticeable with "Enable Parallax" disabled because in that case it always repaints the entire view, but you'll notice when disabling this option.

I think the best fix will be to render this rectangle as its own item, probably controlled by the `MapScene` since we'd have one such item per scene. It would be fancy if this item could render its surrounding area darker, but as a first version I think you can just rely on a `QGraphicsRectItem`. You can call `setZValue` with some positive value to place it above the `MapItem` instances.

In the Properties view, I would suggest to group this property together with the Parallax Origin because they're closely related. I think often people will want to set their Parallax Origin to minus half the viewport size, because that means the parallax layers will align in the top-left of the viewport.

LongSteve:
I have literally this morning been reviewing my game engine tile rendering in order to properly integrate the Tiled parallax factors. This addition will be very welcome for map creation, thanks for working on it. I shall pull this PR now and give it a try.

src/tiled/viewportoverlayitem.cpp
```diff
@@ -0,0 +1,66 @@
...
+    setFlag(QGraphicsItem::ItemIsSelectable, false);
+    setFlag(QGraphicsItem::ItemIsMovable, false);
+    
+    syncWithMapDocument();
+}
+
+void ViewportOverlayItem::syncWithMapDocument()
+{
+    const bool showViewport = Preferences::instance()->showViewport();
+    const QSize viewportSize = mMapDocument->map()->viewportSize();
```
LongSteve: This line currently crashes with mMapDocument being nullptr, presumably caused by line 87 of mapscene.cpp creating a new ViewportOverlayItem passing nullptr to the constructor.  syncWithMapDocument should bail out of mMapDocument is null.

abhinavs1920: Good catch, you’re absolutely right. It was crashing because syncWithMapDocument() was getting called with mMapDocument == nullptr. I’ve added proper null checks and guarded the constructor call. Also cleaned up the memory handling in setMapDocument to use removeItem(). Thanks for spotting this :)

src/tiled/mapview.cpp
```diff
@@ -480,6 +502,133 @@ void MapView::paintEvent(QPaintEvent *event)
     QGraphicsView::paintEvent(event);
 }
 
+void MapView::drawForeground(QPainter *painter, const QRectF &rect)
+{
+    QGraphicsView::drawForeground(painter, rect);
+
+    Preferences *prefs = Preferences::instance();
+    if (!prefs || !prefs->showMapRulers())
```
LongSteve: I think you might have mixed your PRs up slightly.  This PR does not include the code for the rulers, so this test always returns false, and the viewport is never rendered.  

I'm currently just viewing this PR by itself though, and can comment these lines out for testing.

I hope you don't mind me making comments, I hope it is helpful, but if not, please let me know, I am happy to stop bothering you.

LongSteve: Having said that, I cannot seem to get the viewport rectangle to render at all now, even commenting that test out, so maybe something that draws the rectangle has been removed?

abhinavs1920: no problem at all.. I am thankful you are finding issues and helping me out... actually I am working at multiple issues, so I am not sure if I made any mistake while switching branches. Investigating it.. will let u know.. again saying thanks for reviewing :)

abhinavs1920: Heyy @LongSteve , you should use this branch now, please check it once.. rebuilt it then run

LongSteve: Thanks I checked out the latest code, it looks like you're almost there.  The viewport rectangle is rendered, but it's fixed in position at the top left of the map.  I added the following changes to update it's x,y position as the mapview is moved around, so it's always centered.  Sorry this isn't the most ideal way to send diffs, but it shows what and where I added code, and you might want to do the update differently.  I also added an accessor `ViewportOverlayItem *viewportOverlay() const;` to the MapScene.

image

One last thing (sorry!).  The map property values for "Viewport Size" width and height are not saved in the map file (I'm using .tmx).  This looks simple enough to add, just following the same pattern for the Parallax Origin values.

LongSteve: Also only just realised I should have put that code directly into `MapScene::setViewRect` rather than add the accessor, and call it from MapView.  I'm very new to this code, so not necessarily making good decisions.

abhinavs1920: Heyy @LongSteve can you check the implementation, I have now tried and tested the feature, it works fine.. also thanks for your help

LongSteve:
Thanks @abhinavs1920 This PR all checks out, and functionally, the bits you have added are working nicely now for me.  

There were a couple of trivial merge conflicts with the current master branch (preferences.h), but they were just some simple clashes with the changed definition of snapMode.  

The viewport size property is not saved, so it is always 0x0 when you open a map, but that's actually not a big deal for me, I don't need to read that value in my game engine!  

@bjorn may have thoughs on whether this PR can be merged as is, or if the viewport size property saving is a requirement.

abhinavs1920:
Thanks @LongSteve for checking it and testing it out!
Good to know the changes are working as expected.

Regarding the viewport size property, I'm happy to take a look at saving it as well if that ends up being needed. Otherwise I'm fine with leaving it as is :)

bjorn:
@abhinavs1920 I'm sorry, it seems the AI review I was experimenting with has edited your initial comment. This should no longer happen now, but maybe you could revert it back to the version you wrote?

Also I see it left quite some comments. Just note that it might be rather inaccurate, I'm just evaluating it for now.

src/tiled/undocommands.h
```diff
@@ -57,6 +57,7 @@ enum UndoCommands {
     Cmd_ChangeMapStaggerAxis,
     Cmd_ChangeMapStaggerIndex,
     Cmd_ChangeMapTileSize,
+    Cmd_ChangeMapViewportSize,
     Cmd_ChangeObjectGroupColor,
```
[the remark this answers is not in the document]
bjorn: Rather silly comment, since there are no undo command IDs saved into map files.

src/tiled/viewportoverlayitem.h
```diff
@@ -0,0 +1,44 @@
...
+class ViewportOverlayItem : public QGraphicsRectItem
+{
+public:
+    ViewportOverlayItem(MapDocument *mapDocument, QGraphicsItem *parent = nullptr);
+
+    void syncWithMapDocument();
+
+private:
+    MapDocument *mMapDocument;
+};
```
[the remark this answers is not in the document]
bjorn: Wrong, because `MapScene::setMapDocument` re-creates the `ViewportOverlayItem` with an updated pointer, which will also set it to `nullptr` when the last map is closed.

However, I think `ViewportOverlayItem` does not need its own `MapDocument*` member. It only actually uses it in `syncWithMapDocument`, so that function could take the `MapDocument*` as parameter instead.

src/tiled/preferences.cpp
```diff
@@ -372,6 +383,19 @@ void Preferences::setParallaxEnabled(bool enabled)
...
+{
+    setValue(QLatin1String("Interface/ShowViewport"), enabled);
+    emit showViewportChanged(enabled);
+}
+
+void Preferences::setSnapToGrid(bool snapToGrid)
+{
+    setValue(QLatin1String("Interface/SnapToGrid"), snapToGrid);
+    emit snapToGridChanged(snapToGrid);
+}
```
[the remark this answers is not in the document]
bjorn: Correct, this is due to a merge error in 102f212239219be8853c5a77ca5370a66faf1fd4, which re-introduces previously deleted functions. I've fixed this in 07c914e097e96f25b448f7509c1d0103dc4968f4.

src/tiled/mapscene.cpp
```diff
@@ -110,6 +123,22 @@ void MapScene::setMapDocument(MapDocument *mapDocument)
                 this, &MapScene::tilesetReplaced);
     }
 
+    // Update viewport overlay with new map document
+    if (mViewportOverlayItem) {
+        // Remove from scene (scene will delete it)
+        removeItem(mViewportOverlayItem);
+        mViewportOverlayItem = nullptr;
+    }
```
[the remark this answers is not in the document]
bjorn: That's right, this object is leaked on each map change. In fact, I don't think we need to re-create this object at all, especially when we don't store the `MapDocument *` there but just pass that in as parameter to `syncWithMapDocument`.

src/libtiled/map.h
```diff
@@ -174,6 +175,7 @@ class TILEDSHARED_EXPORT Map : public Object
         int compressionLevel = -1;
         QSize chunkSize = QSize(CHUNK_SIZE, CHUNK_SIZE);
         LayerDataFormat layerDataFormat = Base64Zlib;
+        QSize viewportSize;  // Camera bounds / in-game viewport size (0,0 = disabled)
```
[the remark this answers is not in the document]
bjorn: Well, there is indeed an inconsistency there. Actually I would suggest to initialize it as follows for consistency with the `chunkSize` initialization:

```suggestion
        QSize viewportSize = QSize(0, 0);  // Camera bounds / in-game viewport size (empty = disabled)
```

src/tiled/mapview.h
```diff
@@ -85,6 +85,7 @@ class MapView : public QGraphicsView
     bool event(QEvent *event) override;
 
     void paintEvent(QPaintEvent *event) override;
+    void scrollContentsBy(int dx, int dy) override;
```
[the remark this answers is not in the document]
bjorn: Wrong, because the implementation for `scrollContentsBy` is definitely provided. Though they were right about this issue in `preferences.h` because of the merge error that has now been addressed.

src/tiled/viewportoverlayitem.cpp
```diff
@@ -0,0 +1,105 @@
...
+    : QGraphicsRectItem(parent)
+    , mMapDocument(mapDocument)
+{
+    setZValue(10000); // Place above map items
+    setPen(QPen(QColor(255, 0, 0, 180), 2.0));
+    setBrush(Qt::NoBrush);
+    
+    // Item should not be selectable or movable
+    setFlag(QGraphicsItem::ItemIsSelectable, false);
+    setFlag(QGraphicsItem::ItemIsMovable, false);
```
bjorn: I don't think items are selectable or movable by default, so please remove these lines.

src/tiled/viewportoverlayitem.cpp
```diff
@@ -0,0 +1,105 @@
...
+        return;
+    }
+    
+    const bool showViewport = prefs->showViewport();
+    const QSize viewportSize = map->viewportSize();
+    
+    // Validate viewport size (must be positive and reasonable)
+    const bool validSize = !viewportSize.isEmpty() && 
+                          viewportSize.width() > 0 && 
+                          viewportSize.height() > 0 &&
```
bjorn: You can remove these checks because that is exactly what `QSize::isEmpty` checks.

src/tiled/viewportoverlayitem.cpp
```diff
@@ -0,0 +1,105 @@
...
+{
+    // Bail out if no map document
+    if (!mMapDocument) {
+        setVisible(false);
+        return;
+    }
+    
+    // Bail out if map is null
+    Map *map = mMapDocument->map();
+    if (!map) {
```
bjorn: A `MapDocument` always has a valid `Map` instance, because it owns it.

src/tiled/viewportoverlayitem.cpp
```diff
@@ -0,0 +1,105 @@
...
+        setVisible(false);
+        return;
+    }
+    
+    // Get preferences safely
+    Preferences *prefs = Preferences::instance();
+    if (!prefs) {
```
bjorn: `Preferences::instance` will always return a valid pointer. It creates an instance when it doesn't exist yet.

bjorn (commented):
This already works a lot better! But we'll definitely want this property to be saved in the TMX and JSON map formats. I think saving it within the `editorsettings` is a good idea!

Please also have a look at the AI comments that I replied to and didn't resolve. It seems to say something useful about half of the time. :-)

abhinavs1920:
Thanks for the feedback @bjorn :)

I've now added saving of the viewport dimensions to both TMX and JSON formats and implemented it through editorSettings as suggested. I also cleaned up the redundant code and reviewed the AI comments you responded to, addressing the relevant ones.

Let me know if there’s anything else that should be adjusted as per the existing architecture.

LongSteve:
Nice one @abhinavs1920 this is working well for me, thanks for the extra work.  The viewport property is saved and loaded, it renders nicely, changes size with zoom.  I toggled on and off various settings, and exported JSON and everything works as expected.  Very useful indeed for previewing how the parallax layers will render at runtime.

I would suggest only one more tiny thing, which is to fix the compiler warning in:

``mapitem.cpp:396: warning: enumeration value 'ViewportSizeProperty' not handled in switch [-Wswitch]
mapitem.cpp: In member function 'void Tiled::MapItem::documentChanged(const Tiled::ChangeEvent&)':
mapitem.cpp:396:16: warning: enumeration value 'ViewportSizeProperty' not handled in switch [-Wswitch]
  396 |         switch (mapChange.property) {
      |                ^
``

I don't know the details of the coding standard used for the source, but given that's the only compiler warning I get from the entire codebase, I would not want to leave it in there!  My guess to be to simply add a ``case Map::ViewportSizeProperty`` around line 411 in order to not call mapChanged() in this case.

PR #4351, opened 2026-02-22
Added map-bounds clipping options for tile/image/object rendering with map-level clipRect override

### LINKED ISSUE #4153: Option to clip map contents to map bounds
## Problem

When using Image Layers (with repeat X/Y, or with oversized images), or when using oversized tiles, the contents of a map can be much larger than its bounds, causing it to get in the way of other maps in its World. Even when editing such a map without other maps visible, this overspill can misrepresent the map, since in most cases, engines will only show in-bounds content.

## Proposal
A View option to crop the map contents to the map bounds. This should affect at least Image and Tile layers, where it should be relatively straight-forward.

### Objects
For Object Layers, the situation is trickier - Objects don't usually get in the way as much, and it's often desirable to see when an Object is out of bounds even when other layers are clipped and indeed, clipping the other layers may be desirable to help see such Objects better. But, games that use Objects decoratively may also want to clip them to the map bounds. Perhaps there should be a separate View option for Objects, with three options: Show All, Clip to Map Bounds, Hide Out of Bounds. The difference between the latter two would be in how Objects that straddle the map boundary are displayed - clipped, or shown in full. If clipping Objects to map bounds is too complicated, then an option to hide Objects that are completely out of bounds would be fine.

### Other concerns
- The rectangular bounds can be used for non-orthographic maps.
- Parallax should be calculated before the crop, so that parallax cannot cause the map's visuals to extend beyond its bounds.
- Infinite maps don't have defined bounds, but clipping is still useful for those. Infinite maps should be clipped to the bounds Tiled displays for them, i.e. some multiple of the 16x16 chunks Tiled uses internally.
- These options should *not* be tied to Worlds, as clipping map contents can be useful even in stand-alone maps.
- Maps with foreground parallax layers have larger bounds than their "actual" bounds, and users may want to clip to a smaller rectangle than the map bounds. Even if per-layer sizing is added, the map size would probably be the largest of those layers rather than an arbitrary size (which would be convenient for Tiled map loaders, since it would give them a maximum tile layer size right away), then per-layer sizing won't solve the problem of foreground parallax making the clipping region too large. Clipping to the layer bounds also wouldn't help. To accomodate maps with foreground parallax, it would help to have a clipping rect property on the map which, if not set, is the same as the map's current bounding area. The reason it should be a rect rather than a size is to accommodate infinite maps, but it would be useful for some finite maps with "margin" areas too.

JustJ01:
Fixes #4153 

### Description:
This PR adds editor-side clipping controls to keep map visuals constrained, useful for world context maps and oversized/repeating content.

### 1. View Toggle: **Clip Map to Bounds**

Adds a new toggle: View → Clip Map to Bounds

- **On:** Anything drawn outside the map's visible area is hidden.
- **Off:** Maps behave as before (overflow allowed).

Tile Layer and Image Layer rendering now respect clipping when this option is enabled.

---

### 2. View Submenu: **Object Bounds**

Adds a new submenu: View → Object Bounds

This controls how objects behave when they are near or outside the map bounds.

- **Show All**  
  Old behavior. All objects remain visible, including those fully outside the map.

- **Clip to Map Bounds**  
  Objects remain visible, but any portion outside the map bounds is visually clipped.

- **Hide Out-of-Bounds**  
  Objects fully outside the map are hidden.  
  Objects touching or crossing the map boundary remain visible.

Object rendering and visibility now follow the selected mode.

---

### 3. Map-Level Clipping Override (Custom Clipping)

Adds support for a custom map property: `clipRect`
This allows defining a custom clipping rectangle instead of using the full map bounds.

**Primary property name:** `clipRect`  

**Accepted formats:**

- Rect/class-style value with `x, y, width, height`
- 4-number list:  [x, y, width, height]
- String format:  "x,y,width,height"
- If the property is absent or invalid (for example `"hello"`), clipping automatically falls back to the normal map bounds.

---

### 4. Unified Clipping Logic

Tile, image, and object rendering paths now use the same effective clipping bounds, ensuring consistent behavior across all layer types.

---

### 1) Clip Map to Bounds ( Tile + Image layers)
- Open any map
- Add a Image Layer, Browse the image and add the image, Enable Repeat X and Repeat Y
- View -> Clip Map to Bounds, Toggle On or Off

https://github.com/user-attachments/assets/ee378a30-96f9-49e6-ba9d-204458d4f775

### 2) Object Bounds Mode
- Add Object Layer
- Place Objects, one inside the map, one crossing the map border, one fully outside the map 
- VIew -> Object Bounds -> (Select Modes from there)

https://github.com/user-attachments/assets/f1dd0682-10a4-4d1a-bd1b-f13106566569

### 3) Custom Clipping rectangel (clipRect)
- Map -> Map Properties
- In Custom properties add:
        Name: clipRect
        Type: string
        value: (Valid format string mentioned above)

https://github.com/user-attachments/assets/4a978bc8-cad7-4bc5-a94b-f4ba2a0751d7


JustJ01:
Hi @bjorn ,

I've implemented core View clipping for Tile + Image layers, plus object modes (Show All / Clip / Hide Out of Bounds) , persistent preferences and view menu wiring , map-level custom clipping rectangle property (clipRect).


Please review the changes whenever possible.

Items like per-map background fill behavior in World view, world background draw-order policy (mentioned by @eishiya) are not included in this PR. I'm also considering adding dedicated first class UI fields in map properties for clipRect in a follow up.

If the current scope and implementation looks good, I can proceed with those additional features that @eishiya mentioned

src/tiled/mapclipbounds.h
```diff
@@ -0,0 +1,127 @@
...
+}
+
+inline bool variantToDouble(const QVariant &value, double &result)
+{
+    bool ok = false;
+    result = unwrapPropertyValue(value).toDouble(&ok);
+    return ok;
+}
+
+inline std::optional<QRectF> clipRectFromVariant(const QVariant &rawValue)
```
bjorn: This function seems odd to be, accepting all those different formats in which somebody might try to define a rectangle in a string. Was this your idea or did AI write this for you?

I would personally prefer we leave this functionality out of the change initially, and add it as a proper native property later. There's a lot of code involved in supporting this in its current form, from parsing to listening to property changes in the `MapItem`.

src/tiled/mapclipbounds.h
```diff
@@ -0,0 +1,127 @@
+/*
+ * mapclipbounds.h
+ *
+ * This file is part of Tiled.
```
bjorn, line 5: New files should use the standard license header also seen in other files. But in this case I think this file can be removed instead since it is entirely concerned with extracting data from custom properties.

src/tiled/mainwindow.cpp
```diff
@@ -455,8 +466,22 @@ MainWindow::MainWindow(QWidget *parent, Qt::WindowFlags flags)
...
 
+    switch (preferences->objectBoundsVisibility()) {
+    case Preferences::ShowAllObjects:
+        mUi->actionShowAllObjects->setChecked(true);
+        break;
+    case Preferences::ClipObjectsToMapBounds:
+        mUi->actionClipObjectsToMapBounds->setChecked(true);
+        break;
+    case Preferences::HideOutOfBoundsObjects:
+        mUi->actionHideOutOfBoundsObjects->setChecked(true);
```
bjorn: I had suggested to leave this option out in the first version. I think it is potentially confusing because it hides objects without it being entirely clear why, whereas just clipping objects is much easier to understand. The use-case is a little unclear to me as well.

Maybe @eishiya could comment on whether this is a feature they need personally or whether it would be fine to remove it for now.

src/tiled/mapobjectitem.cpp
```diff
@@ -133,6 +139,18 @@ void MapObjectItem::paint(QPainter *painter,
...
+    const bool clipToBounds = boundsVisibility == Preferences::ClipObjectsToMapBounds;
+
+    if (clipToBounds) {
+        QGraphicsItem *mapRoot = this;
+        while (mapRoot->parentItem())
+            mapRoot = mapRoot->parentItem();
+
+        const QRectF clipBounds = mapRectFromItem(mapRoot, Internal::effectiveClipBounds(*renderer));
+        painter->save();
+        painter->setClipRect(clipBounds, Qt::IntersectClip);
```
bjorn: In the interest of performance I think we shouldn't set the clip rect for each of our clipped items  individually. On the other hand, the separate clip settings for objects and other layers don't make it easy to apply clipping globally.

But I think we can at least try to rely on the `QGraphicsItem::ItemClipsChildrenToShape` graphics item flag in `ObjectGroupItem`, where we can set the bounding rect based on the desired clip rect.

Edit: After writing further comments I realize this will not work in the context of parallax layer movement, which would affect our bounding rect. So I'm not sure now what would be the best approach here... I'm thinking it's maybe not worth it to support different clipping options for objects vs. other layers. Clipping works much better and is way more performant if it can be applied at a higher level. I wonder what @eishiya thinks about this because they suggested to treat objects differently.

src/tiled/tilelayeritem.cpp
```diff
@@ -67,7 +69,23 @@ void TileLayerItem::paint(QPainter *painter,
...
+
+        QGraphicsItem *mapRoot = this;
+        while (mapRoot->parentItem())
+            mapRoot = mapRoot->parentItem();
+
+        const QRectF clipBounds = mapRectFromItem(mapRoot, mapBounds);
+        painter->setClipRect(clipBounds, Qt::IntersectClip);
+    }
+
     renderer->drawTileLayer(painter, tileLayer(), option->exposedRect);
```
bjorn, line 76: We can optimize this a little by applying the clip also to the `option->exposedRect`, since that will avoid the processing of coordinates which were clipped.

Ideally we'd use `QGraphicsItem::ItemClipsToShape` flag and clip our `mBoundingRect` instead, but I think this won't work because it will fail to take into account the layer's position changes due to parallax, I'm not sure if we want to be adjusting the `mBoundingRect` based on parallax shift.

src/tiled/mainwindow.cpp
eishiya: Just clipping is fine. I suggested hiding mainly as a potential performance-conscious option, since hiding can be implemented more simply than clipping.

src/tiled/mapobjectitem.cpp
eishiya: If Objects can be clipped without a performance penalty relative to hiding, then I see no benefit to implementing an option to hide them.

src/tiled/mainwindow.cpp
bjorn: Ah, actually clipping should be better for performance reasons since hiding requires toggling individual object's state. Well, provided we can figure out a way to clip all objects using the same rectangle...

Anyway then please remove the "hide out of bounds" option, @JustJ01.

src/tiled/mapobjectitem.cpp
bjorn: @eishiya Here I was trying to ask whether you really needed a separate clipping option for tile/image layers and objects, or whether it would be fine to have one "clip to map" option. That option would be really easy to implement when we didn't need to treat each data type differently, because then we can just enable/disable clipping on the top-level map item.

But due to there being two separate options now, the simple and performant approach can only be used when you enable both.

eishiya: Ah, sorry! Brain's not braining very well today. For my needs, per-layer-type options are not necessary, but would be nice to have. I'd be fine with a single setting, and perhaps if someone else asks for separate options, they could be reconsidered.

Speaking of extra settings: Looking at my issue again, I'm surprised I didn't mention anything about clipping for current map vs other maps in a world. For me, it's often desirable to clip all the maps in the world *except* the one I'm editing.

bjorn:
@JustJ01 In light of the discussions with @eishiya and after my previous code review, I'd like to see the following changes:

* No separate clipping option for tile/image layers vs. objects -> Clipping can then by applied either by the `MapItem` directly or by a dedicated "clip item", which is used as the parent of all layer items. By relying on setting the bounding rect to the area we want to clip in combination with `QGraphicsItem::ItemClipsChildrenToShape`.

* The global "Map Bounds Clipping" option should have three states:
    * No Map Clipping
    * Clip Maps to Bounds
    * Clip Other Maps to Bounds

This should simplify the implementation, avoid performance concerns and I think it's also a little more intuitive.

The following options should not be handled for now, I'd rather see them considered for future changes:

* Map-level clipping override (the need for this is not entirely clear to me yet)
* Clipping map background color (because it affects totally unrelated code and we need to decide whether this needs another option or whether we can apply this based on the above option)

eishiya:
> Map-level clipping override (the need for this is not entirely clear to me yet)

Should I open a separate issue for this, or would it suffice to leave the existing one open?
My use case for custom map bounding rects is that my maps are routinely much larger than their actual "live" area because I need to have room for foreground parallax layers (because they scroll faster than the "main" layers (parallax 1.0), they need more content). In my Worlds, the maps are aligned according to the main layers, so the foreground layers overlap and make it difficult to see . Clipping to map bounds doesn't help the problem at all because all the content is in-bounds - the map bounds are just larger than the part of the map that's actually important in the context of a World. An alternative fix for this issue would be per-layer sizing, but that creates the question of which layer to use as the map bounds, so either way a setting to help deal with parallax>1 layers in maps would be useful.
I can send some screenshots (or even screenshare) on Discord to help understand the problem, if you like.

Edit: Perhaps the middle option should be "Clip **All** Maps to Bounds", for clarity?
Would it be possible to disable the Other maps option and maybe change the text of these options when there are no Worlds loaded? I can see those wordings being confusing to someone who's never used Worlds.

bjorn:
> Should I open a separate issue for this, or would it suffice to leave the existing one open?

I think it would be better to create a new issue for that specific option, and it would indeed be good to see some screenshots to illustrate this need.

> Edit: Perhaps the middle option should be "Clip **All** Maps to Bounds", for clarity?
> Would it be possible to disable the Other maps option and maybe change the text of these options when there are no Worlds loaded? I can see those wordings being confusing to someone who's never used Worlds.

It could get a little strange for this menu to change based on that context, especially since one could officially still have it set to "Clip Other Maps to Bounds" but with the world unloaded.

JustJ01:
Hi @bjorn , I have made the changes as per you requested.

Features implemented:
1. Unified Clipping System
   - Replaced per-layer (tile/image/object) clipping with a single unified system at the MapItem level
   - Clipping is applied at the map level, automatically affecting all layers and objects within that map
2. View Menu: Map Bounds Clipping
   - Adds new submenu: View → Map Bounds Clipping
   - Three mutually exclusive options:
     - No Map Clipping - Default behavior, maps can overflow
     - Clip All Maps to Bounds - All maps are clipped to their visible area
     - Clip Other Maps to Bounds - Only the currently selected map is un-clipped; other maps in world view are clipped
3. World-Aware Clipping
   - When using "Clip Other Maps", the editor detects which map is currently active
   - Switching between maps in a world automatically updates clipping state
   
   

https://github.com/user-attachments/assets/74ef417a-154c-4579-9bbb-66de29ddf23c


https://github.com/user-attachments/assets/38187b56-b62d-4eb9-ae23-fa8302e5dacd

Issue #4116, opened 2024-12-05
"Contiguous" option for Bucket Fill and Magic Wand tools


eishiya:
### Request
It would be nice to have a "contiguous" toggle for the Bucket Fill and Magic Wand, allowing them to affect tiles in non-contiguous region. Currently, non-contiguous fills require first selecting all those tiles and then using Shift+Bucket Fill to fill those areas, but with a "contiguous" toggle, this behaviour could be achieved without any selections, without additional tools.

This toggle should, if possible, be a checkbox with the label "Contiguous" next to it, rather than an icon like the other tool options, as there does not appear to be any iconography associated with this concept. Every program I've ever seen describes this option only as text.

### Justification
Non-contiguous bucket filling is useful for replacing tiles within a layer. For example, one may want it for making similar sets of Automapping rules, for replacing shallow water with deep water, filling flat areas with a pattern/random tiles, and so on. It would be nice to not need extra tools and a modifier key to do this :] If we get Terrain Mode on the bucket fill, non-contiguous bucket filling will get even more useful, allowing quick, terrain-aware replacement of one terrain with another, and selections would likely get in the way of that feature more than plain tile Bucket Fill.

Tiled already has the Select Same Tile tool for non-contiguous selections, but I feel that if Bucket FIll has a contiguous toggle, then so should Magic Wand for consistency. This would make Select Same Tile obsolete.
The benefits of this is that it'll remove a tool from the toolbar and bring the tooling more in line with other software like Aseprite (but further away from GIMP, after which these tools were modelled). The drawback is that if a user needs to frequently alternate between the two selection modes and other tools, it may require extra clicks, since they'll need to first click the Magic Wand Tool, and then check/change the Contiguous option. In my experience in other software, this has not been a problem.
I'll miss the Select Same Tile icon though, it's probably my favourite in Tiled :]

### Comparison to other programs
GIMP works just like Tiled: Separate contiguous and non-contiguous selection tools. Bucket Fill is only Contiguous, but with a non-contiguous selection, you can shift+Bucket Fill to fill it (provided you have the "fill selection" option chosen for the Shift behaviour).
Krita has separate contiguous and non-contiguous selection tools, but a contiguous toggle on its Fill tool. An odd inconsistency.
Aseprite and Photoshop have contiguous toggles on both their Magic Wand and Paint Bucket tools, like what I'm requesting here.

Bjorn noted on Discord,
[...]

PR #4352, opened 2026-02-22
Add Contiguous option to Magic Wand tool (part of #4116)

SIDDHAARTHAA:
This adds a **Contiguous** option to the Magic Wand tool as a first step toward #4116.

With the option enabled (default), the tool keeps the current flood-fill behavior. When disabled, it selects all matching tiles on the layer, similar to the existing Select Same Tile tool. I kept the scope limited to Magic Wand so the approach can be reviewed before continuing with Bucket Fill and the removal of the Select Same Tile tool.

A short demo is attached showing the behavior.
tiled-pr-1 0-

### DIFF
--- a/src/tiled/magicwandtool.cpp
+++ b/src/tiled/magicwandtool.cpp
@@ -28,7 +28,6 @@
 #include "map.h"
 #include "mapdocument.h"
 
-#include <algorithm>
 #include <QCheckBox>
 #include <QToolBar>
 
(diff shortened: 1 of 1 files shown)


SIDDHAARTHAA:
@eishiya @bjorn
I’ve opened this PR as the first step for #4116 .I’d really appreciate any feedback on the approach before I continue with the next steps.

bjorn:
Closing this in favor of merging these changes as part of #4356. Thanks for your efforts, @SIDDHAARTHAA!

PR #4354, opened 2026-02-22
Fix performance issue when painting over existing tiles on large maps

rhythmcache:
ref https://github.com/mapeditor/tiled/issues/4185

I was able to reproduce this issue on a 1024×1024 map using collection based tilesets. Painting over existing tiles with the stamp brush causes significant slowdown.

The issue happens because TileLayer::usedTilesets() was recomputed through drawMargins() during repaint. Since overwriting a tile marked the layer dirty, this triggered a full scan of all tiles repeatedly while dragging the brush.

This change removes the dirty flag full rescan approach and replaces it with incremental tileset reference counting. Tileset usage is now updated in O(1) during setCell(), and bulk operations that bypass setCell() rebuild the reference data when necessary.

### DIFF
--- a/src/libtiled/tilelayer.cpp
+++ b/src/libtiled/tilelayer.cpp
@@ -131,7 +131,6 @@ TileLayer::TileLayer(const QString &name, int x, int y, int width, int height)
     : Layer(TileLayerType, name, x, y)
     , mWidth(width)
     , mHeight(height)
-    , mUsedTilesetsDirty(false)
 {
 }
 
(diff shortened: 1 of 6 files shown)


bjorn:
> The issue happens because TileLayer::usedTilesets() was recomputed through drawMargins() during repaint. Since overwriting a tile marked the layer dirty, this triggered a full scan of all tiles repeatedly while dragging the brush.

This is indeed a rather unfortunate scenario, where the re-computation of `mUsedTilesets` wasn't even necessary most of the time. Both because the old tileset was likely still used somewhere, as well as because even if the set of used tilesets changed, it usually doesn't change the draw margins.

Keeping a reference count is the most straight-forward fix. There will be a bit more overhead in `setCell`, but it reliably avoids the expensive iterating of all tiles in the layer.

I think the implementation could be improved a bit, though.

* I'd suggest to keep only a single container, `QHash<SharedTileset, int> mUsedTilesets`. It means `TileLayer::usedTilesets` should probably return `{ mUsedTilesets.keyBegin(), mUsedTilesets.keyEnd() }`, but I don't think the overhead of constructing a set on-demand will be a problem.

* The  `removeReferencesToTileset` and `replaceReferencesToTileset` don't need to fall back to rebuilding the refcounts. Instead they can perform very similar actions they did before:
    * `removeReferencesToTileset` just removes that entry from the hash map.
    * `replaceReferencesToTileset` takes the refcount from the old tileset out of the hash map and inserts it for the new tileset.

After the above changes, we can remove the unused `rebuildTilesetRefs` and inline `addTilesetRef` and `removeTilesetRef` in `setCell`, since they won't be called from anywhere else anyway.

rhythmcache:
i have pushed an update based on your feedback.
Now using a single QHash<SharedTileset, int> and inlined the refcount updates in setCell. Also removed the rebuild path and updated the bulk operations to adjust the hash directly.

For referencesTileset, I handled the const/non-const mismatch with a qSharedPointerConstCast and added a comment explaining it.

PR #4356, opened 2026-02-23
Add Contiguous Toggle to Bucket Fill and Magic Wand; Remove Select Same Tile Tool

### LINKED ISSUE #4116: "Contiguous" option for Bucket Fill and Magic Wand tools

SIDDHAARTHAA:
Closes #4116 

Adds a "Contiguous" toggle to Magic Wand and Bucket Fill.

Magic Wand:
On (default): selects connected matching tiles
Off: selects all matching tiles on the layer

Bucket Fill:
On (default): fills connected matching tiles
Off: fills all matching tiles on the layer

Includes an Alt modifier to temporarily toggle the mode.

Removes Select Same Tile tool.

magic_wand_tool

tiled-pr-2 0

### DIFF
--- a/src/tiled/magicwandtool.cpp
+++ b/src/tiled/magicwandtool.cpp
@@ -43,6 +43,20 @@ MagicWandTool::MagicWandTool(QObject *parent)
 {
 }
 
+void MagicWandTool::modifiersChanged(Qt::KeyboardModifiers modifiers)
+{
+    AbstractTileSelectionTool::modifiersChanged(modifiers & ~Qt::AltModifier);
+
+    mModifiers = modifiers;
+
+    const bool contiguous = effectiveContiguous();
+
+    updateContiguousCheckBox();
+
+    if (mLastContiguousStatus != contiguous)
+        tilePositionChanged(tilePosition());
+}
+
 void MagicWandTool::tilePositionChanged(QPoint tilePos)
 {
     // Make sure that a tile layer is selected
@@ -55,6 +69,7 @@ void MagicWandTool::tilePositionChanged(QPoint tilePos)
         mMatchCells.clear();
 
     TilePainter tilePainter(mapDocument(), tileLayer);
+    const bool contiguous = effectiveContiguous();
 
     const Cell targetCell = tilePainter.cellAt(tilePos);
     if (!mMatchCells.contains(targetCell))
@@ -64,7 +79,9 @@ void MagicWandTool::tilePositionChanged(QPoint tilePos)
         return mMatchCells.contains(cell);
     };
 
-    if (mContiguous) {
+    mLastContiguousStatus = contiguous;
+
+    if (contiguous) {
         // Flood fill (default magic wand behavior)
         setSelectionPreview(tilePainter.computeFillRegion(tilePos, condition));
     } else {
@@ -91,20 +108,37 @@ void MagicWandTool::populateToolBar(QToolBar *toolBar)
 {
     AbstractTileSelectionTool::populateToolBar(toolBar);
 
-    auto contiguousCheckBox = new QCheckBox(tr("Contiguous"), toolBar);
-    contiguousCheckBox->setChecked(mContiguous);
-    connect(contiguousCheckBox, &QCheckBox::toggled, this, &MagicWandTool::setContiguous);
-    toolBar->addWidget(contiguousCheckBox);
+    mContiguousCheckBox = new QCheckBox(tr("Contiguous"), toolBar);
+    updateContiguousCheckBox();
+    connect(mContiguousCheckBox, &QCheckBox::clicked, this, &MagicWandTool::setContiguous);
+    toolBar->addWidget(mContiguousCheckBox);
 }
 
 void MagicWandTool::setContiguous(bool contiguous)
 {
-    if (mContiguous == contiguous)
+    const bool persistentContiguous = (mModifiers & Qt::AltModifier) ? !contiguous : contiguous;
+
+    if (mContiguous == persistentContiguous)
         return;
 
-    mContiguous = contiguous;
+    mContiguous = persistentContiguous;
+
+    updateContiguousCheckBox();
     mMatchCells.clear();
     tilePositionChanged(tilePosition());
+}
+
+bool MagicWandTool::effectiveContiguous() const
+{
+    return mContiguous != static_cast<bool>(mModifiers & Qt::AltModifier);
+}
+
+void MagicWandTool::updateContiguousCheckBox()
+{
+    if (!mContiguousCheckBox)
+        return;
+
+    mContiguousCheckBox->setChecked(effectiveContiguous());
 }
 
 #include "moc_magicwandtool.cpp"
--- a/src/tiled/magicwandtool.h
+++ b/src/tiled/magicwandtool.h
@@ -27,6 +27,9 @@
 
 #include "tilelayer.h"
 
+#include <QPointer>
+
+class QCheckBox;
 class QToolBar;
 
 namespace Tiled {
@@ -43,6 +46,7 @@ class MagicWandTool final : public AbstractTileSelectionTool
 public:
     MagicWandTool(QObject *parent = nullptr);
 
+    void modifiersChanged(Qt::KeyboardModifiers modifiers) override;
     void languageChanged() override;
     void populateToolBar(QToolBar *toolBar) override;
 
@@ -51,9 +55,14 @@ protected:
 
 private:
     void setContiguous(bool contiguous);
+    bool effectiveContiguous() const;
+    void updateContiguousCheckBox();
 
+    Qt::KeyboardModifiers mModifiers = Qt::NoModifier;
+    bool mLastContiguousStatus = true;
     QVector<Cell> mMatchCells;
     bool mContiguous = true;
+    QPointer<QCheckBox> mContiguousCheckBox;
 };
 
 } // namespace Tiled


SIDDHAARTHAA:
@bjorn @eishiya Please take a look at this at your free time, this is supposed to be the second part of #4116 ( first part has also been implemented, could you look at that too, thanks. )

Let me know if i can proceed with the next steps ;)

src/tiled/bucketfilltool.cpp
```diff
@@ -109,7 +116,31 @@ void BucketFillTool::tilePositionChanged(QPoint tilePos)
...
+                        resultRegion = tileLayer->region(condition);
+
+                        const bool hasEmptyCell = std::any_of(mMatchCells.begin(),
+                                                              mMatchCells.end(),
+                                                              [](const Cell &cell) { return cell.isEmpty(); });
+                        if (hasEmptyCell) {
+                            QRegion emptyRegion = infinite ? tileLayer->bounds() : tileLayer->rect();
+                            emptyRegion -= tileLayer->region();
+                            resultRegion += emptyRegion;
+                        }
```
bjorn: Since you plan to use this code in both the Bucket Fill tool and the Magic Wand tool, I think it's time to move it to `TilePainter`, for example as `TilePainter::computePaintableRegion`.

SIDDHAARTHAA:
>I'm still not really sold on this checkbox option, especially not for the Magic Wand vs. Select Same Tile because them seem like rather different tools to me, but at least for Bucket Fill I can see the use because it makes replacing a tile across the entire map that much faster. But thanks for implementing it in both tools, so I can play around with it and maybe get used to it!

I agree that the use case is most obvious for Bucket Fill. 
But, once a “non-contiguous vs contiguous” concept exists in the editor, it tends to become a general selection/fill behavior rather than something tied to a single tool. Having it behave consistently across tools avoids users needing to mentally map slightly different behaviors for what is essentially the same spatial concept. Even if the Magic Wand use case feels less compelling right now, it keeps the interaction model predictable.

>Would the Alt modifier still be available for both Magic Wang and Bucket Fill tools, in combination with left-click, to quickly toggle the Contiguous mode?

Yes, Alt+Left-click could be used as a temporary Contiguous toggle for both tools without conflicting with existing shortcuts. The checkbox would remain the primary control.

src/tiled/bucketfilltool.cpp
SIDDHAARTHAA: Yess, I noticed that. I’ll move it to `TilePainter`.

SIDDHAARTHAA:
If this direction sounds good, I’ll move the shared logic into `TilePainter::computePaintableRegion` in this PR and use it here. Then I’ll update the Magic Wand PR to rely on that shared function and remove the duplicated logic there. I’ll also proceed with removing the Select Same Tile tool as discussed.
thankyou for your insights @bjorn

SIDDHAARTHAA:
@bjorn I’ve implemented the changes.
I’ve also removed the Select Same Tile tool (the documentation still references it, I’ll update that next).
I also added Alt as a temporary toggle for the Contiguous option, and the checkbox updates live while it’s held.
Once this PR is merged, I’ll update the Magic Wand tool accordingly to use the shared implementation.

src/tiled/bucketfilltool.cpp
```diff
@@ -109,7 +117,17 @@ void BucketFillTool::tilePositionChanged(QPoint tilePos)
...
                     return mMatchCells.contains(cell);
                 };
-                mFillRegion = regionComputer.computePaintableFillRegion(tilePos, condition);
+                if (contiguous) {
+                    mFillRegion = regionComputer.computePaintableFillRegion(tilePos, condition);
+                } else {
+                    const QPoint localPos = tilePos - tileLayer->position();
+                    QRegion resultRegion;
+
+                    if (mapDocument()->map()->infinite() || tileLayer->contains(localPos))
```
dogboydog, line 125: I'm not familiar enough with this part of the code to know if this would be an issue, but if contiguous is off and neither of these conditions apply, resultRegion will be null, right?   Should we default it to `QRegion()` if these conditions don't apply? 

I tried turning contiguous off and clicking on tiles outside of the map's bounds, but didn't notice any errors or anything 

@bjorn

src/tiled/tilepainter.h
```diff
@@ -91,6 +91,19 @@ class TilePainter
     QRegion computePaintableFillRegion(QPoint fillOrigin,
                                        std::function<bool(const Cell &)> condition) const;
 
+    /**
+     * Computes a region containing all cells for which the given \a condition
```
dogboydog, line 95: Is the `\a` here a typo?

dogboydog:
>I'm still not really sold on this checkbox option, especially not for the Magic Wand

I personally find a  contiguous checkbox to be an  intuitive choice for the magic wand tool. Aseprite has the continuous checkbox on the magic wand tool.  I believe Photoshop does as well

Also, tested this out and it seems to work great in all of the scenarios I tried

src/tiled/tilepainter.h
SIDDHAARTHAA: actually, `\a` is a Doxygen parameter marker, not a typo; it highlights argument names in comments/docs. Example: `src/tiled/documentmanager.h:167`.
it's used across the codebase, i followed it

src/tiled/bucketfilltool.cpp
SIDDHAARTHAA: Thankyou for pointing out. I think this is already safe because `resultRegion` default-constructs to an empty `QRegion`.

```cpp
mFillRegion = resultRegion;

if (mFillRegion.isEmpty())
    return;
```

If the bounds check fails, `resultRegion` stays empty, `mFillRegion` becomes empty, and the function exits early

dogboydog: Thanks for the follow up. Makes sense to me.

src/tiled/tilepainter.h
dogboydog: Makes sense, thanks

dogboydog (approved):
The approach here makes sense to me - for this and https://github.com/mapeditor/tiled/pull/4352 , we should update NEWS.md with a summary of the changes. Especially since here we are removing the Select Same Tile  tool.  Besides NEWS.md we will need to update the docs once this is in a release to remove references to the Select Same Tile tool.  Thanks for this enhancement

SIDDHAARTHAA:
I'll update the NEWS.md for this particular PR. However, PR #4352 has some changes that depend on this being merged. So after this PR gets merged, I'll update the NEWS.md for that PR and also update the docs.

src/tiled/tilepainter.h
bjorn: Thanks for your reviews and questions, @dogboydog! Even if in some cases no change is needed, it's good to see @SIDDHAARTHAA able to explain why and you learn a bit in the process as well. :)

src/tiled/bucketfilltool.cpp
```diff
@@ -222,6 +244,44 @@ void BucketFillTool::languageChanged()
...
+{
+    return mContiguous != static_cast<bool>(mModifiers & Qt::AltModifier);
+}
+
+void BucketFillTool::updateContiguousCheckBox()
+{
+    if (!mContiguousCheckBox)
+        return;
+
+    const QSignalBlocker blocker(mContiguousCheckBox);
```
bjorn: Could you switch to using the `triggered` signal when connecting to the checkbox, so that we don't need to use a signal blocker here?

bjorn:
> I'll update the NEWS.md for this particular PR. However, PR #4352 has some changes that depend on this being merged. So after this PR gets merged, I'll update the NEWS.md for that PR and also update the docs.

Actually PR #4352 will be even smaller than it already is with the addition of `TilePainter::computePaintableRegion`. I would suggest to rebase that change onto this branch (adding that change to this PR). It seems to me this should make things easier since it's all very related.

There is a bit of time anyway, since due to Tiled 1.12 being in string freeze now, I would in the end only approve this PR but wait with merging until `1.12` has branched off from `master` (which will happen by 12th of March latest).

SIDDHAARTHAA:
>Actually PR https://github.com/mapeditor/tiled/pull/4352 will be even smaller than it already is

Indeed, I just didn't know I could do that. Thankyou for your word!!

>I would suggest to rebase that change onto this branch

src/tiled/bucketfilltool.cpp
SIDDHAARTHAA: Okay

SIDDHAARTHAA:
I rebased `feature/magic-wand-contiguous` onto `feature/bucket-fill-contiguous` and updated Magic Wand to use the shared `TilePainter` region logic (`computeRegion`) for non-contiguous selection. Merge order should be `feature/bucket-fill-contiguous` first, then `feature/magic-wand-contiguous`.

Thankyou:-)

bjorn:
> Merge order should be `feature/bucket-fill-contiguous` first, then `feature/magic-wand-contiguous`.

If you don't mind I'll just close #4352 now, because these PRs are now the same and this one is where all discussion happened. Thanks for bringing these changes together!

SIDDHAARTHAA:
Indeed, I was about to do it.
Thankyou for your attention

src/tiled/selectsametiletool.cpp
```diff
@@ -1,86 +0,0 @@
...
-
-#include <QGraphicsSceneMouseEvent>
-
-using namespace Tiled;
-
-SelectSameTileTool::SelectSameTileTool(QObject *parent)
-    : AbstractTileSelectionTool("SelectSameTileTool",
-                                tr("Select Same Tile"),
-                                QIcon(QLatin1String(
-                                      ":images/22/stock-tool-by-color-select.png")),
```
bjorn, line 34: Please also remove this image, I don't think we use it anywhere after this removal.

SIDDHAARTHAA: yes, there are many other related things regarding that tool, like translations, manual. should I clean em up?

src/tiled/bucketfilltool.cpp
bjorn: Ah, or `clicked` as is the case for `QCheckBox`, but you found it. :-)

src/tiled/selectsametiletool.cpp
bjorn: The translations will go away as part of the string update that I will do before the next new features release (this could go into Tiled 1.13). I see you've already handled the manual, so I think we got all references now.

SIDDHAARTHAA:
Yes, that piece was missing on my side. I got a bit confused because of the rebase and the related branch work and thought I had already implemented it there, but I hadn’t. I’ve fixed it now and also rebased the branch.

Thankyou!

SIDDHAARTHAA:
@bjorn 
I've updated the PR description to reflect the current behavior (both Magic Wand and Bucket Fill changes). No code changes.

PR #4358, opened 2026-02-24
Scripting: Add optional defaultValue and toolTip params to Dialog add widget methods

Oval17:
Resolves : #4014 

This  adds optional `defaultValue` and `toolTip` parameters to all `Dialog` add-widget methods in the scripting API, making it easier to create fully configured widgets with less code and fewer intermediate variables.

With this change, you can now do it in a single expression:

```js
dialog.addColorButton("My Color", defaultColor, "My explanatory tooltip")
    .colorChanged.connect(function(color) { /* do stuff */ });
```

# ScriptDialog Enhancements

### `src/tiled/scriptdialog.h` & `src/tiled/scriptdialog.cpp`

Added optional parameters to all `add[Widget]` methods:

| Method            | New Parameters |
|-------------------|----------------|
| `addTextInput`    | `toolTip` |
| `addTextEdit`     | `toolTip` |
| `addNumberInput`  | `defaultValue` (double, default `0.0`), `toolTip` |
| `addSlider`       | `defaultValue` (int, default `0`), `toolTip` |
| `addCheckBox`     | `toolTip` |
| `addComboBox`     | `defaultIndex` (int, default `0`), `toolTip` |
| `addColorButton`  | `defaultValue` (`QColor`, default empty), `toolTip` |
| `addFilePicker`   | `defaultValue` (`QString`), `toolTip` |
| `addButton`       | `toolTip` |
| `addLabel`        | `toolTip` |
| `addImage`        | `toolTip` |

###  Backward Compatibility

All new parameters are optional (have default values), ensuring that existing scripts remain fully backward-compatible.

---

### `docs/scripting-doc/index.d.ts`

Updated all TypeScript type declarations to document the new optional parameters, updated one minor typo as well

---

## some impNotes

- Signal handler parameters were intentionally left out of scope, as discussed in the issue.
- `.connect()` remains the cleaner and recommended approach for handling signals.
- All parameters use sensible defaults, so omitting them is completely safe.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -29,6 +29,7 @@
 * Fixed undo behavior after resizing objects certain ways (by Kanishka, #4339)
 * Scripting: Added API for custom property types (with dogboydog, #3971)
 * Scripting: Added TileMap.chunkSize and TileMap.compressionLevel properties
+* Scripting: Added optional defaultValue and toolTip params to Dialog add widget methods (by Oval, #4358)
 * Scripting: Added tiled.session to read and write session properties (by Kanishka, #4345)
 * Scripting: Added MapEditor.selectedTool and MapEditor.tool (#4330)
 * Scripting: Fixed the `fileName` property of map/tileset passed to `FileFormat.write` (by Shuvam Pal, #4359)
--- a/docs/scripting-doc/index.d.ts
+++ b/docs/scripting-doc/index.d.ts
@@ -5694,7 +5694,7 @@ declare class Dialog extends Qt.QWidget {
    * Add a label to the dialog with the given text. A label will always be the
    * first widget in a row.
    *
-   * The optional toolTip parameter sets a tooltip on the label widget.
+   * The optional toolTip parameter sets a tooltip on the label (since 1.12).
    */
   addLabel(labelText: string, toolTip?: string): Qt.QLabel;
 
@@ -5707,7 +5707,7 @@ declare class Dialog extends Qt.QWidget {
   /**
    * Adds an image widget that can display an image in a dialog.
    *
-   * The optional toolTip parameter sets a tooltip on the image widget.
+   * The optional toolTip parameter sets a tooltip on the widget (since 1.12).
    */
   addImage(labelText: string, image: Image, toolTip?: string): ImageWidget;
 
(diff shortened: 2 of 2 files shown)


Oval17:
This is ready for review, inputs and changes @bjorn  @eishiya 
Thanks

src/tiled/scriptdialog.cpp
```diff
@@ -207,65 +210,100 @@ QWidget *ScriptDialog::addSeparator(const QString &labelText)
...
+QWidget *ScriptDialog::addColorButton(const QString &labelText, const QColor &defaultValue, const QString &toolTip)
 {
-    QWidget *colorButton = addDialogWidget(new ColorButton(this), labelText);
-    colorButton->setSizePolicy(QSizePolicy::Fixed, QSizePolicy::Fixed);
-    return colorButton;
+    ColorButton *colorButton = new ColorButton(this);
+    if (defaultValue.isValid())
+        colorButton->setColor(defaultValue);
+    if (!toolTip.isEmpty())
+        colorButton->setToolTip(toolTip);
```
bjorn: Since these two lines are now everywhere before the call to `addDialogWidget`, please add a `widgetToolTip` parameter to `addDialogWidget` to reduce code duplication.

src/tiled/scriptdialog.cpp
```diff
@@ -207,65 +210,100 @@ QWidget *ScriptDialog::addSeparator(const QString &labelText)
...
+        spinBox->setToolTip(toolTip);
+    return addDialogWidget(spinBox, labelText);
 }
 
-QWidget *ScriptDialog::addSlider(const QString &labelText)
+QWidget *ScriptDialog::addSlider(const QString &labelText, int defaultValue, const QString &toolTip)
 {
     QSlider *horizontalSlider = new QSlider(this);
     horizontalSlider->setOrientation(Qt::Horizontal);
+    horizontalSlider->setValue(defaultValue);
```
bjorn, line 244: This one is a little problematic because the value will be clamped between the minimum and maximum value of the slider (the defaults are 0 and 99 according to [QAbstractSlider](https://doc.qt.io/qt-6/qabstractslider.html#QAbstractSlider)). Either we should add at least also the maximum value as parameter or we should make sure to call `setMaximum` with `qMax(defaultValue, horizontalSlider->maximum()` first.

src/tiled/scriptdialog.cpp
Oval17: Added 4th parameter `widgetToolTip` to `addDialogwidget`

src/tiled/scriptdialog.cpp
Oval17: As suggested expanded the maximum to fit the defaultValue before setting it.

bjorn (approved):

PR #4359, opened 2026-02-24
fix: pass source map path to custom export callbacks

### LINKED ISSUE #4172: tiled.registerMapFormat - source map missing value of fileName
**Describe the bug**
When i create new map format i need access to source map file name (to find word where map belongs, as example)

**To Reproduce**
register new map format:
```js
tiled.registerMapFormat("world-lut", {
  name: "LUT World(s) image",
  extension: "png *.bin",
  write: function(p_map, p_fileName) {
    console.log(`the fileName: "${p_map.fileName}"`, typeof (p_map.fileName))
  }
})
```
export, use it to export map - and look at console output:
```
qml: the fileName: "" string
```

**Expected behavior**
in console you see 
```
qml: the fileName: "path_to_map_file.tmj" string
```

**Specifications:**
 - OS: MacOs
 - Tiled Version: 1.11.2

unichronic:
fixes #4172 
`registerMapFormat` exporters could not get the original map file path during export (`map.fileName` was empty in `write/outputFiles` for detached assets).

now we pass the source map path as an optional extra callback argument for scripted map formats:

backward compatibility is preserved as callbacks keep receiving 3 args. source path is only passed when the callback supports it (or opts in via `usesSourceFileName`)

### DIFF
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -19,7 +19,7 @@ on:
     - '.travis.yml'
 
 env:
-  QBS_VERSION: 2.4.1
+  QBS_VERSION: 3.1.1
   CMAKE_VERSION: 3.31
   SENTRY_VERSION: 0.12.8
   SENTRY_ORG: mapeditor
@@ -80,7 +80,7 @@ jobs:
     - name: Install Qt
       uses: jurplel/install-qt-action@v4
       with:
-        version: 6.9.3
+        version: 6.10.1
         arch: linux_gcc_64
         modules: "qtimageformats"
         tools: 'tools_qtcreator'
@@ -211,7 +211,7 @@ jobs:
           version_suffix: "10.13-10.15"
           architectures: x86_64
           cmake_architectures: x86_64
-        - qt_version: 6.9.3
+        - qt_version: 6.10.1
           qt_modules: "qtimageformats"
           version_suffix: "11+"
           architectures: x86_64,arm64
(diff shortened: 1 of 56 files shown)


unichronic:
This is kind of an workaround to fix #4172 but making the fix complete also requires modifying `EditableAsset::fileName()` in `editableasset.cpp` like this:
```
QString EditableAsset::fileName() const
{
    if (document())
        return document()->fileName();

    const Object *obj = object();
    if (obj && obj->typeId() == Object::MapType)
        return static_cast<const Map*>(obj)->fileName;

    return QString();
}
```
I wasn't sure if i should do this. What do you think?

bjorn:
Thanks for trying to solve this, @unichronic!

Unfortunately, this leaves the API confusing because people will just expect the `fileName` property to return something meaningful. It also deviates the JS API from the C++ API.

And the `EditableAsset::fileName` change is indeed too hacky.

But what about we make this `EditableAsset::fileName` virtual so that we can override it for the `EditableMap`? For tilesets I think `EditableTileset::fileName()` may need to access `Tileset::originalTileset`.

unichronic:
so something like this?
```
QString EditableTileset::fileName() const
{
    if (auto doc = tilesetDocument())
        return doc->fileName();

    const auto originalTileset = const_cast<Tileset*>(tileset())->originalTileset();
    return originalTileset ? originalTileset->fileName() : QString();
}
```
and 
```QString fileName() const override;```

bjorn:
> something like this?

Yeah, that's what I had in mind, and it seems to work fine. :-)

I just tweaked it a little to rely on the implementation in the superclass. Thanks for fixing this!

PR #4360, opened 2026-02-24
TilesetDock: Fallback to filename when tileset name is empty

### LINKED ISSUE #4355: TSX Title Shrinks Too Much
[GOA Source (1).zip](https://github.com/user-attachments/files/25502103/GOA.Source.1.zip)

 Describe the bug

When you have a sufficiently high number of TSX files for a map, the title canvas shrinks far too much, making it impossible to see the file name. For my case, this occurs when I have 159 unique TSX files. 

Image

### Steps to reproduce

1. Create a TMX file.
2. Create 159 unique TSX files that will populate the map. (You likely need less)
3. Cry as you are no longer able to see the file names.

### Expected behavior

There should be a minimum size enforced for the QtWidget such that you can always at least see the first 5 characters for a file.

### OS version

Windows 11

### Tiled version

1.11.2

[EDIT] - To add zip to the main post.

SIDDHAARTHAA:
Closes #4355

Fallback to the tileset file name when the tileset name is empty, so TilesetDock tabs never show blank titles.

---

Before:
screenshot-2026-02-24_16-02-17

After:
screenshot-2026-02-24_16-02-03

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -22,6 +22,7 @@
 * Added export plugin for Remixed Dungeon (by Mikhael Danilov, #4158)
 * Added "World > World Properties" menu action (with dogboydog, #4190)
 * Added Delete shortcut to Remove Tiles action by default and avoid ambiguity (#4201)
+* Fixed tileset tabs to fall back to filename in case of unnamed tilesets (by Sid, #4360)
 * Fixed alpha component of tint color not applying correctly to opaque images (by Roland Helmerichs, #4310)
 * Scripting: Added API for custom property types (with dogboydog, #3971)
 * Scripting: Added TileMap.chunkSize and TileMap.compressionLevel properties


MisterMeme95:
Thanks for the update, brother. The only critique is that I'd add a few comments. For a huge project like this, it helps to have a few lines to signify what the function is doing, even if it's pretty simple.

SIDDHAARTHAA:
Got it, thanks for the feedback 🙂 I’ll add some comments to the code shortly.

SIDDHAARTHAA:
I noticed most helper functions in this part of the codebase don’t currently have comments, so I left it consistent for now. If @bjorn approves adding comments here, I’m happy to add them :-)

bjorn:
Yeah I don't think a comment is necessary because the name and implementation of the function already explain quite well what it is doing. I've just changed the code to rely on `displayName` since it seems to serve our purposes better (it's also what is used to set the tab text for the open documents).

There is one remaining issue, where the tab text fails to update for embedded tilesets where the file name of the map the tileset is embedded in changes. But since we're dealing with a rather rare case here anyway, I don't think it's worth the complexity to handle that case. With embedded tilesets, the name will also refresh anyway when switching between maps.

So I think this can be merged, thanks @SIDDHAARTHAA!

SIDDHAARTHAA:
Thankyou :-]

PR #4363, opened 2026-02-25
Avoid assertion on shutdown in debug builds

bjorn:
When the MainWindow is destroyed, SpaceBarEventFilter gets called after MainWindow::mInstance has been set to nullptr.

Small mistake I made in b979729bab3edf06e29b63a092cf8defe6fb6186.


bjorn:
@Oval17 Oh well, we missed to notice an assertion in debug builds.

Oval17:
@bjorn Good catch,
Should have tested in debug mode when reviewing 🤔 will keep that in mind.
Thanks!!

PR #4364, opened 2026-02-25
Fix snapping mode sync across instances

### LINKED ISSUE #3788: Multiple instances of Tiled can cause incorrect Snapping
When using the Offset Layers tool, it always snaps to the grid, regardless of the current Snapping setting. I think this tool should respect the current Snapping option.

For now, it's possible to work around this by adjusting the Offsets directly in the Layer Properties.

Observed in 1.10.1. I suspect this is a regression, though I don't use this tool often enough to have noticed when it happened. This was reported by a user on Discord.

SIDDHAARTHAA:
Closes #3788

This changes snapping preferences to use a single `Interface/SnapMode` enum instead of three separate boolean flags. Legacy `SnapToGrid` / `SnapToFineGrid` / `SnapToPixels` values are migrated on startup with the same priority behavior as before. The snapping menu now updates from `snapMode` consistently and refreshes on window activation / menu open, so multiple Tiled instances stay in sync with `tiled.conf`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -28,6 +28,7 @@
 * Fixed panning with space bar not always working on first click (with Oval, #4338)
 * Fixed undo behavior after resizing objects certain ways (by Kanishka, #4339)
 * Fixed suggesting filename with trailing dot when export filter is unset (by Sid, #4368)
+* Fixed snapping mode sync across instances (by Sid, #4364)
 * Scripting: Added API for custom property types (with dogboydog, #3971)
 * Scripting: Added TileMap.chunkSize and TileMap.compressionLevel properties
 * Scripting: Added optional defaultValue and toolTip params to Dialog add widget methods (by Oval, #4358)


SIDDHAARTHAA:
@bjorn Please review it whenever possible. I've tried my best to write good code :)
I have also rigorously tested it.

src/tiled/mainwindow.cpp
```diff
@@ -412,6 +412,7 @@ MainWindow::MainWindow(QWidget *parent, Qt::WindowFlags flags)
             mTilesetEditor->templatesDock(), &TemplatesDock::tryOpenTemplate);
 
     auto snappingGroup = new QActionGroup(this);
+    snappingGroup->setExclusive(true);
```
bjorn: Exclusive is the default so we don't need this.

src/tiled/mainwindow.cpp
```diff
@@ -564,12 +567,22 @@ MainWindow::MainWindow(QWidget *parent, Qt::WindowFlags flags)
...
+            preferences->setSnapMode(SnapMode::Grid);
+    });
+    connect(mUi->actionSnapToFineGrid, &QAction::toggled, preferences, [preferences](bool checked) {
+        if (checked)
+            preferences->setSnapMode(SnapMode::FineGrid);
+    });
+    connect(mUi->actionSnapToPixels, &QAction::toggled, preferences, [preferences](bool checked) {
+        if (checked)
+            preferences->setSnapMode(SnapMode::Pixels);
+    });
```
bjorn, line 573: Please instead of connecting to the `triggered` signal for each action, connect to the `QActionGroup::triggered` signal of the `actionGroup`.

To get the snap mode from the triggered action, you can do:
```c++
mUi->actionSnapNothing->setData(QVariant::fromValue(SnapMode::None));
```
And in the lambda handling the `triggered` signal:
```c++
connect(snappingGroup, &QActionGroup::triggered, preferences, [preferences](QAction *action) {
    preferences->setSnapMode(action->data().value<SnapMode>());
});
```

src/tiled/mainwindow.cpp
```diff
@@ -423,11 +424,14 @@ MainWindow::MainWindow(QWidget *parent, Qt::WindowFlags flags)
...
     mUi->actionEnableParallax->setChecked(preferences->parallaxEnabled());
-    mUi->actionSnapToGrid->setChecked(preferences->snapToGrid());
-    mUi->actionSnapToFineGrid->setChecked(preferences->snapToFineGrid());
-    mUi->actionSnapToPixels->setChecked(preferences->snapToPixels());
+    updateSnappingActions(preferences->snapMode());
     mUi->actionHighlightCurrentLayer->setChecked(preferences->highlightCurrentLayer());
     mUi->actionHighlightHoveredObject->setChecked(preferences->highlightHoveredObject());
+    connect(preferences, &Preferences::snapModeChanged, this, &MainWindow::updateSnappingActions);
+    connect(mUi->menuSnapping, &QMenu::aboutToShow, this, [this, preferences] {
+        preferences->sync();
```
bjorn: This change adds three calls to `QSettings::sync` in various places, but its documentation says:

> This function is called automatically from QSettings's destructor and by the event loop at regular intervals, so you normally don't need to call it yourself.

So I don't think we need these?

src/tiled/mainwindow.cpp
```diff
@@ -2016,6 +2031,19 @@ void MainWindow::autoMappingWarning(bool automatic)
     }
 }
 
+void MainWindow::updateSnappingActions(SnapMode mode)
+{
+    const QSignalBlocker blockSnapNothing(mUi->actionSnapNothing);
+    const QSignalBlocker blockSnapToGrid(mUi->actionSnapToGrid);
+    const QSignalBlocker blockSnapToFineGrid(mUi->actionSnapToFineGrid);
+    const QSignalBlocker blockSnapToPixels(mUi->actionSnapToPixels);
```
bjorn: If you connect to `triggered` instead of `toggled`, you don't need to use signal blockers because the `triggered` signal is only emitted as a result of user interaction.

src/tiled/mainwindow.cpp
```diff
@@ -965,6 +978,13 @@ void MainWindow::changeEvent(QEvent *event)
     case QEvent::WindowStateChange:
         mUi->actionFullScreen->setChecked(isFullScreen());
         break;
+    case QEvent::ActivationChange:
+        if (isActiveWindow()) {
+            Preferences *preferences = Preferences::instance();
+            preferences->sync();
+            updateSnappingActions(preferences->snapMode());
```
bjorn: Since you're already calling this on `QMenu::aboutToShow`, I don't think we gain anything by also doing it on `ActivationChange` and I think we can also remove the direct call from the `MainWindow` constructor.

bjorn (changes requested):
Thanks for fixing this issue, @SIDDHAARTHAA! I've pushed a bunch of cleanups of stuff you left around which was no longer used and changed `SnapHelper` to use the same enum because it was a bit strange how it still had "snap to pixels" as separate boolean.

But I've also left some comments for you to address! :-)

SIDDHAARTHAA:
Thanks for the review! I addressed all requested points: removed explicit action group exclusivity, switched snap handling to `QActionGroup::triggered` using action `data()` for `SnapMode`, removed manual `QSettings::sync()` calls, removed signal blockers, and dropped the `ActivationChange` refresh path.

src/tiled/mainwindow.cpp
SIDDHAARTHAA: okayy

src/tiled/mainwindow.cpp
```diff
@@ -416,18 +415,23 @@ MainWindow::MainWindow(QWidget *parent, Qt::WindowFlags flags)
...
     mUi->actionShowObjectReferences->setChecked(preferences->showObjectReferences());
     mUi->actionShowTileAnimations->setChecked(preferences->showTileAnimations());
     mUi->actionShowTileCollisionShapes->setChecked(preferences->showTileCollisionShapes());
     mUi->actionEnableParallax->setChecked(preferences->parallaxEnabled());
-    mUi->actionSnapToGrid->setChecked(preferences->snapToGrid());
-    mUi->actionSnapToFineGrid->setChecked(preferences->snapToFineGrid());
-    mUi->actionSnapToPixels->setChecked(preferences->snapToPixels());
     mUi->actionHighlightCurrentLayer->setChecked(preferences->highlightCurrentLayer());
     mUi->actionHighlightHoveredObject->setChecked(preferences->highlightHoveredObject());
+    connect(preferences, &Preferences::snapModeChanged, this, &MainWindow::updateSnappingActions);
```
bjorn: I think we also don't need this connection, since we're always calling `updateSnappingActions` when the menu opens anyway.

bjorn:
Thanks @SIDDHAARTHAA!

SIDDHAARTHAA:
My pleasure ;)

PR #4367, opened 2026-02-26
feat: warn about missing fonts in Issues view for maps and templates

### LINKED ISSUE #3548: Tiled should warn when loading Text Objects with missing fonts
When opening maps that contain Text Objects that refer to a font that Tiled cannot find, Tiled has good behaviour: it uses a default placeholder font, and keeps the reference. However, the user is not informed why this is happening, and this can be confusing when the user *thinks* they have the font (e.g. if they have "MyFont Regular" instead of "MyFont Light").

I think when opening Maps and Templates with Text Objects with missing fonts, Tiled should show a warning letting the user know that the font could not be found, and this warning should include the exact name of the font as it appears in the Object. If there's room, it might be good to also mention that the font reference will be saved as-is or that a placeholder font will be displayed, to alleviate fears of data loss.

SIDDHAARTHAA:
closes #3548

Adds missing font warnings for text objects via the **Issues** view when opening maps/templates.

- Reports the exact missing font name.
- Mentions fallback behavior and that font references are preserved.
- Uses `JumpToObject` for map objects (double-click warning to jump).
- Adds equivalent warning handling for template text objects.

@bjorn If there are any changes needed, I’m happy to update it.

---

image

### DIFF
--- a/src/tiled/templatesdock.cpp
+++ b/src/tiled/templatesdock.cpp
@@ -338,7 +338,7 @@ void TemplatesDock::checkTileset()
     }
 
     if (MapObject *object = dummyObject(); object && object->shape() == MapObject::Text) {
-        const auto availableFamilies = availableFontFamiliesCaseFolded();
+        const auto availableFamilies = availableFontFamilies();
         const QString missingFamily = missingTextObjectFontFamily(*object, availableFamilies);
 
         if (!missingFamily.isEmpty()) {
--- a/src/tiled/textobjectfontwarninghelper.cpp
+++ b/src/tiled/textobjectfontwarninghelper.cpp
@@ -47,19 +47,13 @@ static bool isGenericFontFamily(const QString &family)
     return genericFamilies.contains(family.toCaseFolded());
 }
 
-QSet<QString> availableFontFamiliesCaseFolded()
+QStringList availableFontFamilies()
 {
-    QSet<QString> families;
-    const QStringList fontFamilies = QFontDatabase().families();
-
-    for (const QString &fontFamily : fontFamilies)
-        families.insert(fontFamily.toCaseFolded());
-
-    return families;
+    return QFontDatabase().families();
 }
 
 QString missingTextObjectFontFamily(const MapObject &mapObject,
-                                    const QSet<QString> &availableFamilies)
+                                    const QStringList &availableFamilies)
 {
     if (mapObject.shape() != MapObject::Text)
         return {};
@@ -68,7 +62,7 @@ QString missingTextObjectFontFamily(const MapObject &mapObject,
     if (family.isEmpty() || isGenericFontFamily(family))
         return {};
 
-    return availableFamilies.contains(family.toCaseFolded()) ? QString() : family;
+    return availableFamilies.contains(family, Qt::CaseInsensitive) ? QString() : family;
 }
 
 } // namespace Tiled
--- a/src/tiled/mapdocument.cpp
+++ b/src/tiled/mapdocument.cpp
@@ -1650,7 +1650,7 @@ void MapDocument::checkIssues()
               this);
     }
 
-    const auto availableFamilies = availableFontFamiliesCaseFolded();
+    const auto availableFamilies = availableFontFamilies();
     for (Layer *layer : map()->objectGroups()) {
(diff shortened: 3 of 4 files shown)


src/tiled/textobjectfontwarninghelper.cpp
```diff
@@ -0,0 +1,74 @@
...
+}
+
+QString missingTextObjectFontFamily(const MapObject &mapObject,
+                                    const QSet<QString> &availableFamilies)
+{
+    if (mapObject.shape() != MapObject::Text)
+        return {};
+
+    const QString family = mapObject.textData().font.family().trimmed();
+    if (family.isEmpty() || isGenericFontFamily(family))
```
bjorn, line 62: Did you check whether we can just use `QFont::exactMatch` rather than checking manually whether the family is known?

Unfortunately there are a bunch of reasons beyond the family that can cause `QFont::exactMatch` to return `false`, and it does not tell us what doesn't match, but I think it's worth checking whether we can use it anyway.

src/tiled/templatesdock.cpp
```diff
@@ -295,37 +300,63 @@ void TemplatesDock::refreshDummyObject()
...
+        } else {
+            mFixTilesetButton->setVisible(false);
+            mDescriptionLabel->setVisible(false);
+        }
     } else {
         mFixTilesetButton->setVisible(false);
         mDescriptionLabel->setVisible(false);
     }
+
+    if (MapObject *object = dummyObject()) {
```
bjorn:
```suggestion
    if (MapObject *object = dummyObject(); object && object->shape() == MapObject::Text) {
```
We'll want to add this because templated text objects are rather rare and we'll avoid calling `availableFontFamiliesCaseFolded` then.

SIDDHAARTHAA: good catch, I'll add that

src/tiled/textobjectfontwarninghelper.cpp
SIDDHAARTHAA: For the font detection, I was avoiding `QFont::exactMatch()` because it can return false for reasons beyond the family, but I agree it’s worth checking whether it’s good enough here before doing the manual family-based lookup. I also hadn’t considered the foundry-qualified family case, so I’ll revisit that part.

SIDDHAARTHAA:
I looked into [`QFontInfo::exactMatch()`](https://doc.qt.io/qt-6/qfontinfo.html#exactMatch), and I don’t think it’s the right check for this warning.

The warning is specifically about whether the referenced font family can be found, while `exactMatch()` is broader and can return false when other requested font attributes don’t match exactly. Since Tiled text objects also store pixel size, bold/italic, underline, strikeout and kerning, using it here would risk false positives for “missing font” warnings.

So I think a family-availability check is still the better fit here. Your point about foundry-qualified family names is valid though, so I’ve tightened the helper around Qt’s reported family names directly rather than relying on the extra case-folded set. Qt also notes that [`QFontDatabase::families().contains()`](https://doc.qt.io/qt-6/qfontinfo.html) is a faster, but less thorough alternative.

bjorn:
> Since Tiled text objects also store pixel size, bold/italic, underline, strikeout and kerning, using it here would risk false positives for “missing font” warnings.

Right, I realize it checks those things as well, but I wondered if this would be actually an issue because maybe there is no reason for any of these not to match exactly? I thought it might be worth checking, though I guess it'll remain tricky because it might behave differently on different systems.

SIDDHAARTHAA:
I tested `QFontInfo::exactMatch()` locally, and on this system it does give false positives for this use case.

I ran a small probe on Fedora 43 / Qt 6.10.2 across the installed non-generic families from `QFontDatabase().families()`, checking plain/bold/italic/bold+italic cases. Out of 229 families / 916 checks, 36 came back with `exactMatch() == false` even though the family itself was present. The failures I saw were cases where Qt resolved the request to a different reported family name, for example `JetBrainsMono NF` resolving to `JetBrainsMono Nerd Font`.

So I don’t think `exactMatch()` is suitable as the sole check for this warning, since it would still risk false positives for “missing font”. I think the narrower family-availability check remains the better fit here.

PR #4368, opened 2026-02-26
Fix Export As trailing dot when export filter is unset

### LINKED ISSUE #3520: Extension error in export function
"Export as..." puts two dots before file extension:

- filename..tsx
- filename..tsj

SIDDHAARTHAA:
closes #3520

Fixes `Export As...` generating filenames with a trailing dot when no export filter is available (e.g. unset session preference), which could lead to outputs like `name..tsx`.

Changes:
- In export filename suggestion, only append `.` + extension when the parsed extension is non-empty.
- Align tileset export filter session key to `tileset.lastUsedExportFilter` so the correct preference is reused.

### DIFF
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -19,7 +19,7 @@ on:
     - '.travis.yml'
 
 env:
-  QBS_VERSION: 2.4.1
+  QBS_VERSION: 3.1.1
   CMAKE_VERSION: 3.31
   SENTRY_VERSION: 0.12.8
   SENTRY_ORG: mapeditor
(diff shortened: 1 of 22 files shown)


src/tiled/mainwindow.cpp
```diff
@@ -149,13 +149,13 @@ ExportDetails<Format> chooseExportDetails(const QString &fileName,
 
         QRegularExpression extensionFinder(QLatin1String("\\(\\*\\.([^\\)\\s]*)"));
         QRegularExpressionMatch match = extensionFinder.match(selectedFilter);
-        const QString extension = match.captured(1);
+        const QString extension = match.captured(1).trimmed();
```
bjorn:
```suggestion
        const QString extension = match.captured(1);
```
Let's not trim this because the capture in the regular expression already excludes whitespace.

PR #4369, opened 2026-02-26
Fix: Keyboard shortcuts for objects in full screen mode

### LINKED ISSUE #4201: Bugs / Inconsistencies with Fullscreen, Keybindings, Status Bar, Vertical Sorting.

PoonamMehan:
Closes #4201

### DIFF
--- a/src/tiled/abstractobjecttool.cpp
+++ b/src/tiled/abstractobjecttool.cpp
@@ -46,6 +46,7 @@
 #include <QKeyEvent>
 #include <QMenu>
 #include <QMessageBox>
+#include <QShortcut>
 #include <QToolBar>
 #include <QUndoStack>
 
@@ -157,14 +158,6 @@ void AbstractObjectTool::keyPressed(QKeyEvent *event)
     case Qt::Key_PageDown:  lower(); return;
     case Qt::Key_Home:      raiseToTop(); return;
     case Qt::Key_End:       lowerToBottom(); return;
-    case Qt::Key_X:         flipHorizontally(); return;
-    case Qt::Key_Y:         flipVertically(); return;
-    case Qt::Key_Z:
-        if (event->modifiers() & Qt::ShiftModifier)
-            rotateLeft();
-        else
-            rotateRight();
-        return;
     case Qt::Key_D:
         if (event->modifiers() & Qt::ControlModifier) {
             duplicateObjects();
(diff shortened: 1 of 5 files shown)


PoonamMehan:
Okay, I'll try that force push on the branch

PoonamMehan:
@bjorn I solved this using the QShortcut approach you suggested. 
QShortcut instances are parented to the main window.
Also, the console dock was calling setFocus() on its input line via visibilityChanged, which fired on any visibility change, including fullscreen transitions. This caused the selected object/tile to lose focus and the cursor to jump to the console input box whenever switching to fullscreen with the console open.
So I fixed it by only requesting focus when the user explicitly opens the console (via the terminal button in the status bar or View → Views and Toolbars → Console). This is done by connecting to QAction::triggered instead of QAction::toggled triggered.
Would love a review when you get a chance :)

PR #4371, opened 2026-02-27
Add texture filtering preference (#3801)

### LINKED ISSUE #3801: Add option to use smooth scaling regardless of current scale
**Problem**

Currently Tiled determines whether to perform smooth scaling when rendering images based on the current zoom level. When it is 100%, 200%, 300%, etc., nearest-neighbor scaling is used and smooth scaling is only used when scaling down or for values in between.

In general, this makes the map look fine both when it is scaled down as well as when it is scaled up (especially pixel art). However, this logic predates the support for resized and rotated tile objects. When resizing or rotating tile objects, it can be desired to apply a smooth rendering algorithm regardless of the view scale.

**Solution**

The solution might be to add an explicit "Texture filtering" option (not sure about name), either in View menu or Preferences. Possible values could include "Nearest Neighbor", "Smooth" and "Auto" (the current behavior). In addition, when we keep "Auto" as the default behavior, it might be a good idea to enhance it a little, so that scaled-down tile objects are smoothly rendered.

This new option should also affect the "Export as Image" action.

The `tmxrasterizer` tool already applies smooth rendering by default and provides a `--no-smoothing` parameter to disable it.

keshav0479:
Fixes #3801

This adds a Texture filtering preference under Edit --> Preferences --> Interface with three options:

* Auto - current behavior (smooth only at fractional zoom levels)
* Nearest Neighbor - always sharp/pixelated
* Smooth - always interpolated
The main motivation is that rotated or resized tile objects look jagged at native zoom levels (100%, 200%) because Auto forces nearest-neighbor there. With the Smooth option, users can get clean rendering regardless of zoom.

The preference applies to the map view, tileset view, wang color view, and export-as-image. Preview/thumbnail contexts (minimap, stamp model) are left always-smooth intentionally.

https://github.com/user-attachments/assets/2bf46f3c-6a51-41d7-ba6d-ac926a4ff44d

### DIFF
--- a/src/tiled/exportasimagedialog.cpp
+++ b/src/tiled/exportasimagedialog.cpp
@@ -109,7 +109,7 @@ ExportAsImageDialog::~ExportAsImageDialog()
 
 static bool smoothTransform(qreal scale)
 {
-    return scale != qreal(1) && scale < qreal(2);
+    return Preferences::instance()->smoothTransform(scale != qreal(1) && scale < qreal(2));
 }
 
 void ExportAsImageDialog::accept()
@@ -170,7 +170,9 @@ void ExportAsImageDialog::accept()
         renderFlags |= MiniMapRenderer::DrawGrid;
     if (session::includeBackgroundColor)
         renderFlags |= MiniMapRenderer::DrawBackground;
-    if (session::useCurrentScale && smoothTransform(mCurrentScale))
+
+    if (Preferences::instance()->smoothTransform(session::useCurrentScale
+                                                  && smoothTransform(mCurrentScale)))
         renderFlags |= MiniMapRenderer::SmoothPixmapTransform;
 
     MapRenderer *renderer = mMapDocument->renderer();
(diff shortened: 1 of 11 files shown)


dogboydog:
I tested this out a bit and it seems to work well. I actually had never noticed the original behavior of applying smoothing between 101% and 199%, since I'm usually way more zoomed in than that.  I might not be the target audience here since  my games are pixel art which would generally only use Nearest Neighbor, but I'd say the change looks good overall.

src/tiled/exportasimagedialog.cpp
```diff
@@ -110,7 +110,15 @@ ExportAsImageDialog::~ExportAsImageDialog()
 
 static bool smoothTransform(qreal scale)
 {
-    return scale != qreal(1) && scale < qreal(2);
+    switch (Preferences::instance()->textureFiltering()) {
```
bjorn: This change has a `switch (Preferences::instance()->textureFiltering())` in 4 places. Since they all depend on the `Preferences` now, maybe we should have a single helper function there to determine whether smoothing should be applied.

keshav0479: Good idea, moved the switch into a `Preferences::smoothTransform(bool)` helper so each call site just passes its auto-mode condition. Also rebased and added the NEWS entry.

PR #4372, opened 2026-02-27
Implement Backspace key support for polygon/polyline creation

dhruv1955:
## Summary

Add Backspace key support to remove the last placed point during polygon and polyline creation, without canceling the entire shape.

## Problem

When drawing polygons or polylines, users had no way to undo just the last placed point. The only option was Escape, which canceled the entire shape — forcing a restart even for a single mis-click.

## Solution

Handle `Qt::Key_Backspace` in `keyPressed()` when `state() == CreatingObject`:

* Cancels the operation if only the starting point remains
* Removes the last point (or first, when extending from the beginning)
* Creating mode: updates polygon directly via `setPolygon()`
* Extending mode: pushes `ChangePolygon` onto the undo stack for proper history
* Calls `synchronizeOverlayObject()` to refresh the visual preview

## Changes

* File: `src/tiled/createpolygonobjecttool.cpp`
* Function: `keyPressed()` — previously only had TODO comments

## Testing

1. Select Insert Polygon tool (P), click 3–4 points, press Backspace — last point removed
2. Press Backspace repeatedly until 1 point remains — operation cancels
3. Extend an existing polyline, add points, press Backspace — point removed with undo history entry

## Notes

Resolves the first TODO in `keyPressed()`.
The second TODO (Shift+Enter to finish as polygon) is left for a follow-up.

### DIFF
--- a/docs/manual/objects.rst
+++ b/docs/manual/objects.rst
@@ -342,8 +342,12 @@ right-clicking those nodes and choosing "Extend Polyline", or by
 switching to the :ref:`insert-polygon-tool` tool and clicking on either
 end of an already selected polyline.
 
+.. raw:: html
+
+   <div class="new">New in Tiled 1.12</div>
+
 While creating or extending a polygon/polyline with the
-:ref:`insert-polygon-tool`, you can press ``Backspace`` to remove the
+:ref:`insert-polygon-tool` tool, you can press ``Backspace`` to remove the
 previously added point.
 
 


src/tiled/createpolygonobjecttool.cpp
```diff
@@ -113,8 +114,40 @@ void CreatePolygonObjectTool::deactivate(MapScene *scene)
...
     // TODO: Modifier for finishing as polygon (Shift+Enter)
+
+    // Backspace removes the last added point (or first, when extending from the beginning)
+    if (event->key() == Qt::Key_Backspace && state() == CreatingObject) {
+        MapObject *newObject = mNewMapObjectItem->mapObject();
+        QPolygonF currentPolygon = newObject->polygon();
+
+        // If only the starting point remains, cancel the whole operation
+        if (currentPolygon.size() <= 1) {
+            cancelNewMapObject();
```
bjorn, line 130: This canceling of the whole operation only works correctly in `Creating` mode. In extending mode, it is too late and leaves a polygon object with a single point, which can't be interacted with.

I'm not sure what we should do in this case. I think there are two options:

* If we stop "extending" an object while it has less than 2 points, or attempt to shorten it further, automatically delete it (can be done with `RemoveMapObjects`).
* Ignore <kbd>Backspace</kbd> in "extending" mode when there are only 2 points remaining.

dhruv1955:
Implemented the requested follow-up in 4204895f.

- Ignore Backspace in extending mode when only two points remain,
  preventing creation of an invalid 1-point polyline.
- Creating mode behavior remains unchanged (Backspace cancels when
  only the starting point remains).

Also added:
- Release note in NEWS.md
- Manual updates in:
  - docs/manual/objects.rst (Working with Objects → Edit Polygons)
  - docs/manual/keyboard-shortcuts.rst (Insert Polygon shortcuts)

bjorn:
Thanks for addressing this feature, @dhruv1955!

Regarding the AI code review, it's something I'm now experimenting with as explained in https://github.com/mapeditor/tiled/issues/4399. But I just noticed that this AI even edited your original description of this PR, which I think is not OK! So I will be looking for the option to disable that (edit: I've reverted your comment to the original).

Btw, your descriptions look very much written by AI as well. Please note that for GSoC we do not accept AI-written proposals and AI-written code. I've now made this explicit on our [GSoC 2026](https://github.com/mapeditor/tiled/wiki/GSoC-2026) wiki page.

PR #4374, opened 2026-02-28
feature: tiled.open() and Asset.save() in scripts executed via CLI scripts

### LINKED ISSUE #4095: Scripting: `tiled.open()` and `Asset.save()` in scripts executed via CLI
When running scripts via command line interface, there is no good way to open and save Tiled assets. Such scripts are responsible for opening any Assets mentioned in their arguments, or may choose which Assets they'd like to open (e.g. all the Assets in a Project), but the only method available to open Assets seems to be via MapFormat/TilesetFormat. The only other option is to run a scripted exporter via `--export-map/tileset`, where Tiled will handle loading the (single) initial asset for you.

In addition, it's a common task to modify some asset via a script and to then save it, but once again, the only option to save an Asset in a script run via CLI is MapFormat/TilesetFormap, one can't simply `map.save()` like one can in the GUI.

Allowing more of the GUI-only methods to be usable in CLI mode would make it much easier to write scripts that can work both via CLI and via GUI. Aside from making the aforementioned methods available, it would also be convenient if there was some way to have Tiled load the Asset(s) listed in parameters instead of just passing the paths along to the script, and allowing scripts to use `tiled.activeAsset` and/or `tiled.openAssets` to get at those loaded Assets; this would allow many scripts written for the GUI to work via CLI without any changes, since those properties are the main way scripts choose the Assets to work on.

As a counterpart to `tiled.open()`, `tiled.close()` should also be made available, so that memory use can be kept in check for scripts that process many Assets.

This is not a critical feature since all the functionality is technically available via MapFormat/TilesetFormat. Making these GUI-only features available via CLI will just make scripting more straightforward rather than more capable. However, though it seems minor, I think this difference in approach to opening and saving files is significant enough that it discourages people from writing scripts that can work both in the GUI and via CLI. I think we'd see more widely useful scripts if the differences between the two modes were restricted to truly GUI-specific functionality like Automapping.

PoonamMehan:
This PR intrduces EditorManager as a GUI-agnostic document/session manager. DocumentManager focuses only on GUI concerns such as tabs, editors, dialogs, and window state.
 EditorManager owns the list of open documents, the QUndoGroup, and core file operations (open, save, close, reload), so asset session logic is centralized and reusable with or without the GUI.

DocumentManager was refactored to delegate all document/session behavior to EditorManager.
Core scripting and data types (for example ScriptModule, Document, and EditableAsset) now use EditorManager instead of DocumentManager for session access.

This ensures tiled.open(), tiled.openAssets, asset.save(), and tiled.close()/tiled.reload() all use the same backend in both GUI and CLI modes.

Closes #4095

### DIFF
--- a/src/tiled/editormanager.cpp
+++ b/src/tiled/editormanager.cpp
@@ -65,11 +65,6 @@ void EditorManager::deleteInstance()
     mInstance = nullptr;
 }
 
-EditorManager::~EditorManager()
-{
-    mInstance = nullptr;
-}
-
 Document *EditorManager::currentDocument() const
 {
     if (mCurrentIndex < 0 || mCurrentIndex >= mDocuments.size())


bjorn:
Hmm, this looks really promising, but there's something odd about the naming. What we should have is this:

* `DocumentManager`: Manages open documents but does not do any UI stuff.
* `EditorManager`: Manages editors (UI), the tab widget, etc.

Judging by your description and the current implementation, it's been done the other way around. Maybe you could address this by swapping the names of these classes and the related file names?

Also I see currently both `DocumentManager` and `EditorManager` store a `QVector<DocumentPtr> mDocuments`. But this should be necessary only in the `DocumentManager` (if using the correct naming). The `EditorManager` can always access the `DocumentManager`, just not the other way around.

Issue #3636, opened 2023-03-28
Automapping: Wildcard input index


eishiya:
tl;dr version: Wildcard input indices, allowing certain input layers to be shared among all the input indices of a rule.

A couple of times I've seen people get confused about input indices in a reasonable way: they think that if they have an input "without" an index, it'll be part of all indices.

For example:
> input1_Detail
> input2_Detail
> input_Ground

They think input_Ground will behave like two identical input1_Ground and input2_Ground layers.

Of course, the truth is that an empty index is just another index. However, it would be quite convenient if it was possible to mark certain input layers as being used for all indices, using a wildcard index.
It's fairly common in my rules that address multiple layers that only inputs from one layer need to have independent indices, and inputs from other layers are the same for all indices, necessitating a bunch of duplication. A wildcard index would avoid this.

While I think the usefulness of this is clear enough, what the wildcard should *look* like is less clear:
- `*`, e.g. `input*_foo`: this is widely used for wildcards already, and it's not likely anyone is using this as an actual index already. However, people may expect this to perform partial string matching, they might expect `inputA*` to match `inputAbove`, for example.
- Empty index, e.g. `input_foo`: this would probably be intuitive to new users as the fact that an empty index is an actual separate index seems to confuse folks. But, existing Automapping rules may include combinations of the empty index and named indices and expect them to all be separate indices. I know I have some rules like that.
- some reserved keyword, such as `ALL`, e.g. `inputALL_foo`: this would avoid the drawbacks of the above approaches and isn't likely anyone's using something like "ALL" as an index already, but it's longer and not language-neutral - but neither is "input", of course.

I'd love to hear people's preferences on different ways to do the wildcard.

eishiya:
Now that `output`s with an empty index are always output as of #3909, perhaps a similar change could be made for `input`? Changing the meaning of an empty input index may break existing rules, but probably not much more than the output change did, and it would at least make for consistent/comparable behaviour for empty indices.

AndreaBellocci:
Hi, is this issue still relevant? I'd like to try to address it

bjorn:
@AndreaBellocci Yes, this issue is still relevant. You can have a look at #3909 because it made a related change to the output index, though right now I'm not sure if that helps figure out the implementation for the same behavior of the input index.

In any case the decision is not to use an explicit wildcard like `*` in the index, but rather change the behavior for the empty index (`input_[layer-name]`), the second option suggested above by eishiya.

[...]

PR #4375, opened 2026-03-01
Automapping: empty input index (#3636)

AndreaBellocci:
Implemented wildcard input sets as explained in #3636 
I updated the documentation on automapping, though it may be nice to include an example showing the new behaviour.

### DIFF
--- a/src/tiled/automapper.cpp
+++ b/src/tiled/automapper.cpp
@@ -379,6 +379,7 @@ bool AutoMapper::setupRuleMapLayers()
 
     QString error;
 
+    // Loop over all layers in the map and parse them
     for (Layer *layer : mRulesMap->allLayers()) {
         if (layer->isGroupLayer() || layer->isImageLayer())
             continue;
@@ -451,7 +452,7 @@ bool AutoMapper::setupRuleMapLayers()
 
         const int layerNameStartPosition = ruleMapLayerName.indexOf(QLatin1Char('_'));
 
-        // both 'rule' and 'output' layers will require and underscore and
+        // both 'input' and 'output' layers will require and underscore and
         // rely on the correct position detected of the underscore
         if (layerNameStartPosition == -1) {
             error += tr("Did you forget an underscore in layer '%1'?").arg(ruleMapLayerName);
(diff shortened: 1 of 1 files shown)


src/tiled/automapperwrapper.cpp
```diff
@@ -59,21 +59,26 @@ AutoMapperWrapper::AutoMapperWrapper(MapDocument *mapDocument,
...
                              [&] (const TileLayer *tileLayer) { return autoMapper->ruleLayerNameUsed(tileLayer->name()); }))
                 continue;
         }
 
+        // Automap
         autoMapper->autoMap(region, appliedRegionPtr, context);
 
-        if (appliedRegionPtr) {
+        if (appliedRegionPtr)
+        {
```
bjorn: Please don't reformat the code!

src/tiled/automappingmanager.cpp
```diff
@@ -79,7 +80,7 @@ void AutomappingManager::autoMap()
     }
 
     autoMapInternal(region, nullptr);
-}
+} // autoMap
```
bjorn: Please don't add comments saying which function ends here!

src/tiled/automapper.h
```diff
@@ -165,7 +166,7 @@ struct RuleMapSetup
     // The properties that should be copied to the target map, when any rule
     // matches for these output layers.
     QHash<const Layer*, Properties> mOutputLayerProperties;
-};
+}; // RuleMapSetup
```
bjorn: Please don't add comments about which class or struct ends here.

src/tiled/automapper.cpp
```diff
@@ -43,71 +43,71 @@ namespace Tiled {
 
 static int wrap(int value, int bound)
 {
-    return (value % bound + bound) % bound;
+	return (value % bound + bound) % bound;
```
bjorn: Please don't replace space indentation with tabs!

bjorn:
@AndreaBellocci I understand the vast code reformatting might have been some VS code accident. But it makes it very hard to see what are the actually relevant changes. Please either revert the code changes or re-apply the relevant changes based on a new branch, whatever is easier for you.

Thanks in any case for looking into this feature!

AndreaBellocci:
My apologies for this, I will fix it as soon as possible. Out of curiosity: do you have (or do you plan to use) precommits files to ensure code style doesn't change?

bjorn:
> Out of curiosity: do you have (or do you plan to use) precommits files to ensure code style doesn't change?

While that might be nice, I have so far not found an automated code formatter that would not make lots of undesired changes to the existing code. If introduced, such a tool should only affect changed lines or be very conservative in the changes that it applies.

Maybe consider why the changes you've made happened, so we might put instructions somewhere to avoid this for future contributors. We generally want to avoid changes to unrelated code.

PR #4382, opened 2026-03-03
Fixed bugs about Japanese & Chinese input methods:

### LINKED ISSUE #3403: Japanese IMEs do not work in the "open file in project" widget (Windows)
**Describe the bug**
The text entry field in the "open file in project" widget does not seem to support entry via Google Japanese IME or Microsoft IME. I only use it for Japanese but I imagine it might affect other languages that work similarly. 

I tried modifying the source code for filteredit.cpp to `setAttribute(Qt::WA_InputMethodEnabled)` in the constructor after doing some googling but that wasn't enough to fix this. 

Not sure if this is platform specific but had @eishiya  and myself confirm on Windows. 

**To Reproduce**
1. Install Google IME or Microsoft IME  Japanese input. 
2. Go to 'Open Files in Project'
3. Switch to Japanese input 
4. Attempt to enter text in Japanese
5. Observe that text cannot be entered

**Expected behavior**
This field should allow text to be entered in Japanese.

**Media**
When the "open file in project" widget is focused, google IME reports being disabled
image
IME working in another text field in Tiled
image

Note: 
We might want to test out some of the widgets in the new Dialog functionality and see if any of them suffer the same problem, and if so apply the solution there too. 

**Specifications:**
 - OS: Windows 10
 - Tiled Version: 1.9.0

cortexA233:
Fixed https://github.com/mapeditor/tiled/issues/3403

Add setAttribute(Qt::WA_InputMethodEnabled); to LocatorWidget, to avoid the problem of non-Latin input methods like Chinese or Japanese IME.

Relevant bugs of Qt: https://qt-project.atlassian.net/browse/QTBUG-83490

### DIFF
--- a/src/tiled/locatorwidget.cpp
+++ b/src/tiled/locatorwidget.cpp
@@ -36,6 +36,8 @@
 #include <QApplication>
 #include <QCollator>
 #include <QDir>
+#include <QGuiApplication>
+#include <QInputMethod>
 #include <QKeyEvent>
 #include <QPainter>
 #include <QScrollBar>
@@ -271,7 +273,10 @@ LocatorWidget::LocatorWidget(LocatorSource *locatorSource,
     , mResultsView(new ResultsView(this))
 {
     setAttribute(Qt::WA_DeleteOnClose);
+    setAttribute(Qt::WA_InputMethodEnabled);
     setFrameStyle(QFrame::StyledPanel | QFrame::Plain);
+    setAutoFillBackground(true);
+    setFocusPolicy(Qt::StrongFocus);
 
     mLocatorSource->setParent(this);        // take ownership of source
 
@@ -286,6 +291,7 @@ LocatorWidget::LocatorWidget(LocatorSource *locatorSource,
     mFilterEdit->setFilteredView(mResultsView);
     mFilterEdit->setClearTextOnEscape(false);
     mFilterEdit->setFont(scaledFont(mFilterEdit->font(), 1.5));
+    mFilterEdit->setAttribute(Qt::WA_InputMethodEnabled);
 
     setFocusProxy(mFilterEdit);
     mResultsView->setFocusProxy(mFilterEdit);
@@ -315,6 +321,12 @@ void LocatorWidget::setVisible(bool visible)
 
     if (visible) {
         setFocus();
+        activateWindow();
+
+        // Ensure the text field gets focus directly, so input methods can
+        // attach to it reliably.
+        mFilterEdit->setFocus(Qt::ShortcutFocusReason);
+        QGuiApplication::inputMethod()->update(Qt::ImEnabled);
 
         if (!mFilterEdit->text().isEmpty())
             mFilterEdit->clear();
--- a/src/tiled/locatorwidget.cpp
+++ b/src/tiled/locatorwidget.cpp
@@ -289,7 +289,6 @@ LocatorWidget::LocatorWidget(LocatorSource *locatorSource,
     mFilterEdit->setFilteredView(mResultsView);
     mFilterEdit->setClearTextOnEscape(false);
     mFilterEdit->setFont(scaledFont(mFilterEdit->font(), 1.5));
-    mFilterEdit->setAttribute(Qt::WA_InputMethodEnabled);
 
     setFocusProxy(mFilterEdit);
     mResultsView->setFocusProxy(mFilterEdit);
--- a/src/tiled/locatorwidget.cpp
+++ b/src/tiled/locatorwidget.cpp
@@ -321,6 +321,8 @@ void LocatorWidget::setVisible(bool visible)
         setFocus();
         activateWindow();
 
+        // just a fallback to make sure the input method will be enabled in some situation
+        // like changed the focus object when showing this widget
         QGuiApplication::inputMethod()->update(Qt::ImEnabled);
         if (!mFilterEdit->text().isEmpty())
             mFilterEdit->clear();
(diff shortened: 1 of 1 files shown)


dogboydog:
image
 Looks like it fixes the issue based on my testing  :+1:  Thanks! 
 
 I also tested on Ubuntu and was able to use my Japanese IME there

bjorn:
That's great, thanks for looking into this @cortexA233!

However, I see a lot of changes to get this to work, and I wonder if they are really all necessary?

* Now we have two `setFocus` calls.
* There is `setAutoFillBackground`, is it related?
* Do we need `Qt::StrongFocus` policy, and if so, why?
* Is `Qt::WA_InputMethodEnabled` really needed on both `LocatorWidget` and `mFilterEdit`?
* Do we really need the explicit `QGuiApplication::inputMethod()->update(Qt::ImEnabled);` call (I expect so, just asking)

There is one comment, but I guess there could be some more (though no need to add anything obvious).

cortexA233:
> That's great, thanks for looking into this @cortexA233!
> 
> However, I see a lot of changes to get this to work, and I wonder if they are really all necessary?
> 
> * Now we have two `setFocus` calls.
> * There is `setAutoFillBackground`, is it related?
> * Do we need `Qt::StrongFocus` policy, and if so, why?
> * Is `Qt::WA_InputMethodEnabled` really needed on both `LocatorWidget` and `mFilterEdit`?
> * Do we really need the explicit `QGuiApplication::inputMethod()->update(Qt::ImEnabled);` call (I expect so, just asking)
> 
> There is one comment, but I guess there could be some more (though no need to add anything obvious).

Hi, @bjorn , thanks for your reply!

For your question 1 and 2, I reviewed the code and I will delete unrelated ```setAutoFillBackground``` call, and redundant ```setFocus``` call. sorry for the miss.

For ```Qt::StrongFocus```, In this case the LocatorWidget is a popup widget so it's unnecessary indeed, but I think if we changed it into other class such as a ```QDialog``` (In fact, I firstly thought this issue is a ```Qt::Popup``` bug, so I once changed LocatorWidget into a ```QDialog```), it will allow users to use tab shortcut to focus on it. Of course, if you think it's unnecessary, I can delete it.

About question 4 ```Qt::WA_InputMethodEnabled```, I found when I delete the ```setAttribute(Qt::WA_InputMethodEnabled)``` for ```LocatorWidget``` and only keep the call for ```FilterEdit```, it still works. At first I tried only add ```setAttribute(Qt::WA_InputMethodEnabled)``` to the constructor of ```FilterEdit```, but it didn't work. So I just call ```setAttribute(Qt::WA_InputMethodEnabled)``` for both FilterEdit and LocatorWidget. Now I know I just didn't realized that I shouldn't do such operation in the constructor of a Qt widget component. It's just a remnant of my stupid brute-force enumeration, I will delete it :)

And the ```QGuiApplication::inputMethod()->update(Qt::ImEnabled)``` call is also necessary for now. yesterday I tried several times but it seems I have call this to make it works.

I will commit the redundant code removal I mentioned above, thanks for your time and reply!

cortexA233:
> * Do we really need the explicit `QGuiApplication::inputMethod()->update(Qt::ImEnabled);` call (I expect so, just asking)

Hi, bjorn! I have reviewed and done some test again, and about this ```QGuiApplication::inputMethod()->update(Qt::ImEnabled);``` call, for now it's indeed unnecessary. I've checked Qt's source code, this call would refresh the cache state of the **current focused object**. 
[`QInputMethod()::update()`](https://github.com/qt/qtbase/blob/v6.10.2/src/gui/kernel/qinputmethod.cpp#L284)
image
In my first commit, I have a redundant ```setFocus()``` call for the ```mFilterEdit``` widget, this may cause (not 100% reproducable) a different query result for the ```objectAcceptsInputMethod()``` and make IME disabled, in that case, the ```update()``` call would be a error fallback to make sure the IME would be enabled. Since now we only have the original one ```setFocus()``` call, this ```update()``` call should be unnecessary. But since it's just a fallback call and won't change the right enabled state into a wrong result, so I think I would keep this call and leave a comment in this line. Or I can also delete it if you want :)

Thanks again for your review and reply!

bjorn:
@dogboydog Could you do another test with the reduced patch?

Apart from the cleanups by @cortexA233, I've reduced it further since the flag that was being set was already set according to my debug line, so the `setAttribute(Qt::WA_InputMethodEnabled)` call on the `mFilterEdit` should have just been doing an early return.

I'm also interested in knowing whether each of the remaining three changes are actually necessary.

dogboydog:
I'll try to do so later today

dogboydog:
It still works on Ubuntu for me as of [2ec40d5](https://github.com/mapeditor/tiled/pull/4382/commits/2ec40d5bd62aa175eee88cc1bcb97354a19294de) ( having the  locator widget up blocks screenshots for me, but i could type in the field in japanese) . The windows build failed / timed out but hopefully the fix is pretty cross platform

bjorn:
@dogboydog I've re-triggered the Windows builds and they succeeded now. It would be nice if you could test it also there, because I think this feature has a high chance of being different per platform. But no rush.

Unfortunately, if you rely on these builds it means you can't test whether it works with an even smaller patch.

@cortexA233, since you indicated the `update()` should be unnecessary, I'd prefer to leave it out. It would just be good to verify the fix still works, but I don't personally use input methods so I'm not sure how to test it.

dogboydog:
I can try the windows build later on. On Linux I'm set up to  compile Tiled so I could try a modified patch locally, but I only have a Windows VM now so it's easier to use the CI builds there

dogboydog:
image
Still appears to work on Windows

PR #4383, opened 2026-03-03
Adds optional rulers along the map edges showing tile coordinates

### LINKED ISSUE #4155: Map with numbers - preview
Due map is starting from 0,0 and user operating on some logic as shift square, enemy etc, is good to see map cordinations without operate mouse and cursor - right now user must check on down belt the coordination.

Looking something like "virtual" sprites with numbers under or above images as transparent (option to set alpha), all in grid.
Due possible long map, the numbers can be to big, by this require a number size.

Something like "axes" can be helpful around map as numbers above with cross line to spy coordination of tile.
More cognitive can be grid line difference, eg. units 5, 10 as solid, other type line style or color (for configuration) - if have big map is hard to recognize cords and especially in debug process.

abhinavs1920:
Closes #4155

Helpful when debugging or working with larger maps.

**What it does:**
- Shows X coordinates on top, Y coordinates on left
- Rulers stay fixed at viewport edges when you pan/zoom
- Uses your existing grid settings (major/fine grid) for tick marks
- Toggle with View → Show Map Rulers or Ctrl+R

**Why rulers instead of numbers on every tile:**
Less cluttered, better performance, and you can actually see your tiles.

**Implementation:**
Renders in `MapView::drawForeground()` so it stays viewport-fixed. Major grid lines get labels, fine grid lines get small ticks. Updates when you change grid preferences.

image

### DIFF
--- a/src/tiled/mapview.cpp
+++ b/src/tiled/mapview.cpp
@@ -30,6 +30,7 @@
 #include "pannableviewhelper.h"
 #include "preferences.h"
 #include "rulerwidget.h"
+#include "rulerwidget.h"
 #include "tileanimationdriver.h"
 #include "utils.h"
 #include "zoomable.h"
@@ -129,8 +130,6 @@ MapView::MapView(QWidget *parent)
     // Corner square covering the top-left intersection of the two rulers
     mRulerCorner = new QWidget(this);
     mRulerCorner->setAttribute(Qt::WA_TransparentForMouseEvents);
-    // Background is painted by the ruler widgets on either side; just needs
-    // to be opaque so the map doesn't show through the corner gap.
     mRulerCorner->setAutoFillBackground(true);
 
     // Update view when preferences change
@@ -543,11 +542,6 @@ void MapView::paintEvent(QPaintEvent *event)
     QGraphicsView::paintEvent(event);
 }
 
-void MapView::drawForeground(QPainter *painter, const QRectF &rect)
-{
-    QGraphicsView::drawForeground(painter, rect);
-}
-
 void MapView::hideEvent(QHideEvent *event)
 {
     // Disable panning when the view gets hidden in any way
@@ -561,6 +555,7 @@ void MapView::resizeEvent(QResizeEvent *event)
         updateSceneRect(s->sceneRect());
 
     QGraphicsView::resizeEvent(event);
+    updateRulerGeometry();
     updateRulerGeometry();
 }
 
--- a/src/tiled/mapview.h
+++ b/src/tiled/mapview.h
@@ -34,6 +34,7 @@ class MapObject;
 class MapDocument;
 class MapScene;
 class PannableViewHelper;
+class RulerWidget;
 class TileAnimationDriver;
 class Zoomable;
 
@@ -89,7 +90,6 @@ protected:
     bool event(QEvent *event) override;
 
     void paintEvent(QPaintEvent *event) override;
-    void drawForeground(QPainter *painter, const QRectF &rect) override;
     void hideEvent(QHideEvent *) override;
     void resizeEvent(QResizeEvent *event) override;
 
--- a/src/tiled/rulerwidget.cpp
+++ b/src/tiled/rulerwidget.cpp
@@ -27,6 +27,8 @@
 
 #include <QPainter>
 
+#include <cmath>
+
 using namespace Tiled;
 
 RulerWidget::RulerWidget(Orientation orientation, MapView *mapView)


abhinavs1920:
Heyy @bjorn , added the screenshot in the PR.. please check this feature once you have time.
I am excited as it will help a lot of users :)

src/tiled/mainwindow.ui
```diff
@@ -338,6 +339,17 @@
...
   </action>
+  <action name="actionShowMapRulers">
+   <property name="checkable">
+    <bool>true</bool>
+   </property>
+   <property name="text">
+    <string>Show Map &amp;Rulers</string>
+   </property>
+   <property name="shortcut">
+    <string notr="true">Ctrl+R</string>
```
bjorn: The <kbd>Ctrl+R</kbd> shortcut conflicts with the hidden "Reload Tileset Images" shortcut. I think we can use <kbd>Ctrl+Shift+R</kbd>, which is also what GIMP uses.

src/tiled/mapview.cpp
```diff
@@ -118,6 +120,28 @@ MapView::MapView(QWidget *parent)
...
         updatePanningDriverState();
     });
+
+    // Update view when preferences change
+    Preferences *prefs = Preferences::instance();
+    connect(prefs, &Preferences::showMapRulersChanged, this, [this] {
+        if (scene())
+            scene()->invalidate(sceneRect(), QGraphicsScene::ForegroundLayer);
+    });
+    connect(prefs, &Preferences::gridColorChanged, this, [this] {
```
bjorn: I don't think we should use the grid color on the ruler, because that setting is there to adjust the grid to be visible on top of your graphics, whereas the rulers would have a UI-based background color.

src/tiled/mapview.cpp
```diff
@@ -480,6 +504,145 @@ void MapView::paintEvent(QPaintEvent *event)
...
+void MapView::drawForeground(QPainter *painter, const QRectF &rect)
+{
+    QGraphicsView::drawForeground(painter, rect);
+
+    Preferences *prefs = Preferences::instance();
+    if (!prefs || !prefs->showMapRulers())
+        return;
+
+    MapScene *scene = mapScene();
+    if (!scene || !scene->mapDocument())
```
bjorn: The return value of `scene->mapDocument()` is checked again later, so we don't need to check it here.

src/tiled/mapview.cpp
```diff
@@ -480,6 +504,145 @@ void MapView::paintEvent(QPaintEvent *event)
     QGraphicsView::paintEvent(event);
 }
 
+void MapView::drawForeground(QPainter *painter, const QRectF &rect)
+{
+    QGraphicsView::drawForeground(painter, rect);
+
+    Preferences *prefs = Preferences::instance();
+    if (!prefs || !prefs->showMapRulers())
```
bjorn: We don't need to check `prefs` because `Preferences::instance` always returns a `Preferences` instance (it creates one if it doesn't exist yet).

src/tiled/mapview.cpp
```diff
@@ -480,6 +504,145 @@ void MapView::paintEvent(QPaintEvent *event)
...
+    MapScene *scene = mapScene();
+    if (!scene || !scene->mapDocument())
+        return;
+
+    MapDocument *mapDocument = scene->mapDocument();
+    if (!mapDocument)
+        return;
+        
+    const Map *map = mapDocument->map();
+    if (!map)
```
bjorn: A `MapDocument` always owns a `Map` instance, so we don't need to check this. It also always has a `MapRenderer`, so we don't need to check that one below either.

src/tiled/mapview.cpp
```diff
@@ -480,6 +504,145 @@ void MapView::paintEvent(QPaintEvent *event)
...
+    if (!map->infinite()) {
+        visibleTileRect = visibleTileRect.intersected(QRect(0, 0, map->width(), map->height()));
+    }
+
+    // Use grid preferences for tick spacing
+    const QSize gridMajor = prefs->gridMajor();
+    const int gridFine = prefs->gridFine();
+    
+    // Avoid division by zero
+    if (gridMajor.width() <= 0 || gridMajor.height() <= 0 || gridFine <= 0)
```
bjorn: I had my major grid set to 0, which prevented me from seeing the rulers at all. I think you need to still draw the rulers and prevent the division by zero some other way (see for example the check for major lines in `OrthogonalRenderer::drawGrid`).

src/tiled/mapview.cpp
```diff
@@ -480,6 +504,145 @@ void MapView::paintEvent(QPaintEvent *event)
...
+        return;
+    
+    // Styling - use light background with dark text for good contrast
+    const QColor bgColor = QColor(240, 240, 240, 230);  // Light gray, semi-transparent
+    const QColor textColor = QColor(60, 60, 60);         // Dark gray text
+    const QColor tickColor = QColor(120, 120, 120);      // Medium gray for ticks
+    const QColor borderColor = QColor(180, 180, 180);    // Light border
```
bjorn: Consider finding suitable colors in the map view's `palette()`.

src/tiled/mapview.cpp
```diff
@@ -480,6 +504,145 @@ void MapView::paintEvent(QPaintEvent *event)
...
+
+    // Draw Y axis ticks and labels
+    for (int y = visibleTileRect.top(); y <= visibleTileRect.bottom(); ++y) {
+        QPointF tilePos = renderer->tileToPixelCoords(0, y);
+
+        if (tilePos.y() < viewportRect.top() || tilePos.y() > viewportRect.bottom())
+            continue;
+
+        bool isMajor = (y % gridMajor.height()) == 0;
+        bool isFine = (y % gridFine) == 0;
```
bjorn: Hmm, this doesn't make sense. The fine grid is a smaller grid than the tile grid. I think we can leave it out for now. We can just use minor ticks for any line which isn't on the major grid.

src/tiled/mapview.cpp
```diff
@@ -627,6 +790,12 @@ void MapView::mouseMoveEvent(QMouseEvent *event)
     QGraphicsView::mouseMoveEvent(event);
     mLastMousePos = event->globalPos();
     mLastMouseScenePos = mapToScene(viewport()->mapFromGlobal(mLastMousePos));
+    
+    // Trigger foreground redraw if rulers are enabled to prevent text erasure
```
bjorn: This doesn't work reliably and we also don't want to be repainting the entire view on every mouse move.

abhinavs1920:
I just tested again the rulers, they work well! But one thing I noticed that when zoomed out significantly, the tile coordinate labels can overlap since every tile gets a tick mark. Should we skip drawing labels (or entire ticks) when they'd be too close together in screen space?

Also skipping will keep the number size readable.

image

PR #4384, opened 2026-03-03
Add optional autosave feature with configurable intervals

### LINKED ISSUE #3133: Add option to autosave files
It would be nice to have the option to auto-save maps and tilesets with some interval after each change. This can prevent data loss, but can also be simply convenient to make changes immediately available without having to trigger a save manually.

The auto-saving should apply to all currently not auto-saved assets: Maps, Tilesets and Worlds.

Suggestion: Add a combo-box with text "Autosave interval", with the following options:

* Disabled
* 1 second
* 5 seconds
* 1 minute

Alternatively, a text edit could be added where the user can fill in the exact amount of seconds to wait after a change. Something like:

[ ] Autosave after making changes: [ 1 seconds ]

Feature requested at https://itch.io/post/4512127.

abhinavs1920:
Closes #3133

Adds an optional autosave feature that automatically saves your work after you stop making changes. Helps prevent data loss without being intrusive.

**What it does:**
- New "Autosave interval" setting in Preferences → General → Saving and Loading
- Options: Disabled (default), 1/5/10/15/30/60 minutes
- Only saves after you stop editing for the specified time (debouncing)
- Works for maps, tilesets, and worlds

**Why debouncing:**
The timer restarts with each change, so it only saves after you've been idle for the full interval. This avoids interrupting your workflow with constant saves while still protecting your work.

**What it won't do:**
- Won't save untitled files (you need to save them once first)
- Won't save read-only files
- Won't create backup/recovery files (that's a different feature)

**Since this feature can be customized a lot, I am open for discussions and more ideas :)**

### DIFF
--- a/src/tiled/documentmanager.cpp
+++ b/src/tiled/documentmanager.cpp
@@ -595,7 +595,23 @@ int DocumentManager::insertDocument(int index, const DocumentPtr &document)
     connect(documentPtr, &Document::saved, this, &DocumentManager::onDocumentSaved);
 
     // Connect to undo stack changes for autosave (any command pushed means document was modified)
-    connect(documentPtr->undoStack(), &QUndoStack::indexChanged, this, &DocumentManager::documentModified);
+    connect(documentPtr->undoStack(), &QUndoStack::indexChanged, this, [=] (int index) {
+        Q_UNUSED(index)
+        
+        if (!documentPtr->isModified())
+            return;
+
+        // Only autosave documents that have a file name
+        if (documentPtr->fileName().isEmpty())
+            return;
+
+        int interval = Preferences::instance()->autosaveInterval();
+        if (interval <= 0)
+            return;
+
+        mDocumentsPendingAutosave.insert(documentPtr);
+        mAutosaveTimer->start(interval * 1000);
+    });
 
     mTabBar->insertTab(index, QString());
     updateDocumentTab(documentPtr);
@@ -1498,86 +1514,25 @@ void DocumentManager::autosaveIntervalChanged(int seconds)
 {
     // Stop any pending autosave
     mAutosaveTimer->stop();
-    mDocumentsPendingAutosave.clear();
 
     if (seconds <= 0)
         return;
-
-    for (const auto &doc : mDocuments) {
-        if (doc->isModified() && !doc->fileName().isEmpty()) {
-            mDocumentsPendingAutosave.insert(doc.data());
-        }
-    }
 
     if (!mDocumentsPendingAutosave.isEmpty())
         mAutosaveTimer->start(seconds * 1000);
 }
 
-void DocumentManager::documentModified()
-{
-    auto *undoStack = qobject_cast<QUndoStack*>(sender());
-    if (!undoStack)
-        return;
-
-    Document *document = nullptr;
-    for (const auto &doc : mDocuments) {
-        if (doc->undoStack() == undoStack) {
-            document = doc.data();
-            break;
-        }
-    }
-
-    if (!document)
-        return;
-
-    if (!document->isModified())
-        return;
-
-    // Only autosave documents that have a file name
-    if (document->fileName().isEmpty())
-        return;
-
-    int interval = Preferences::instance()->autosaveInterval();
-    if (interval <= 0)
-        return;
-
-    mDocumentsPendingAutosave.insert(document);
-
-    mAutosaveTimer->start(interval * 1000);
-}
-
 void DocumentManager::performAutosave()
 {
-
     QSet<Document*> documentsToSave = mDocumentsPendingAutosave;
     mDocumentsPendingAutosave.clear();
 
     for (Document *document : documentsToSave) {
-        // Check if document still exists in our list
-        bool found = false;
-        for (const auto &doc : mDocuments) {
-            if (doc.data() == document) {
-                found = true;
-                break;
-            }
-        }
-
-        if (!found) {
-            qDebug() << "Autosave: Document no longer exists, skipping";
+        if (!document->isModified())
             continue;
-        }
 
-        if (!document->isModified()) {
+        if (document->isReadOnly())
             continue;
-        }
-
-        if (document->fileName().isEmpty()) {
-            continue;
-        }
-
-        if (document->isReadOnly()) {
-            continue;
-        }
 
         // Perform the save
         saveDocument(document);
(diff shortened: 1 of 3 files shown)


abhinavs1920:
Heyy @bjorn , @eishiya I have implemented the autosave feature.. let me know your thoughts once you are available.

src/tiled/preferencesdialog.cpp
```diff
@@ -68,6 +68,14 @@ PreferencesDialog::PreferencesDialog(QWidget *parent)
     mUi->styleCombo->setItemData(0, Preferences::SystemDefaultStyle);
     mUi->styleCombo->setItemData(1, Preferences::TiledStyle);
 
+    mUi->autosaveCombo->addItem(tr("Disabled"), 0);
+    mUi->autosaveCombo->addItem(tr("1 minute"), 60);
+    mUi->autosaveCombo->addItem(tr("5 minutes"), 300);
+    mUi->autosaveCombo->addItem(tr("10 minutes"), 600);
+    mUi->autosaveCombo->addItem(tr("15 minutes"), 900);
+    mUi->autosaveCombo->addItem(tr("30 minutes"), 1800);
+    mUi->autosaveCombo->addItem(tr("60 minutes"), 3600);
```
bjorn: Rather than having each of these be its own translatable string, please use `QCoreApplication::translate`, putting `%n` in the string somewhere and using the `n` parameter to have it choose the right plural form.

Please also add "1 second" and "5 seconds" though. I think we can remove "15 minutes" and "60 minutes" to keep the list from being too long. I don't think 15 adds much value in addition to 10 minutes and I think 60 minutes is way too long.

src/tiled/preferencesdialog.cpp
```diff
@@ -96,6 +104,13 @@ PreferencesDialog::PreferencesDialog(QWidget *parent)
     connect(mUi->naturalSorting, &QCheckBox::toggled,
             preferences, &Preferences::setNaturalSorting);
 
+    connect(mUi->autosaveCombo, qOverload<int>(&QComboBox::currentIndexChanged), this, [=] (int index) {
```
bjorn: I don't think we need the `qOverload`.

```suggestion
    connect(mUi->autosaveCombo, &QComboBox::currentIndexChanged, this, [=] (int index) {
```

src/tiled/documentmanager.cpp
```diff
@@ -582,6 +594,9 @@ int DocumentManager::insertDocument(int index, const DocumentPtr &document)
     connect(documentPtr, &Document::changed, this, &DocumentManager::onDocumentChanged);
     connect(documentPtr, &Document::saved, this, &DocumentManager::onDocumentSaved);
 
+    // Connect to undo stack changes for autosave (any command pushed means document was modified)
+    connect(documentPtr->undoStack(), &QUndoStack::indexChanged, this, &DocumentManager::documentModified);
```
bjorn: I'd suggest to use a lambda here that captures the `documentPtr`. It should be guaranteed to stay valid because if it were deleted, the `QUndoStack` would be deleted along with it. That way, we don't need to search for the document by the undo stack pointer in `DocumentManager::documentModified`.

src/tiled/documentmanager.cpp
```diff
@@ -1476,4 +1494,94 @@ void DocumentManager::abortMultiDocumentClose()
...
+void DocumentManager::performAutosave()
+{
+
+    QSet<Document*> documentsToSave = mDocumentsPendingAutosave;
+    mDocumentsPendingAutosave.clear();
+
+    for (Document *document : documentsToSave) {
+        // Check if document still exists in our list
+        bool found = false;
+        for (const auto &doc : mDocuments) {
```
bjorn: You're already removing documents from `mDocumentsPendingAutosave` in `DocumentManager::closeDocumentAt`, so I don't think this check is necessary.

src/tiled/documentmanager.cpp
```diff
@@ -1476,4 +1494,94 @@ void DocumentManager::abortMultiDocumentClose()
...
+        if (!found) {
+            qDebug() << "Autosave: Document no longer exists, skipping";
+            continue;
+        }
+
+        if (!document->isModified()) {
+            continue;
+        }
+
+        if (document->fileName().isEmpty()) {
```
bjorn: Since `saveDocument` also checks this, we don't need another check here.

src/tiled/documentmanager.cpp
```diff
@@ -1476,4 +1494,94 @@ void DocumentManager::abortMultiDocumentClose()
...
+    // Stop any pending autosave
+    mAutosaveTimer->stop();
+    mDocumentsPendingAutosave.clear();
+
+    if (seconds <= 0)
+        return;
+
+    for (const auto &doc : mDocuments) {
+        if (doc->isModified() && !doc->fileName().isEmpty()) {
+            mDocumentsPendingAutosave.insert(doc.data());
```
bjorn: I don't think we need to clear this set and then re-build it.

In fact, I wonder if we need to keep track of the set of modified documents at all? Since we're only using a single autosave timer, it should be enough to just start that timer and when it triggers, save any modified documents.

abhinavs1920:
Simplified autosave based on the review:

Updated plural handling using QCoreApplication::translate with %n
Added 1s and 5s options, removed longer unused intervals

Also addressed all other minor suggestions like simplifying connections, using a lambda for documentModified, and cleaning up redundant logic.

Thanks @bjorn, this made the implementation much cleaner :)

eishiya:
The point of using `%n` is that a single string can handle all the number forms (which are more complex than just singular and plural in some languages), so there should not be separate "1 minute" and "%n minutes" strings. I'm not 100% sure and hopefully Bjorn will correct me, but I think it should be "%n minutes" for both, just with `1` as the parameter to `%n` in the 1 minute option. The localisation system will handle using the correct word (e.g. "minute", "minutes") depending on the parameter and the language.

Edit: or rather, "%n minute(s)" for consistency with other strings of this type.

abhinavs1920:
You're right! I initially thought we needed separate strings for singular vs plural, but after reading the [Qt docs](https://doc.qt.io/qt-6/i18n-source-translation.html#handle-plural-forms), I now understand that %n handles all number forms automatically based on language rules. Updated to use "%n second(s)" and "%n minute(s)" consistently for all values.

Bit of overthinking :/

Issue #4063, opened 2024-09-28
Tiled not unloading files stored in RAM


cowardlycamper32:
**Describe the bug**
While loading a large sum of image files, I noticed memory usage had hit 90%+ and so closed the file i was currently working on. Memory usage did not change.

**To Reproduce**
1. Load large amount of files
2. Open Task Manager, or similar system viewer
3. Close Tab with large amount of files
4. Observe memory usage not changing

**Expected behavior**
Expected memory usage to decrease when closing tab and deloading the images.

**Media**
N/A

**Specifications:**
 - OS: Windows 11
 - Tiled Version: 1.11.0

eishiya:
You closed the tab of the file, but did you make sure the file wasn't actually used by some other document, and therefore still loaded? Tilesets remain loaded if they're included in any open map, and I think Maps remain loaded if their World is loaded.

cowardlycamper32:
I havent used any of the assets yet so the tiles shouldnt be used elsewhere. I understand the reopen closed tab part, but from further testing, memory usage seems to increse slightly upon closing a tab

bjorn:
Currently the image cache used by Tiled indeed never gets cleared, so if Tiled eats lots of memory due to having loaded many images, the only way to reduce memory usage is by restarting Tiled.

cortexA233:
Hi, @bjorn , can I introduce a configurable cache size limit (maybe LRU-based?), or just add a manual "clear cache" button to solve this issue?
I’d be happy to implement either of these 2 thoughts if these direction make sense to you.

bjorn:
@cortexA233 I'm not sure a limit works best in our case, since it is meant to avoid loading the same image multiple times. If we limit the cache we can get into this situation:

* Image A is loaded and cached
* Image B is loaded and cached, cache is full so image A is evicted (but it remains in memory, since it is still in use)
* Image A is loaded and cached again -> **now the same image exists in memory twice**

The above could make the memory situation worse. So if we're going to evict anything from the cache, we need to limit this to those images which are no longer referenced apart from their entry in the cache. This is something we should ideally do with a small delay.

I see you've already done quite a bit of work towards this, so I've assigned you the issue. I think we can use `QImage::isDetached` to check for images that are candidates for eviction (it checks whether the internal reference count is 1). The main issue is that this function is marked "internal" and as such does not appear in the documentation, but it's a public non-deprecated function so maybe it's fine if we use it.

PR #4386, opened 2026-03-04
[WIP] Add a optional image cache limit, fix issue #4063

cortexA233:
Try to fix https://github.com/mapeditor/tiled/issues/4063
Add a image cache limit, with a LRU-based cache maintenance strategy, and flexible image cache limit option in preference dialog

### DIFF
--- a/src/libtiled/imagelayer.cpp
+++ b/src/libtiled/imagelayer.cpp
@@ -32,6 +32,7 @@
 #include "imagecache.h"
 
 #include <QBitmap>
+#include <QImage>
 
 using namespace Tiled;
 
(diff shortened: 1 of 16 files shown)


MollahHamza:
Looks good overall. Just wondering how do we ensure images still referenced by open documents aren’t evicted prematurely?

cortexA233:
> Looks good overall. Just wondering how do we ensure images still referenced by open documents aren’t evicted prematurely?

I think it's just a question of trade-off, it's unavoidable that some images referenced by other files would be evicted (for example, the user have a project with 3 different tmx map, and each map includes tiles from 1000+ different tsx tilesets), in that case ```ImageCache::loadImage()``` function will reload the images from the disk, so the final output should be correct. Maybe I can make this feature to be optional (I will add a checkbox for this immediately)

cortexA233:
Hi @bjorn , when I working on this PR, I found that the image-cache limit won't solve the RAM issue if many maps are opened/referenced at the same time (likely due to map object/reference lifetimes, not only image cache entries).

Would it make sense to discuss guardrails such as:
  
  - a configurable limit on simultaneously opened maps, or warning prompts when map count/memory crosses a threshold, 
  - a more aggressive resource release strategy for background maps? Like when map count/memory crosses a threshold, automatically delete the image references of oldest unused maps. 
 
I’m not sure which direction best fits the project, so I’d really value your guidance. If this sounds useful, I can draft a new PR focus on the issue. Thanks!

PR #4387, opened 2026-03-04
Add "Property Value" checkbox to custom enums and class

### LINKED ISSUE #4118: Add a flag so custom classes and custom enums are only usable as class members
**Feature request description**
As far as I know, it is currently not possible to select which custom enums we see when adding a new property.
We always see all available enums, while some of them could only be used through custom classes (and not directly added as properties).
It would be nice to be able to select which enum to show or to hide.

When creating a new custom enum, add the possibility to hide it when adding new properties, ie. to not use it as a property value.

It could be done the same as for custom classes, by adding a `Property value` tickbox:
image

For JSON export, we could use the same `useAs` array but only take into account the `property` value.

abhinavs1920:
Closes #4118

## Summary

This PR implements adding a "Property Value" checkbox to custom enums, matching the existing functionality for custom classes. This checkbox controls whether a type appears in the top-level property dropdown, while all types remain usable as class members regardless of the checkbox state.

### 1. Extended Enum Type with Usage Flags
- Added `usageFlags` field to `EnumPropertyType` (defaults to `0x001` for backward compatibility)
- Added `isPropertyValueType()` method to check if enum can be used as top-level property
- Updated JSON serialization to export/import `useAs` array for enums

### 2. Updated UI
- Added "Property Value" checkbox to enum editor in `PropertyTypesEditor`
- Positioned consistently with class editor layout
- Connected checkbox to update enum's `usageFlags`

### 3. Updated Filtering Logic
- Modified `possiblePropertyValues()` to check `isPropertyValueType()` for both classes AND enums
- Clarified in comments that filtering only applies to top-level properties, not class members

### DIFF
--- a/src/libtiled/properties.cpp
+++ b/src/libtiled/properties.cpp
@@ -504,13 +504,8 @@ QVariantList possiblePropertyValues(const ClassPropertyType *parentClassType)
         // Filter types based on PropertyValueType flag for top-level properties.
         // Note: This filtering does NOT apply to class member selection - all types
         // are always usable as class members regardless of this flag.
-        if (propertyType->isClass()) {
-            if (!static_cast<const ClassPropertyType&>(*propertyType).isPropertyValueType())
-                continue;
-        } else if (propertyType->isEnum()) {
-            if (!static_cast<const EnumPropertyType&>(*propertyType).isPropertyValueType())
-                continue;
-        }
+        if (!parentClassType && !propertyType->isPropertyValueType())
+            continue;
 
         values.append(propertyType->wrap(propertyType->defaultValue()));
     }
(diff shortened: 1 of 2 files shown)


src/libtiled/properties.cpp
```diff
@@ -501,10 +501,16 @@ QVariantList possiblePropertyValues(const ClassPropertyType *parentClassType)
         if (parentClassType && !parentClassType->canAddMemberOfType(propertyType.data()))
             continue;
 
-        // Avoid suggesting classes not meant to be used as property value
-        if (propertyType->isClass())
+        // Filter types based on PropertyValueType flag for top-level properties.
+        // Note: This filtering does NOT apply to class member selection - all types
+        // are always usable as class members regardless of this flag.
```
bjorn, line 506: This comment communicates the correct intended behavior, but the code continues to apply filtering to class member selected as before. If we want to ignore the "property value type" flag for class member selection, we should skip these checks when `parentClassType` is set.

PR #4392, opened 2026-03-05
TiledQuick: Add border and grid over mapitem.

UltraDagon:
Draws a simple 1px-width border around `MapItem` using the Qt Scene Graph. The border is drawn on top of the map, and the color is changeable within `tiledquick/qml/main.qml`.

Draws a grid across the map. The color is changeable within `tiledquick/qml/main.qml`. The grid's dash length, space length after dash, and thickness are changeable within `libtiledquick/grid.frag`.

This is my first pull request, so I chose a small feature and wanted to practice mirroring the existing style and standards. This problem has also helped me gain some experience with QtQuick, and I plan on implementing the mouse-following grid-snapping object found in the default Tiled next.

As part of my attempt to imitate the existing style, I have kept my in-code comments to a minimum. Please let me know if including more comments is the better practice, and I can go back and add them.

Any and all feedback is appreciated!

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,5 +1,5 @@
 clone_depth: 200
-version: "1.12.{build}"
+version: '1.12.{build}'
 image:
   - Visual Studio 2022
 
@@ -107,9 +107,9 @@ after_build:
 
 artifacts:
   - name: Installer
-    path: "Tiled-*.msi"
+    path: 'Tiled-*.msi'
   - name: Archive
-    path: "tiled-*.7z"
+    path: 'tiled-*.7z'
 
 deploy:
   - provider: FTP
@@ -121,30 +121,30 @@ deploy:
     folder: update.mapeditor.org/snapshots-win
     artifact: /.*\.(msi|7z)/
     on:
-      branch: snapshot
-      push_release: true
+        branch: snapshot
+        push_release: true
 
   - provider: Webhook
     url: https://app.signpath.io/API/v1/670574d6-49bf-4a0c-824a-3ae977077079/Integrations/AppVeyor?ProjectKey=tiled&SigningPolicyKey=test-signing
     authorization:
-      secure: 4E7IuM1Ftvdkx43gsqI3tUWb6tcvrfKB22sa6DsWaBzT+zv3DqceDCk0qc/JqTO+Er5UQNYwolmWBzMdys6fVA==
+        secure: 4E7IuM1Ftvdkx43gsqI3tUWb6tcvrfKB22sa6DsWaBzT+zv3DqceDCk0qc/JqTO+Er5UQNYwolmWBzMdys6fVA==
     on:
-      appveyor_repo_tag: false
-      branch: master
-      push_release: true
+        appveyor_repo_tag: false
+        branch: master
+        push_release: true
 
   - provider: Webhook
     url: https://app.signpath.io/API/v1/670574d6-49bf-4a0c-824a-3ae977077079/Integrations/AppVeyor?ProjectKey=tiled&SigningPolicyKey=release-signing
     authorization:
-      secure: 4E7IuM1Ftvdkx43gsqI3tUWb6tcvrfKB22sa6DsWaBzT+zv3DqceDCk0qc/JqTO+Er5UQNYwolmWBzMdys6fVA==
+        secure: 4E7IuM1Ftvdkx43gsqI3tUWb6tcvrfKB22sa6DsWaBzT+zv3DqceDCk0qc/JqTO+Er5UQNYwolmWBzMdys6fVA==
     on:
-      branch: snapshot
-      push_release: true
+        branch: snapshot
+        push_release: true
 
   - provider: Webhook
     url: https://app.signpath.io/API/v1/670574d6-49bf-4a0c-824a-3ae977077079/Integrations/AppVeyor?ProjectKey=tiled&SigningPolicyKey=release-signing
     authorization:
-      secure: 4E7IuM1Ftvdkx43gsqI3tUWb6tcvrfKB22sa6DsWaBzT+zv3DqceDCk0qc/JqTO+Er5UQNYwolmWBzMdys6fVA==
+        secure: 4E7IuM1Ftvdkx43gsqI3tUWb6tcvrfKB22sa6DsWaBzT+zv3DqceDCk0qc/JqTO+Er5UQNYwolmWBzMdys6fVA==
     on:
-      appveyor_repo_tag: true
-      push_release: true
+        appveyor_repo_tag: true
+        push_release: true
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,5 +1,5 @@
 clone_depth: 200
-version: '1.12.{build}'
+version: "1.12.{build}"
 image:
   - Visual Studio 2022
 
@@ -17,7 +17,7 @@ environment:
   matrix:
     - QT_VERSION: 6.10.2
       QT_ARCH: win64_msvc2022_64
-      QT_MODULES: qtimageformats
+      QT_MODULES: qtimageformats qtshadertools
       QTDIR: C:\Qt\6.10.2\msvc2022_64
       PYTHONHOME: C:\Python312-x64
       DEFAULT_PROFILE: MSVC2022-x64
@@ -39,7 +39,7 @@ environment:
       CXX: i686-w64-mingw32-g++.exe
     - QT_VERSION: 6.10.2
       QT_ARCH: win64_mingw
-      QT_MODULES: qtimageformats
+      QT_MODULES: qtimageformats qtshadertools
       QTDIR: C:\Qt\6.10.2\mingw_64
       PYTHONHOME: C:\Python312-x64
       MINGW: C:\Qt\Tools\mingw1310_64
@@ -107,9 +107,9 @@ after_build:
 
 artifacts:
   - name: Installer
-    path: 'Tiled-*.msi'
+    path: "Tiled-*.msi"
   - name: Archive
-    path: 'tiled-*.7z'
+    path: "tiled-*.7z"
 
 deploy:
   - provider: FTP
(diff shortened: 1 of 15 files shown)


UltraDagon:
Update: Added Grid over mapitem

Created the following files to make the new GPU-rendered grid feature:
- mapgriditem.h/cpp: The QT item that houses and updates the mapgridmaterial, which draws the grid using a shader.
- mapgridmaterial.h/cpp: Builds and passes values into the grid shader.
- grid.vert: The vertex shader that defines the bounding box for the fragment shader.
- grid.frag: The fragment shader that defines what pixels should change at which positions. Contains the math for how the grid is defined.

Other files were updated for the following reasons:
- Add the grid QT item into main.qml.
- Register and support the grid QT item and newly created files.
- Support the use of the qtshadertools module.

This addition was much larger than my last one, and I am definitely expecting to have gotten some of it wrong. Any and all feedback is appreciated!

I added a few comments pointing out areas I'd like feedback on, if possible. Thank you!

src/libtiledquick/libtiledquick.qbs
```diff
@@ -3,11 +3,11 @@ import qbs.Utilities
 DynamicLibrary {
     targetName: "tiledquick"
     builtByDefault: false
-    condition: Utilities.versionCompare(Qt.core.version, "6.5") >= 0
+    condition: Qt.core && Utilities.versionCompare(Qt.core.version, "6.5") >= 0
```
UltraDagon, line 6: This is to avoid a crash when running the codeql-analysis test, as Qt.core was previously being compared while undefined (since it's running an older version of Qt).

If there is a better way to handle this, please let me know.

src/libtiledquick/libtiledquick.qbs
```diff
@@ -36,6 +36,10 @@ DynamicLibrary {
     }
 
     files: [
+        "mapgriditem.cpp",
```
UltraDagon, line 41: Not sure if it's best practice to keep my newly created files at the top or bottom of this list, so I just kept them where the autogenerator put them.

src/libtiledquick/libtiledquick.qbs
```diff
@@ -55,6 +61,17 @@ DynamicLibrary {
...
 
+    Group {
+        name: "Shaders"
+        files: [
+            "grid.vert",
+            "grid.frag",
+        ]
+        fileTags: ["qt.shadertools.qsb"]
+
+        overrideTags: false
```
UltraDagon: This line is here due to a strange issue where it would not compile the shaders unless this line was present. Would love a short explanation if possible.

src/libtiledquick/libtiledquick.qbs
```diff
@@ -65,6 +82,8 @@ DynamicLibrary {
         cpp.includePaths: exportingProduct.sourceDirectory
     }
 
+    Depends { name: "Qt.quick" }
```
UltraDagon: Whoops! This does not need to be here. I'm pretty sure Qt's code auto-generator thing made it. (removed in subsequent push)

src/tiledquick/qml/main.qml
```diff
@@ -160,6 +160,23 @@ ApplicationWindow {
...
+                anchors.fill: mapItem
+
+                color: "black"
+            }
+
+            Tiled.MapGridItem {
+                id: mapGriditem
+                anchors.fill: mapItem
+
+                gridSize: mapItem.pixelToTileCoords(width, height);
```
UltraDagon: For this property, I am not sure if this is the best way to get this information. Please let me know if there is a better way to get the grid size (X tiles by Y tiles) or tile size (in pixels).

appveyor.yml
UltraDagon, line 1: The changes in this file, aside from adding qtshadertools on lines 20 and 42, were all due to a VSCode extension auto-reformatting the file on save. I can restore it to the older version's formatting. Just let me know if it is necessary.

src/libtiledquick/libtiledquick.qbs
bjorn: The order is supposed to be alphabetical, but the existing list already wasn't ordered correctly. I'd suggest we fix this while we're adding stuff.

src/libtiledquick/libtiledquick.qbs
bjorn: That's a bit odd because `Qt.core` should be available. Maybe we could add a separate `Depends { name: "Qt.core"; }` line below instead, but I guess this check is better since it may also avoid some unhelpful errors when even Qt core isn't found.

src/libtiledquick/libtiledquick.qbs
bjorn: I don't think it's necessary to explicitly specify the file tags, and maybe doing so was overriding the [automatic tags](https://doc.qt.io/qbs/qml-qbsmodules-qt-shadertools.html#relevant-file-tags) that the `Qt.shadertools` module was specifying? I think we can juts delete both `overrideTags` and the `fileTags` properties.

src/tiledquick/qml/main.qml
bjorn: I guess there isn't yet. I think we'll need to somehow expose the loaded map as a `EditableMap` instance, which would make the necessary properties accessible from QML.

appveyor.yml
bjorn: Yeah it would be nice to make such changes in a separate commit, if they're making the file more consistent.

bjorn (changes requested):
This is quite a nice change, @UltraDagon, thanks for looking into these things!

I wondered a little about the purpose of the `MapBorderItem` as opposed to a standard `Rectangle` item, but I see the latter does not have a "cosmetic" mode so its border width would need to be inverse-adjusted to the scale to render at 1px. The specialized `MapBorderItem` might be the better solution.

I've tweaked the `MapGridItem`'s fragment shader a little to place the grid on the edge between the tiles rather than on the tile. It really only makes difference in case of a width > 1 but it seems like this way we also didn't need the specific checks to disable the leftmost and topmost lines.

UltraDagon:
Thank you for the feedback, Bjorn! I will work to get these changes finished within a few days.

UltraDagon:
I ended up using mapItem's mMap to get the tile size information, and from that, I was able to get the grid size in tiles.

src/libtiledquick/mapgriditem.cpp
```diff
@@ -0,0 +1,109 @@
...
+    vertices[2].set(0, height(), 0, 1);
+    vertices[3].set(width(), height(), 1, 1);
+
+    auto *material = static_cast<MapGridMaterial *>(gridNode->material());
+    material->mColor = mColor;
+    material->mScale = mScale;
+    material->mPixelWidth = width();
+    material->mPixelHeight = height();
+    material->mTileWidth = width() / mGridSize.x();
+    material->mTileHeight = height() / mGridSize.y();
```
bjorn: This may divide by zero when `mGridSize` isn't set. But I think more interesting is that `MapGridItem` actually needs the tile size, not the grid size. So I'd suggest to rename `mGridSize` to `mTileSize` and set it accordingly, so that here we can just assign that to the material.

src/libtiledquick/mapgridmaterial.cpp
```diff
@@ -0,0 +1,76 @@
...
+{
+    return new MapGridShader();
+}
+
+int MapGridMaterial::compare(const QSGMaterial *other) const
+{
+    auto *m = static_cast<const MapGridMaterial *>(other);
+    return (mColor == m->mColor &&
+            qFuzzyCompare(mScale, m->mScale) &&
+            qFuzzyCompare(mPixelWidth, m->mPixelWidth) ? 0 : 1);
```
bjorn: I'm not sure this is doing what the `compare` function should be doing. The Qt docs talk about returning -1, 0 or 1, to establish a rendering order for materials that reduces state switches. Then again I'm also not entirely sure at the moment which properties are relevant in that case.

Why did you choose these properties and left out `mPixelHeight`, `mTileWidth` and `mTileHeight`?

src/tiledquick/qml/main.qml
```diff
@@ -160,6 +160,23 @@ ApplicationWindow {
...
+                anchors.fill: mapItem
+
+                color: "black"
+            }
+
+            Tiled.MapGridItem {
+                id: mapGriditem
+                anchors.fill: mapItem
+
+                gridSize: Qt.point(width / mapItem.tileSize().width, height / mapItem.tileSize().height);
```
bjorn: Alright, so I just noticed the `MapGridMaterial` really just needs the tile size, so it's extra odd to do this calculation here. This should just be:
```suggestion
                tileSize: mapItem.tileSize();
```
Once you've made the other suggested change. :-)

src/tiledquick/qml/main.qml
```diff
@@ -216,7 +233,7 @@ ApplicationWindow {
                     } else {
                         var mapRelativeCoords = singleFingerPanArea.mapToItem(mapItem, singleFingerPanArea.mouseX, singleFingerPanArea.mouseY)
                         var tileCoords = mapItem.screenToTileCoords(mapRelativeCoords.x, mapRelativeCoords.y)
-                        Math.floor(tileCoords.x) + ", " + Math.floor(tileCoords.y)
+                        Math.floor(tileCoords.x) + ", " + Math.floor(tileCoords.y);
```
bjorn:
```suggestion
                        Math.floor(tileCoords.x) + ", " + Math.floor(tileCoords.y)
```
Just suggesting to revert addition of semicolon. I'm open to changing to coding style, but then we'll do it consistently. :-)

bjorn (changes requested):
I've pushed some further tweaks and left you some more feedback. It's not quite ready for merging yet, but getting close. :-)

src/libtiledquick/mapgridmaterial.cpp
UltraDagon: I had forgotten to add the rest of the comparable properties. Good catch, thank you!

src/tiledquick/qml/main.qml
UltraDagon: Sorry about that! I use that text area for testing and left it there by mistake.

bjorn:
@UltraDagon Hmm, you've opened this PR from your `master` branch and are now conflating too many unrelated changes. I think adding grid and border in the same PR was still fine, but let's close that first and then open a new PR for the `EditableMap` related change. I can try to help if you're unfamiliar with managing branches in git, let me know on Discord.

UltraDagon:
Successfully split the maploader/editablemap changes into it's own pull request ([#4520](https://github.com/mapeditor/tiled/pull/4520))!

UltraDagon:
Would you like me to add the copyright headers to all files added in the pull request?

bjorn:
> Would you like me to add the copyright headers to all files added in the pull request?

I've added them to the source files mostly for consistency. I think now only the shaders do not have them, but you could add a brief comment to them as well. I personally wouldn't put the full copyright header there (it's actually a rather old habit by now I guess, which I haven't extended to QML files either). I think we could also merge this change as-is now.

bjorn (approved):

UltraDagon:
My apologies, I misread the commit and thought that only mapgridmaterial.h/cpp had the copyright headers. I agree that it would make more sense to only have the headers in the source files and leave them out of the other files. It should be good to merge, and I'll update my other PR to change the QML to use the new editablemap property.

PR #4393, opened 2026-03-05
Add drag-to-toggle for layer visibility and lock icons

### LINKED ISSUE #2864: Easier showing/hiding of layers by dragging over eye icons
To quote eishiya from the [Tiled forum](https://discourse.mapeditor.org/t/quality-of-life-feature-suggestions/4619/3):

> In Photoshop, I can quickly show/hide several layers by clicking on one visibility icon and dragging to others, and it’ll set their visibility to be the same:
> 
> eye-icon
> 
> I often try to do this in Tiled out of habit, but it drags the layer in the layer stack instead. It would be nice if Tiled had this same feature, or at least didn’t drag the layer if I click+drag on the eye or lock icons rather than the rest of the layer.

I totally agree the eyes should work like this. Also worth considering in combination with #1995.

keshav0479:
Fixes #2864

You can now click on an eye or lock icon and drag across rows to toggle multiple layers at once - same behavior as Photoshop. Dragging sets all crossed layers to the same state as the first toggle, and the whole drag undoes in one Ctrl+Z.

For the merge logic, I went with bjorn's suggested approach - added a `mergeable` flag and custom `mergeWith` on `SetLayerVisible`/`SetLayerLocked`, with a protected `append()` and `values()` on `ChangeValue` to support it.

Normal single-click toggling and layer reordering are unaffected.

I'm looking into extending this to ObjectsView too - it uses a different undo path (`ChangeMapObject`) so it'll need a bit of a different approach. Will add as a follow-up commit if it works out cleanly.

https://github.com/user-attachments/assets/3f1b3ba1-1c00-4d13-8632-f9f21ee463c5

### DIFF
--- a/src/tiled/layerdock.cpp
+++ b/src/tiled/layerdock.cpp
@@ -280,6 +280,11 @@ void LayerView::mousePressEvent(QMouseEvent *event)
         return;
     }
 
+    if (event->modifiers()) {
+        QTreeView::mousePressEvent(event);
+        return;
+    }
+
     Layer *layer = layerForProxyIndex(proxyIndex);
     if (!layer) {
         QTreeView::mousePressEvent(event);
--- a/src/tiled/objectsview.cpp
+++ b/src/tiled/objectsview.cpp
@@ -32,9 +32,12 @@
 #include <QGuiApplication>
 #include <QHeaderView>
 #include <QKeyEvent>
+#include <QMouseEvent>
 #include <QMenu>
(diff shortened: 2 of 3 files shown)


keshav0479:
Also got it working in the Objects panel, same drag behavior, works for both object rows and layer rows.
Ready for Review :)

https://github.com/user-attachments/assets/7a948769-b7f3-4b62-b6cf-aae4ae56ae99

PR #4394, opened 2026-03-05
Fix the placeholder tiles selection bug

### LINKED ISSUE #3498: It should not be possible to select "non-existing" Tiles in Collection of Images.
For end users, this almost feels buggy:

Strange selection of Objects

I can select 3 "empty" Tile slots, where nothing is there and I can assign them with values ("xxx") like in the GIF. But obviously I'm not assigning them with values, but the very last Object (the leek in the example) is assigned. It would be less irritating for the User to not be able to select these empty slots at all.

cortexA233:
Fixed https://github.com/mapeditor/tiled/issues/3498
Now when users click a placeholder cell, the current selected cell will be deselected, instead of staying the selection on the placeholder cell.
GIF 2026-3-5 14-57-28
GIF 2026-3-5 14-56-38

### DIFF
--- a/src/tiled/tileseteditor.cpp
+++ b/src/tiled/tileseteditor.cpp
@@ -582,12 +582,7 @@ void TilesetEditor::currentChanged(const QModelIndex &index)
 {
     if (index.isValid()) {
         auto model = static_cast<const TilesetModel*>(index.model());
-        auto newTile = model->tileAt(index);
-        if (newTile && newTile->imageShape().isEmpty()) {
-            setCurrentTile(nullptr);
-        } else {
-            setCurrentTile(newTile);
-        }
+        setCurrentTile(model->tileAt(index));
     } else {
         setCurrentTile(nullptr);
     }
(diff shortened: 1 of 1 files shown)


bjorn:
Thanks for looking into this issue!

Unfortunately, I don't think this is the right approach to fixing this. We don't need the Properties view to become empty, keeping the last selected object visible there is consistent with all other types of selectable objects.

I tried doing the following instead, to simply disable the selection for these non-tile items in the `TilesetModel`:

```diff
 Qt::ItemFlags TilesetModel::flags(const QModelIndex &index) const
 {
     Qt::ItemFlags defaultFlags = QAbstractListModel::flags(index);
     defaultFlags |= Qt::ItemIsDropEnabled;
 
     if (index.isValid())
         defaultFlags |= Qt::ItemIsDragEnabled;
 
+    if (!tileAt(index))
+        defaultFlags &= ~Qt::ItemIsSelectable;
+
     return defaultFlags;
 }
```
But this still didn't stop keyboard navigation and generally didn't prevent those indexes from becoming the "current" one. Then I tried overriding `TilesetModel::index` to not return valid indexes for these locations, and that seems to have worked. I've opened #4512 with this alternative fix.

Issue #4396, opened 2026-03-06
Export as Image gives no error when image save fails


PrateekSingh070:
### Describe the bug

When using File → Export as Image, if the image file cannot be saved (due to file permissions, an invalid path, or an unsupported file extension), the dialog closes without showing any error. The user gets no indication that the export failed. This is inconsistent with map and tileset export paths, which do show error dialogs on failure.

The root cause is in `exportasimagedialog.cpp` — the return value of `QImage::save()` is not checked.

### Steps to reproduce

1. Open any map in Tiled
2. Go to File → Export as Image...
3. Choose a read-only directory as the save location (or use an unsupported file extension like .xyz)
4. Click Save
5. The dialog closes silently — no image file is created and no error is shown

### Expected behavior

An error dialog should appear informing the user that the image could not be saved, similar to how other export operations (map export, tileset export) report errors when writing fails.

### OS version

Windows 10

### Tiled version

Latest

kunal649:
hey! Ive submitted a PR that fixes this ...
PR: #4397

@bjorn , sir can you review it, it's a minor fix !

PrateekSingh070:
@bjorn, sir is this pr merged or should I solve this issue

kunal649:
both PR's are reviewed but not merged ~ 
#4397 
#4409

bjorn:
Closing now that #4397 is merged. :-)

PR #4397, opened 2026-03-06
Show Error dialog when export-to-image fails to save

kunal649:
Fixes an Issue #4396 where File → Export as Image would silently close the dialog if the image could not be saved, giving the user no feedback about the failure. 

The return value of QImage::save() was not being checked. this commit wraps the call in an error check and shows a warning dialog if saving fails, consistent with how other export operations handle errors in Tiled. 

thanks :-) 

Screenshot 2026-03-06 115817

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -29,6 +29,7 @@
 * Fixed undo behavior after resizing objects certain ways (by Kanishka, #4339)
 * Fixed suggesting filename with trailing dot when export filter is unset (by Sid, #4368)
 * Fixed snapping mode sync across instances (by Sid, #4364)
+* Fixed missing error message when 'Export as Image' fails (by kunal649, #4397)
 * Scripting: Added API for custom property types (with dogboydog, #3971)
 * Scripting: Added TileMap.chunkSize and TileMap.compressionLevel properties
 * Scripting: Added optional defaultValue and toolTip params to Dialog add widget methods (by Oval, #4358)


src/tiled/exportasimagedialog.cpp
```diff
@@ -204,7 +204,13 @@ void ExportAsImageDialog::accept()
 
         miniMapRenderer.renderToImage(image, renderFlags);
 
-        image.save(fileName);
+        if (!image.save(fileName)) {
+             QMessageBox::warning(this,
+                         tr("Export Failed"),
+                         tr("The image could not be saved to the specified location. "
+                            "Please check the file path and permissions."));
```
bjorn: It's a good idea to add some error reporting here. We can actually do even better by using `QImageWriter imageWriter(fileName)`. When `imageWriter.write(image)` fails we can use `imageWriter.errorString()` to get the failure reason and include that in the popup message.

kunal649: didn't know about it, and yeah its better! 
~ I've updated it.
image

bjorn:
Thanks @kunal649! I've merged this despite the string freeze for Tiled 1.12, because I think an English error is still better than nothing at all. :-)

kunal649:
I'll keep that in mind next time 🙂‍↕️

bjorn:
Well, there's no need for you to keep the string freeze in mind I guess. Not sure why I mentioned it, but I'm holding off on merging some changes because they add new strings to the UI at least.

PR #4400, opened 2026-03-06
fix parenting of tileset filter edit widget

dogboydog:
Bug introduced in #4239 . On some systems this causes a small window to temporarily appear as Tiled starts up


bjorn:
> Kody Review Complete

I've turned off this useless message now.

bjorn:
> MCP Connection Failed

I've asked about this issue on Kodus support channel, because I have no idea what I can do to fix this. But seems like they're not active now.

bjorn:
Thanks @dogboydog for your effort in finding out what caused this issue and for submitting the fix! And I'm sorry for the AI mess!

There was a rather strange failure in the CI build but I will ignore this for now since it has nothing to do with this fix.

PR #4401, opened 2026-03-06
Fix point object bounding box in multi-object selections

### LINKED ISSUE #3784: Multi-object selection bounding box wrong for Points
When a Point Object is part of a multi-object selection, the bounding box of the selection takes into account the entire shape of the displayed "pin" shape, rather than just the actual point. This can make the bounding box larger than expected, affecting snapping and rotation behaviour. The behaviour is inconsistent with how Points behave on their own.

In addition to that, the shape taken into account is the "pin" shape 100% zoom, so at other zooms the bounding box doesn't even match the visuals:
At 100%:
image
At 200%:
image

I expect Point objects to be treated as **points** even in multi-object selections, and for the "pin" visualisation to have no effect on the combined bounding box and therefore snapping and rotation when they are part of a multi-object selection.

Observed in a 2023.07.07 nightly on Windows 10, but I suspect this behaviour is older and not OS-dependent.

MollahHamza:
Fixes #3784

The objectBounds() function was using renderer->shape(object) for Point objects, which returns the visual pin shape rather than the actual point position. This caused the selection bounding box to be larger than expected, affecting snapping and rotation behavior.

Fix by returning a zero-size rect at the point's screen position, consistent with how Point objects are treated geometrically.

Before:
Screenshot 2026-03-06 221013

After:
Screenshot 2026-03-06 221340

### DIFF
--- a/src/tiled/objectselectiontool.cpp
+++ b/src/tiled/objectselectiontool.cpp
@@ -924,7 +924,8 @@ static QRectF objectBounds(const MapObject *object,
             return transform.map(screenPolygon).boundingRect();
         }
         case MapObject::Point: {
-            return transform.mapRect(renderer->shape(object).boundingRect());
+            const QPointF pos = renderer->pixelToScreenCoords(object->position());
+            return transform.mapRect(QRectF(pos, QSizeF(0, 0)));
         }
         case MapObject::Polygon:
         case MapObject::Polyline: {


bjorn:
@MollahHamza Thanks for this fix, it looks good and worked fine in my tests!

Btw, I would suggest that you leave the GitHub option to allow the maintainer to push to your branch enabled. I usually use that as an opportunity to push minor tweaks without going through review comments as well as to update the `NEWS.md` file as part of the merge. This time I'll just remember to update it while cherry-picking the fix to the `1.12` branch. :-)

MollahHamza:
@bjorn Thankss. 
Also I turned on edit for maintainers on all my PRs and will do it in other future PRs too.

PR #4403, opened 2026-03-06
Add plain text fallback when reading clipboard map data

sam-coolshrestha:
### Summary

Add a plain text fallback when reading clipboard map data.

### Problem

`ClipboardManager::update()` enables the Paste option when plain text exists in the clipboard. However, `ClipboardManager::map()` previously only attempted to read the custom text/tmx MIME type, which could lead to Paste being enabled while no data is actually read.

### Solution

If text/tmx data is unavailable, the code now falls back to reading plain text from the clipboard and attempts to parse it as TMX data.

### Result

This keeps clipboard behavior consistent and allows valid TMX XML copied as plain text (for example from external tools or browsers) to be parsed correctly.

### DIFF
--- a/src/tiled/clipboardmanager.cpp
+++ b/src/tiled/clipboardmanager.cpp
@@ -80,14 +80,23 @@ ClipboardManager *ClipboardManager::instance()
 std::unique_ptr<Map> ClipboardManager::map() const
 {
     const auto mimeData = mClipboard->mimeData();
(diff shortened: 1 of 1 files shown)


bjorn:
Closing this because it has the same problems I've outlined at https://github.com/mapeditor/tiled/pull/4289#issuecomment-3583092315 (which have also already been addressed on that PR).

PR #4404, opened 2026-03-06
Add warning log when clipboard TMX parsing fails

sam-coolshrestha:
Adds a warning log when parsing TMX data from the clipboard fails.

This helps with debugging situations where clipboard data cannot be converted into a Map instance.

### DIFF
--- a/src/tiled/clipboardmanager.cpp
+++ b/src/tiled/clipboardmanager.cpp
@@ -95,13 +95,6 @@ std::unique_ptr<Map> ClipboardManager::map() const
 
     return map;
 }
-
-    if (!map) {
-        qWarning() << "Failed to parse TMX data from clipboard";
-    }
-
-    return map;
-}
 /**
  * Sets the given map on the clipboard.
  */


src/tiled/clipboardmanager.cpp
```diff
@@ -85,9 +86,14 @@ std::unique_ptr<Map> ClipboardManager::map() const
         return nullptr;
 
     TmxMapFormat format;
-    return format.fromByteArray(data);
-}
+    auto map = format.fromByteArray(data);
 
+    if (!map) {
+        qWarning() << "Failed to parse TMX data from clipboard";
```
bjorn: While this may be visible to developers, the standard error output is not visible on macOS and Windows and even on Linux only to those who launched Tiled from a terminal. If you want to make this more useful, I think we should at least:

* Make sure this warning is reported in the UI (can be done with `Tiled::WARNING`, see existing usages.
* Include the actual error from `format.errorString()` in the message.

src/tiled/clipboardmanager.cpp
```diff
@@ -77,6 +77,7 @@ ClipboardManager *ClipboardManager::instance()
  * Retrieves the map from the clipboard. Returns null when there was no map or
  * loading failed.
  */
+
```
bjorn, line 80: Please revert the addition of this empty line. :-)

sam-coolshrestha: Hi! I've pushed the requested changes. The warning now uses Tiled::WARNING
and includes format.errorString(). Please let me know if anything else
needs improvement.

bjorn: There's still an empty line added here by the PR.

src/tiled/clipboardmanager.cpp
```diff
@@ -85,9 +86,15 @@ std::unique_ptr<Map> ClipboardManager::map() const
         return nullptr;
 
     TmxMapFormat format;
-    return format.fromByteArray(data);
-}
+    auto map = format.fromByteArray(data);
 
+    if (!map) {
+        Tiled::WARNING << "Failed to parse TMX data from clipboard:"
+                       << format.errorString();
```
bjorn, line 93: This does not compile, see the error in the builds done by GitHub:

```
ERROR: /home/runner/work/tiled/tiled/src/tiled/clipboardmanager.cpp: In member function ‘std::unique_ptr<Tiled::Map> Tiled::ClipboardManager::map() const’:
/home/runner/work/tiled/tiled/src/tiled/clipboardmanager.cpp:92:16: error: ‘WARNING’ is not a member of ‘Tiled’
   92 |         Tiled::WARNING << "Failed to parse TMX data from clipboard:"
      |                ^~~~~~~
```
In addition to needing an `#include "logginginterface.h` for `Tiled::WARNING`, this is not how `Tiled::WARNING` is supposed to be used. Please check its other usages.

Issue #3860, opened 2023-12-19
Zooming is wonky outside map bounds


eishiya:
When zooming in/out, Tiled restricts the map view position so that it's not too far from the map bounds. This means that if a map has Objects far outside of the map bounds, it becomes very annoying to zoom on them, because Tiled keeps resetting the view closer to the map bounds.

**Steps to Reproduce**
1. Create a Map. Leave the tile layer empty, or only populated in a small area.
2. Create an Object Layer and add Objects far away outside the map bounds (or the tiles, in the case of an infinite map).
3. Attempt to zoom in with the scroll wheel on the distant Objects.
4. Observe that the map view frames an area closer to the map bounds, potentially nowhere near where you were trying to zoom.

**Expectation**
Tiled should zoom in on this distant area just the same as on anything within the map bounds.

In this gif, the map is infinite, but you can see the calculated map bounds as the green dot in the upper left.
![tiled_zoomweirdness2](https://github.com/mapeditor/tiled/assets/70545646/35e66283-9534-4503-bb10-a23c712382a5)

**Proposed changes**
When zooming, Tiled should not try to force the view to be within some bounds at all. If the user is zooming on some location, it's because they want to see it. Only apply the limits when Tiled initiates the zoom, e.g. when opening a map.
Alternatively, Tiled should take Objects and Image Layers (and their offsets) into account when calculating the effective bounds for the view, rather than only using the tile layer bounds.

**Specifications:**
Windows 10, Tiled 1.10.2 (me) and Windows 8.1, Tiled 1.10.1 (mingw, who reported this on Discord). I am sure this is a much older issue.

Clemapfel:
Bump, I am also running into this issue on infinite maps. It is quite annoying when using tiled to edit objects far outside the tile bounds, which happens commonly when using tiled for polygon editing. 

As a workaround, you can add a tile to the top left and bottom right of all objects, then undo placing both tiles. This causes zoom to work correctly, but only until the map is reloaded, I assume because the placing of tiles allocates the chunks for the infinite maps, and zooming only works correctly if the cursor is inside the bounds of a chunk. 

Because behavior changes after the reload, even though the state of the map in terms of all tiles and objects is the same before and after since we used undo to remove the workaround-tiles, makes me consider this a bug.

kunal649:
@bjorn i was looking into this & i think ive found the culprit. 
[...]

Issue #4216, opened 2025-07-27
Properties editor becomes hard to read because of color contrast on Breeze Dark


MatusGuy:
### Describe the bug

Part of the new properties editor is illegible in the Breeze Dark colorscheme:

<img width="307" height="686" alt="Image" src="https://github.com/user-attachments/assets/33b30621-8a08-46bb-ac18-f83be75a40ac" />

It could be that the wrong color roles for the color scheme are being used.

### Steps to reproduce

Set the system colorscheme to Breeze Dark and the Tiled theme to Native

### Expected behavior

Be legible

### OS version

openSUSE Tumbleweed Linux

### Tiled version

2025.06.13

MatusGuy:
It appears this only happens with the Fusion Qt style, as it looks fine with Oxygen

kunal649:
turns out  this isn't just a linux/breeze dark issue ,  it reproduced on Windows 11 with : system dark + Native theme.
 headers showing up in reddish color (my sys accent is red by the way)
<img width="212" height="400" alt="Image" src="https://github.com/user-attachments/assets/8d0dba78-83b4-4dd3-a168-c607bac3c0c3" />

I traced it to PropertyLabel::setHeader  (line 864) in propertyeditorwidgets.cpp & here
`setBackgroundRole(header ? QPalette::Dark: QPalette::NoRole);`
`setForegroundRole(header ? QPalette::BrightText: QPalette::NoRole);`

seems wrong semantic roles for this use case.
the root cause is that BrightText has no strict cross-platform contract.  Qt only defines it as "contrasting against Dark" but leaves the actual color to each platform's theme. so it fails differently everywhere: invisible on Breeze Dark, reddish on Windows dark mode.
i landed on two options  : 
`Button` + `ButtonText` ( okay but semantically wrong i guess) 
`AlternateBase` + `Text`  Qt's dedicated role for "subtly distinct backgrounds" that's explicitly never accent-colored, and Text is documented as readable against it.

happy to open a PR if looks good
<img width="1233" height="852" alt="Image" src="https://github.com/user-attachments/assets/ec0ee258-e121-4251-b166-3329219da2c5" />

bjorn:
> seems wrong semantic roles for this use case.
> the root cause is that BrightText has no strict cross-platform contract. Qt only defines it as "contrasting against Dark" but leaves the actual color to each platform's theme. so it fails differently everywhere: invisible on Breeze Dark, reddish on Windows dark mode.

[...]

PR #4409, opened 2026-03-08
wonky-zooming-fixed

kunal649:
closes : issue #3860 
I mentioned in issue #3860 comment, things i consider before making the change. 
https://github.com/mapeditor/tiled/issues/3860#issuecomment-4018748683

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -19,6 +19,7 @@
 * Made the shortcut for current tool switch to previous tool (#4280)
 * Allow canceling Select Same Tile, Magic Wand and Bucket Fill operations with right-click and Escape
 * Allow dragging over multiple tiles with Select Same Tile, Magic Wand and Bucket Fill tools (#4276)
+* Allow zooming in on areas outside the map bounds (by kunal649, #3860)
 * Don't switch to Edit Polygons tool on double-click with Alt pressed
 * Adjust world map position when resizing a map with offset (#4270)
 * Added export plugin for Remixed Dungeon (by Mikhael Danilov, #4158)


bjorn:
@kunal649 Thanks for this! It's exactly what @eishiya asked for in #3860 and after testing it a little I guess it's the easiest fix. The fact that the user can now zoom in outside of the map boundaries does not seem like a serious issue, though I had previously prevented this intentionally to avoid disorientation.

It would probably still be good to also expand the scrollable area based on objects placed outside the map boundaries, but that's a very different change that can be done in another PR. Since it involves finding a bounding box that includes object placement it's also not so easy to do this efficiently.

kunal649:
didn't realize the restriction was intentional, but good to know testing confirmed it's fine either way !!
for the scrollable area expansion i can look into it, but if there's anything more pressing before the 1.12 release you'd want fixed,, happy to pick that up instead.

bjorn:
> didn't realize the restriction was intentional, but good to know testing confirmed it's fine either way !!

I think it makes sense for tile layers at least, since you can't anyway place anything outside of them (well, except for infinite maps). A similar thing happens when zooming in/out in image editors, where the image is usually always kept at least partially visible.

Well, we can always reconsider this when we do actually expand the scrollable area based on objects as well.

> if there's anything more pressing before the 1.12 release you'd want fixed,, happy to pick that up instead.

For 1.12 the most pressing thing is simply to release it, since it's been way too long since the last release. But I'm trying to still include fixes that seem fine. It would also be nice if we could avoid the regression in #4216. Maybe I'll still find some time to look into that, though.

kunal649:
> Well, we can always reconsider this when we do actually expand the scrollable area based on objects as well.

yeah, that's fair. 

> For 1.12 the most pressing thing is simply to release it, since it's been way too long since the last release. But I'm trying to still include fixes that seem fine. It would also be nice if we could avoid the regression in https://github.com/mapeditor/tiled/issues/4216. Maybe I'll still find some time to look into that, though.

this issue was under my radar last week too, but skipped it assuming it was Linux-only. i left my findings in a comment there ~  [#4216-comment](https://github.com/mapeditor/tiled/issues/4216#issuecomment-4044668797)

PR #4411, opened 2026-03-08
Fix/runtime translation

### LINKED ISSUE #3443: Localization entries are not applied completely
**Describe the bug**
Not all translations are applied immediately after switching the language.

**To Reproduce**
For the following steps it is assumed the ui language is set to 'English'.
1. Start Tiled
2. Open Preferences menu (Edit > Preferences)
3. Switch to 'Interface'
4. Select a language different than 'English' (best one with an (almost) complete translation)

Result: Most texts are changed, but not all. (See image below)

**Expected behavior**
All texts are changed.

**Media**
tiled_directlyAfterSettingLanguage
Left side: How it looks directly after changing the language
Right side: After restarting Tiled with the correct display

**Specifications:**
 - OS: Windows 11
 - Tiled Version: 1.9.0

SIDDHAARTHAA:
Fixes runtime language-switch gaps where some UI strings only updated after restart.

Added missing `LanguageChange` / `Preferences::languageChanged` retranslation hooks in affected widgets, models, and action handlers, and refreshed dynamic property labels in `PropertiesWidget`.

closes #3443

### DIFF
--- a/src/tiled/brokenlinks.cpp
+++ b/src/tiled/brokenlinks.cpp
@@ -106,8 +106,10 @@ void BrokenLinksModel::languageChanged()
 {
     emit headerDataChanged(Qt::Horizontal, 0, columnCount() - 1);
 
-    if (rowCount() > 0)
-        emit dataChanged(index(0, 2), index(rowCount() - 1, 2));
+    // link type column is affected by language change
+    const int rows = rowCount();
+    if (rows > 0)
+        emit dataChanged(index(0, 2), index(rows - 1, 2));
 }
 
 void BrokenLinksModel::setDocument(Document *document)
--- a/src/tiled/brokenlinks.cpp
+++ b/src/tiled/brokenlinks.cpp
@@ -27,6 +27,7 @@
 #include "mainwindow.h"
 #include "map.h"
 #include "objectgroup.h"
+#include "preferences.h"
 #include "replacetemplate.h"
 #include "replacetileset.h"
 #include "session.h"
@@ -41,6 +42,7 @@
 
 #include <QPushButton>
 #include <QDialogButtonBox>
+#include <QEvent>
 #include <QFileDialog>
 #include <QFileInfo>
 #include <QHeaderView>
@@ -99,6 +101,13 @@ BrokenLinksModel::BrokenLinksModel(QObject *parent)
     : QAbstractListModel(parent)
     , mDocument(nullptr)
 {
+    connect(Preferences::instance(), &Preferences::languageChanged,
+            this, [this] {
+                emit headerDataChanged(Qt::Horizontal, 0, 2);
+
+                if (rowCount() > 0)
+                    emit dataChanged(index(0, 2), index(rowCount() - 1, 2));
+            });
 }
 
 void BrokenLinksModel::setDocument(Document *document)
@@ -411,11 +420,9 @@ BrokenLinksWidget::BrokenLinksWidget(BrokenLinksModel *brokenLinksModel, QWidget
                                     Qt::Horizontal,
                                     this))
 {
-    mTitleLabel->setText(tr("Some files could not be found"));
-    mDescriptionLabel->setText(tr("One or more referenced files could not be found. You can help locate them below."));
     mDescriptionLabel->setWordWrap(true);
 
-    mLocateButton = mButtons->addButton(tr("Locate File..."), QDialogButtonBox::ActionRole);
+    mLocateButton = mButtons->addButton(QString(), QDialogButtonBox::ActionRole);
     mLocateButton->setEnabled(false);
 
     QFont font = mTitleLabel->font();
@@ -461,6 +468,27 @@ BrokenLinksWidget::BrokenLinksWidget(BrokenLinksModel *brokenLinksModel, QWidget
     // For some reason a model reset doesn't trigger the selectionChanged signal,
     // so we need to handle that explicitly.
     connect(brokenLinksModel, &BrokenLinksModel::modelReset, this, &BrokenLinksWidget::selectionChanged);
+
+    retranslateUi();
+}
+
+void BrokenLinksWidget::changeEvent(QEvent *event)
+{
+    QWidget::changeEvent(event);
+
+    if (event->type() == QEvent::LanguageChange)
+        retranslateUi();
+}
+
+void BrokenLinksWidget::retranslateUi()
+{
+    mTitleLabel->setText(tr("Some files could not be found"));
+    mDescriptionLabel->setText(tr("One or more referenced files could not be found. You can help locate them below."));
+
+    if (auto ignoreButton = mButtons->button(QDialogButtonBox::Ignore))
+        ignoreButton->setText(tr("Ignore"));
+
+    selectionChanged();
 }
 
 void BrokenLinksWidget::clicked(QAbstractButton *button)
--- a/src/tiled/brokenlinks.cpp
+++ b/src/tiled/brokenlinks.cpp
@@ -104,7 +104,7 @@ BrokenLinksModel::BrokenLinksModel(QObject *parent)
 
 void BrokenLinksModel::languageChanged()
 {
-    emit headerDataChanged(Qt::Horizontal, 0, 2);
+    emit headerDataChanged(Qt::Horizontal, 0, columnCount() - 1);
 
     if (rowCount() > 0)
         emit dataChanged(index(0, 2), index(rowCount() - 1, 2));
@@ -542,6 +542,8 @@ void BrokenLinksWidget::selectionChanged()
                 mLocateButton->setText(tr("Open Tileset..."));
             break;
         }
+    } else {
+        mLocateButton->setText(tr("Locate File..."));
     }
 }
 
--- a/src/tiled/brokenlinks.cpp
+++ b/src/tiled/brokenlinks.cpp
@@ -27,7 +27,6 @@
 #include "mainwindow.h"
 #include "map.h"
 #include "objectgroup.h"
-#include "preferences.h"
 #include "replacetemplate.h"
 #include "replacetileset.h"
 #include "session.h"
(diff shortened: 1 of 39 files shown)


src/tiled/brokenlinks.cpp
```diff
@@ -99,6 +101,13 @@ BrokenLinksModel::BrokenLinksModel(QObject *parent)
     : QAbstractListModel(parent)
     , mDocument(nullptr)
 {
+    connect(Preferences::instance(), &Preferences::languageChanged,
+            this, [this] {
+                emit headerDataChanged(Qt::Horizontal, 0, 2);
+
+                if (rowCount() > 0)
+                    emit dataChanged(index(0, 2), index(rowCount() - 1, 2));
+            });
```
bjorn: I'd like to avoid direct connections to `Preferences::languageChanged` when possible, because the `QEvent::LanguageChange` allows responding to this in a less dependent way. For `BrokenLinksModel`, we can add a `BrokenLinksModel::languageChanged` and call it from `BrokenLinksWidget::changeEvent`.

src/tiled/commanddatamodel.cpp
```diff
@@ -37,6 +37,15 @@ static const char *commandMimeType = "application/x-tiled-commandptr";
...
 {
+    connect(Preferences::instance(), &Preferences::languageChanged,
+            this, [this] {
+                emit headerDataChanged(Qt::Horizontal, 0, 2);
+
+                const int rows = rowCount();
+                const int columns = columnCount();
+                if (rows > 0 && columns > 0)
+                    emit dataChanged(index(0, 0), index(rows - 1, columns - 1));
+            });
```
bjorn: Here too, please introduce `CommandDataModel::languageChanged` and call it from the `CommandsEdit`, which as a widget will receive the `QEvent::LanguageChange`.

Also, while you're at it, could you remove the unused `CommandDataModel` instance from `CommandManager`? I seem to have forgotten it in a refactor.

src/tiled/commandmanager.cpp
```diff
@@ -78,6 +78,8 @@ CommandManager::CommandManager()
 
     connect(ProjectManager::instance(), &ProjectManager::projectChanged,
             this, &CommandManager::updateActions);
+    connect(preferences, &Preferences::languageChanged,
+            this, &CommandManager::retranslateUi);
```
bjorn: This function is already called from `MainWindow::retranslateUi`.

src/tiled/layermodel.cpp
```diff
@@ -44,6 +45,11 @@ LayerModel::LayerModel(QObject *parent):
 {
     mTileLayerIcon.addFile(QLatin1String(":images/32/layer-tile.png"));
     mObjectGroupIcon.addFile(QLatin1String(":images/32/layer-object.png"));
+
+    connect(Preferences::instance(), &Preferences::languageChanged,
+            this, [this] {
+                emit headerDataChanged(Qt::Horizontal, 0, 2);
+            });
```
bjorn: Please trigger this from `LayerView` based on `QEvent::LanguageChange`.

src/tiled/locatorwidget.cpp
```diff
@@ -323,6 +324,16 @@ void LocatorWidget::setVisible(bool visible)
     }
 }
 
+void LocatorWidget::changeEvent(QEvent *event)
+{
+    QFrame::changeEvent(event);
+
+    if (event->type() == QEvent::LanguageChange) {
+        mFilterEdit->setPlaceholderText(mLocatorSource->placeholderText());
+        setFilterText(mFilterEdit->text());
```
bjorn: I think we don't need this in the `LocatorWidget` because it is a popup and as such it's not possible to change the language while it is open. Or did you notice a string failing to update here?

src/tiled/mapdocumentactionhandler.cpp
```diff
@@ -235,6 +236,9 @@ MapDocumentActionHandler::MapDocumentActionHandler(QObject *parent)
 
     updateActions();
     retranslateUi();
+
+    connect(Preferences::instance(), &Preferences::languageChanged,
+            this, &MapDocumentActionHandler::retranslateUi);
```
bjorn: This is already called from `MainWindow::retranslateUi`.

src/tiled/mapobjectmodel.cpp
```diff
@@ -81,6 +82,11 @@ MapObjectModel::MapObjectModel(QObject *parent)
     , mObjectGroupIcon(QLatin1String(":/images/16/layer-object.png"))
 {
     mObjectGroupIcon.addFile(QLatin1String(":images/32/layer-object.png"));
+
+    connect(Preferences::instance(), &Preferences::languageChanged,
+            this, [this] {
+                emit headerDataChanged(Qt::Horizontal, 0, ColumnCount - 1);
+            });
```
bjorn: Please trigger this from `ObjectsView` (or `ObjectsDock`) based on `QEvent::LanguageChange`.

src/tiled/newversiondialog.cpp
```diff
@@ -67,6 +65,24 @@ NewVersionDialog::~NewVersionDialog()
     delete ui;
 }
 
+void NewVersionDialog::changeEvent(QEvent *event)
+{
+    QDialog::changeEvent(event);
+
+    if (event->type() == QEvent::LanguageChange) {
```
bjorn: We don't need this in the `NewVersionDialog` because it is a modal dialog, so it's not possible to change the language while it is open.

src/tiled/propertieswidget.cpp
```diff
@@ -3092,6 +3092,8 @@ bool PropertiesWidget::event(QEvent *event)
     }
     case QEvent::LanguageChange:
         retranslateUi();
+        if (mDocument)
+            currentObjectChanged(mDocument->currentObject());
```
bjorn: Since you've already adjusted a bunch of widgets used in the properties view to handle `QEvent::LanguageChange`, is this really necessary?

src/tiled/shortcutsettingspage.cpp
```diff
@@ -93,6 +94,11 @@ ActionsModel::ActionsModel(QObject *parent)
             this, &ActionsModel::actionChanged);
     connect(ActionManager::instance(), &ActionManager::actionsChanged,
             this, [this] { mDirty = mConflictsDirty = true; refresh(); });
+    connect(Preferences::instance(), &Preferences::languageChanged,
+            this, [this] {
+                emit headerDataChanged(Qt::Horizontal, 0, 2);
+                refresh();
+            });
```
bjorn: Please trigger this from `ShortcutSettingsPage` based on `QEvent::LanguageChange`.

src/tiled/stampactions.cpp
```diff
@@ -76,6 +77,9 @@ StampActions::StampActions(QObject *parent) : QObject(parent)
 
     setEnabled(false);
     languageChanged();
+
+    connect(Preferences::instance(), &Preferences::languageChanged,
+            this, &StampActions::languageChanged);
```
bjorn: This is already called from the various tools that use `StampActions`.

src/tiled/templatesdock.cpp
```diff
@@ -64,6 +65,8 @@ TemplatesDock::TemplatesDock(QWidget *parent)
     , mRedoAction(new QAction(this))
     , mMapScene(new MapScene(this))
     , mMapView(new MapView(this))
+    , mObjectTemplate(nullptr)
+    , mPropertiesDock(nullptr)
```
bjorn: Good catch that these were not getting initialized! Just please move the initialization to the header file.

src/tiled/tilestampmodel.cpp
```diff
@@ -21,12 +21,17 @@
...
 
 namespace Tiled {
 
 TileStampModel::TileStampModel(QObject *parent)
     : QAbstractItemModel(parent)
 {
+    connect(Preferences::instance(), &Preferences::languageChanged,
+            this, [this] {
+                emit headerDataChanged(Qt::Horizontal, 0, 1);
+            });
```
bjorn: Could we trigger this from `TileStampsDock::retranslateUi`?

bjorn (changes requested):
Thanks for catching so many cases where strings were not getting updated! However, there are also many places covered which don't need to be:

* Popups and dialogs don't need this, because you can't change the language while they are open.
* In some cases, the strings were already getting updated through other means. This makes me wonder how you searched for missing cases.

And I'd like to avoid connecting to `Preferences::languageChanged` where possible, and I think it is basically always possible because in the end there's always a widget involved in displaying a string, which will get the `QEvent::LanguageChange`.

I see there were three existing cases of connecting to `languageChanged`, but I might look into changing those sooner or later because I think it doesn't set the best example. :-)

SIDDHAARTHAA:
Thanks for the detailed feedback @bjorn , that makes sense.

You’re right that my initial sweep overreached. I started from a static search (`tr(...)`, constructor-set text/tooltips, model `headerData()/data()`), then added missing local hooks, but that did not sufficiently account for existing retranslation paths through parent widgets (like `MainWindow::retranslateUi`) and it also included popup/dialog cases that don’t need live updates.

I’ve now adjusted this to follow the pattern you described:
- removed the new direct `Preferences::languageChanged` connections,
- moved model refreshes behind widget-driven `QEvent::LanguageChange`,
- and dropped popup/dialog-only retranslation changes.

I’m currently testing the updated version and will commit/push shortly once verified.

src/tiled/brokenlinks.cpp
SIDDHAARTHAA: done

src/tiled/commanddatamodel.cpp
SIDDHAARTHAA: done

bjorn (approved):
Thanks @SIDDHAARTHAA! I've submitted some additional tweaks, and I think this is ready to be merged now. :-)

PR #4412, opened 2026-03-09
tmxrasterizer: add --search-path option and stdout output support

### LINKED ISSUE #3867: tmxrasterizer as a git diff tool: override search path, and allow explicitly writing to stdout
Two little features would make it much easier to plug `tmxrasterizer` into git as a map differ:

1. An argument for overriding the initial search path.  The problem is, when git diffs two versions of the same file, it has to put at least one of them in a temporary file, and now it doesn't know where to find any of its tilesets.  I know the original full path, but I have no way to provide it.

2. Writing the image to stdout.  This isn't strictly necessary since tempfiles are always an option, but it did surprise me a little that giving `-` as the outfile path literally created a file named `-`  :)

MollahHamza:
Fixes #3867                                                                                                             
                                                                                                                          
Adds two features to make tmxrasterizer more useful as a git diff tool:                                                 
                  
 --search-path <path>                                                                                                    
Overrides the directory used to resolve tileset and image references. Useful when rasterizing a temporary copy of a map
file (e.g. from git diff) where the file has been relocated away from its assets.                                       
                  
Stdout output                                                                                                           
Passing - as the output filename writes the PNG to stdout, enabling pipe-based workflows:
tmxrasterizer map.tmx - | compare - other.png diff.png                                                                  
                                                        
The search path support is implemented via a new readMap(fileName, searchPath, error) overload in mapformat.cpp, which  
passes the custom path to MapReader using its existing QIODevice-based API. The --search-path option also applies when  
rendering worlds.

### DIFF
--- a/src/tmxrasterizer/main.cpp
+++ b/src/tmxrasterizer/main.cpp
@@ -105,6 +105,9 @@ int main(int argc, char *argv[])
                           { QStringLiteral("frame-duration"),
                             QCoreApplication::translate("main", "Duration of each frame in milliseconds, defaults to 100."),
                             QCoreApplication::translate("main", "number") },
+                          { QStringLiteral("search-path"),
+                            QCoreApplication::translate("main", "Override the search path used to resolve tileset and image references. Useful when the map file is in a temporary location (e.g. during git diff)."),
+                            QCoreApplication::translate("main", "path") },
                       });
     parser.addPositionalArgument(QStringLiteral("map|world"), QCoreApplication::translate("main", "Map or world file to render."));
     parser.addPositionalArgument(QStringLiteral("image"), QCoreApplication::translate("main", "Image file to output."));
@@ -185,6 +188,9 @@ int main(int argc, char *argv[])
             exit(1);
         }
     }
+
+    if (parser.isSet(QLatin1String("search-path")))
+        w.setSearchPath(parser.value(QLatin1String("search-path")));
 
     return w.render(fileToOpen, fileToSave);
 }
--- a/src/tmxrasterizer/main.cpp
+++ b/src/tmxrasterizer/main.cpp
@@ -106,7 +106,7 @@ int main(int argc, char *argv[])
                             QCoreApplication::translate("main", "Duration of each frame in milliseconds, defaults to 100."),
                             QCoreApplication::translate("main", "number") },
                           { QStringLiteral("search-path"),
-                            QCoreApplication::translate("main", "Override the search path used to resolve tileset and image references. Useful when the map file is in a temporary location (e.g. during git diff)."),
+                            QCoreApplication::translate("main", "Override the search path used to resolve tileset and image references."),
                             QCoreApplication::translate("main", "path") },
                       });
     parser.addPositionalArgument(QStringLiteral("map|world"), QCoreApplication::translate("main", "Map or world file to render."));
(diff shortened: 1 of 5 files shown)


MollahHamza:
Hey @bjorn,
Kody flagged that the new readMap overload hardcodes MapReader, bypassing the plugin system for non-TMX formats. Should I add a fallback to findSupportingMapFormat first, or is this acceptable given the TMX-focused use case?

src/tmxrasterizer/main.cpp
```diff
@@ -105,6 +105,9 @@ int main(int argc, char *argv[])
                           { QStringLiteral("frame-duration"),
                             QCoreApplication::translate("main", "Duration of each frame in milliseconds, defaults to 100."),
                             QCoreApplication::translate("main", "number") },
+                          { QStringLiteral("search-path"),
+                            QCoreApplication::translate("main", "Override the search path used to resolve tileset and image references. Useful when the map file is in a temporary location (e.g. during git diff)."),
```
bjorn: This is a little too verbose to include in `--help`. I think we can just state what it does here:
```suggestion
                            QCoreApplication::translate("main", "Override the search path used to resolve tileset and image references."),
```

src/tmxrasterizer/tmxrasterizer.cpp
```diff
@@ -174,7 +176,8 @@ int TmxRasterizer::render(const QString &fileName,
     // If we're not rendering a world, load the map once and create a renderer
     if (!fileName.endsWith(QLatin1String(".world"), Qt::CaseInsensitive)) {
         QString errorString;
-        map = readMap(fileName, &errorString);
+        map = mSearchPath.isEmpty() ? readMap(fileName, &errorString)
+                                    : readMap(fileName, mSearchPath, &errorString);
```
bjorn: I see this pattern in three places now. I think that's why it would be preferable if we could just do this:
```suggestion
        map = readMap(fileName, mSearchPath, &errorString);
```
And the `readMap` function internally would choose `QFileInfo(fileName).absolutePath()` in case the passed in `mSearchPath` is empty.

bjorn (changes requested):
> Kody flagged that the new readMap overload hardcodes MapReader, bypassing the plugin system for non-TMX formats. Should I add a fallback to findSupportingMapFormat first, or is this acceptable given the TMX-focused use case?

Yeah, in my original comment at https://github.com/mapeditor/tiled/issues/3867#issuecomment-1887032701 I had already noted that an extension of the interface would be necessary, because indeed this way the change is hardcoded for TMX files, even though tmxrasterizer also supports other formats and it would be nice to support this feature especially also for the JSON format.

I also wrote "For other formats, I guess providing a default implementation that ignores the custom search path would be fine for now, so we don't need to change all the implementations." There I meant that we would add the following virtual function to `MapFormat`:

```c++
    virtual std::unique_ptr<Map> read(const QString &fileName, const QString &path)
    { Q_UNUSED(path) read(fileName); }
```

Since it has a default implementation that just ignores the search path, we wouldn't need to modify all plugins that implement this interface for now, but could choose to do this only for TMX and JSON.

MollahHamza:
Hey @bjorn, I addressed the review changes                                                                                       
                                                                                                                          
  - Added MapFormat::read(fileName, searchPath) virtual with a default implementation that ignores the search path        
  - readMap with search path now tries plugins first, falls back to TMX reader                                            
  - Shortened --search-path help text                                                                                     
  - Removed the repeated mSearchPath.isEmpty() ternary empty search path now defaults to the file's directory internally

PR #4413, opened 2026-03-09
Add preference for default object color 

### LINKED ISSUE #3861: Allow changing default Object colour
### Problem

The default grey Object colour can be hard to see when placed against some tiles, particularly ones with a lot of grey. This is especially an issue when editing and checking Tile Collisions.

Here, for example, it can be hard to tell whether the tile has had collision added or not, especially when you want to check quickly:
image

### Proposal
It would be nice if the default Object colour could be customised, similarly to how the grid colour can be changed.

### Existing alternatives
Currently, one can increase the Object line thickness, which can help sometimes, but can make the object cover up important artwork otherwise.
Another option is to use a custom class that sets a different colour. However, this gets needlessly tedious when you don't otherwise need to use a class, such as for most collisions.

(The visibility issue was reported by Lumin and the screenshot is also from them, I'm just posting this for them.)

MollahHamza:
Fixes #3861                                                                                                             
                                                                                                                          
Adds a "Default object color" setting to Preferences → Interface, placed alongside the other color settings (Grid color, Background fade color). The color defaults to gray, matching the previous hardcoded behavior, and updates all map objects live when changed.                                                                                              
                  
The default color is stored as a static in libtiled and set from Preferences on startup and on change, keeping libtiled independent of the preferences system.

### DIFF
--- a/src/libtiled/mapobject.cpp
+++ b/src/libtiled/mapobject.cpp
@@ -41,6 +41,18 @@
 
 namespace Tiled {
 
+static QColor sDefaultObjectColor = Qt::gray;
+
+QColor MapObject::defaultColor()
+{
+    return sDefaultObjectColor;
+}
+
+void MapObject::setDefaultColor(const QColor &color)
+{
+    sDefaultObjectColor = color;
+}
+
 TextData::TextData()
     : font(QStringLiteral("sans-serif"))
 {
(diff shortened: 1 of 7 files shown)


MollahHamza:
Addressed Kody's feedback added a validity check in defaultObjectColor() to fall back to the default gray if the  stored value is malformed.

PR #4415, opened 2026-03-10
Fix empty nested class properties not being cleaned up (#3998)

### LINKED ISSUE #3998: Cleaning up empty nested classes
Nested class properties remain even if they do not have a single subproperty. They are visible in the interface and then saved in the json map. They appear if you clear all nested properties manually but forget to delete the nested class itself. They also remain after `setProperty(path, undefined)` is executed.

It would be convenient if they were cleaned automatically.

Oval17:
1 -> When all members of a top-level class property are reset, the empty class container is now automatically removed but only when that property is defined in the object's assigned class.

2 -> Previously, the cleanup in setClassPropertyValue was intentionally skipped at depth 1 (allowReset = false) to avoid removing manually-added class properties. This fix threads an `allowTopLevelReset` flag through from `Object::setProperty` , which sets it to true only when the top-level key exists in `classType()->members`. Manually added properties (not defined in the object's class) are still preserved.

Fixes #3998

Ready for Review and Inputs @bjorn

### DIFF
--- a/src/libtiled/object.cpp
+++ b/src/libtiled/object.cpp
@@ -124,12 +124,20 @@ QVariantMap Object::inheritedProperties() const
 bool Object::setProperty(const PropertyPath &path, const QVariant &value)
 {
     // Allow removing an empty top-level class property only when it is defined
-    // in the object's class, so that manually added properties are preserved.
+    // in the object's class and the class's own value for it is empty. If the
+    // class has a non-empty default, removing the override would revert to that
+    // non-empty inherited value, which is not the intent.
     bool allowTopLevelReset = false;
     if (path.size() > 1) {
         const auto &topKey = std::get<QString>(path.first());
-        if (const ClassPropertyType *type = classType())
-            allowTopLevelReset = type->members.contains(topKey);
+        if (const ClassPropertyType *type = classType()) {
+            if (type->members.contains(topKey)) {
+                const auto classValue = type->members.value(topKey);
+                allowTopLevelReset = !classValue.isValid() ||
+                    (classValue.userType() == propertyValueId() &&
+                     classValue.value<PropertyValue>().value.toMap().isEmpty());
+            }
+        }
     }
     return setPropertyMemberValue(mProperties, path, value, allowTopLevelReset);
 }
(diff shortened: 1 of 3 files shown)


bjorn:
Hmm, it seems this patch really only changes the behavior of `Object::setProperty`, but that function is only used in `EditableObject::setPropertyImpl` (scripting API), and then only when an object does not have an undo stack associated with it. It makes me wonder how you tested this?

For the new behavior to work from the UI, we probably need to change `Document::setPropertyMember`, which is what the `SetProperty` undo command uses. We probably also need to change `VariantMapProperty::setMemberValue` accordingly because it applies the change to the local `mValue` (a copy of the properties of the object) before emitting its signals (based on which `CustomProperties::setPropertyValue` creates the undo command).

I realize the whole setup is a little involved.

There is another problematic aspect. We don't want to automatically reset a top-level class value when the class value inherited from the object's class is not an empty value. Because when we reset the top-level value when it's empty, it would revert back to inheriting the non-empty value. So at least in that case, the user needs to explicitly reset the top-level value if that it what they want.

Oval17:
Hi @bjorn 
I trie to addressed going one by one 
1.` Object::setPropert` : our PR previous check type->members.contains(topKey) was too broad. We now also verify that the class's stored value for that key is empty (either invalid or PropertyValue wrapping {}). If the class has non-empty defaults like lets say { hp: 50 }, removing the override would wrongly revert to inheriting those defaults.

2. `Document::setPropertyMembe` This is the actual UI/undo path (SetProperty undo command calls this). Added the same allowTopLevelReset logic, passes it to setNestedPropertyValue, and calls removeProperty instead of setProperty when the top-level value becomes invalid after cleanup.

3. `VariantMapProperty::setMemberValue` umm This updates the local mValue copy before emitting memberValueChanged (which creates the undo command). Now passes allowTopLevelReset to setPropertyMemberValue based on mSuggestions (the inherited values). Also fixed updateModifiedRecursively to fall back to mSuggestions.value(topLevelName) when the key was removed from mValue, so sub-properties correctly show inherited values.

Is this correct direction?

Oval17:
Have added comments as well in PR...

PR #4416, opened 2026-03-10
Fix scripts checkbox and icon overlapping issue

### LINKED ISSUE #3464: Scripting: Actions can't have checkboxes
If an Action is `checkable`, its icon is displayed instead of the checkbox, with a button-like look to indicate whether the Action is `checked`:
Default icon:
image
Custom icon:
image

When an action with an icon is unchecked, it's hard to see that it even *can* be checked. The default checkbox does a much better job of communicating that an action is checkable.

This only affects scripted actions, as none of the native checkable actions have icons.
Since script authors can simply choose to not assign an icon, it's probably fine that those with custom icons are displayed this way, allowing authors to essentially implement custom toggle styles if they wish. However, there's no way to *not* have an icon since iconless actions are given the extension icon by default! As such, there's no way to display a regular checkbox.

Possible solutions:
- Do not show the default extension icon on Actions that are `checkable`
- Make the default extension icon small and have it overlap the corner of the checkbox (and possibly any assigned icon as well).

saurabh-zz007:
Fixes #3464

**The Issue:**
If an Action is checkable, its icon is displayed instead of the checkbox, with a button-like look to indicate whether the Action is checked.

**The Fix:**
Updated the UI engine to support side-by-side dual rendering of both the checkbox and the icon:
- Removed the `if (menuItem->icon.isNull())` restriction in `drawControl` (`CE_MenuItem`).
- Updated `sizeFromContents` (`CT_MenuItem`) to widen the menu item bounds when both elements are present to prevent text clipping.
- Decoupled the icon rendering rect (`vCheckRect`) from the checkbox rect, applying a `dualOffset` to shift the icons to the right.

**Testing:**
Tested locally on Windows 11 under the Tiled Fusion theme.

**Results**
Screenshot 2026-03-20 212157

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,4 +59,3 @@ gmon.out
 # Vim files
 *.swp
 *.swo
-.vs/
--- a/src/tiled/tiledproxystyle.cpp
+++ b/src/tiled/tiledproxystyle.cpp
@@ -706,25 +706,25 @@ void TiledProxyStyle::drawControl(ControlElement element,
             if (
                 qobject_cast<const QComboBox*>(widget) ||
                 (option->styleObject && option->styleObject->property("_q_isComboBoxPopupItem").toBool()))
-                ignoreCheckMark = true;
+                ignoreCheckMark = true; //ignore the checkmarks provided by the QComboMenuDelegate
 
-            if (!ignoreCheckMark) {
+            if (!ignoreCheckMark) { 
                 checkcol = qMax<int>(defaultCheckWidth, menuItem->maxIconWidth) + dualOffset;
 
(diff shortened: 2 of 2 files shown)


bjorn:
Hmm, now the menu gets extra space for icons/checkboxes even when not necessary. Here's a menu with checkboxes but with no icons:

image

Or a menu with icons and no checkboxes:

image

Could the extra space only appear when both are present? Is seems this is what you're intending to do in the `sizeFromContents` change, but it is not working. I think you'll need to do something with `QStyleOptionMenuItem::menuHasCheckableItems`.

PR #4417, opened 2026-03-10
fix(mainwindow): handle per-monitor DPI changes in changeEvent (#4357)

### LINKED ISSUE #4357: Only some parts of Tiled properly respond to changing windows with a different Display Scale setting
### Describe the bug

This might be related to #2835 (or maybe is the same thing? It seems a bit different from an outside view, but unsure if that's actually the case).

When moving Tiled between a primary monitor with 150% display scale and a secondary monitor with 100% display scale, the fonts and padding seem to resize appropriately, but the UI size/padding/etc. doesn't reflect the changes.

Example image of:
* Maximized notepad window on the primary monitor in the background, as a baseline size reference
* A smaller notepad window also on the primary monitor, for size/behavior reference of another program's behavior
* Tiled on the primary monitor

Image

Followup example:
* Maximized notepad window still on primary monitor
* Smaller notepad window moved to secondary monitor
* Tiled moved to secondary monitor

Image

You'll notice that Notepad feels the same size (relatively speaking) in isolation, but the window size is actually much smaller compared to when it was on the primary monitor.

Meanwhile with Tiled, the fonts are smaller, but almost everything else is basically the same size. This makes the window itself feel much larger despite technically being the same size, and also makes it seem like most UI elements have an unusually large amount of inner padding.

---

I'm not sure if this is actually Tiled or Qt. I tested with the Windows 64bit Qt6 build, but an interesting thing to note is that the Windows 32bit Qt5 build has a similar problem with slight differences: the outer window size *was* correctly adjusted, and some other things seem closer to what I would expecte too. However, some parts appear to be worse since they didn't account for the window size changing, like the sidebars.

Primary monitor example (conditions matching the first example from before):
Image

Secondary monitor example (conditions matching the second example from before):
Image

### Steps to reproduce

Move Tiled from one monitor to the other.

At least on my computer, you can see the adjustments take place when it gets a little over halfway across the monitor boundary.

### Expected behavior

Every aspect of Tiled to feel appropriately rescaled on the new monitor, not just some parts.

### OS version

Windows 11, 64-bit

### Tiled version

CI build from 2/20/2026 (second from last commit in #4308, commit e794b0a, from action run 22233411625)

Aayushmaan-24:
Closes #4357

When moving the Tiled window between monitors with different display scaling
(e.g. 150% → 100%), fonts rescale correctly but many UI elements (padding,
sidebars, toolbars, status bar, docks) do not relayout properly, leading to
inconsistent appearance and oversized inner spacing.

Changes:
- Expanded `MainWindow::changeEvent` to detect `QEvent::ScreenChangeInternal`,
  `QEvent::StyleChange`, and `QEvent::ApplicationStateChange`.
- Forced relayout with `adjustSize()` and `updateGeometry()` on central widget,
  toolbars, status bar, and all dock widgets.
- Added `qDebug()` logging for verification.

Tested using `QT_SCALE_FACTOR=1.0` and `1.5` on Linux (simulates high/low DPI).
Limited direct multi-monitor reproduction on single-screen setup, but follows
Qt best practices for dynamic DPI handling.

Looking forward to feedback!

### DIFF
--- a/src/tiled/mainwindow.cpp
+++ b/src/tiled/mainwindow.cpp
@@ -95,6 +95,8 @@
 #include <QUndoStack>
 #include <QUndoView>
 #include <QVariantAnimation>
+#include <QDebug>
+#include <QScreen>
 
 #ifdef Q_OS_WIN
 #if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
@@ -958,14 +960,65 @@ void MainWindow::closeEvent(QCloseEvent *event)
 void MainWindow::changeEvent(QEvent *event)
 {
     QMainWindow::changeEvent(event);
+
     switch (event->type()) {
     case QEvent::LanguageChange:
         mUi->retranslateUi(this);
         retranslateUi();
         break;
+
     case QEvent::WindowStateChange:
         mUi->actionFullScreen->setChecked(isFullScreen());
         break;
+
+    // ── NEW: Handle DPI/monitor scaling changes ──
+    case QEvent::ScreenChangeInternal:   // Primary event for moving between monitors
+    case QEvent::StyleChange:            // Often sent after DPI change (style adapts)
+    case QEvent::ApplicationStateChange: // Sometimes triggered in scaling transitions
+        {
+            qDebug() << "DPI/monitor change detected - forcing UI relayout"
+                     << "devicePixelRatio:" << devicePixelRatioF()
+                     << "logicalDpiX:" << logicalDpiX()
+                     << "logicalDpiY:" << logicalDpiY();
+
+            // Force relayout of central area (map view, etc.)
+            if (centralWidget()) {
+                centralWidget()->adjustSize();
+                centralWidget()->updateGeometry();
+            }
+
+            // Toolbars (icons/text sizes may need refresh)
+            for (QToolBar *toolbar : allToolBars()) {
+                toolbar->adjustSize();
+                toolbar->updateGeometry();
+            }
+
+            // Status bar
+            if (statusBar()) {
+                statusBar()->adjustSize();
+            }
+
+            // Dock widgets / side panels (layers, properties, project, etc.)
+            for (QDockWidget *dock : findChildren<QDockWidget*>()) {
+                if (dock->widget()) {
+                    dock->widget()->adjustSize();
+                    dock->widget()->updateGeometry();
+                }
+            }
+
+            // Optional: full window refresh (helps if overall size feels wrong)
+            updateGeometry();
+
+            // If custom views (map/tileset) need extra nudge:
+            if (mMapEditor) {
+                if (MapView *view = mMapEditor->currentMapView()) {
+                    view->updateGeometry();
+                    view->viewport()->update();
+                }
+            }
+        }
+        break;
+
     default:
         break;
     }
--- a/src/tiled/utils.cpp
+++ b/src/tiled/utils.cpp
@@ -377,26 +377,25 @@ void saveGeometry(QWidget *widget)
 
 int defaultDpi()
 {
-    static int dpi = []{
-        if (const QScreen *screen = QGuiApplication::primaryScreen())
-            return static_cast<int>(screen->logicalDotsPerInchX());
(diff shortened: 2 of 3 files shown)


bjorn:
I think the main problem in adjusting to DPI changes is rather the static `scale` variable in this function in `utils.cpp`:

```c++
qreal defaultDpiScale()
{
    static qreal scale = []{
        if (const QScreen *screen = QGuiApplication::primaryScreen())
            return screen->logicalDotsPerInchX() / 96.0;
        return 1.0;
    }();
    return scale;
}
```

Due to this, many sizes will only adjust to the initial DPI of the primary screen, rather than being able to adjust dynamically when Tiled is moved from one screen to another.

I didn't test these changes yet, but it does not look like they address this issue. Your changes explicitly trigger a bunch of geometry and size updates, but that should not be necessary. I'm not entirely sure how dynamic adjustment to different DPIs should be handled, though. Most applications deal with this by applying a scaling factor on top of their rendering, but this often looks ugly.

> Tested using `QT_SCALE_FACTOR=1.0` and `1.5` on Linux (simulates high/low DPI).

You will probably need a multi-screen setup with differently configured screens to test this. Just testing with `QT_SCALE_FACTOR` does not suffice to see dynamic adjustment to different DPIs.

Aayushmaan-24:
Thanks for the detailed feedback, that makes sense.

You were right that the root cause is defaultDpiScale() caching the DPI of the primary screen once and then never updating. My earlier changes in MainWindow::changeEvent only forced geometry/size updates on various widgets and didn’t actually fix the underlying scaling logic.

I’ve now pushed a follow‑up that:

Removes the static caching and makes DPI queries dynamic:
defaultDpi() and defaultDpiScale() now compute their values from QGuiApplication::primaryScreen() on each call instead of using a static cached value.
Introduces a per‑screen API:
defaultDpiScale now has the signature qreal defaultDpiScale(const QScreen *screen = nullptr);
Callers that only know about the global state can still use defaultDpiScale() with no arguments (which falls back to the current primary screen).
Callers that know which QScreen they are on can pass that in to get a truly per‑monitor scale factor.
Makes dpiScaled(...) use the dynamic scale:
dpiScaled(qreal) now multiplies by defaultDpiScale() (without caching), so all existing uses that go through dpiScaled pick up the updated DPI value automatically.
This doesn’t yet wire per‑monitor QScreen* everywhere, but it removes the global static that was preventing any dynamic DPI adjustment and provides the API needed to make specific code paths fully per‑monitor aware.

On my side I can only test on a single‑monitor Linux/Chromebook setup (I used QT_SCALE_FACTOR=1.0 and 1.5 to sanity‑check that icons/handles etc. still scale correctly and that there are no regressions). I don’t currently have access to a Windows multi‑monitor setup with mixed DPI, so I can’t reproduce the original issue exactly.

If you (or someone with such a setup) can test a build from this branch and let me know how Tiled behaves when moving between a 150% primary monitor and a 100% secondary monitor, I’d be happy to iterate further and, if needed, start passing the appropriate QScreen* into the places that use defaultDpiScale for per‑monitor correctness.

bjorn:
Since you can't test this and your contribution still clearly can't work just based on code review, I'm closing this. A few concrete issues:

* Now `dpiScaled` no longer has a static variable, but it was using a static variable for performance reasons. We don't want to look up the primary screen and its scale very often. A solution may involve caching the result but invalidating it based appropriate signals or events called from Qt (like `QEvent::ScreenChangeInternal`, but as the name indicates, this one in particular is an internal event).

* Even with `dpiScaled` looking up the current value each time, it's not going to affect all the previous calls. Tiled does many calls to this function during startup to set spacing, margin, icon sizes, etc. in various places and none of those are going to pick up the new value automatically.

I think for supporting multi-monitor setups properly, the solution is likely going to be to adjust the general UI scale to the current monitor rather than adjusting all spacings, margins, etc. One problem with a UI scale is that the current map view scale would need to compensate for it. Another is that the current widget style includes a lot of 1-pixel lines and borders which doesn't work nicely for scaled UI (something to keep in mind for #3597).

PR #4420, opened 2026-03-12
Add (automatic) support for TileMapLayer export #4106

### LINKED ISSUE #4106: Use TileMapLayer for Godot exports as opposed to deprecated TileMapLayer
**Describe the bug**
Tiled exports to Godot using the `TileMap` node which has been deprecated (https://docs.godotengine.org/en/stable/tutorials/migrating/upgrading_to_godot_4.3.html#id4). 

I think the change on the plugin is actually quite small, it currently produces this:

```
[node name="01" type="Node2D"]

[node name="TileMap" type="TileMap" parent="."]
tile_set = ExtResource("TileSet_0")
format = 2
layer_0/name = "Ground"
layer_0/tile_data =  ...
```

We want:
```
[node name="01" type="Node2D"]

[node name="Ground" type="TileMapLayer" parent="."]
tile_set = ExtResource("TileSet_0")
use_parent_material = true
tile_map_data = PackedByteArray("
```

**To Reproduce**
1. Export a map to Godot
2. Open map in Godot
3. See deprecation warning for the use of a `TileMap` node

**Expected behavior**
The export should not produce a deprecation warning.

**Specifications:**
 - OS: macOS
 - Tiled Version: 1.11.0

YusefZahran:
closes #4106 

This PR introduces exporting to Godot's post 4.3 TileMapLayer node, manually checking the godot project's version and deciding to export either using the old TileMap node format or the new TileMapLayer node format, if using 4.3 or later.

This also now separates each tiled map layer into separate TileMapLayers in Godot. It also limits each layer to using only one tileset (as that's how TileMapLayer works in godot).

Changes were only made to the to the tscnplugin.cpp file. Any notes would be appreciated!

### DIFF
--- a/src/plugins/tscn/tscnplugin.cpp
+++ b/src/plugins/tscn/tscnplugin.cpp
@@ -751,7 +751,6 @@ static void writeTileMap(const Map *map, QFileDevice *device,const QString tiles
 // Write TileMapLayer. Writes each layer as a separate TileMapLayer
 static void writeTileMapLayer(const Map *map, QFileDevice *device, AssetInfo &assetInfo)
 {
-
     int layerIndex = 0;
     for (const auto layer : std::as_const(assetInfo.layers)) {
 
@@ -760,8 +759,8 @@ static void writeTileMapLayer(const Map *map, QFileDevice *device, AssetInfo &as
                                            "Separate the tilesets into seperate layers."));
         }
 
-        device->write(formatByteString("[node name=\"%2\" type=\"TileMapLayer\""
-                                       " parent=\".\"]\n",
+        device->write(formatByteString("[node name=\"%2\" type=\"TileMapLayer\" "
+                                       "parent=\".\"]\n",
                                        sanitizeQuotedString(layer->name())));
 
         if (layer->resolvedProperty("zIndex").isValid()) {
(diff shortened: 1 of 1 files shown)


src/plugins/tscn/tscnplugin.cpp
```diff
@@ -874,6 +1090,12 @@ bool TscnPlugin::write(const Map *map, const QString &fileName, Options options)
         AssetInfo assetInfo = collectAssets(map);
         auto tilesetResPath = map->propertyAsString("tilesetResPath");
 
+        QString projectFilePath = assetInfo.resRoot + "/project.godot";
```
Taffer, line 1093: I'm no expert (just nosing around!), but does this require the `project.godot` to be in the same directory as the file you're exporting? For example, would exporting a map to `maps/` still work?

YusefZahran: Yes it'll still work fine,  you can export your map anywhere inside your godot project directory and its sub-directories. This line just fetches the `project.godot` file to figure out if it should export `TileMap` nodes or `TileMapLayer` nodes depending on the project's engine version.

Taffer: Awesome, thanks for verifying!

PR #4421, opened 2026-03-12
Add PrimitivePropertyType with visual color indicators

### LINKED ISSUE #4169: Feature: Allow custom properties to be assigned a unique color to show inside of tileview/mapview/tileset preview windows
I love how flexible Tiled is with custom properties. It allows me to keep all my map related configs in one editor. My biggest gripe with tile is it's extremely hard to visually see what tiles have custom properties and which dont.

It'd be amazing if we could assign a color value to each property type (optional maybe?) and use that color to outline the tile on the mapview/tileview/tileset preview window. Also, show the color when the tile is selected next to the custom property itself.

Can't think of any other than using a different map editor. For example, I have an "interactable" property which allows me to configure if the tile is interactable in game or not. Currently, it's impossible to quickly see all the interactable tiles in your tileset. 

It doesn't need to be a color either, just some sort of visual representation to see. I picked color since it would be easy to have several colors showing several properties by using a dashed outline with different colors.

kanishka0411:
Fixes #4169 
## Summary

Implements the initial version of property visualization as discussed . Introduces a new ```PrimitivePropertyType```that can be assigned a visual color, which is then displayed as a colored circle overlay on tiles in the tileset view.

## What this PR does

- **New property type**: ```PrimitivePropertyType```with configurable storage type (bool, int, float, string, color) and an optional visual color
- **Editor UI**: "Add Primitive" button in the Custom Types Editor with a color picker and storage type selector
- **Tileset visualization**: Colored circle indicators drawn in the top-left corner of tiles that have properties of a primitive type with a visual color set
- **Scripting support**: Primitive property type exposed to the JavaScript API
- **Auto-refresh**: Tileset view updates when property type colors change
- **Tests**: JSON round-trip, default values, and export/import tests

## How to test

1. Open a project → View → Custom Types Editor
2. Click "Add Primitive", name it (e.g., `Interactable`), pick a color
3. Assign the primitive-typed property to a tile
4. A colored circle appears on the tile in the tileset view

### DIFF
--- a/src/libtiled/properties.cpp
+++ b/src/libtiled/properties.cpp
@@ -483,17 +483,23 @@ void initializeMetatypes()
     QMetaType::registerConverter<QString, FilePath>(&FilePath::fromString);
 }
 
+QVariantList possiblePrimitiveValues()
+{
+    return {
+        QVariant(false),                        // bool
+        QVariant(QColor()),                     // color
+        QVariant(0.0),                          // float
+        QVariant::fromValue(FilePath()),        // file
+        QVariant(0),                            // int
+        QVariant::fromValue(ObjectRef()),       // object
+        QVariant(QString()),                    // string
+    };
+}
+
 QVariantList possiblePropertyValues(const ClassPropertyType *parentClassType)
 {
-    QVariantList values;
+    QVariantList values = possiblePrimitiveValues();
 
-    values.append(false);                               // bool
-    values.append(QColor());                            // color
-    values.append(0.0);                                 // float
-    values.append(QVariant::fromValue(FilePath()));     // file
-    values.append(0);                                   // int
-    values.append(QVariant::fromValue(ObjectRef()));    // object
-    values.append(QString());                           // string
     values.append(QVariant(QVariantList()));            // list
 
     for (const auto &propertyType : Object::propertyTypes()) {
@@ -510,17 +516,6 @@ QVariantList possiblePropertyValues(const ClassPropertyType *parentClassType)
     }
 
     return values;
-}
-
-QVariantList possiblePrimitiveValues()
-{
-    return {
-        QVariant(false),                        // bool
-        QVariant(0),                            // int
-        QVariant(0.0),                          // float
-        QVariant(QString()),                    // string
-        QVariant::fromValue(QColor()),          // color
-    };
 }
 
 } // namespace Tiled
--- a/src/tiled/propertytypeseditor.cpp
+++ b/src/tiled/propertytypeseditor.cpp
@@ -214,16 +214,8 @@ PropertyTypesEditor::PropertyTypesEditor(QWidget *parent)
             this, [this] { addPropertyType(PropertyType::PT_Class); });
     auto primitiveMenu = new QMenu(this);
     for (const auto &value : possiblePrimitiveValues()) {
-        QString displayName;
-        switch (value.userType()) {
-        case QMetaType::Bool:    displayName = tr("Bool"); break;
-        case QMetaType::Int:     displayName = tr("Int"); break;
-        case QMetaType::Double:  displayName = tr("Float"); break;
-        case QMetaType::QString: displayName = tr("String"); break;
-        case QMetaType::QColor:  displayName = tr("Color"); break;
-        default: continue;
-        }
-        primitiveMenu->addAction(displayName, this, [this, value] {
+        const QIcon icon = PropertyTypesModel::iconForProperty(value);
+        primitiveMenu->addAction(icon, userTypeName(value), this, [this, value] {
             addPropertyType(PropertyType::PT_Primitive, value);
         });
     }
--- a/src/tiled/propertytypesmodel.cpp
+++ b/src/tiled/propertytypesmodel.cpp
@@ -166,7 +166,7 @@ QModelIndex PropertyTypesModel::addNewPropertyType(PropertyType::Type type, cons
         propertyType = std::make_unique<EnumPropertyType>(name);
         break;
     case PropertyType::PT_Primitive:
-        propertyType = std::make_unique<PrimitivePropertyType>(name, defaultValue.isValid() ? defaultValue : QVariant(false));
+        propertyType = std::make_unique<PrimitivePropertyType>(name, defaultValue);
         break;
     }
 


saurabh-zz007:
Hey @kanishka0411 , I was looking through the Linux build failure in your pr. It looks like it might be a header issue, I noticed QAction (and possibly std::make_unique) are being used, but their headers (`#include <QAction>` and `#include <memory>`) might be missing. Also, did you make sure to add the new source files to the CMakeLists.txt? Sometimes Linux is way more strict than windows in these things.

Hope this helps!

src/libtiled/propertytype.h
```diff
@@ -176,6 +178,39 @@ class TILEDSHARED_EXPORT ClassPropertyType final : public PropertyType
...
+public:
+    enum StorageType {
+        BoolValue,
+        IntValue,
+        FloatValue,
+        StringValue,
+        ColorValue
+    };
+
+    StorageType storageType = BoolValue;
```
bjorn: I think rather than storing this, we could just store `QVariant defaultValue`. This then covers the type and additionally would give us a place to store a different default value (which we can create an edit widget for in the UI).

src/libtiled/propertytype.cpp
```diff
@@ -471,6 +478,83 @@ void ClassPropertyType::setUsageFlags(int flags, bool value)
...
+
+ExportValue PrimitivePropertyType::toExportValue(const QVariant &value, const ExportContext &context) const
+{
+    return PropertyType::toExportValue(value, context);
+}
+
+QVariant PrimitivePropertyType::toPropertyValue(const QVariant &value, const ExportContext &context) const
+{
+    return PropertyType::toPropertyValue(value, context);
+}
```
bjorn: Since `toExportValue` and `toPropertyValue` just call the superclass, we can leave them out entirely.

src/libtiled/propertytype.cpp
```diff
@@ -471,6 +478,83 @@ void ClassPropertyType::setUsageFlags(int flags, bool value)
...
+    auto json = PropertyType::toJson(context);
+    json.insert(QStringLiteral("storageType"), storageTypeToString(storageType));
+    if (visualColor.isValid())
+        json.insert(QStringLiteral("visualColor"), visualColor.name(QColor::HexArgb));
+    return json;
+}
+
+void PrimitivePropertyType::initializeFromJson(const QJsonObject &json)
+{
+    storageType = storageTypeFromString(json.value(QStringLiteral("storageType")).toString());
```
bjorn: If we store `QVariant value` as member, I think we can use the `type` and `value` fields here together with `ExportValue` to reuse the existing property loading code and don't need `PrimitivePropertyType::storageTypeFromString` and `PrimitivePropertyType::storageTypeToString`.

src/tiled/propertytypeseditor.cpp
```diff
@@ -208,6 +212,8 @@ PropertyTypesEditor::PropertyTypesEditor(QWidget *parent)
             this, [this] { addPropertyType(PropertyType::PT_Enum); });
     connect(mAddClassPropertyTypeAction, &QAction::triggered,
             this, [this] { addPropertyType(PropertyType::PT_Class); });
+    connect(mAddPrimitivePropertyTypeAction, &QAction::triggered,
+            this, [this] { addPropertyType(PropertyType::PT_Primitive); });
```
bjorn: I think it would be good if this action opened a menu so the user picks a specific primitive type when creating the property type. This can be implemented by setting a menu on the action.

src/tiled/propertytypeseditor.cpp
```diff
@@ -1034,6 +1054,62 @@ void PropertyTypesEditor::addEnumProperties()
...
+void PropertyTypesEditor::addPrimitiveProperties()
+{
+    mPrimitiveColorButton = new ColorButton(mUi->groupBox);
+    mPrimitiveColorButton->setToolTip(tr("Visual Color"));
+    mPrimitiveColorButton->setSizePolicy(QSizePolicy::Fixed, QSizePolicy::Preferred);
+    connect(mPrimitiveColorButton, &ColorButton::colorChanged,
+            this, &PropertyTypesEditor::primitiveColorChanged);
+
+    mPrimitiveStorageTypeCombo = new QComboBox(mUi->groupBox);
+    mPrimitiveStorageTypeCombo->addItems({ tr("Bool"), tr("Int"), tr("Float"), tr("String"), tr("Color") });
```
bjorn: Maybe we could reuse `possiblePropertyValues` here, or separate the primitive values into another function, which we can call directly from here.

src/tiled/propertytypeseditor.cpp
```diff
@@ -1034,6 +1054,62 @@ void PropertyTypesEditor::addEnumProperties()
...
+        if (index != -1)
+            setPrimitiveStorageType(static_cast<PrimitivePropertyType::StorageType>(index));
+    });
+
+    auto nameAndColor = new QHBoxLayout;
+    nameAndColor->addWidget(mNameEdit);
+    nameAndColor->addWidget(mPrimitiveColorButton);
+
+    mDetailsLayout->addRow(tr("Name"), nameAndColor);
+    mDetailsLayout->addRow(tr("Save as"), mPrimitiveStorageTypeCombo);
```
bjorn:
```suggestion
    mDetailsLayout->addRow(tr("Type"), mPrimitiveStorageTypeCombo);
```

src/tiled/propertytypesmodel.cpp
```diff
@@ -239,7 +246,8 @@ bool PropertyTypesModel::checkTypeNameUnused(const QString &name) const
 QString PropertyTypesModel::nextPropertyTypeName(PropertyType::Type type) const
 {
     const auto baseText = type == PropertyType::PT_Enum ? tr("Enum")
-                                                        : tr("Class");
+                         : type == PropertyType::PT_Primitive ? tr("Primitive")
+                         : tr("Class");
```
bjorn: Now it would really make sense to use a `switch` to initialize `baseText`.

src/tiled/tilesetview.cpp
```diff
@@ -163,6 +166,44 @@ void TileDelegate::paint(QPainter *painter,
     if (mTilesetView->markAnimatedTiles() && tile->isAnimated())
         drawFilmStrip(painter, targetRect);
 
+    // Draw property visual indicators
+    {
+        const auto &propertyTypes = *ProjectManager::instance()->project().propertyTypes();
```
bjorn: We can use `Object::propertyTypes()` to get the currently active property types.

src/tiled/tilesetview.cpp
```diff
@@ -163,6 +166,44 @@ void TileDelegate::paint(QPainter *painter,
...
+
+        for (auto it = properties.constBegin(); it != properties.constEnd(); ++it) {
+            if (it.value().userType() != propertyValueId())
+                continue;
+
+            const auto propertyValue = it.value().value<PropertyValue>();
+            if (propertyValue.typeId == 0)
+                continue;
+
+            const auto *type = propertyTypes.findTypeById(propertyValue.typeId);
```
bjorn:
```suggestion
            const auto *type = propertyValue.type();
```
(also there is no need to check `typeId` for 0 above)

src/tiled/tilesetview.cpp
```diff
@@ -323,6 +364,9 @@ TilesetView::TilesetView(QWidget *parent)
     connect(prefs, &Preferences::showTilesetGridChanged,
             this, &TilesetView::setDrawGrid);
 
+    connect(prefs, &Preferences::propertyTypesChanged,
+            this, [this] { viewport()->update(); });
```
bjorn, line 364: We'll also need to update the viewport, or ideally just the specific tiles, when properties are added, removed or changed (see relevant signals in `Document`).

It isn't the easiest thing to handle. One example of updating stuff based on property changes are the object reference arrows which are updated by the `ObjectSelectionItem`, which connects to all these signals in its constructor.

tests/properties/test_properties.cpp
```diff
@@ -448,6 +485,103 @@ void test_Properties::loadEnumInNestedClass()
...
+
+    auto interactable2 = types2.findPropertyValueType(QStringLiteral("Interactable"));
+    QVERIFY(interactable2);
+    QCOMPARE(interactable2->type, PropertyType::PT_Primitive);
+    auto &prim2 = static_cast<const PrimitivePropertyType&>(*interactable2);
+    QCOMPARE(prim2.storageType, PrimitivePropertyType::BoolValue);
+    QCOMPARE(prim2.visualColor, QColor(Qt::red));
+}
+
+void test_Properties::primitiveDefaultValues()
```
bjorn: I don't think this is a meaningful test.

kanishka0411:
Hey @bjorn, thanks for the review!  i have made the changes based on your suggestions.
* The main change is swapping out the StorageType enum for a QVariant defaultValue like you suggested
* The "Add Primitive" button now pops up a menu to pick the type, and I got rid of the storage type combo from the details panel since it's no longer needed.
* Also simplified the tileset view to use propertyValue.type() directly and hooked up the document signals so indicators update in real time.

Let me know if there's anything else you'd like me to change :-)

src/tiled/propertytypeseditor.cpp
```diff
@@ -212,8 +212,30 @@ PropertyTypesEditor::PropertyTypesEditor(QWidget *parent)
...
-            this, [this] { addPropertyType(PropertyType::PT_Primitive); });
+    auto primitiveMenu = new QMenu(this);
+    for (const auto &value : possiblePrimitiveValues()) {
+        QString displayName;
+        switch (value.userType()) {
+        case QMetaType::Bool:    displayName = tr("Bool"); break;
+        case QMetaType::Int:     displayName = tr("Int"); break;
+        case QMetaType::Double:  displayName = tr("Float"); break;
+        case QMetaType::QString: displayName = tr("String"); break;
+        case QMetaType::QColor:  displayName = tr("Color"); break;
```
bjorn: Please use the same functions as in `AddValueProperty::createEditor` to get the names and icons for each type:

```c++
    for (const auto &value : values) {
        const QIcon icon = PropertyTypesModel::iconForProperty(value);
        typeBox->addItem(icon, userTypeName(value), value);
    }
```

src/libtiled/properties.cpp
```diff
@@ -512,6 +512,17 @@ QVariantList possiblePropertyValues(const ClassPropertyType *parentClassType)
...
 }
 
+QVariantList possiblePrimitiveValues()
+{
+    return {
+        QVariant(false),                        // bool
+        QVariant(0),                            // int
+        QVariant(0.0),                          // float
+        QVariant(QString()),                    // string
+        QVariant::fromValue(QColor()),          // color
```
bjorn: I think this should include all primitive types, including `file` and `object`. And then we can change `possiblePropertyValues` above to call this function to initialize `values` before adding the list type and the custom types.

src/tiled/propertytypesmodel.cpp
```diff
@@ -166,7 +166,7 @@ QModelIndex PropertyTypesModel::addNewPropertyType(PropertyType::Type type)
         propertyType = std::make_unique<EnumPropertyType>(name);
         break;
     case PropertyType::PT_Primitive:
-        propertyType = std::make_unique<PrimitivePropertyType>(name);
+        propertyType = std::make_unique<PrimitivePropertyType>(name, defaultValue.isValid() ? defaultValue : QVariant(false));
```
bjorn: `defaultValue` should always be valid, right?

src/tiled/tilesetview.cpp
```diff
@@ -392,6 +388,14 @@ void TilesetView::setTilesetDocument(TilesetDocument *tilesetDocument)
...
         connect(mTilesetDocument, &TilesetDocument::tilesAdded, this, &TilesetView::refreshColumnCount);
         connect(mTilesetDocument, &TilesetDocument::tilesRemoved, this, &TilesetView::refreshColumnCount);
+        connect(mTilesetDocument, &Document::propertyAdded,
+                this, [this] { viewport()->update(); });
+        connect(mTilesetDocument, &Document::propertyRemoved,
+                this, [this] { viewport()->update(); });
+        connect(mTilesetDocument, &Document::propertyChanged,
+                this, [this] { viewport()->update(); });
+        connect(mTilesetDocument, &Document::propertiesChanged,
+                this, [this] { viewport()->update(); });
```
bjorn, line 398: That's a little heavy-handed, but I realize anything more optimal is rather difficult at the moment.

If we add more property-based visuals we'll probably want to introduce a unified signal that gives us the old and the new value, instead of added, removed and changed (where an "invalid" value can be used to signal an absence).

src/libtiled/properties.cpp
kanishka0411: Good idea! Added file and object to ```possiblePrimitiveValues()```, and ```possiblePropertyValues()``` now calls it to initialize the list before adding list and custom types.

src/tiled/propertytypesmodel.cpp
kanishka0411: You're right, removed the check. It's always valid since it comes from ```possiblePrimitiveValues()```

src/tiled/tilesetview.cpp
kanishka0411: Agreed, it's the simplest approach for now. A unified signal with old/new values would be a nice improvement down the line if we add more property-based visuals

src/tiled/propertytypeseditor.cpp
kanishka0411: Done! Now using ```userTypeName(value)``` and ```PropertyTypesModel::iconForProperty(value)``` in a loop, same pattern as ```AddValueProperty::createEditor```

PR #4423, opened 2026-03-12
fix(tilesetview): refresh tile layout on tileset image resize (#3923)

### LINKED ISSUE #3923: Tile Animation Editor may not show additional tiles if source image resized
**Describe the bug**
If the source image underlying a tileset is updated, Tiled will generally reflect that.  However, if a vertically larger image is loaded in, the tiles on the bottom will not appear in Tile Animation Editor (the movie camera icon) until the tileset is closed and re-opened.

**To Reproduce**
1. Create a tileset with a given size.
2. Replace the underlying image with one that is taller.
3. Click on the Tile Animation Editor (movie icon)
4. Attempt to select tile from the new section at the bottom

**Expected behavior**
New tiles are available.

**Media**
N/A

**Specifications:**
Verified using Tiled 1.10.2 on MacOS.

Aayushmaan-24:
Closes #3923

**Problem**  
When replacing a tileset source image with a taller version (additional rows), the Tile Animation Editor's tileset view did not display the new bottom tiles until the tileset document was closed and reopened — even though the main tileset view updated correctly.

**Changes**  
This commit ensures the TilesetView properly reacts to tileset changes:
- Added listener in TilesetView for `TilesetDocument::tilesetChanged(Tileset*)`
- When the tileset changes, call `scheduleDelayedItemsLayout()` and `refreshColumnCount()` to force correct relayout and update the visible tile grid.

- Simplified `TileAnimationEditor::tilesetChanged()` to:
  - Reset the current tile
  - Update the tileset view background
  - Forward `tilesetChanged()` to its `TilesetModel`
  - Rely on the updated `TilesetView` handling for layout refresh (no redundant view refresh needed)

No regressions observed in core tileset behavior or animation editing workflow.


commit-msg.txt
bjorn, line 1: Please don't commit the `commit-msg.txt` file. :-)

Aayushmaan-24: Done, deleted and pushed again.

bjorn:
Hmm, I've tested this and the fix is not working for me. It can't really work, because when a tileset image is reloaded (which is done by the `TilesetManager`), it emits `TilesetManager::tilesetImagesChanged`. There is no emit of `TilesetDocument::tilesetChanged`, which this change hooks the `TilesetView` to.

Also, calling `scheduleDelayedItemsLayout` and `refreshColumnCount` on the view are not the right functions to make sure any newly added tiles become visible. We instead need `TilesetModel::tilesetChanged` to be called.

Finally, the note about "Simplified `TileAnimationEditor::tilesetChanged`" is weird, because this PR does not change it, and the existing function already does what it was supposedly simplified to.

Are you wasting my time with LLM fails?

Unfortunately, this code is a real mess. Ideally we'd have a single `TilesetModel` instance for each tileset, but this is currently not possible due to the way the dynamic wrapping is implemented. The simple fix for now is to connect the `TileAnimationEditor` to `TilesetManager::tilesetImagesChanged` as well, so I'll push a patch that does just that (done in #4508).

Issue #3833, opened 2023-10-19
Status Bar pixel coord is rounded rather than floored for orthogonal maps


howprice:
**Describe the bug**
The integer pixel coordinates displayed in the Tiled status bar are incorrect for object layers. The numbers appear to be rounded to nearest rather than floored. This can lead to a lot of confusion when debugging placements by moving the mouse and looking at the coord.

**To Reproduce**
1. New Map
2. Orthogonal
3. Place a tile at (0,0) in the default tile layer in which each pixel is obvious e.g. checkerboard pattern, zoom in on it.
4. Add an Object Layer
5. Select the object layer
6. Select *Select Objects (S*
7. Move the mouse pointer around in the top left pixel. It can be seen that the Pixel Coord in the Status Bar changes from (0,0), (0,1), (1,0), and (1,1) depending on which quadrant of the pixel the mouse is pointing at.

**Expected behavior**
In Step 7 above, the Pixel Coordinate in the Status bar should always read (0,0). It would make sense to floor the float/double pixel coord instead of rounding it in this case.

More discussion here: https://discord.com/channels/524610627545595904/524610627545595906/1164211436206182532

**Media**
![image](https://github.com/mapeditor/tiled/assets/13691253/00e93623-693e-4af6-891b-9bdc90d7969d)

**Specifications:**
 - OS: Windows 10
 - Tiled Version: 2023.08.03

howprice:
> Is this issue actually specific to orthogonal maps? Is it floored for hexagonal, etc maps? I suspect it's true for all map types, it's just that for non-staggered Isometric maps in particular, it's less relevant.

I only have experience with orthogonal maps I'm afraid. I'm not sure how any changes would affect other map types.

kunal649:
took a look at this,  traced it to QPointF::toPoint() in mouseMoved() in both abstractobjecttool.cpp and abstractworldtool.cpp. toPoint() rounds to nearest which explains the flickering. tile coords right below in the same function already use qFloor correctly, so pixel coords should do the same.
it also covers all map types since these are base classes ,,, and interestingly the staggered renderer test already uses qFloor for the same conversion, so this aligns with the existing pattern throughout the codebase.

ive opened a PR : [PR](https://github.com/mapeditor/tiled/pull/4426)
open to suggestions if im missing something : )

PR #4426, opened 2026-03-12
Fix status bar pixel coords being rounded instead of floored

kunal649:
I've mentioned what change might be needed in issue #3833 comment : [https://github.com/mapeditor/tiled/issues/3833#issuecomment-4049757604](https://github.com/mapeditor/tiled/issues/3833#issuecomment-4049757604)

As pointer is in top-left pixel, it's floored : 0, 0 (0, 0) and stopped flickering around when we move it in single pixel. 
Screenshot (81)


bjorn:
Thanks for fixing this, @kunal649! (and let me go hide somewhere for leaving this issue open for almost 3 years...)

kunal649:
haha, better late than not floored 😄. thanks for the merge!

PR #4433, opened 2026-03-14
Add preference for world object label visibility

### LINKED ISSUE #4078: Add option for viewing all object names in world

Tiled doesn't have option for viewing all object names when you building map using world, only current selected map objects are displayed.

example map:
image

Add new option for rendering object names when `world` are loaded:
1. only selected (default)
2. nearest to selected
3. all

Build your map in single file.

Krish2882005:
Closes #4078

**Summary**
Adds options to display object names across multiple maps when working in a World. 

**Changes**
* Added `WorldObjectLabelVisibility` preferences and UI toggles (`Active Map Only`, `Nearest Maps`, `All Maps`).
* Background maps now only process labels, suspending other updates to prevent CPU overhead.

**How to test**
1. Open a World with multiple maps containing named objects.
2. Toggle `View -> Show World Object Names` modes and verify label visibility on adjacent/distant maps.

### DIFF
--- a/src/tiled/mainwindow.cpp
+++ b/src/tiled/mainwindow.cpp
@@ -463,6 +463,26 @@ MainWindow::MainWindow(QWidget *parent, Qt::WindowFlags flags)
     connect(objectLabelVisibilityGroup, &QActionGroup::triggered,
             this, &MainWindow::labelVisibilityActionTriggered);
 
+    QActionGroup *worldObjectLabelVisibilityGroup = new QActionGroup(this);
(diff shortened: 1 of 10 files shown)


bjorn:
Thanks for working on this, @Krish2882005!

I'm not sure this makes sense as an entirely separate setting. I see this is what @boruok asked for originally, but I don't see a good reason not to integrate this with the other label options:

* No Labels
* Only Selected
* All Objects on Map
* All Objects in World

I don't think we need the option to only show labels for "nearby" maps in the world, I think this follows naturally from the (currently hidden) option to only show nearby maps in combination with the "All Objects in World" option.

Krish2882005:
I’ll make the changes once my exams are over 👍

Issue #1723, opened 2017-09-06
Allow deleting object templates from a template group


bjorn:
Currently you can't delete a object template from a template group. This is a missing feature that unfortunately @thabetx didn't get around to within the timeline of his project.

unichronic:
hi @bjorn !
Can be annoying for projects that use templates a lot to go and do it outside Tiled. Having these actions inbuilt within TIled would be a nice feature to have. I'd be interested to work on this.
but then this change could land a little on the bigger side I think.
what would you suggest? breaking it down into multiple PRs or one single PR would be better?

bjorn:
> what would you suggest? breaking it down into multiple PRs or one single PR would be better?

I think it would be a good idea to break this down into multiple PRs. You might want to start with just adding "Delete template".

Note that in the meantime there is no "Templates" view anymore, it got merged into the "Project" view, which shows the files in a project, including template files. There is still the "Template Editor", but I guess the suggested actions don't really have a place there.

Also note that these file-based actions (delete, rename, move) are in fact relevant for all Tiled assets. Users also want to be able to rename or move maps, tilesets or images while keeping existing references intact. It would be good if you could plan for this in your implementation, though it's alright to focus on one asset type at first. Most of the problem is probably to determine which other assets in the project are affected by a deletion/rename/move (since Tiled has no "dependency manager" yet) and then to actually load/save them with updated references (or maybe only load if needed, and rely on user to save when they want to).

I'll assign this issue to you now. Please let me know if you'd like to create any sub-issues from the above list or if you have any questions!

unichronic:
I think creating subissues for each of the actions is enough.

PR #4437, opened 2026-03-16
Add functionality for Delete Template

unichronic:
ref #1723 
I have added the delete functionality for templates. Please check it out if the changes are alright.

menu

Currently after deleting a template it takes 10-15s to disappear from the file list during Folders referesh using `FileSystemWatcher`.
Do you think that's an issue?

### DIFF
--- a/src/libtiled/templatemanager.cpp
+++ b/src/libtiled/templatemanager.cpp
@@ -83,6 +83,11 @@ ObjectTemplate *TemplateManager::loadObjectTemplate(const QString &fileName, QSt
     return objectTemplate;
 }
 
+void TemplateManager::deleteTemplate(const QString &fileName)
+{
+    mWatcher->removePath(fileName);
+}
+
 void TemplateManager::pathsChanged(const QStringList &paths)
 {
     for (const QString &fileName : paths) {
--- a/src/tiled/projectdock.cpp
+++ b/src/tiled/projectdock.cpp
@@ -23,8 +23,12 @@
 #include "actionmanager.h"
 #include "addremovetileset.h"
 #include "documentmanager.h"
+#include "layer.h"
+#include "mapdocument.h"
 #include "mapdocumentactionhandler.h"
 #include "mapeditor.h"
+#include "mapobject.h"
+#include "objectgroup.h"
 #include "objecttemplate.h"
 #include "preferences.h"
 #include "projectmanager.h"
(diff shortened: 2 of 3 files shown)


src/tiled/projectdock.cpp
```diff
@@ -345,6 +355,35 @@ void ProjectView::restoreExpanded(const QModelIndex &parent)
...
+                    }
+                }
+                if (inUse) break;
+            }
+            if (inUse) break;
+        }
+        if (inUse) break;
+    }
+
+    if (inUse) {
```
bjorn: Consider making a local static `isTemplateInUse` function, so you can just `return true` instead of setting and checking an `isUse` variable.

src/libtiled/templatemanager.cpp
```diff
@@ -83,6 +83,14 @@ ObjectTemplate *TemplateManager::loadObjectTemplate(const QString &fileName, QSt
     return objectTemplate;
 }
 
+void TemplateManager::removeTemplate(const QString &fileName)
+{
+    if (auto templ = mObjectTemplates.take(fileName)) {
+        mWatcher->removePath(fileName);
+        delete templ;
```
bjorn: Of course, we can't actually delete the template instance when it is still in use, even if the user might be fine with this. By deleting this object, any references to this template will become roaming pointers and can cause crashes or random broken behavior later.

Currently, templates are never unloaded, which is what avoids such issues. I wonder if it hurts to just keep them in memory even if we delete the file.

src/tiled/projectdock.cpp
unichronic: done

src/libtiled/templatemanager.cpp
unichronic: done

unichronic:
> Yes, this is not ideal. It's rather longer than I would expect since the FileSystemWatcher sets mChangedPathsTimer.setInterval(500), meaning it should only take half a second until the list is refreshed.

yes, the `FileSystemWatcher` triggers instantly, but the time to remove depends on the size the project folder since the whole file tree is refreshed everytime. I tried with smaller smaller folder, it was removed instantly. So I think unless the project folder has too many items it won't take that much time (I had my project folder as one of my root folders xD). 

Addressed the other comments.

unichronic:
hello @bjorn any updates on this?

Issue #3282, opened 2022-02-21
Swapping between maps jiggles the selected layer and groups


Macksaur:
### Problem

When I am moving around multiple maps and copy/pasting pieces as source material and I swap back to the target map I find all the groups closed and the layer I was on forgotten and instead replaced with a "best guess" at a layer with the same name.

```
Map A:
- Group1
    - Objects
    - ActualTarget <-- this layer is selected
- Group0
    - Objects

Map B:
- Group1
    - Objects <-- this layer will be selected in order to copy objects
```

Swapping between `Map A -> Map B` and swapping back to `Map A` after selecting some objects from `Map B` it places me in `Group0/Objects` when I return to `Map A` instead of `Group1/Objects`, not that it matters because I had selected the layer `Group1/ActualTarget` before I left `Map A`, intending to paste there.

### Goal

I want to be able to edit multiple maps simultaneously with complex group structures and copy/paste between them onto different layers without having to constantly find the layer I was previously on.

1) Remember the open/closed status of group layers when switching maps
2) Remember the selected layer when swapping between maps
3) If none of the above, make a better guess at the layer including group structure?

bjorn:
> 1. Remember the open/closed status of group layers when switching maps

This feature is currently missing.

> 2\. Remember the selected layer when swapping between maps

That's what Tiled does by default. The exception is when using a world to jump from one map to another, in which case it tries to be smart by keeping the same layer(s) selected, which it does by looking for layers with the same name.

> 3\. If none of the above, make a better guess at the layer including group structure?

I guess indeed the way forward would be to make the search for matching layers take into account group layers.

UliAbo:
I don't how complex this is all to implement, but **point 1 "Remember the open/closed status of group layers when switching maps" would be very helpful** for me as I have a lot of groups and using at least 2 TMX files in parallel. And this drives me crazy opening the groups again after switching... 

Would it be possible to implement at least point 1 "soon-ish", as I could imagine the other two are more complex?

bjorn:
@UliAbo Sure, I'll have a look at it.

bjorn:
@UliAbo Done for Tiled 1.8.5 or 1.9, whichever comes first. You can [get a build from here](https://github.com/mapeditor/tiled/actions/runs/2089450912) in a moment. :-)

UliAbo:
That's cool @bjorn 
I will give it a try today!

UliAbo:
[...]

PR #4438, opened 2026-03-16
Add preference to disable layer synchronization in World Tool

kunal649:
As discussed in issue #3282, the automatic layer synchronization (`switchToDocumentAndHandleSimiliarTileset`) actively overwrites user's selected layer on target map ,, which breaks the copy/paste workflow between maps. 
Through this PR, i introduced a toggle in the Preferences (Interface --> Behavior) to give users control over this behavior. 

following changes made: 
1. Added `syncLayersInWorldTool` boolean to `Preferences`. 
2. Added a checkbox to `preferencesdialog.ui` and linked it in `preferencesdialog.cpp`.
3. Updated `MapItem::mouseReleaseEvent` and `WorldMoveMapTool::mouseReleased` to check this preference. And if unchecked, it falls back to standard `switchToDocument` logic, preserving the individual layer state of each map.

Fixes : #3282 
Below is the proof of successfully working. Let me know if I missed something. 

https://github.com/user-attachments/assets/2caa7f60-5b19-425b-9856-414cf573df7e

tgl

### DIFF
--- a/src/tiled/mapitem.cpp
+++ b/src/tiled/mapitem.cpp
@@ -335,9 +335,16 @@ void MapItem::mouseReleaseEvent(QGraphicsSceneMouseEvent *event)
         MapView *view = static_cast<MapView*>(event->widget()->parent());
         QRectF viewRect { view->viewport()->rect() };
         QRectF sceneViewRect = view->viewportTransform().inverted().mapRect(viewRect);
-        DocumentManager::instance()->switchToDocumentAndHandleSimiliarTileset(mMapDocument.data(),
-                                                                              sceneViewRect.center() - pos(),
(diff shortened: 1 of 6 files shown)


kunal649:
Initially, I was trying to fix the sync logic itself!! I wrote a helper function to make `switchToDocumentAndHandleSimiliarTileset` account for nested groups by building and comparing absolute paths (like..., Group1/objects).
however, while testing the copy/paste workflow, I realized that making the algorithm "smarter" didn't solve the root UX issue. like, even if the sync is 100% correct, it's still annoying if you just want to keep different layers clicked on different maps.
So eventually  i scrapped the path-matching overhead in favor of this simple bypass toggle : ) 
 (pardon my grammar)

PR #4439, opened 2026-03-17
Adds initial mcp server

jackboyla:


bjorn:
Most of this PR seems to be adding Python code, but it does all seem to be optional? Or does the Python code achieve something which couldn't be done with just a Tiled extension?

If your MCP extension works without the Python code, I think it could be a good example for how to provide something like this for Tiled and it would be nice to publish it somewhere separately.

Eventually I should have a look at providing that `tiled.setTimeout` function...

PR #4440, opened 2026-03-17
Fix ability to change properties after deselecting current object

bjorn:
When deselecting the current object, its properties generally remain visible in the Properties view. But when changing any of the properties, the changes were only applied to "selected objects".

Now, when there is still a `MapObject` set as the current object, changing its properties will apply to it regardless of whether it is still selected.


bjorn:
How could I forget to include that one in the 1.12.1 release... Oh well, there will surely be a 1.12.2.

PR #4449, opened 2026-03-19
Reduce animated tile marker opacity during Wang editing

wheevu:
Animated tiles keep their film-strip corner marker while editing Wang/Terrain sets in the tileset view. Since the Wang overlay also uses the tile corner area, the marker can compete visually with that information.

This adjusts the marker opacity only while Wang editing is active. The normal tileset browsing is unchanged.


bjorn:
> review requested due to automatic review settings

Hmm, not sure what happened here, I don't recall enabling automatic review settings.

Anyway, I really like this idea, but I didn't like the trailing whitespace and removed empty lines, so I've reverted those changes. I also thought that if we're reducing opacity in terrain edit mode, we should reduce it properly, since 0.5 opacity still left the animation marker rather visible, so I've reduced it further to 0.3.

Thanks for this nice improvement, @wheevu!

PR #4455, opened 2026-03-22
Fix object reference arrows not hiding with per-object visibility

### LINKED ISSUE #4451: objects reference arrows still draw despite object toggled hidden
### Describe the bug

Image

Notice how all objects are hidden but it's still drawing the object reference arrows .

Here is the full map file if needed:
<details>
<summary>tiled json file</summary>

```
{ "compressionlevel":-1,
 "height":1,
 "infinite":false,
 "layers":[
        {
         "data":[0],
         "height":1,
         "id":1,
         "name":"Tile Layer 1",
         "opacity":1,
         "type":"tilelayer",
         "visible":true,
         "width":1,
         "x":0,
         "y":0
        },
    
        {
         "draworder":"topdown",
         "id":3,
         "name":"Object Layer 2",
         "objects":[
                {
                 "height":0,
                 "id":4,
                 "name":"point",
                 "opacity":1,
                 "point":true,
                 "rotation":0,
                 "type":"",
                 "visible":true,
                 "width":0,
                 "x":24,
                 "y":0
                }],
         "opacity":1,
         "type":"objectgroup",
         "visible":true,
         "x":0,
         "y":0
        },
    
        {
         "draworder":"topdown",
         "id":2,
         "name":"Object Layer 1",
         "objects":[
                {
                 "height":24,
                 "id":1,
                 "name":"obj1",
                 "opacity":1,
                 "properties":[
                        {
                         "name":"anims",
                         "type":"list",
                         "value":[
                                {
                                 "type":"string",
                                 "value":"foo"
                                },
                            
                                {
                                 "type":"int",
                                 "value":3
                                },
                            
                                {
                                 "type":"bool",
                                 "value":false
                                },
                            
                                {
                                 "type":"object",
                                 "value":2
                                },
                            
                                {
                                 "type":"object",
                                 "value":3
                                }]
                        },
                    
                        {
                         "name":"test",
                         "type":"int",
                         "value":1
                        }],
                 "rotation":0,
                 "type":"",
                 "visible":true,
                 "width":24,
                 "x":8,
                 "y":8
                },
            
                {
                 "height":0,
                 "id":2,
                 "name":"obj2",
                 "opacity":1,
                 "rotation":0,
                 "type":"",
                 "visible":true,
                 "width":0,
                 "x":0,
                 "y":0
                },
            
                {
                 "height":0,
                 "id":3,
                 "name":"obj3",
                 "opacity":1,
                 "rotation":0,
                 "type":"",
                 "visible":true,
                 "width":0,
                 "x":-8,
                 "y":24
                }],
         "opacity":1,
         "type":"objectgroup",
         "visible":true,
         "x":0,
         "y":0
        }],
 "nextlayerid":4,
 "nextobjectid":5,
 "orientation":"orthogonal",
 "renderorder":"right-up",
 "tiledversion":"1.12.0",
 "tileheight":32,
 "tilesets":[],
 "tilewidth":32,
 "type":"map",
 "version":"1.11",
 "width":1
}
```
</details>

### Steps to reproduce

1. Create an object with reference arrows to other objects
2. Toggle their visibility off

### Expected behavior

Object reference arrows should be hidden if the object visibility is also hidden. Turning visibility of the entire layer off seems to turn off the object reference arrows though

### OS version

archlinux on sway/wayland

### Tiled version

1.12

Aayushmaan-24:
Object reference arrows were correctly hidden when hiding an entire layer, but remained visible when toggling the visibility of individual objects.

The root cause was in addRemoveObjectReferences(): it checked objectGroup->isHidden() to skip hidden layers, but never checked object->isVisible() for individual objects. Additionally, when a MapObjectsChanged event was emitted for a visibility toggle, only syncOverlayItems() was called (which only syncs positions), so the arrows were never removed.

Fix:
- In addRemoveObjectReferences(), skip invisible source objects and skip creating arrows that point to invisible target objects.
- In changeEvent(), when MapObjectsChanged carries VisibleProperty, call addRemoveObjectReferences() to rebuild arrows.

Fixes #4451

### DIFF
--- a/src/tiled/objectselectionitem.cpp
+++ b/src/tiled/objectselectionitem.cpp
@@ -411,9 +411,15 @@ void ObjectSelectionItem::changeEvent(const ChangeEvent &event)
     case ChangeEvent::LayerChanged:
         layerChanged(static_cast<const LayerChangeEvent&>(event));
         break;
-    case ChangeEvent::MapObjectsChanged:
-        syncOverlayItems(static_cast<const MapObjectsChangeEvent&>(event).mapObjects);
+    case ChangeEvent::MapObjectsChanged: {
+        const auto &mapObjectsChange = static_cast<const MapObjectsChangeEvent&>(event);
+        if (mapObjectsChange.properties & MapObject::VisibleProperty) {
+            if (Preferences::instance()->showObjectReferences())
+                addRemoveObjectReferences();
+        }
+        syncOverlayItems(mapObjectsChange.mapObjects);
         break;
+    }
     case ChangeEvent::MapObjectsAdded:
         objectsAdded(static_cast<const MapObjectsEvent&>(event).mapObjects);
         break;
@@ -931,7 +937,13 @@ void ObjectSelectionItem::addRemoveObjectReferences()
                 continue;
 
             for (MapObject *object : objectGroup->objects()) {
+                if (!object->isVisible())
+                    continue;
+
                 forEachObjectReference(object->properties(), [&] (ObjectRef ref) {
+                    MapObject *targetObject = DisplayObjectRef(ref, mMapDocument).object();
+                    if (targetObject && !targetObject->isVisible())
+                        return;
                     ensureReferenceItem(object, ref);
                 });
             }


src/tiled/objectselectionitem.cpp
```diff
@@ -931,7 +937,13 @@ void ObjectSelectionItem::addRemoveObjectReferences()
                 continue;
 
             for (MapObject *object : objectGroup->objects()) {
+                if (!object->isVisible())
+                    continue;
+
                 forEachObjectReference(object->properties(), [&] (ObjectRef ref) {
+                    MapObject *targetObject = DisplayObjectRef(ref, mMapDocument).object();
+                    if (targetObject && !targetObject->isVisible())
+                        return;
```
bjorn, line 946: If we're going to hide based on target object visibility, it will be easier to do it in `ensureReferenceItem`. However, I'm not entirely sure it's helpful. We also don't hide arrows pointing at objects in hidden layers (only arrows pointing _from_ hidden layers).

src/tiled/objectselectionitem.cpp
bjorn, line 993: There's a check on `object->isVisible()` missing here.

src/tiled/objectselectionitem.cpp
```diff
@@ -411,9 +411,15 @@ void ObjectSelectionItem::changeEvent(const ChangeEvent &event)
     case ChangeEvent::LayerChanged:
         layerChanged(static_cast<const LayerChangeEvent&>(event));
         break;
-    case ChangeEvent::MapObjectsChanged:
-        syncOverlayItems(static_cast<const MapObjectsChangeEvent&>(event).mapObjects);
+    case ChangeEvent::MapObjectsChanged: {
+        const auto &mapObjectsChange = static_cast<const MapObjectsChangeEvent&>(event);
+        if (mapObjectsChange.properties & MapObject::VisibleProperty) {
+            if (Preferences::instance()->showObjectReferences())
+                addRemoveObjectReferences();
```
bjorn, line 418: This full refresh based on any toggle of visibility may be a little heavy. I think this is another reason to not base reference visibility on the target object, because in that case we could instead call the `addRemoveObjectReferences` overload that takes a specific object, for each of the changed objects only.

bjorn (changes requested):
Thanks for trying to address this issue!

As-is, the change is not entirely complete and I think it's not a good idea to take target object visibility into account. See inline comments.

PR #4465, opened 2026-03-24
docs: clarify GPL is version 2 or later (COPYING)

### LINKED ISSUE #2018: [Clariication Required] What version of GPL is the GPL part of Tiled is ?
Hello,
I'm trying to make a package for your program for the Void Linux, which needs to know what version of GPL 2 the GPL part of the codebase uses, aka, if its GPL2 only or GPL 2 and later.

I'm unsure about this part of the license so it could help if you clarify it.

PrateekSingh070:
Clarifies that Tiled is licensed under "GPL v2 or later" in `COPYING`.

The project's actual licensing (GPL-2.0-or-later, matching `README.md` and source file headers) wasn't reflected in `COPYING`, which just included the bare GPL v2 text. This adds a short header above the license text making the "or later" intent explicit, so anyone opening `COPYING` sees the same licensing statement that's in the source headers.

Branch rebased onto current master; one-file change only.

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -15,3 +15,16 @@ tmxviewer-java       util/java/tmxviewer-java  BSD 2-clause license
 
 The full text of each license is provided in the files LICENSE.GPL
 and LICENSE.BSD.
+
+GPL version (2 only vs 2 or later)
+----------------------------------
+The Tiled application and its bundled plugins (see table above) are
+licensed under the GNU General Public License **version 2 or later**.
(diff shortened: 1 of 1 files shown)


PrateekSingh070:
Hi @bjorn,
Just a gentle follow-up on this docs clarification PR.
I kept it intentionally small to avoid any merge friction, and I’m happy to revise wording if you prefer a different phrasing in COPYING.
Thanks.

PR #4466, opened 2026-03-25
Fix "Detach templates" not working when export target is the same file

### LINKED ISSUE #4456: "Detach templates" does not work with "Repeat last export on save" when exporting to the same file.
### Describe the bug

With both "Detach templates" and "Repeat last export on save" options enabled, when last export was to the same file that is being edited - templates do not detach, you still get ` <object id="59" template="chest.tx" x="15" y="94"/>` all over the file.

### Steps to reproduce

1. Create and save map in the "Map.tmx" file.
3. Open it, go into the preferences, set "Detach templates" and "Repeat last export on save".
4. Now export this map into itself "Map.tmx"
5. Now add some templates on the map.
6. Now press Ctrl+S to save the map(Which should perform last export and trigger template detaching).
7. Open "Map.tml" in the text editor and observe that templates did no detach, you have "template" objects everywhere.

### Expected behavior

Templates should detach.

### OS version

Windows 10

### Tiled version

1.12.0

Hijanhv:
## Summary

Fixes #4456 — when **Detach template instances** and **Repeat last export on save** are both enabled and the export target is set to the same file as the map itself, templates were not being detached on subsequent saves.

**Root cause (two interacting bugs):**

1. `ExportHelper::prepareExportMap` always clears `exportFileName`/`exportFormat` from the exported clone — so the file on disk loses the export settings. On the next reload triggered by the file watcher, those settings are gone, causing `exportDocument` to early-exit with an empty export filename.

2. The export write updates the file's on-disk timestamp, but `mLastSaved` is only updated during the regular save. The file watcher therefore treats the export write as an external change and auto-reloads the document, which replaces the in-memory map (including its export settings) with the freshly-loaded one that has no export settings.

**Fix:**

- In `ExportHelper::prepareExportMap`, preserve `exportFileName` and `exportFormat` in the exported file when the export target and the map file resolve to the same canonical path. This restores persistence across reloads and Tiled restarts.
- Add `Document::refreshLastSaved(fileName)` and call it after a successful export to the same file (in `exportDocument`, `exportMapAs`, and `exportTilesetAs`). This prevents the file watcher from treating the export write as an external change and triggering an unnecessary auto-reload.

## Test plan

- [ ] Enable **Detach template instances** and **Repeat last export on save** in preferences
- [ ] Open a map, do **File → Export As…** targeting the same `.tmx` file
- [ ] Add template instances to the map, then press Ctrl+S
- [ ] Open the saved file in a text editor — template objects should be fully detached (no `template=` attribute)
- [ ] Press Ctrl+S again — templates should continue to be detached on each save
- [ ] Close and reopen Tiled, reload the file — export-on-save should still work (export settings persisted in file)
- [ ] Verify that exporting to a *different* file still clears export settings from the exported file (no regression)

### DIFF
--- a/src/tiled/exporthelper.cpp
+++ b/src/tiled/exporthelper.cpp
@@ -24,6 +24,8 @@
 #include "objectgroup.h"
 #include "wangset.h"
 
+#include <QFileInfo>
+
 namespace Tiled {
 
 /**
@@ -106,8 +108,15 @@ const Map *ExportHelper::prepareExportMap(const Map *map, std::unique_ptr<Map> &
     // Make a copy to which export options are applied
     exportMap = map->clone();
 
-    // We don't want to save the export options in the exported file
-    if (hasExportSettings) {
+    // We don't want to save the export options in the exported file,
+    // unless we're exporting to the same file (to preserve the export settings
+    // so they can be picked up again on the next save)
+    const auto canonicalMapFile = QFileInfo(map->fileName).canonicalFilePath();
+    const auto canonicalExportFile = QFileInfo(map->exportFileName).canonicalFilePath();
+    const bool isSelfExport = !canonicalMapFile.isEmpty()
+            && canonicalMapFile == canonicalExportFile;
+
+    if (hasExportSettings && !isSelfExport) {
         exportMap->exportFileName.clear();
         exportMap->exportFormat.clear();
     }
--- a/src/tiled/mainwindow.cpp
+++ b/src/tiled/mainwindow.cpp
@@ -1342,6 +1342,10 @@ bool MainWindow::exportDocument(Document *document)
 
             if (exportFormat->write(map, exportFileName, exportHelper.formatOptions())) {
                 statusBar()->showMessage(tr("Exported to %1").arg(exportFileName), 3000);
+                // When exporting to the same file, update the saved timestamp
+                // to avoid triggering an unnecessary auto-reload
+                if (exportFileName == document->fileName())
+                    document->refreshLastSaved(exportFileName);
                 return true;
             }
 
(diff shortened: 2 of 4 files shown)


src/tiled/exporthelper.cpp
```diff
@@ -106,8 +108,15 @@ const Map *ExportHelper::prepareExportMap(const Map *map, std::unique_ptr<Map> &
...
 
-    // We don't want to save the export options in the exported file
-    if (hasExportSettings) {
+    // We don't want to save the export options in the exported file,
+    // unless we're exporting to the same file (to preserve the export settings
+    // so they can be picked up again on the next save)
+    const auto canonicalMapFile = QFileInfo(map->fileName).canonicalFilePath();
+    const auto canonicalExportFile = QFileInfo(map->exportFileName).canonicalFilePath();
+    const bool isSelfExport = !canonicalMapFile.isEmpty()
+            && canonicalMapFile == canonicalExportFile;
```
bjorn, line 117: I don't think this is the right way to do it, because it compares the previously used export file rather than the current export target. You'd need to pass the current export target into the `prepareExportMap` function.

src/tiled/mainwindow.cpp
```diff
@@ -1342,6 +1342,10 @@ bool MainWindow::exportDocument(Document *document)
 
             if (exportFormat->write(map, exportFileName, exportHelper.formatOptions())) {
                 statusBar()->showMessage(tr("Exported to %1").arg(exportFileName), 3000);
+                // When exporting to the same file, update the saved timestamp
+                // to avoid triggering an unnecessary auto-reload
+                if (exportFileName == document->fileName())
+                    document->refreshLastSaved(exportFileName);
```
bjorn, line 1348: This is a bit hacky and then even duplicated in 4 places.

I'm also not sure if the reload is really unnecessary when exporting to the file that is loaded? I guess when the user does this, they expect their export options (like detached templates) to affect their current file, and since the export options are only applied to a clone of the map, the reload in this case is necessary to replace the loaded map with the version that has those options applied.

bjorn (changes requested):
I'm not sure we actually want to disable the reload, and the way to check for "self-export" is slightly broken as such. See inline comments.

Issue #3372, opened 2022-05-24
Double clicking a .tx file should open it in Tiled.app


hotdogsoup-nl:
**Describe the bug**
Double clicking a .tx file should open it in Tiled.app

**To Reproduce**
1. open a Finder window. 
2. double click on a .tx file

***Expected behavior**
Tiled.app should open the template file in the Template editor.

**Work around**
By holding Option+CMD the .tx file can be dragged & dropped onto the Tiled.app icon. Now the file is opened correctly.

**Specifications:**
 - macOS 12.3.1
 - Tiled Version: 1.8.5

bjorn:
Yeah, that's a good thing to look into. I'm not sure if it's enough to just add a line with `tx` to [the Info.plist file](https://github.com/mapeditor/tiled/blob/7f7589e54d824fc28949f5ad914a8e5b2377f7cd/src/tiledapp/Info.plist#L22), but that'd be a start at least. Probably `tsx`, as well as `tmj`, `tsj` and `tj` (more specific extensions supported by the JSON format) should be added there as well.

eishiya:
It would be nice if the Windows installer also associated those other formats with Tiled, ideally optionally. Looks like only TMX is registered:
https://github.com/mapeditor/tiled/blob/3d8724f91ba53d7c1d245b5a8f778662a7aa5d0d/dist/win/installer.wxs#L142-L147

It looks like in some versions of Windows (or at least in some combinations of settings?), manual registrations display a different, less clear icon, so even though it's fairly easy to make the associations manually after installing, it would be nice to have the installer make them.
Screenshot from Mitsuru Tenma on Discord: https://discord.com/channels/524610627545595904/524610627545595906/981968829473239040 (I'll edit this post with a copy of the image if I get their permission to do so.)
On my Windows manual associations look fine, but I can see why they'd be suboptimal on theirs. I couldn't replicate their icon appearance via various settings, so it may be a version difference, in which case, the ugliness is there to stay xP

bjorn:
Implemented for macOS by #4469. Windows installer additions still pending...

Issue #2885, opened 2020-08-10
Tileset file extension (.tsx) conflicts with Typescript JSX files


craxal:
This is a minor issue, but worth considering. Many systems already recognise .tsx files as TypeScript files that can be compiled to .jsx files. This means the system will open the .tsx files in a code editor by default instead of Tiled. Changing the default application is not a good solution for systems that have TypeScript and Tiled tileset files (such as mine).

I suggest changing the default file extension for tileset files. Maybe something like .tmsx?

bjorn:
Puh, Tiled has been using `.tsx` extension for over 12 years and TypeScript comes along and now we shall have to find another name? You might be right, but I feel a little angry about it. :P

For now I'd suggest you just save your external tilesets as `.xml`, which is an alternative supported by Tiled. You can't associate such a file with Tiled either, but at least it avoids it being categorized as TypeScript. For opening files, you're anyway using a project and just typing `Ctrl+P` followed by a file name snippet, right?

craxal:
@bjorn For the record, I don't blame you for feeling angry! The alternative extension is fine, especially since I usually only open maps directly instead of tilesets.

BrendanMulcahy:
This issue has been pretty annoying in Visual Studio 2019 C++ project using Tiled. If you accidentally open one of the tsx files in VS, Intellisense will turn on its Typescript compiler and it seems impossible to disable unless you have a csproj, which we dont. Normal building via cmake still works but seeing hundreds of typescript errors is distracting.

Our solution was to rename all the .tsx files to .xml and rebuild our .tmx files.  After closing and re-opening VS, it seems to have resolved the issue.

Biffyn:
A bit late to the party here, but for those using VS Code, you can easily map file associations. This can be done at the user, workspace or project level by mapping `.tsx` to `xml` in your settings.

I do this at the project level in a `settings.json` file as I'm a front-end dev by trade, so only want this at the project level. 

**.vscode/settings.json**:
```json
"files.associations": {
  "*.tsx": "xml"
}
```

TheBoneJarmer:
> A bit late to the party here, but for those using VS Code, you can easily map file associations. This can be done at the user, workspace or project level by mapping `.tsx` to `xml` in your settings.
> 
> I do this at the project level in a `settings.json` file as I'm a front-end dev by trade, so only want this at the project level.
> 
> **.vscode/settings.json**:
> 
> "files.associations": {
>   "*.tsx": "xml"
> }

[...]

PR #4469, opened 2026-03-25
Update Info.plist with Tiled file types

### LINKED ISSUE #4450: MacOS package does not declare a UTI
I've contemplated the idea of creating a Spotlight importer and/or a Quick Look generator for Tiled maps. However, this means I need a way to declare that they can recognize a Tiled map. It can _probably_ handle matching on file extension, but the usual recommended way of checking file types is with a Uniform Type Identifier.

But Tiled's Info.plist only declares tmx, and only by extension, which means it's automatically assigned a dynamic UTI – so it could be different on every machine (I think). It doesn't even mention the other file types it can open.

This section of the Info.plist would need to be extended:

```xml
<key>CFBundleDocumentTypes</key>
<array>
	<dict>
		<key>CFBundleTypeExtensions</key>
		<array>
			<string>tmx</string>
		</array>
		<key>CFBundleTypeIconFile</key>
		<string>tmx-icon-mac</string>
		<key>CFBundleTypeName</key>
		<string>Tiled Map</string>
		<key>CFBundleTypeRole</key>
		<string>Editor</string>
	</dict>
</array>
```

It would need to add the `LSItemContentTypes` array key in the dict, as well as similar dict entries for other file types.

This is also the reason why double-clicking a tsx file doesn't by default open it in Tiled, and why trying to force the issue gives the warning that MacOS is not sure if Tiled can open tsx files. I thought I saw an open issue for that earlier, but I can't find it now.

JarvisX-Z:
Fix macOS UTI declarations for Tiled file types
This PR improves macOS file type recognition by properly declaring Uniform Type Identifiers (UTIs).
Changes:
Added LSItemContentTypes for .tmx and .tsx
Added UTExportedTypeDeclarations
Added support for .tsx files (tilesets)
Ensured valid and clean Info.plist structure
Result:
macOS can reliably identify Tiled file types
Double-clicking .tsx files should open correctly
Improves integration with Spotlight and Quick Look
Fixes #4450

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,6 +9,7 @@
 * Scripting: Added MapObject.resolvedClassName() (by MatusGuy, #4529)
 * Fixed crash when the selection becomes empty while starting a move (#4536)
 * Linux: Added file associations for .tmj, .tsj, .tiled-project and .world files (by miffe, #4550)
+* macOS: Declared UTIs and file associations for all supported Tiled formats (with Jyotish, #4469)
 * snap: Updated to Qt 6
 * Raised the minimum Qt version to 6.2 (drops Qt 5 support)
 
--- a/src/tiledapp/Info.plist
+++ b/src/tiledapp/Info.plist
@@ -2,14 +2,19 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+
 	<key>CSResourcesFileMapped</key>
 	<true/>
+
 	<key>NSHumanReadableCopyright</key>
 	<string>Copyright 2008-2020 Thorbjørn Lindeijer</string>
+
 	<key>NSHighResolutionCapable</key>
 	<true/>
+
 	<key>CFBundleDocumentTypes</key>
 	<array>
+
 		<dict>
 			<key>CFBundleTypeName</key>
 			<string>Tiled Map</string>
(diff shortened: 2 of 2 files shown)


CelticMinstrel (commented):
Nice! That "conforms to XML" should also let the Finder know that BBEdit can open tmx and tsx files.

I _think_ there were also more file extensions used by the JSON map format? I'm not quite sure; I can't find where I saw those referenced before. Still, recognizing the primary tmx and tsx is already a big improvement.

JarvisX-Z:
Thanks for the feedback!
Good point about additional file extensions for the JSON map format — I’ll look into that and update the PR if needed.
Also, I don’t currently have access to a macOS environment to test these changes locally. I’ve based the implementation on the documentation and existing patterns, so I’d really appreciate your help with testing on macOS.

CelticMinstrel (commented):
That might be difficult, unless there's an artifact I can download from the CI or something – I'm not set up to build Tiled at the moment and am probably missing most of the dependencies.

bjorn:
@CelticMinstrel There are artifacts available for each change. For PRs, click the "Checks" tab at the top, then click on "Build Packages" in the sidebar, and scroll down to "Artifacts". :-)

CelticMinstrel (commented):
Sorry for the delay. It looks like the artifact from this PR works – I can drag-and-drop a .tsx file onto the app and it opens.

But see my comment on the source issue as well.

bjorn:
I've added support for the remaining formats, so now this PR also closes #3372.

bjorn:
Verified the built universal app on macOS 15.7.7 (Intel), registered via `lsregister`. Seven of the eight declared extensions resolve to their new `org.mapeditor.tiled.*` UTIs and open in Tiled: `.tmx`, `.tmj`, `.tsj`, `.tx`, `.tj`, `.tiled-project` and `.world`.

The exception is `.tsx`, and one data point is worth recording here: with a TypeScript-aware editor installed (VS Code / Zed / etc.), `.tsx` resolves to `com.microsoft.typescript` (conforming to `public.text`), and because our tileset document type only claims `org.mapeditor.tiled.tsx` via `LSItemContentTypes`, Tiled is not even offered as a candidate opener for `.tsx`, let alone the default. So the exported UTI does not win `.tsx` back on such machines. This is the known #2885 conflict; the new insight is just that declaring the UTI does not change it. The other seven, including `.tsj`, work reliably.

PR #4476, opened 2026-03-27
Add command variables %exportfile and %exportpath

IoriBranford:


src/tiled/command.cpp
```diff
@@ -74,10 +75,17 @@ static QString replaceVariables(const QString &string, bool quoteValues = true)
         const QString &fileName = document->fileName();
         QFileInfo fileInfo(fileName);
         const QString mapPath = fileInfo.absolutePath();
+
+        const QString &exportFileName = document->lastExportFileName();
+        QFileInfo exportFileInfo(exportFileName);
+        const QString exportPath = QFileInfo(exportFileName).absolutePath();
```
bjorn:
```suggestion
        const QString exportPath = QFileInfo(exportFileName).absolutePath();
```
(also consider doing the same for the other `fileInfo` which is only used once, for consistency)

src/tiled/command.cpp
```diff
@@ -35,6 +35,7 @@
 #include <QProcess>
 #include <QStandardPaths>
 #include <QUndoStack>
+#include <qfileinfo.h>
```
bjorn: The typical way to include Qt headers is by their forwards:

```suggestion
#include <QFileInfo>
```
Also, please keep these includes sorted.

bjorn (commented):
@IoriBranford That's a nice addition. Could you add a line to the `NEWS.md` file about this as well?

bjorn (approved):

Issue #2849, opened 2020-06-26
[feature] ci + homebrew integration


cruzdanilo:
it would be nice if the travis ci script could create a new merge request automatically on [homebrew-cask](https://github.com/Homebrew/homebrew-cask) repository (or something like that) every new release.

bjorn:
Definitely! Do you have an example of a project already doing this? Or could you help with setting that up?

It doesn't need to be triggered from Travis CI of course, it can also use GitHub Actions.

PoonamMehan:
We could use, official Homebrew action (https://github.com/Homebrew/actions)
Maintained by the Homebrew team, so it's reliable long-term. It works by checking for new versions on a schedule (cron job), which means there's always a small delay between a Tiled release and the Homebrew update request being sent.

Another approach, eugenesvk's fork (https://github.com/eugenesvk/action-homebrew-bump-cask)
This one reacts immediately when a new release tag is published, making it faster and more direct. The downside is that it's maintained by a single person, which creates a dependency risk if the repo stops being actively maintained or patched over time.

Happy to hear thoughts on which direction makes more sense for the project, and can you assign me this issue? @bjorn 
> message from discord, pasting it here for tracking purposes

bjorn:
@PoonamMehan I'll assign the issue to you, thank you for looking into this!

To me it sounds like eugenesvk will be the better option. We're not doing Tiled releases that often, so a cron job will always either check too frequently or take too long to update. Maybe it would even be best to set up this action as a manual trigger. It just needs to be easy but it can be part of the release process to trigger that job.

PoonamMehan:
I have made a PR for the workflow file. On your end, you'll have to create a custom personal GitHub access token with the 'public_repo' and 'workflow' scopes. And add this token to repo's secrets as `HOMEBREW_BUMP_TOKEN`.

@bjorn

PR #4478, opened 2026-03-27
fix(ci): replace third-party action with native brew bump-cask-pr command

PoonamMehan:
Alternative approach to #2849 

Revised the merge-request-homebrew.yml workflow to replace the third-party eugenesvk/action-homebrew-bump-cask action with a direct brew bump-cask-pr command. The workflow now uses HOMEBREW_GITHUB_API_TOKEN for authentication, correctly resolves the tag via github.ref_name instead of github.ref because itreturned the full ref path (refs/tags/v1.11.2) rather than just the tag name, and strips the v prefix to pass a clean version number. A regex validation step has also been added to catch invalid version inputs early.

This approach eliminates the dependency on an external action.


bjorn (approved):
Thanks @PoonamMehan, that looks fine so I'll give it a testing run! :-)

bjorn:
[The test](https://github.com/mapeditor/tiled/actions/runs/25163485455) reported:
```
Error: Whoops, the tiled cask has its version update
```
That seems promising, let's see what happens when we release 1.12.2. :-)

bjorn:
Oh well, it's now failing with the following:
```
Error: Whoops, the tiled cask has its version update
pull requests automatically opened by BrewTestBot every ~3 hours!
We'd still love your contributions, though, so try another one
that is excluded from autobump list (i.e. it has 'no_autobump!'
method or 'livecheck' block with 'skip'.)
```
So since Tiled is automatically checked for a new release every 3 hours, it refuses out manual bump PR. Well, I guess then we can drop our workflow again, heh. I'm not sure if it's worth asking to disable BrewTestBot, seems fine to rely on that if it checks so often.

Turns out that was the error all along, but the way the error was highlighted threw me off last time:

image

PR #4481, opened 2026-03-28
Auto select property name text in rename dialog

Praagya26:
### **Summary**

Automatically select the full property name when opening the rename dialog for custom properties.

### **Problem**

Previously, the cursor was placed at the end of the property name, requiring manual selection before typing a new name.

### **Fix**

Use QTimer::singleShot(0) to select all text in the internal QLineEdit once the dialog is shown.

BEFORE: 
image

AFTER:
image

### **Result**

Improves rename workflow by allowing users to immediately overwrite the property name.

### DIFF
--- a/src/tiled/propertieswidget.cpp
+++ b/src/tiled/propertieswidget.cpp
@@ -19,8 +19,7 @@
  */
 
 #include "propertieswidget.h"
-#include <QTimer>
-#include <QLineEdit>
+
 #include "actionmanager.h"
 #include "changeimagelayerproperty.h"
 #include "changelayer.h"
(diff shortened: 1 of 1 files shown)


src/tiled/propertieswidget.cpp
```diff
@@ -2902,6 +2903,10 @@ void PropertiesWidget::renameProperty(const QString &name)
     dialog->setInputMode(QInputDialog::TextInput);
     dialog->setLabelText(QCoreApplication::translate("Tiled::PropertiesDock", "Name:"));
     dialog->setTextValue(name);
+    QTimer::singleShot(0, dialog, [dialog]() {
+        if (auto lineEdit = dialog->findChild<QLineEdit*>())
+            lineEdit->selectAll();
```
bjorn, line 2908: Did you try just calling this function after the `dialog->open()` call below, rather than using `QTimer::singleShot`?

Praagya26: I used `QTimer::singleShot` to wait until the dialog is fully shown before accessing the line edit. I'll try calling it right after `dialog->open()` as suggested and update the PR based on the result.

Praagya26: @bjorn I've moved the selection logic to run right after dialog->open() using a queued invocation. This works as expected without needing QTimer::singleShot.

bjorn: I guess something went wrong with pushing this change, because in 410682da054f0edab9cc3e80c0a816c9463b8937 a stray submodule pointer appeared instead. I'll have a look at fixing this up.

bjorn:
Hmm, on Linux the name is already being selected, so I'll need to verify this chance on Windows.

bjorn:
Hmm, when testing the Tiled 1.12.1 release built against Qt 6.10.2 on Windows 11 (25H2) I also can't reproduce the original issue. I'm going to close this for now, but feel free to reopen with additional information needed to reproduce the original issue.

PR #4484, opened 2026-03-31
Add tint color support for tiles and tile objects #3950 

### LINKED ISSUE #3950: Allow tinting of a tile/object
I am using Tiled also as an "object editor" for my game. My workflow is:
- Create a tileset with a collection of images
- Add an image to the tileset (e.g. the first frame of the idle animation of a game object like the player)
- Use collision editor to create Box2D fixture definitions
- Assign a class to the tile to add game properties to it, like damage, speed, ...

One of the properties is a color because it is possible to tint a specific game object. From what I know, this is currently not possible.
You can set a color for the class, which is not correct in my case because it is a general class for all game objects.
You can set a color on an object layer which is also not correct because I need a color for a specific object that is related to a specific tile of a tileset.

Ideally, we can have a color attribut for a class with a specific key that will be used for tinting and which is then also considered for the tile rendering in Tiled (tile in tileset + object on objectLayer).
Maybe "tintColor" as a key or just "tint".

Since you can theoretically have multiple color attributes for a class, Tiled somehow needs to know which color it should use for tinting. That's why a special key might be useful or a separate property type "tint color" that can only be added once to a class.

I think there are no alternatives?

YusefZahran:
Closes #3950 

Additions:
This PR introduces tinting for individual tiles through the tileset view, and tile objects for tilemaps.
Currently as of this PR, the tint color of both tiles and tile objects, and the Layer's tint color multiply together (respectively depending on type of layer). The issue thread had discussion about how the tints should affect each other, but I've implemented them to multiply - for now - as I think actually using the feature will give a better idea about how this part should be handled.

Missing:
While currently saving work fine for both tsx and json files, there's no support for exports yet (for example, exporting the tint color as modulate property for godot tiles).

Looking forward for feedback on this. :)

### DIFF
--- a/src/libtiled/isometricrenderer.cpp
+++ b/src/libtiled/isometricrenderer.cpp
@@ -352,7 +352,7 @@ void IsometricRenderer::drawMapObject(QPainter *painter,
         QRectF bounds = { pixelToScreenCoords(object->position()), object->size() };
         bounds.translate(-alignmentOffset(bounds, object->alignment(map())));
 
-        CellRenderer(painter, this, object->objectGroup()->effectiveTintColor())
+        CellRenderer(painter, this, object->effectiveTintColor())
                 .render(cell, bounds.topLeft(), bounds.size());
 
         if (testFlag(ShowTileObjectOutlines)) {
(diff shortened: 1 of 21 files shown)


hiive:
I would love to see this feature merged.

PR #4485, opened 2026-03-31
Add initial glTF export plugin scaffold

PrateekSingh070:
Scaffold only — not ready for review as an exporter. This PR adds the plumbing a glTF / .glb map format plugin would need (plugin.json, qbs, the `Tiled::WritableMapFormat`-derived class, name filter, output-file enumeration) and nothing else.

Scope after cleanup:

- `src/plugins/gltf/` — seven new files
- `src/plugins/plugins.qbs` — one line registering the plugin

The plugin is disabled by default (`condition: false` in `gltf.qbs`, and `"defaultEnable": false` in `plugin.json`), so this has no effect on default builds. The exporter's `write()` currently returns `false` with a "glTF export is not implemented yet (GSoC 2026 work in progress)" error, which is the honest state rather than a half-working writer.

I'm keeping this open mainly as evidence attached to the GSoC 2026 proposal for mapeditor/tiled#2741. If the proposal is accepted, I'll reopen the actual exporter work as reviewable chunks (vertex/index buffers, embedded textures, .glb container, validator integration) rather than as one large PR. Happy to close this if maintainers would rather not carry an inert scaffold branch.

Earlier iterations of this branch touched unrelated files (a CI workflow, NEWS.md, a doc reference page, and two unrelated Tiled widgets). All of that has been dropped; the branch has been rebased onto current master and squashed to one commit containing only the scaffold files.

### DIFF
--- a/src/plugins/gltf/gltf.qbs
+++ b/src/plugins/gltf/gltf.qbs
@@ -1,0 +1,14 @@
+TiledPlugin {
+    condition: false   // WIP scaffold — not built by default. See #2741.
+
+    cpp.defines: base.concat(["GLTF_LIBRARY"])
+
+    files: [
+        "gltfexporter.cpp",
+        "gltfexporter.h",
+        "gltf_global.h",
+        "gltfplugin.cpp",
+        "gltfplugin.h",
+        "plugin.json",
+    ]
+}
--- a/src/plugins/gltf/gltf_global.h
+++ b/src/plugins/gltf/gltf_global.h
@@ -1,0 +1,29 @@
+/*
+ * glTF Tiled Plugin
+ * Copyright 2026, Prateek Singh <prateeksingh765000017@gmail.com>
+ *
+ * This file is part of Tiled.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the Free
+ * Software Foundation; either version 2 of the License, or (at your option)
+ * any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include <QtCore/qglobal.h>
+
+#if defined(GLTF_LIBRARY)
+#  define GLTFSHARED_EXPORT Q_DECL_EXPORT
+#else
+#  define GLTFSHARED_EXPORT Q_DECL_IMPORT
+#endif
--- a/src/plugins/gltf/gltfexporter.cpp
+++ b/src/plugins/gltf/gltfexporter.cpp
@@ -1,0 +1,74 @@
+/*
+ * glTF Tiled Plugin
+ * Copyright 2026, Prateek Singh <prateeksingh765000017@gmail.com>
+ *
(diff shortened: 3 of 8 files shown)


PrateekSingh070:
Small follow-up pushed on this draft PR: I split the write path into a dedicated GltfExporter helper so the plugin entry point stays minimal and future cgltf-based export logic can be added behind that interface.

PrateekSingh070:
Another small follow-up pushed: the plugin now implements outputFiles() and centralizes .gltf/.glb container detection in GltfExporter, so expected output planning is in place before actual writer logic lands.

PrateekSingh070:
Pushed another incremental step: .gltf export now writes a minimal placeholder output (JSON + empty .bin) through SaveFile, while .glb still returns a clear not-implemented message. This keeps the PR small but makes the write path concrete.

PrateekSingh070:
Pushed another incremental update: placeholder .gltf output now includes Tiled map metadata in root extras (orientation, map/tile dimensions, infinite flag, plus stagger/hex fields where relevant).

PrateekSingh070:
Small follow-up pushed: placeholder output now includes a named root node and scene entry (derived from output file base name), so the generated glTF has a minimal but inspectable node hierarchy in viewers.

PrateekSingh070:
Follow-up pushed: root extras now includes an export block with placeholder metadata (container, placeholder, exportedAtUtc) so generated files are self-describing while the exporter is still in early stages.

PrateekSingh070:
Another incremental update pushed: placeholder output now includes explicit empty glTF resource tables (bufferViews, accessors, materials, meshes) and marks extras.export.hasGeometry = false, making the file structure ready for the next geometry-wiring steps.

PrateekSingh070:
Milestone update: placeholder export now emits a concrete fixed quad mesh (positions/uvs/indices) with real .bin data and connected bufferViews/accessors/materials/meshes in the generated .gltf. This is still fixed geometry, before map traversal logic.

PrateekSingh070:
Incremental follow-up pushed: the placeholder quad is now sized from map->tileWidth() / map->tileHeight() (with fallback handling), so generated geometry reflects map dimensions even before tile-layer traversal is added.

PrateekSingh070:
Hi @bjorn,

Quick status update on the glTF exporter draft work:

- I’ve been intentionally building this in small reviewable steps (plugin scaffold -> exporter helper -> placeholder glTF/bin output -> basic mesh wiring).
- Current output is still placeholder-level by design, but now includes map metadata and a minimal map-aware quad path to validate structure incrementally.
- I’m avoiding big jumps so each step can be reviewed/refined before moving to full tile-layer traversal.

If you’d like, I can pause at this stage and split the next work into whichever order you prefer (for example: first non-empty tile placement vs. full orthogonal layer traversal).

Thanks.

PrateekSingh070:
Follow-up pushed: placeholder quad node translation is now derived from map content by locating the first non-empty cell in visible tile layers and translating the mesh node to that tile position (still a single-quad placeholder, before full traversal).

PrateekSingh070:
Milestone update: exporter now emits one quad per non-empty cell for the first visible non-empty tile layer (orthogonal-only scope for now). This replaces the single-quad placeholder while keeping UV/material behavior intentionally simple for incremental review.

PrateekSingh070:
Milestone update: UVs are now derived per tile from tileset atlas parameters (tileId, columns, margin, spacing, image size) for the first visible non-empty orthogonal tile layer. This keeps one-layer scope, but moves output from generic UVs toward recognizable tile mapping.

PrateekSingh070:
Addressed one review concern: src/plugins/gltf/plugin.json now sets defaultEnable to alse so the exporter is not exposed by default while implementation is still in progress.

PR #4486, opened 2026-03-31
Fix: prevent whitespaces-only property names by trimming input

Praagya26:
Fix: Prevent whitespace-only property names

This PR improves validation when renaming properties by ensuring that user input is properly sanitized before being applied.

### **Problem**
Previously, the rename dialog accepted names that:
- Contained only whitespace (e.g. "   ")
- Had leading/trailing spaces (e.g. "  name  ")
- Resulted in unintended or inconsistent property names

This could lead to:
- Invisible or confusing property names (whitespace-only)
- Duplicate-like names that differ only by spaces
- Poor UX when managing properties

**### Solution:**
- Trim the input string using `QString::trimmed()`
- Reject names that become empty after trimming
- Prevent redundant renames when the trimmed name matches the existing name
- Apply the trimmed value for valid renames

**### Examples:**
 Input                  Previous Behavior        New Behavior        

 `"   "`                    Accepted                       Rejected        
`"  name  "`           Saved as `"  name  "`     Saved as `"name"`  
`"name"` (same)   Renamed again             Ignored           
`"newName"`        Renamed                      Renamed 
 
**### Impact:**
- Improves data consistency
- Prevents invalid or misleading property names
- Enhances overall user experience

This complements my previous PR #4481 related to property renaming and ensures more consistent behavior.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 ### Unreleased
 
 * Added 'Collapse All' action and 'Only Expand to Current' mode to Project view (with rhythmcache, #4346)
+* Added whitespace trimming for custom property names (with Praagya26, #4486)
 * Reduced animated tile marker opacity during terrain editing (by Huy Vũ, #4449)
 * Fixed ability to change properties after deselecting the current object (#4440)
 * Fixed Properties view getting stuck with updates disabled (#4506)
--- a/src/tiled/propertieswidget.cpp
+++ b/src/tiled/propertieswidget.cpp
@@ -2904,22 +2904,15 @@ void PropertiesWidget::renameProperty(const QString &name)
     dialog->setTextValue(name);
     dialog->setWindowTitle(QCoreApplication::translate("Tiled::PropertiesDock", "Rename Property"));
 
-    connect(dialog, &QInputDialog::textValueSelected, this, [=](const QString &newName) {
-        QString trimmedName = newName.trimmed();
-
-        if (trimmedName.isEmpty())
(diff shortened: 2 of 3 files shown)


kunal649:
Nice catch Pragya. I noticed we have same whitespace inconsistency in `RenameTileset` in `tilesetchanges.cpp` too, 
[https://github.com/Praagya26/tiled/blob/f13f3f1e4befde3596aee9ea18ba88f16b595434/src/tiled/tilesetchanges.cpp#L33](https://github.com/Praagya26/tiled/blob/f13f3f1e4befde3596aee9ea18ba88f16b595434/src/tiled/tilesetchanges.cpp#L33)

would be great if u add `.trimmed( )` guard there as well : )

Edit : I found same in Edit commands dialog, (what if we have a centralized base class for naming validation across the project 🤔)

Praagya26:
Thanks for pointing this out! @kunal649  I'll add the '.trimmed()' guard in 'tilesetchanges.cpp' as well. A centralized base class for naming validation sounds like a great improvement. it would make the codebase more consistent. surely will look for that!

bjorn (approved):
I think this change is good because it just touches the relevant UI rather than generally enforcing this rule in lower level code (like in #4495). However, it only touched the rename property dialog, so I've pushed a commit that also covers the UI flow for adding new properties. It also reduces the diff a little by keeping `newName` for the new name, renaming the untrimmed parameter instead.

Thanks for helping with this, @Praagya26!

PR #4494, opened 2026-04-03
CI: Updated Qt 6 builds to Qt 6.11.1

bjorn:


bjorn:
Even though it compiles fine, there appears to be a new crash when loading JS modules which wasn't there yet with the Qt 6.10.2 builds. So this needs further testing.

bjorn:
It seems like the following script can be used to reproduce the problem:

[slow-quadtree.zip](https://github.com/user-attachments/files/26479271/slow-quadtree.zip)

Probably due to this line:

```js
import { readFileSync } from 'fs';
```

 'fs' is a Node.js built-in that doesn't exist in Qt's JS engine. In Qt 6.10.2, `QJSEngine::importModule()` handles this gracefully with an error. In Qt 6.10.3, it segfaults in `QV4::ExecutionEngine::loadModule()`.

Oval17:
Hi @bjorn 
I went through this , the segfault in `QV4::ExecutionEngine::loadModule() `is triggered when a `.mjs` extension file contains import statements for bare specifiers (e.g. import { readFileSync } from 'fs') as you mentioned abouve.
Qt 6.10.2 returned an error gracefully; Qt 6.10.3 crashes instead.

The approach I want to take(before upgrading to Qt 6.10.3 in CI) is to pre-scan each `.mjs` file for bare specifiers before passing it to `QJSEngine::importModule()`. If any are found, we bail out early with a `Tiled::ERROR` message rather than letting the engine crash:

Module '`slow-quadtree.mjs`' uses bare specifiers which are not supported in Tiled  (Node.js built-ins are unavailable): fs

Does this approach look correct to you?

bjorn:
@Oval17 It really seems like a regression in Qt which should first of all be reported with a small app that reproduces the issue. I think working this around in Tiled is secondary, because we don't need to move to 6.10.3 in a rush and people wouldn't generally have scripts with such problematic lines in them.

Oval17:
@bjorn 
I tried to file bug in QT but I am unable to it says you dont have access/authorization.
I used this link : [bugreports.qt.io](http://bugreports.qt.io/)

Btw this is the format of Issue I found online and have made according to that template:

**Component:** Qt Quick / QML Engine (or Qt QML → QJSEngine)

**Summary:**

QJSEngine::importModule() segfaults on bare specifiers in Qt 6.10.3 (regression from 6.10.2)

**Description:**

In Qt 6.10.2, calling QJSEngine::importModule() on a module that imports
a bare specifier (e.g. a Node.js built-in like 'fs') returns an error
object gracefully.

In Qt 6.10.3, the same call crashes in QV4::ExecutionEngine::loadModule()
with a segmentation fault.

**Minimal reproducer:**

**main.cpp:**

 ```
 #include <QCoreApplication>
  #include <QJSEngine>
  #include <QFile>

  int main(int argc, char *argv[]) {
      QCoreApplication app(argc, argv);
      QJSEngine engine;
      engine.installExtensions(QJSEngine::AllExtensions);
      QJSValue result = engine.importModule("test.mjs");
      qDebug() << result.toString();
      return 0;
  }
```

**test.mjs:**

```
  import { readFileSync } from 'fs';
  export function hello() { return readFileSync('x.txt', 'utf8'); }
```

**Expected:** `importModule()` returns an error object (as in Qt 6.10.2)

**Actual:**   Crash in `QV4::ExecutionEngine::loadModule()`

Affects version: 6.10.3
Works in version: 6.10.2

**Severity:** Regression

bjorn:
I'm not seeing the crash on Qt 6.11.0, but I tracked down the change that introduced it and it might not have been part of that release yet. I've commented about the crash there: https://codereview.qt-project.org/c/qt/qtdeclarative/+/699373/comment/4b9de962_af44b6ec/

bjorn:
Reported as https://qt-project.atlassian.net/browse/QTBUG-145967 (btw, it doesn't crash for me when using `QJSEngine`, I had to use `QQmlEngine`).

I had to remove the broken script on my local machine now that Fedora has updated to Qt 6.10.3, heh.

bjorn:
Updated to jump straight to Qt 6.11.1, since 6.10.3 suffers from QTBUG-145967. This also brings the fix for QTBUG-143708, a possible crash when restoring dock layouts. Rebased onto master, no toolchain changes needed.

PR #4495, opened 2026-04-05
Fix: trim whitespace in tileset name when renaming

Praagya26:
### **Problem:**
1. Renaming a tileset allowed whitespace-only names.
2. Leading and trailing spaces were not trimmed, resulting in inconsistent naming.

### **Fix:**
Applied QString::trimmed() to sanitize input:
In tilesetchanges.cpp using newName.trimmed()
In tilesetdocument.cpp using name.trimmed()

This ensures:
Whitespace-only names are rejected.
Leading and trailing spaces are removed before further processing.

### **Result:**
Prevents invalid or empty tileset names.
Improves consistency and reliability of naming behavior.

### **Future Work:**
As suggested by @kunal649, consider introducing a CENTRALIZED VALIDATION UTILITY.
This would help enforce consistent naming rules across the codebase and reduce duplication.

### DIFF
--- a/src/tiled/tilesetdocument.cpp
+++ b/src/tiled/tilesetdocument.cpp
@@ -297,7 +297,7 @@ void TilesetDocument::removeMapDocument(MapDocument *mapDocument)
 
 void TilesetDocument::setTilesetName(const QString &name)
 {
-    mTileset->setName(name);
+    mTileset->setName(name.trimmed());
     emit tilesetNameChanged(mTileset.data());
 
     for (MapDocument *mapDocument : mapDocuments())
--- a/src/tiled/tilesetchanges.cpp
+++ b/src/tiled/tilesetchanges.cpp
@@ -36,7 +36,7 @@ RenameTileset::RenameTileset(TilesetDocument *tilesetDocument,
                                                "Change Tileset Name"))
     , mTilesetDocument(tilesetDocument)
     , mOldName(tilesetDocument->tileset()->name())
-    , mNewName(newName)
+    , mNewName(newName.trimmed())
 {}
 
 void RenameTileset::undo()


kunal649:
😅 you can actually just push extra commits to your existing branch for #4486. It’ll update that PR automatically so we can keep all these whitespace fixes in one place for Bjorn to review. It saves everyone some clicks if we group them! (let's see what bjorn's take on naming validation idea, though implementing that won't be a cup of tea )

src/tiled/tilesetdocument.cpp
```diff
@@ -297,7 +297,7 @@ void TilesetDocument::removeMapDocument(MapDocument *mapDocument)
 
 void TilesetDocument::setTilesetName(const QString &name)
 {
-    mTileset->setName(name);
+    mTileset->setName(name.trimmed());
```
bjorn, line 300: Since `setTilesetName` is only called from `RenameTileset`, this change seems superfluous. The only thing it changes, is that the previous name will also be trimmed on undoing a tileset rename. But I don't think that's what we want.

bjorn (changes requested):
> Whitespace-only names are rejected

This change does nothing to reject whitespace-only names. For that, the trim would need to be applied before constructing a `RenameTileset` command, for example before the `isEmpty` check in `NewTilesetDialog::updateOkButton`.

But as-is, you can no longer type a tileset name with a space in the Name field in the Properties view, for example, because the space is immediately trimmed. The automatic trimming of the name might also be unexpected when setting the name from scripts.

So I'd suggest to implement this only as a UI change, so that it can properly respond:

* In `NewTilesetDialog::updateOkButton`, check whether the string is empty after trimming. Probably also in `NewTilesetDialog::nameEdited`.
* Trim the name in `NewTilesetDialog::tryAccept`.
* If we want to disallow trailing and leading whitespace in the Properties view, we should probably use a `QValidator`. Currently the `StringProperty` does not support this, but it could support setting a custom `TrimValidator` that accepts trailing whitespace as `Intermediate` and trims it in `fixup`.

PR #4499, opened 2026-04-12
Add pixi-tiledmap support details to documentation

riebel:
Added information about pixi-tiledmap as a loader and renderer for Tiled Maps in Pixi.JS.


bjorn:
Thanks for this useful addition and for maintaining it!

PR #4504, opened 2026-04-18
Fix transparent tiles being saved with tile ID instead of 0

Lillious:
## Summary
Transparent tiles in tilesets are now correctly serialized as GID 0 (empty) instead of their actual tile ID when
saving maps.

## Problem
When a tile in a tileset is completely transparent (all pixels have zero alpha), painting with it and saving the
map would incorrectly save that tile's ID instead of 0, resulting in invisible tiles appearing on the map.

## Solution
Added a transparency check in `GidMapper::cellToGid()` that:
- Extracts each tile's region from a tileset image
- Checks if all pixels are completely transparent (alpha = 0)
- Returns GID 0 for fully transparent tiles instead of their actual ID

This fix applies to all export formats (JSON, TMX, etc.) since they all use the GID mapper for serialization.

## Testing
- Paint with a transparent tile from a tileset
- Save the map to JSON
- Verify that the painted areas have GID 0 instead of the transparent tile's ID

### DIFF
--- a/src/libtiled/gidmapper.cpp
+++ b/src/libtiled/gidmapper.cpp
@@ -133,6 +133,27 @@ unsigned GidMapper::cellToGid(const Cell &cell) const
     if (i == i_end) // tileset not found
         return 0;
 
+    // Check if the tile is completely transparent, treat as empty
+    Tile *tile = tileset->tileAt(cell.tileId());
+    if (tile) {
+        const QPixmap &img = tile->image();
+        if (!img.isNull() && img.hasAlphaChannel()) {
+            // Extract just this tile's region and check if all pixels are transparent
+            QPixmap tilePixmap = img.copy(tile->imageRect());
+            QImage imgData = tilePixmap.toImage();
+
+            bool allTransparent = true;
+            for (int y = 0; y < imgData.height() && allTransparent; ++y) {
+                for (int x = 0; x < imgData.width() && allTransparent; ++x) {
+                    if (qAlpha(imgData.pixel(x, y)) > 0)
+                        allTransparent = false;
+                }
+            }
+            if (allTransparent)
+                return 0;
+        }
+    }
+
     unsigned gid = i.key() + cell.tileId();
     if (cell.flippedHorizontally())
         gid |= FlippedHorizontallyFlag;


dogboydog:
This doesn't seem like it's necessarily desirable for all users. It's not "incorrect" to save transparent tiles with their actual IDs. After all,  they are tiles in the tileset. There may be users that use particular transparent tiles to have a special meaning in their engine. This PR would break that. I think a better solution for you would be to write a script that removes any transparent tiles from your map.

dogboydog:
I did not mean to close it, sorry

Lillious:
> This doesn't seem like it's necessarily desirable for all users. It's not "incorrect" to save transparent tiles with their actual IDs. After all,  they are tiles in the tileset. There may be users that use particular transparent tiles to have a special meaning in their engine. This PR would break that. I think a better solution for you would be to write a script that removes any transparent tiles from your map. 

It may not seem "incorrect" but I do not see a valid reason for placed tiles with no actual content to have tile IDs. I do not believe this change would actually break anything since a tile ID of 0 would result in the same exact visual and functional result as one with an ID but with no content.

I'm open to having a discussion about how this would break something but I simply don't see it.

dogboydog:
Some people use a layer of tiles as "data tiles" where the tile in the tileset have custom properties used to indicate logic in their game. It's possible someone could use a transparent tile for this. I just don't think we can assume transparent tiles shouldn't be written out to the map 

Sorry the close PR button is easy to hit on mobile

Lillious:
> Some people use a layer of tiles as "data tiles" where the tile in the tileset have custom properties used to indicate logic in their game. It's possible someone could use a transparent tile for this. I just don't think we can assume transparent tiles shouldn't be written out to the map 
> 
> 
> 
> Sorry the close PR button is easy to hit on mobile

That could be the case but you simply can't identify where this tile is because it has no content. There is no way for someone to know where it is visually because the editor doesn't even show it as a selectable tile.

Lillious:
> > Some people use a layer of tiles as "data tiles" where the tile in the tileset have custom properties used to indicate logic in their game. It's possible someone could use a transparent tile for this. I just don't think we can assume transparent tiles shouldn't be written out to the map 
> > 
> > 
> > 
> > Sorry the close PR button is easy to hit on mobile
> 
> That could be the case but you simply can't identify where this tile is because it has no content. There is no way for someone to know where it is visually because the editor doesn't even show it as a selectable tile.

For example, if you place a tile with no content and you use the selection tool, drag the selection area over it, copy and paste for placement, that tile with no content is not included.

zlago:
consider this: a tileset is edited to use a transparent image on accident
it could be any one of reasons, image editor autosave, messing up which layers get exported, deleting the wrong tile(s) before saving the image, any of these could happen
you probably would not want to lose progress because you did not notice your tiled map has missing tiles before saving it, especially if you did not set up version control (or just dont commit very frequently, i certainly dont)
also consider tilesets with an invalid image (because you havent installed qaseprite yet), you really wouldnt want forgetting a plugin to erase the entire map

Lillious:
> consider this: a tileset is edited to use a transparent image on accident
> 
> it could be any one of reasons, image editor autosave, messing up which layers get exported, deleting the wrong tile(s) before saving the image, any of these could happen
> 
> you probably would not want to lose progress because you did not notice your tiled map has missing tiles before saving it, especially if you did not set up version control (or just dont commit very frequently, i certainly dont)
> 
> also consider tilesets with an invalid image (because you havent installed qaseprite yet), you really wouldnt want forgetting a plugin to erase the entire map

I just don't see someone accidentally doing this as it requires a chain of incorrect steps. Plus the undo functionality exists. it doesn't remove invalid images, just completely transparent tiles which the program already assigns as nothing.

If the selection tool can't detect and copy/paste these transparent tiles why does saving treat them differently?

bjorn:
I'm closing this because even if the idea is great, the solution can't possibly be to analyze pixel data in `cellToGid`, that's just going to be way too slow (it's even copying each tile image in order to analyze it and it's not doing any caching, but that is not to say I would find this approach acceptable when caching was added).

I think a proper solution involves making sure that such empty tiles do not become placeable tiles in the first place. We can have a tileset option (it could even be enabled by default) that skips the creation of tiles when it detects no contents at that location in the tileset image. Another solution would be to implement this as an export option.

PR #4505, opened 2026-04-19
Add agent-oriented Tiled bridge and orchestrator

Gyejoon:
## Summary
- add a standalone `tiledagentbridge` C++ bridge that reuses `libtiled` for map and tileset open/save, typed edit commands, validation, and undo/redo over JSON-RPC stdio
- add a Node.js/TypeScript orchestrator with workspace/session management, HTTP/JSON endpoints, MCP tools, and a minimal debug UI
- add bridge and orchestrator tests, including a real child-process integration test against the bridge binary

## Test Plan
- [x] `cd tools/tiled-agent-orchestrator && npm run build`
- [x] `cd tools/tiled-agent-orchestrator && npm test`
- [x] `ctest --test-dir build/tiledagentbridge --output-on-failure`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,6 +53,10 @@ util/java/tmxviewer-java/target
 util/java/libtiled-java/target
 /util/java/target/
 
+# Node.js
+node_modules/
+tools/tiled-agent-orchestrator/dist/
+
 # Linux perf/debugging
 gmon.out
 
--- a/src/tiledagentbridge/CMakeLists.txt
+++ b/src/tiledagentbridge/CMakeLists.txt
@@ -1,0 +1,49 @@
+cmake_minimum_required(VERSION 3.21)
+
+project(tiledagentbridge LANGUAGES CXX)
+
+set(CMAKE_AUTOMOC ON)
+set(CMAKE_AUTORCC ON)
+set(CMAKE_AUTOUIC ON)
+set(CMAKE_CXX_STANDARD 17)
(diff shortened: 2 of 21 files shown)


dogboydog:
I just want to note my opposition to adding any gen AI focused features to Tiled. I am just a contributor though so it's not up to just me.

bjorn:
This approach seems to duplicate a large amount of editor functionality (the gist of this PR seems to be the `bridgeengine.cpp` at ~1000 LoC) which would become much larger as further functionality is added.

Apart from this duplication, the LLM functioning essentially as a parallel editor may not be ideal for users who only want to offload part of their work to the LLM? I'm not entirely sure, but at least it would need to know about unsaved changes in that case.

I think we can avoid lots of duplication as well as provide an API that can more easily be kept up to date by basing the MCP on the existing scripting API. I'm not sure however how we can enable interaction with an existing Tiled instance. Maybe by implementing the [Streamable HTTP](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#streamable-http) transport?

While I do think it's interesting to enable LLMs to control Tiled to automate parts of user's workflows, it's not something I will personally spend time on at the moment. But I'm fine with others experimenting with this.

bjorn:
Also, just adding a reference here, but I think #4439 already did the approach of implementing an MCP as Tiled extension, but they also added a big chunk of Python code with the same problems as in `bridgeengine.cpp`, with it essentially duplicating all editor functionality.

PR #4515, opened 2026-05-02
Added mapreader tileset tests

ryankurz0303:
This is for a project for my software testing college class.


bjorn:
@ryankurz0303 It's a good idea to introduce tests for this, but I think the tests are missing some kind of verification that the embedded or external tilesets are actually loaded. As-is, all tests only verify that the map itself was loaded. Would you be interested in expanding the verification in the new tests?

PR #4520, opened 2026-05-17
Added editablemap property to tiledquick's maploader

UltraDagon:
This adds an editablemap property to maploader, through which you can access the map that maploader currently has. Eventually, this should be able to replace the existing map property, the mapref class, and multiple functions within the class, if my understanding is correct.

Added sizeChanged notify signal to make some editablemap q_properties available in tiledquick/qml/main.qml

Moved the Map EditableMap::*map() {} function inside of editablemap's class body to avoid a dll import warning. Let me know if this needs to be reverted and if there is a better method.


bjorn:
@UltraDagon This PR needs a rebase on the latest `master` branch now that your previous PR is merged. Feel free to ask for help on Discord if you're unsure how to do this.

src/tiled/editablemap.h
```diff
@@ -211,12 +211,17 @@ class EditableMap final : public EditableAsset
...
+    {
+        return static_cast<Map*>(object());
+    }
+
     MapDocument *mapDocument() const;
 
     QSharedPointer<Document> createDocument() override;
 
 signals:
+    void sizeChanged();
```
bjorn: Eventually we will need such signals, but is it really necessary to add them even if you're not emitting the signal anywhere? I think the lack of a signal is just something the IDE warns about, but shouldn't make the property unusable.

src/libtiledquick/maploader.cpp
```diff
@@ -56,6 +57,8 @@ void MapLoader::setSource(const QUrl &source)
     const bool errorDiff = m_error != error;
 
     m_map = std::move(map);
+    // Currently duplicates m_map, once mapref and mapitem are replaced m_map should be moved rather than cloned
```
bjorn: I think we can do this as part of this PR? The `EditableMap` instance should be able to replace the `Map` instance, since it just wraps it.

src/libtiledquick/maploader.cpp
```diff
@@ -30,6 +30,7 @@ using namespace TiledQuick;
 MapLoader::MapLoader(QObject *parent)
     : QObject(parent)
     , m_map(nullptr)
+    , m_editableMap(nullptr)
```
bjorn: A `std::unique_ptr` will be `nullptr` by default so let's leave this out, you can remove the `m_map` line as well.

src/tiled/editablemap.h
```diff
@@ -211,12 +211,17 @@ class EditableMap final : public EditableAsset
     void setSelectedLayers(const QList<QObject*> &layers);
     void setSelectedObjects(const QList<QObject*> &objects);
 
-    Map *map() const;
+    Map *map() const
+    {
+        return static_cast<Map*>(object());
```
bjorn: Hmm, not immediately sure why this would be necessary, it shouldn't make any difference.

bjorn (commented):
Yeah it looks like the rebase went through fine. I left some comments.

src/tiled/editablemap.h
UltraDagon: I removed the NOTIFY parts of the Q_PROPERTY and the sizedChanged signal. The map still works upon changing tmx files, but does output this to the terminal:
`QQmlExpression: Expression qrc:/main.qml:175:17 depends on non-bindable properties:
    Tiled::EditableMap::tileWidth
    Tiled::EditableMap::tileHeight`

src/tiled/editablemap.h
UltraDagon: I have (nearly) restored it to its original state:
`Map *map() const;` -> `inline Map *map() const;`

Without adding the inline keyword, it would output this warning during compilation:
> warning: 'Tiled::Map* Tiled::EditableMap::map() const' redeclared without dllimport attribute after being referenced with dll linkage
411 | inline Map *EditableMap::map() const

src/tiled/editablemap.h
bjorn: Yeah, we will need to add the signal, but I think it's better to have this warning than to silence it with a non-functional one. :-)

src/tiled/editablemap.h
bjorn: Alright, Windows-only issue I think, but it probably makes sense now that the class is marked for export. Indeed in such cases let's just add the `inline` keyword then.

bjorn:
> Eventually, this should be able to replace the existing map property and the mapref and mapitem classes, if my understanding is correct.

Yes, I think it should, but please don't wait with that but try to do this as part of this PR. Remove `MapRef` and make the `MapItem` take an `EditableMap*` instead.

src/libtiledquick/mapitem.h
```diff
@@ -43,16 +43,19 @@ class TILEDQUICK_SHARED_EXPORT MapItem : public QQuickItem
 {
     Q_OBJECT
 
-    Q_PROPERTY(TiledQuick::MapRef map READ map WRITE setMap RESET unsetMap NOTIFY mapChanged)
+    Q_PROPERTY(Tiled::Map *map READ map NOTIFY mapChanged)
+    Q_PROPERTY(Tiled::EditableMap *editableMap READ editableMap WRITE setEditableMap RESET unsetEditableMap NOTIFY mapChanged)
```
bjorn: Please just call this property `map` and remove the `Q_PROPERTY` for `Tiled::Map*`, since that type is not known to the QML engine anyway. Just having `mapChanged` should be enough, we don't also need `editableMapChanged`.

bjorn (changes requested):
@UltraDagon This PR is still not properly rebased on the latest `master`. After rebase, there should not be merge commits like "Merge branch 'master' into maploader-editablemap" (4fcc6d9465b28a6e09cfdff7087628822d788dda) on your PR branch.

Since you seem to be running into problems with this, feel free to ping me on Discord if you need help, maybe we can go over this together with screen sharing.

UltraDagon:
Follow up on our earlier discussion about `Depends { name: "libtilededitor" }` vs `Depends { name: "libtiled" }` in tiledquickplugin.qbs:
It seems like tiledquickplugin does need libtilededitor because of libtiledquick plugin needing maploader.h, which needs editablemap.h:
`tiled/src/libtiledquick/mapitem.h:24:10: fatal error: editablemap.h: No such file or directory`
I'm not sure if there is a way to get around this. For now, I'm going to leave it as-is to allow tiledquick to compile.

bjorn:
> It seems like tiledquickplugin does need libtilededitor because of libtiledquick plugin needing maploader.h, which needs editablemap.h

Alright, so that's a dependency we should export from `libtiledquick`. Instead of adding an explicit `libtilededitor` dependency  to `tiledquickplugin`, you can try adding `Depends { name: "libtilededitor" }` to the `Export` item in `libtiledquick.qbs`.

PR #4529, opened 2026-05-23
Fix empty `className` for template-instantiated objects

### LINKED ISSUE #4528: `className` is empty for template-initialized objects
### Describe the bug

Given an object A initialized by a template with a non-empty class name, `A.className` is empty.

### Steps to reproduce

1. Create an object, give it any class name.
2. Save it as a template.
3. Instantiate the template (by dragging the object from the Project view)
4. Attempt to access the object via the scripting API and check its `className` property

### Expected behavior

In the case of the class name remaining unchanged, the template initialized object should have a `className` property equal to the class name of the template, as is shown in the properties view for the object as placeholder text.

### OS version

openSUSE Tumbleweed Linux x64

### Tiled version

1.12.1

MatusGuy:
This adds a new method to `MapObject`, `resolvedClassName`. It falls back to the template classname in case the object classname is empty.

Fixes #4528

However, the new method is missing documentation.


bjorn:
@MatusGuy Thanks for your help with adding this!

PR #4530, opened 2026-05-27
(WIP) TiledQuick New Hardware Renderer Preference Setting

UltraDagon:
Adds a preference to the interface preference tab to toggle on/off the new TiledQuick hardware renderer.

Currently, enabling the preference sets the viewport in MapView to be a QWidget container of a QQuickView object that renders a test QtQuick QML file. It may be possible to use a QQuickWidget instead, though I ran into some issues doing that and opted for the container method.

Most of the code uses the existing OpenGL preference code as the basis, so it's definitely possible that my MapEditor and MapView additions are not optimal.


bjorn:
First of all, this change needs to be rebased on my latest `master` branch.

But second, `MapView::setUseNewHardwareRenderer` is the wrong level to apply the setting. Because the `MapView` instance itself is the `QGraphicsView` that we're trying to replace. You need to be looking into replacing the `MapView` instances (managed by `MapEditor`) with `QQuickWidget` instances.

Because there is so much existing code relying directly on `MapView` instances though, I think it would be fine to keep those around in parallel, and use the "new hardware renderer" option first of all only to control which of the views is visible in the `mWidgetStack`.

We could instantiate a `QQuickWidget` for each map when the option is enabled, keeping track of it with a `QHash<MapDocument*, QQuickWidget*> mQuickViewForMap` in parallel to the existing `mWidgetForMap`. Alternatively, I think we could try to use a single `QQuickWidget` instance and switch the currently displayed map from within Qt Quick based which map is currently selected. The latter option would consume much less resources when many maps are open, which making map switching a little slower (if it's too slow though, a single `QQuickWidget` could also maintain multiple `MapItem` instances internally, making only the current one visible).

UltraDagon:
I messed up the rebase and think it will be easier to make a new pull request, since there was only one commit.

Duplicate PR: #4537

bjorn:
> I messed up the rebase and think it will be easier to make a new pull request, since there was only one commit.

For next time, just note that there is no need to open a new pull request. Once you have the rebase done properly, or cherry-picked or redid the change or whatever, just force-push to the PR branch (`nhr/preference-setting` in this case) to update the PR.

PR #4534, opened 2026-05-28
Add configurable world grid

kanishka0411:
Adds an optional world grid for loaded worlds.

The grid size is stored on the world as `gridWidth` and `gridHeight`, so it's saved in the `.world` file, and can be set from the World Properties dialog. Visibility is kept as a user preference via View → Show World Grid, similar to the existing Show Grid action.

The grid is drawn in `MapScene::drawBackground` as a dashed, semi-transparent world-space guide. Since it's drawn as scene background, it can show through transparent parts of maps. Kept this as a simple first version so the world grid stays visually distinct from the per-map tile grid.


src/libtiled/world.cpp
```diff
@@ -360,6 +368,10 @@ bool World::save(World &world, QString *errorString)
         document.insert(QLatin1String("properties"), properties);
     document.insert(QLatin1String("type"), QLatin1String("world"));
     document.insert(QLatin1String("onlyShowAdjacentMaps"), world.onlyShowAdjacentMaps);
+    if (world.gridWidth > 0 && world.gridHeight > 0) {
```
bjorn: Let's store both if either of them is set.

src/tiled/worldpropertiesdialog.cpp
```diff
@@ -42,9 +46,47 @@ WorldPropertiesDialog::WorldPropertiesDialog(WorldDocumentPtr world, QWidget *pa
...
+
+    // Avoid one undo step per keystroke while typing a value.
+    ui->gridWidthSpinBox->setKeyboardTracking(false);
+    ui->gridHeightSpinBox->setKeyboardTracking(false);
+
+    refreshGridFromWorld();
+
+    connect(ui->gridWidthSpinBox, qOverload<int>(&QSpinBox::valueChanged),
+            this, &WorldPropertiesDialog::pushGridSizeCommand);
+    connect(ui->gridHeightSpinBox, qOverload<int>(&QSpinBox::valueChanged),
```
bjorn: You don't need `qOverload` here, since there is only one `valueChanged` signature with the deprecated QString-version disabled.

bjorn (commented):
I think this looks pretty good. I would suggest to use a `QSize` to store the grid size, which generally shortens the code a little, but I realize I didn't do this myself either with the `tileWidth` and `tileHeight` values.

It would be nice if you could adjust the `WorldMoveMapTool` to also use this new world grid for snapping.

src/libtiled/world.cpp
kanishka0411: updated 👍

src/tiled/worldpropertiesdialog.cpp
kanishka0411: ah right, removed it

src/tiled/mainwindow.ui
```diff
@@ -338,6 +340,22 @@
...
+   <property name="text">
+    <string>Show &amp;World Grid</string>
+   </property>
+  </action>
+  <action name="actionSnapToWorldGrid">
+   <property name="checkable">
+    <bool>true</bool>
+   </property>
+   <property name="text">
+    <string>Snap to World Grid</string>
```
bjorn: Let's move this to the "Snapping" menu, with a separator between this and the other snap options, but call it "Snap Maps to World Grid" for clarity. The View menu is getting a little too large.

src/tiled/changeworld.h
```diff
@@ -89,6 +90,22 @@ class SetMapRectCommand : public QUndoCommand
     QRect mPreviousRect;
 };
 
+class SetWorldGridCommand : public QUndoCommand
+{
+public:
+    SetWorldGridCommand(WorldDocument *worldDocument, QSize gridSize);
+
+    void undo() override { setGridSize(mPreviousSize); }
+    void redo() override { setGridSize(mSize); }
```
bjorn, line 99: I've noticed each change to the world grid triggers a new undo command. You've alleviated this somewhat by calling `setKeyboardTracking(false)`, but this doesn't address the same problem when using the scroll wheel or up/down buttons to change the grid values.

I think the fix should be similar to `SetMapRectCommand` above, to give this command an ID and implement `mergeWith`.

src/tiled/worldmovemaptool.cpp
```diff
@@ -87,16 +87,9 @@ void WorldMoveMapTool::keyPressed(QKeyEvent *event)
     if (!document || !mapCanBeMoved(document) || mDraggingMap)
         return;
 
-    const bool moveFast = modifiers & Qt::ShiftModifier;
```
bjorn, line 90: Please keep this variable and check `moveFast` below. The variable helps keep the code readable without needing comments.

src/tiled/worldpropertiesdialog.cpp
```diff
@@ -42,9 +46,46 @@ WorldPropertiesDialog::WorldPropertiesDialog(WorldDocumentPtr world, QWidget *pa
     ui->propertiesWidget->propertiesView()->widget()->setContentsMargins(0, halfSpacing, 0, halfSpacing);
 
     ui->propertiesWidget->setDocument(mWorldDocument.get());
+
+    // Avoid one undo step per keystroke while typing a value.
+    ui->gridWidthSpinBox->setKeyboardTracking(false);
+    ui->gridHeightSpinBox->setKeyboardTracking(false);
```
bjorn: I think rather than setting up this UI entirely manually, it would be nice to use the new properties framework, same way as done in the `ProjectPropertiesDialog`, but we can also leave this as-is for now, I can try to transition this dialog later.

bjorn (changes requested):
Works nicely and I think you've made some nice choices, addressing some small existing issues as well. I made a few comments on minor points, which I can also help with tomorrow.

src/tiled/mainwindow.ui
kanishka0411: Moved it into the Snapping menu with a separator and renamed it "Snap Maps to World Grid"

src/tiled/changeworld.h
kanishka0411: Done. Gave it an id and a mergeWith like SetMapRectCommand

src/tiled/worldpropertiesdialog.cpp
kanishka0411: Sounds good, I'll leave it as-is for now then👍

bjorn: I tried this and while it reduced the code a little bit, I don't really like the result in the UI, so I'll leave this be for now. It looked like this:

image

bjorn: Alright, so I've just used two `IntProperty` instances now. :-)

image

bjorn (approved):

bjorn:
@kanishka0411 Thanks for this improvement! Looking forward to your next PR! :-)

PR #4537, opened 2026-05-30
TiledQuick New Hardware Renderer Preference Setting

UltraDagon:
Adds a preference to the interface preference tab to toggle on/off the new TiledQuick hardware renderer. This feature currently uses test_renderer.qml. In another PR, TiledQuick will be used to render the maps.

Adds a QHash to the MapEditor class named mQuickWidgetForMap, which is similar to mWidgetForMap, that stores QQuickWidgets.

Adds multiple if/else statements to change behavior depending on whether or not the preference is enabled. These can later be contained in an interface called ViewInterface that houses both MapViews and QQuickWidgets.

Upon toggling the preference, it switches out mWidgetStack's loaded MapViews/QQuickWidgets for their counterparts.

(This is a duplicate of #4530, which has been closed)

### DIFF
--- a/src/tiled/libtilededitor.qbs
+++ b/src/tiled/libtilededitor.qbs
@@ -620,7 +620,6 @@ DynamicLibrary {
     }
 
     Export {
-
         Depends { name: "cpp" }
         Depends { name: "libtiled" }
         Depends { name: "qtsingleapplication" }
--- a/src/tiled/libtilededitor.qbs
+++ b/src/tiled/libtilededitor.qbs
@@ -13,6 +13,7 @@ DynamicLibrary {
     Depends { name: "Qt.openglwidgets"; condition: Qt.core.versionMajor >= 6; required: false }
     Depends { name: "Qt.dbus"; condition: qbs.targetOS.contains("linux") && project.dbus; required: false }
     Depends { name: "Qt.gui-private"; condition: qbs.targetOS.contains("windows") && Qt.core.versionMajor >= 6 }
+    Depends { name: "Qt.quickwidgets" }
 
     cpp.includePaths: {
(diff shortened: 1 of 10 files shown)


UltraDagon:
Copying over Bjorn's comment from #4530
> First of all, this change needs to be rebased on my latest master branch.
> 
> But second, MapView::setUseNewHardwareRenderer is the wrong level to apply the setting. Because the MapView instance itself is the QGraphicsView that we're trying to replace. You need to be looking into replacing the MapView instances (managed by MapEditor) with QQuickWidget instances.
> 
> Because there is so much existing code relying directly on MapView instances though, I think it would be fine to keep those around in parallel, and use the "new hardware renderer" option first of all only to control which of the views is visible in the mWidgetStack.
> 
> We could instantiate a QQuickWidget for each map when the option is enabled, keeping track of it with a QHash<MapDocument*, QQuickWidget*> mQuickViewForMap in parallel to the existing mWidgetForMap. Alternatively, I think we could try to use a single QQuickWidget instance and switch the currently displayed map from within Qt Quick based which map is currently selected. The latter option would consume much less resources when many maps are open, which making map switching a little slower (if it's too slow though, a single QQuickWidget could also maintain multiple MapItem instances internally, making only the current one visible).

UltraDagon:
All should be in the right place now, aside from the interface needed for the MapViews and QuickWidgets.

For now, I named all of the QQuickWidget objects quickWidget as opposed to quickView to avoid confusion between the QtQuickWidgets/QtQuickWidget object and the QtQuick/QQuickView object. If quickView is the preferred name, I am fine with renaming all quickWidget occurrences to quickView.

PR #4540, opened 2026-06-03
add preference for repeated shortcuts toggling to previous tool

### LINKED ISSUE #4490: Add option to disable keyboard shortcuts switching to previous tool

The new feature introduced in #4280 is rarely useful to me and I often trigger it inadvertently. I now have to look at (or remember) which tool I currently have selected whenever I decide to use any tool.

I would like to have an option to disable the feature introduced in #4280, and I'm willing to work on it.

_No response_

dogboydog:
Fix  #4490

 I defaulted it to false  because I don't personally prefer this behavior either and we had a few users complain, so I thought most  users familiar with the app might expect it to be off . I can change it to default to true if desired

### DIFF
--- a/src/tiled/preferences.cpp
+++ b/src/tiled/preferences.cpp
@@ -691,6 +691,17 @@ void Preferences::setNaturalSorting(bool enabled)
     emit naturalSortingChanged(enabled);
 }
 
+bool Preferences::repeatShortcutForPreviousTool() const
+{
+    return get("Project/RepeatShortcutForPreviousTool", false);
+}
+
+void Preferences::setRepeatShortcutForPreviousTool(bool enabled)
+{
+    setValue(QLatin1String("Project/RepeatShortcutForPreviousTool"), enabled);
+    emit naturalSortingChanged(enabled);
+}
+
 void Preferences::addToRecentFileList(const QString &fileName, QStringList& files)
 {
     // Remember the file by its absolute file path (not the canonical one,
--- a/src/tiled/preferences.cpp
+++ b/src/tiled/preferences.cpp
@@ -693,13 +693,12 @@ void Preferences::setNaturalSorting(bool enabled)
 
 bool Preferences::repeatShortcutForPreviousTool() const
 {
-    return get("Project/RepeatShortcutForPreviousTool", false);
+    return get("Interface/RepeatShortcutForPreviousTool", false);
 }
 
 void Preferences::setRepeatShortcutForPreviousTool(bool enabled)
 {
-    setValue(QLatin1String("Project/RepeatShortcutForPreviousTool"), enabled);
-    emit naturalSortingChanged(enabled);
+    setValue(QLatin1String("Interface/RepeatShortcutForPreviousTool"), enabled);
 }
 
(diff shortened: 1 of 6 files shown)


src/tiled/preferences.cpp
```diff
@@ -691,6 +691,17 @@ void Preferences::setNaturalSorting(bool enabled)
     emit naturalSortingChanged(enabled);
 }
 
+bool Preferences::repeatShortcutForPreviousTool() const
+{
+    return get("Project/RepeatShortcutForPreviousTool", false);
```
bjorn: I would suggest to put this in "Interface/" instead.

src/tiled/preferences.cpp
```diff
@@ -691,6 +691,17 @@ void Preferences::setNaturalSorting(bool enabled)
...
+}
+
+void Preferences::setRepeatShortcutForPreviousTool(bool enabled)
+{
+    setValue(QLatin1String("Project/RepeatShortcutForPreviousTool"), enabled);
+    emit naturalSortingChanged(enabled);
```
bjorn: Wrong signal being emitted. Can just be removed because this setting does not need a signal.

src/tiled/toolmanager.cpp
```diff
@@ -262,7 +262,7 @@ void ToolManager::createShortcuts(QWidget *parent)
                 if (!action->isChecked()) {
                     // Select the tool associated with this shortcut
                     action->trigger();
-                } else {
+                } else if ( Preferences::instance()->repeatShortcutForPreviousTool() ) {
```
bjorn: Style nit:
```suggestion
                } else if (Preferences::instance()->repeatShortcutForPreviousTool()) {
```

src/tiled/preferencesdialog.ui
```diff
@@ -273,6 +273,16 @@
             </property>
            </widget>
           </item>
+            <item row="8" column="0" colspan="2">
```
bjorn: Indentation is off.

src/tiled/toolmanager.cpp
```diff
@@ -237,7 +237,7 @@ void ToolManager::retranslateTools()
  * This is done to make sure the shortcuts can still be used even when the
  * actions are only added to a tool bar and this tool bar is hidden.
  *
- * In addition, this allows us to implement a toggle behavior: if the tool is
+ * In addition, this allows us to optionally implement a toggle behavior: if the tool is
```
bjorn: I'd say the behavior is optional, not the implementation. :-)
```suggestion
 * In addition, this allows us to implement the optional toggle behavior: if the tool is
```

bjorn (changes requested):
Thanks for addressing this need, @dogboydog! Indeed it is clearly necessary judging by the feedback, and I agree it seems better to default this feature to "off".

I've written some feedback in comments that I could also address myself now before merging. Also needs a `NEWS.md` entry.

bjorn (approved):

dogboydog:
Oops, sorry for missing those. Thanks

PR #4544, opened 2026-06-09
New Hardware Renderer: Load TiledQuick's QML instead of test_renderer.qml

UltraDagon:
This replaces the existing source of MapEditor's quickWidget, test_renderer.qml, with tiledquick/qml/widget.qml, which is a new QML file using the main body from tiledquick/qml/main.qml and excluding all ApplicationWindow-specific features.


UltraDagon:
Copying over Bjorn's comments from Discord:
> I don't think you'll want to use Tiled Quick as the QQuickWidget' source. As you noticed this already won't work due to the use of ApplicationWindow: Tiled Quick is its own application. Instead, you can take from Tiled Quick the parts that you can use in the map view: the pannable view based on MapItem, MapBorderItem, MapGridItem, etc.

> But try to set it up in a way that doesn't make Tiled load the QML from the Tiled Quick resource file. It should just have the QML in its own resource file.

UltraDagon:
This last push should finish addressing everything from the Discord comment. widget.qml holds all of the used TiledQuick parts within a rectangle that can be imported into a QQuickWidget. It's packed along with DragArea.qml into the widget.qrc resource file, which is accessed in libtilededitor.qbs.

Two questions:
I put widget.qml and widget.qrc in the tiledquick/qml folder since I believe it is supposed to contain the GUI files. Is this the right place for them, and is widget.qml a descriptive enough name?

Since widget.qml is a nearly identical copy of main.qml's body, should main.qml's body be replaced with widget.qml so that updating widget.qml will update both the main Tiled editor and the TiledQuick application?

bjorn:
> I put widget.qml and widget.qrc in the tiledquick/qml folder since I believe it is supposed to contain the GUI files. Is this the right place for them, and is widget.qml a descriptive enough name?

That is not the right place. I think you can just add your QML files to the existing generated `tilededitor.qrc`, which you have already done with the `test_renderer.qml` file. I'd suggest you create a `qml` folder there with a `MapView.qml` in it (which is your `widget.qml`, but from which you should remove the `Tiled.MapLoader` (since the map is already loaded and the `EditableMap*` should be provided by the `MapEditor`) and the `Settings` (since we store view position and scale in the session) and the `RowLayout` at the bottom).

> Since widget.qml is a nearly identical copy of main.qml's body, should main.qml's body be replaced with widget.qml so that updating widget.qml will update both the main Tiled editor and the TiledQuick application?

It's identical now because you copied it, but it should be clear that the file will look rather different going forward. I don't think it makes sense to share anything at this point, beyond what we put in the `tiledquickplugin`, which is used by both Tiled Quick and the new map view. If we want to share some QML components, probably we could put them in the `tiledquickplugin` somehow, but I think it's too early for this.

bjorn:
@UltraDagon Btw, I've dropped the Qt 5 builds from CI on `master` and have just merged that into the `new-hardware-renderer` branch, so you could rebase this PR on top.

Though that won't fix the CI failures, which are due to "Product 'tiled' depends on 'tiledquickplugin'", and the latter requires Qt 6.5 rather than 6.2. But we should try to avoid this direct dependency anyway.

src/tiled/mapeditor.cpp
```diff
@@ -355,9 +356,38 @@ void MapEditor::addDocument(Document *document)
...
 
+    QQuickItem *rootObject = quickWidget->rootObject();
+
+    if (rootObject)
+    {
+        QObject *mapItem = rootObject->findChild<QObject*>("mapItem");
+
+        if (mapItem)
+        {
+            EditableMap *editableMap = new EditableMap(mapDocument->map());
```
bjorn: This creates a read-only `EditableMap` wrapper around the `Map*`, however, the `MapDocument` already has an `EditableMap` instance, and you should just use that one:
```suggestion
            EditableAsset *editableMap = mapDocument->editable();
```

src/tiled/resources/qml/mapview.qml
```diff
@@ -0,0 +1,167 @@
...
+        property string mapsFolder
+        property alias mapSource: mapLoader.source
+
+        // TODO: Eventually should be remembered for each map
+        property alias mapScale: mapContainer.scale
+        property alias mapX: mapContainer.x
+        property alias mapY: mapContainer.y
+    }
+
+    // TODO: Remove mapLoader from mapview.qml
```
bjorn: What's stopping you? :-)

bjorn (changes requested):
I've pushed some changes to fix the QML resource setup. I'd prefer to just let them be part of the existing auto-generated .qrc file, rather than maintaining one manually as is currently done for Tiled Quick.

bjorn:
The macOS Qt 6.2 and Ubuntu Qt 6.4 builds are failing because of this error:
```
ERROR: /home/runner/work/tiled/tiled/src/tiledapp/tiledapp.qbs:6:1 Error while handling product 'tiled':
/home/runner/work/tiled/tiled/src/tiledapp/tiledapp.qbs:6:1 Product 'tiled' depends on 'tiledquickplugin',
/home/runner/work/tiled/tiled/src/tiledquickplugin/tiledquickplugin.qbs:3:1 but product 'tiledquickplugin' is disabled.
```
That's because the Tiled Quick Plugin depends on Qt >= 6.5. The fix should probably be to make this dependency optional and disable the Qt Quick view for those builds, but let's not worry about that for now. I'll disable these builds on this branch instead.

(also please don't forget to rebase this PR on the latest `new-hardware-renderer` branch)

src/tiled/resources/qml/mapview.qml
UltraDagon: Haha, I wanted to make sure I had my current work pushed before heading to bed in case there were any issues. I'll get that finished today.

src/tiled/resources/qml/mapview.qml
```diff
@@ -0,0 +1,160 @@
...
+import QtQuick.Dialogs
+import org.mapeditor.Tiled as Tiled
+import QtCore
+
+Rectangle {
+    id: widget
+    anchors.fill: parent
+    visible: true
+
+    Settings {
```
bjorn: Please remove this `Settings` instance.

src/tiled/resources/qml/mapview.qml
```diff
@@ -0,0 +1,160 @@
...
+
+                visibleArea: {
+                    var scale = mapContainer.scale
+                    Qt.rect(-mapContainer.x / scale,
+                            -mapContainer.y / scale,
+                            mapView.width / scale,
+                            mapView.height / scale);
+                }
+
+                // TODO: Remove if not needed later.
```
bjorn: Please remove now rather than later. :-)

src/tiled/resources/qml/mapview.qml
```diff
@@ -0,0 +1,160 @@
...
+    RowLayout {
+        anchors.bottom: parent.bottom;
+
+        Label {
+            text: {
+                var mapRelativeCoords = singleFingerPanArea.mapToItem(mapItem, singleFingerPanArea.mouseX, singleFingerPanArea.mouseY)
+                var tileCoords = mapItem.screenToTileCoords(mapRelativeCoords.x, mapRelativeCoords.y)
+                Math.floor(tileCoords.x) + ", " + Math.floor(tileCoords.y)
+            }
+        }
```
bjorn: Let's remove this `RowLayout` because it is the Tiled Quick status bar, and we already have a status bar as part of the Tiled window.

src/tiled/libtilededitor.qbs
```diff
@@ -556,6 +556,8 @@ DynamicLibrary {
         "utils.h",
         "variantmapproperty.cpp",
         "variantmapproperty.h",
+        "viewinterface.cpp",
+        "viewinterface.h",
```
bjorn: This interface is a good idea, but please name it `MapViewInterface` because `ViewInterface` seems a little too generic.

src/tiled/mapeditor.cpp
```diff
@@ -352,18 +356,32 @@ void MapEditor::addDocument(Document *document)
...
-    QQuickWidget *quickWidget = new QQuickWidget(mWidgetStack);
-    quickWidget->setSource(QUrl(QStringLiteral("qrc:/test_renderer.qml")));
+    engine->rootContext()->setContextProperty(QStringLiteral("mapItemMap"), editableMap);
+    quickWidget->setSource(QUrl(QStringLiteral("qrc:/qml/mapview.qml")));
     quickWidget->setResizeMode(QQuickWidget::SizeRootObjectToView);
 
-    mQuickWidgetForMap.insert(mapDocument, quickWidget);
+    if (quickWidget->status() == QQuickWidget::Error) {
+        for (const QQmlError &error : quickWidget->errors()) {
+            qDebug() << "QML Error:" << error.toString();
```
bjorn: Hm, aren't errors already logged somewhere?

src/tiled/viewinterface.cpp
bjorn, line 1: Please don't forget to add the copyright headers to new files.

src/tiled/viewinterface.h
```diff
@@ -0,0 +1,31 @@
...
+
+class ViewInterface : public QObject
+{
+    Q_OBJECT
+
+public:
+    ViewInterface(QWidget *parent = nullptr);
+    ~ViewInterface() override;
+
+    QWidget* getWidget() const;
```
bjorn: Coding style nitpick:
```suggestion
    QWidget *getWidget() const;
```
Please adjust for the other functions and their definitions as well.

src/tiled/viewinterface.cpp
```diff
@@ -0,0 +1,40 @@
...
+{
+    return mMapView.get();
+}
+
+QQuickWidget* ViewInterface::quickWidget() const
+{
+    return mQuickWidget.get();
+}
+
+inline bool ViewInterface::quickEnabled() const
```
bjorn: This function doesn't need to be a member of `ViewInterface`, can be static:
```suggestion
static bool quickEnabled()
```
(but does need to be moved up to before its first use, I'd recommend just after the headers have been included)

src/tiled/mapeditor.cpp
UltraDagon: I'm not sure of how to see the QML errors in the Qt Creator, but I removed this section as it was only there for debugging.

bjorn (approved):

bjorn:
@UltraDagon Nicely done to fix the Qt < 6.5 builds by making that part optional. Looking forward to your next PR! :-)

PR #4545, opened 2026-06-16
Show resize handles on maps in the world view

kanishka0411:
First piece of the drag-handle map resize feature.For now this just shows the 8 resize handles on the active map's corners and edges (no resize yet)

### DIFF
--- a/src/tiled/worldmovemaptool.cpp
+++ b/src/tiled/worldmovemaptool.cpp
@@ -35,6 +35,7 @@
 #include "zoomable.h"
 
 #include <QApplication>
+#include <QGraphicsView>
 #include <QKeyEvent>
 #include <QMenu>
 #include <QToolBar>
@@ -47,6 +48,36 @@
 using namespace Tiled;
 
 namespace Tiled {
+
+namespace {
+
+// which edges each resize handle moves, in the same order as
+// setSelectionScreenRect (corners and edge midpoints)
+struct HandleEdges { bool left, right, top, bottom; };
+
+const HandleEdges handleEdges[] = {
+    { true,  false, true,  false },     // top-left
+    { false, false, true,  false },     // top
+    { false, true,  true,  false },     // top-right
+    { true,  false, false, false },     // left
+    { false, true,  false, false },     // right
+    { true,  false, false, true  },     // bottom-left
+    { false, false, false, true  },     // bottom
+    { false, true,  false, true  },     // bottom-right
+};
+
+Qt::CursorShape cursorForHandle(int handle)
+{
+    switch (handle) {
+    case 0: case 7: return Qt::SizeFDiagCursor;     // top-left / bottom-right
+    case 2: case 5: return Qt::SizeBDiagCursor;     // top-right / bottom-left
+    case 1: case 6: return Qt::SizeVerCursor;       // top / bottom
+    case 3: case 4: return Qt::SizeHorCursor;       // left / right
+    default:        return Qt::ArrowCursor;
+    }
+}
+
+} // namespace
 
 WorldMoveMapTool::WorldMoveMapTool(QObject *parent)
     : AbstractWorldTool("WorldMoveMapTool",
@@ -72,6 +103,7 @@ void WorldMoveMapTool::keyPressed(QKeyEvent *event)
     case Qt::Key_Right: moveBy = QPointF(1, 0); break;
     case Qt::Key_Escape:
         abortMoving();
+        abortResizing();
         return;
     default:
         AbstractWorldTool::keyPressed(event);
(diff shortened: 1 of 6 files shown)


src/tiled/worldmovemaptool.cpp
```diff
@@ -131,17 +163,89 @@ void WorldMoveMapTool::moveMap(MapDocument *document, QPoint moveBy)
...
+
+    // the content only shifts when the left or top edge is the one being moved
+    mResizeOffset = QPoint(edges.left ? newWidth - map->width() : 0,
+                           edges.top ? newHeight - map->height() : 0);
+    mResizeNewSize = QSize(newWidth, newHeight);
+
+    // preview the result, matching what resizeMap() will produce
+    const QPoint topLeft(mResizeStartWorldRect.left() - mResizeOffset.x() * tileWidth,
+                         mResizeStartWorldRect.top() - mResizeOffset.y() * tileHeight);
+    const QRect previewRect(topLeft, QSize(newWidth * tileWidth, newHeight * tileHeight));
```
bjorn: Please see if you can use the `MapRenderer` to make sure the preview size matches also isometric and other orientations (see how this rect is calculated elsewhere).

src/tiled/worldmovemaptool.cpp
```diff
@@ -131,17 +163,89 @@ void WorldMoveMapTool::moveMap(MapDocument *document, QPoint moveBy)
...
+    if (mDraggingMap || mResizingMap)
         return;
 
     if (event->button() == Qt::LeftButton && mapCanBeMoved(targetMap())) {
-        // initiate drag action
+        // initiate a resize when one of the handles is pressed
+        QTransform viewTransform;
+        if (QWidget *viewport = event->widget())
+            if (auto view = qobject_cast<QGraphicsView*>(viewport->parent()))
+                viewTransform = view->transform();
```
bjorn: Please use a consistent method for getting the view transform, and consider making a helper functions for it. I think I prefer the `views.first()` in `mouseMoved`.

src/tiled/worldmovemaptool.cpp
```diff
@@ -157,8 +261,23 @@ void WorldMoveMapTool::mousePressed(QGraphicsSceneMouseEvent *event)
...
+
+        // show a resize cursor while hovering one of the handles
+        const auto views = mapScene()->views();
+        const QTransform viewTransform = views.isEmpty() ? QTransform()
+                                                         : views.first()->transform();
+        const int handle = resizeHandleAt(pos, viewTransform);
+        const Qt::CursorShape shape = (handle != -1) ? cursorForHandle(handle)
+                                                     : Qt::ArrowCursor;
+        if (cursor().shape() != shape)
+            setCursor(shape);
```
bjorn: We already have `refreshCursor` and I'd suggest to move this logic there, so all cursor logic lives in one place (track a hovered-handle index).

src/tiled/worldmovemaptool.cpp
```diff
@@ -48,6 +49,36 @@ using namespace Tiled;
...
+    { false, true,  false, true  },     // bottom-right
+};
+
+Qt::CursorShape cursorForHandle(int handle)
+{
+    switch (handle) {
+    case 0: case 7: return Qt::SizeFDiagCursor;     // top-left / bottom-right
+    case 2: case 5: return Qt::SizeBDiagCursor;     // top-right / bottom-left
+    case 1: case 6: return Qt::SizeVerCursor;       // top / bottom
+    case 3: case 4: return Qt::SizeHorCursor;       // left / right
```
bjorn: Consider introducing a classic enum for describing the handle indexes, to reduce the amount of magic numbers.

bjorn (changes requested):
It appears to work pretty well! I've added some comments on the implementation. :-)

bjorn:
@kanishka0411 The following block of code appears three times now (up from two):
```c++
        DocumentManager *manager = DocumentManager::instance();
        MapView *view = manager->viewForDocument(mapDocument());
        const QRectF viewRect { view->viewport()->rect() };
        const QRectF sceneViewRect = view->viewportTransform().inverted().mapRect(viewRect);
```
Please introduce a helper function for this. Maybe it can even include the `forceCenterOn` call because that appears to be the thing it is always used for in the end.

Also consider splitting out a `startResizing` from `WorldMoveMapTool::mousePressed` and an `finishResizing` from `WorldMoveMapTool::mouseReleased` (and introducing a `finishMoving` while you're at it), to keep the code a little easier to read.

In testing I noticed that the handles work for any hovered map, but the mouse cursor is only correctly updated when hovering the handles of the currently active map.

When zooming in, all resize handles seem to be off by one pixel in one or both directions:

image

Please still have a look at the proposed refactors and the above issues. :-)

bjorn:
@kanishka0411 I've done a few direct tweaks rather than going through another feedback iteration.

I think this change is almost ready to be merged now, but since it adds a new feature, could you add a line to the `NEWS.md` file?

We'll also want to update `docs/manual/worlds.rst` under "Editing Worlds" about resizing maps and mention the world grid on that page, but updating the manual can be done in a separate PR. Maybe one you can still finish this week?

kanishka0411:
Sure! I'll open a separate PR for the manual update today:)

bjorn (approved):

PR #4547, opened 2026-06-18
New Hardware Renderer: Add tool brush to mapview.qml

UltraDagon:
This adds Tiled's existing basic tool functionality to the New Hardware Renderer's mapview.qml.

Adds a toolBrush MapItem to mapview.qml.
Adds a new RegionOverlay.qml element used to highlight some region of the map.
Adds a selectedRegion RegionOverlay for tile-selection-based tools.
Adds a brushRegion RegionOverlay for the ToolBrush's edit preview region.
Adds mouse functionality to singleFingerPanArea in mapview.qml.
Changes mapItem to update upon tool usage.
Adds a new EditableMap constructor, allowing for the usage of SharedMaps.
Adds sizeChange and tileSizeChanged signals to EditableMap.

Lesser changes:
Abstracted setRegion and translateRegion functions in BrushItem. Not required for this PR, but may be useful in the future to signal changes at the BrushItem level.

### DIFF
--- a/src/libtiledquick/libtiledquick.qbs
+++ b/src/libtiledquick/libtiledquick.qbs
@@ -73,8 +73,6 @@ DynamicLibrary {
         cpp.includePaths: exportingProduct.sourceDirectory
     }
 
-    Depends { name: "Qt.core" }
-
     install: !qbs.targetOS.contains("darwin")
     installDir: {
         if (qbs.targetOS.contains("windows"))
--- a/src/libtiledquick/libtiledquick.qbs
+++ b/src/libtiledquick/libtiledquick.qbs
@@ -38,6 +38,8 @@ DynamicLibrary {
         "mapitem.h",
         "maploader.cpp",
         "maploader.h",
+        "regionoverlay.cpp",
+        "regionoverlay.h",
         "tiledquick_global.h",
         "tilelayeritem.cpp",
         "tilelayeritem.h",
@@ -70,6 +72,8 @@ DynamicLibrary {
 
         cpp.includePaths: exportingProduct.sourceDirectory
     }
+
+    Depends { name: "Qt.core" }
 
     install: !qbs.targetOS.contains("darwin")
     installDir: {
--- a/src/libtiledquick/mapitem.cpp
+++ b/src/libtiledquick/mapitem.cpp
@@ -42,6 +42,13 @@ void MapItem::setMap(Tiled::EditableMap *editableMap)
 
     mEditableMap = editableMap;
     mMap = editableMap ? editableMap->map() : nullptr;
+
+    if (editableMap)
+    {
+        Tiled::MapDocument *mapDocument = editableMap->mapDocument();
+        connect(mapDocument, &Tiled::MapDocument::regionChanged, this, &MapItem::repaintRegion);
+    }
+
     refresh();
     emit mapChanged();
 }
@@ -169,4 +176,10 @@ void MapItem::refresh()
 
     const QRect rect = mRenderer->mapBoundingRect();
     setImplicitSize(rect.width(), rect.height());
+}
+
+void MapItem::repaintRegion(const QRegion &region, Tiled::TileLayer *tileLayer)
+{
+    // TODO: update only the region edited
+    refresh();
 }
--- a/src/libtiledquick/mapitem.cpp
+++ b/src/libtiledquick/mapitem.cpp
@@ -30,6 +30,7 @@ using namespace TiledQuick;
 
 MapItem::MapItem(QQuickItem *parent)
     : QQuickItem(parent)
+    , mMap(nullptr)
     , mEditableMap(nullptr)
 {
 }
--- a/src/libtiledquick/mapitem.cpp
+++ b/src/libtiledquick/mapitem.cpp
@@ -192,7 +192,7 @@ void MapItem::repaintRegion(const QRegion &, Tiled::TileLayer *tileLayer)
         if (tileLayer == tileLayerItem->layer()) {
             // TODO: Update only the region edited
             tileLayerItem->update();
-            continue;
+            break;
         }
     }
 }
--- a/src/libtiledquick/mapitem.cpp
+++ b/src/libtiledquick/mapitem.cpp
@@ -30,8 +30,6 @@ using namespace TiledQuick;
 
 MapItem::MapItem(QQuickItem *parent)
     : QQuickItem(parent)
-    , mMap(nullptr)
-    , mEditableMap(nullptr)
 {
 }
 
--- a/src/libtiledquick/mapitem.cpp
+++ b/src/libtiledquick/mapitem.cpp
@@ -30,6 +30,7 @@ using namespace TiledQuick;
 
 MapItem::MapItem(QQuickItem *parent)
     : QQuickItem(parent)
+    , mEditableMap(nullptr)
 {
 }
 
@@ -187,7 +188,7 @@ void MapItem::refresh()
 void MapItem::repaintRegion(const QRegion &, Tiled::TileLayer *tileLayer)
 {
     for (TileLayerItem *tileLayerItem : std::as_const(mTileLayerItems)) {
-        if (tileLayer == tileLayerItem->getLayer()) {
+        if (tileLayer == tileLayerItem->layer()) {
             // TODO: Update only the region edited
             tileLayerItem->update();
             continue;
--- a/src/libtiledquick/mapitem.cpp
+++ b/src/libtiledquick/mapitem.cpp
@@ -40,15 +40,20 @@ void MapItem::setMap(Tiled::EditableMap *editableMap)
     if (mEditableMap == editableMap)
         return;
 
-    mEditableMap = editableMap;
-    mMap = editableMap ? editableMap->map() : nullptr;
+    if (mEditableMap && mEditableMap->mapDocument()) {
+        Tiled::MapDocument *oldMapDocument = mEditableMap->mapDocument();
+        disconnect(oldMapDocument, &Tiled::MapDocument::regionChanged, this, &MapItem::repaintRegion);
+        disconnect(oldMapDocument, &Tiled::MapDocument::mapResized, this, &MapItem::refresh);
+    }
 
-    if (editableMap && editableMap->mapDocument())
-    {
+    if (editableMap && editableMap->mapDocument()) {
         Tiled::MapDocument *mapDocument = editableMap->mapDocument();
         connect(mapDocument, &Tiled::MapDocument::regionChanged, this, &MapItem::repaintRegion);
         connect(mapDocument, &Tiled::MapDocument::mapResized, this, &MapItem::refresh);
     }
+
+    mEditableMap = editableMap;
+    mMap = editableMap ? editableMap->map() : nullptr;
 
     refresh();
     emit mapChanged();
--- a/src/libtiledquick/mapitem.cpp
+++ b/src/libtiledquick/mapitem.cpp
@@ -43,7 +43,7 @@ void MapItem::setMap(Tiled::EditableMap *editableMap)
     mEditableMap = editableMap;
     mMap = editableMap ? editableMap->map() : nullptr;
 
-    if (editableMap)
+    if (editableMap && editableMap->mapDocument())
     {
         Tiled::MapDocument *mapDocument = editableMap->mapDocument();
         connect(mapDocument, &Tiled::MapDocument::regionChanged, this, &MapItem::repaintRegion);
@@ -178,8 +178,13 @@ void MapItem::refresh()
     setImplicitSize(rect.width(), rect.height());
 }
 
-void MapItem::repaintRegion(const QRegion &region, Tiled::TileLayer *tileLayer)
+void MapItem::repaintRegion(const QRegion &, Tiled::TileLayer *tileLayer)
 {
-    // TODO: update only the region edited
-    refresh();
+    for (TileLayerItem *tileLayerItem : std::as_const(mTileLayerItems)) {
+        if (tileLayer == tileLayerItem->getLayer()) {
+            // TODO: Update only the region edited
+            tileLayerItem->update();
+            continue;
+        }
+    }
 }
--- a/src/libtiledquick/mapitem.cpp
+++ b/src/libtiledquick/mapitem.cpp
@@ -47,6 +47,7 @@ void MapItem::setMap(Tiled::EditableMap *editableMap)
     {
         Tiled::MapDocument *mapDocument = editableMap->mapDocument();
         connect(mapDocument, &Tiled::MapDocument::regionChanged, this, &MapItem::repaintRegion);
+        connect(mapDocument, &Tiled::MapDocument::mapResized, this, &MapItem::refresh);
     }
 
     refresh();
--- a/src/libtiledquick/regionoverlay.cpp
+++ b/src/libtiledquick/regionoverlay.cpp
@@ -20,24 +20,55 @@
 
 #include "regionoverlay.h"
 
-#include <QSGMaterial>
-//#include "regionoverlaymaterial"
+#include <QPainterPath>
+#include <QApplication>
+#include <QPalette>
 
 using namespace TiledQuick;
 
 RegionOverlay::RegionOverlay(QQuickItem *parent)
-    : QQuickItem{parent}
+    : QQuickItem(parent)
+    , mTileSize(0, 0)
+    , mRegion(QRegion())
+    , mColor(QApplication::palette().highlight().color())
 {
-
 }
 
 RegionOverlay::~RegionOverlay() = default;
 
-QSGNode *RegionOverlay::updatePaintNode(QSGNode *node, QQuickItem::UpdatePaintNodeData *)
+QList<QPolygonF> RegionOverlay::polygons() const
 {
-    auto regionNode = static_cast<QSGGeometryNode *>(node);
+    QPainterPath path;
 
-    return regionNode;
+    for (const QRect &r : mRegion.rects())
+        path.addRect(r);
+
+    QList<QPolygonF> polygons = path.simplified().toFillPolygons();
+
+    QTransform transform;
+    transform.scale(mTileSize.x(), mTileSize.y());
+
+    for (QPolygonF &polygon : polygons)
+        polygon = polygon * transform;
+
+    return polygons;
+}
+
+QColor RegionOverlay::strokeColor() const
+{
+    return mColor;
+}
+
+QColor RegionOverlay::fillColor() const
+{
+    QColor fillColor = mColor;
+    fillColor.setAlpha(64);
+    return fillColor;
+}
+
+QPointF RegionOverlay::tileSize() const
+{
+    return mTileSize;
 }
 
 void RegionOverlay::setTileSize(const QPointF &tileSize)
@@ -47,55 +78,18 @@ void RegionOverlay::setTileSize(const QPointF &tileSize)
 
     mTileSize = tileSize;
     emit tileSizeChanged();
-    update();
 }
 
-QPointF RegionOverlay::tileSize() const
+QRegion RegionOverlay::region() const
 {
-    return mTileSize;
+    return mRegion;
 }
 
-void RegionOverlay::setScale(const qreal &scale)
+void RegionOverlay::setRegion(const QRegion &region)
 {
-    if (mScale == scale)
+    if (mRegion == region)
         return;
 
-    mScale = scale;
-    emit scaleChanged();
-    update();
-}
-
-qreal RegionOverlay::scale() const
-{
-    return mScale;
-}
-
-void RegionOverlay::setValidColor(const QColor &color)
-{
-    if (mValidColor == color)
-        return;
-
-    mValidColor = color;
-    emit validColorChanged();
-    update();
-}
-
-QColor RegionOverlay::validColor() const
-{
-    return mValidColor;
-}
-
-void RegionOverlay::setInvalidColor(const QColor &color)
-{
-    if (mInvalidColor == color)
-        return;
-
-    mInvalidColor = color;
-    emit invalidColorChanged();
-    update();
-}
-
-QColor RegionOverlay::invalidColor() const
-{
-    return mInvalidColor;
+    mRegion = region;
+    emit regionChanged();
 }
--- a/src/libtiledquick/regionoverlay.cpp
+++ b/src/libtiledquick/regionoverlay.cpp
@@ -33,6 +33,7 @@ RegionOverlay::RegionOverlay(QQuickItem *parent)
     , mMapRect(QRect())
     , mValidColor(QApplication::palette().highlight().color())
     , mInvalidColor(QColor(255,0,0))
+    , mRegionAlpha(64)
 {
 }
 
@@ -46,7 +47,7 @@ QColor RegionOverlay::validStrokeColor() const
 QColor RegionOverlay::validFillColor() const
 {
     QColor fillColor = mValidColor;
-    fillColor.setAlpha(64);
+    fillColor.setAlpha(mRegionAlpha);
     return fillColor;
 }
 
@@ -58,7 +59,7 @@ QColor RegionOverlay::invalidStrokeColor() const
 QColor RegionOverlay::invalidFillColor() const
 {
     QColor fillColor = mInvalidColor;
-    fillColor.setAlpha(64);
+    fillColor.setAlpha(mRegionAlpha);
     return fillColor;
 }
 
@@ -104,6 +105,20 @@ void RegionOverlay::setMapRect(const QRect &rect)
     emit mapRectChanged();
 }
 
+int RegionOverlay::regionAlpha() const
+{
+    return mRegionAlpha;
+}
+
+void RegionOverlay::setRegionAlpha(const int &alpha)
+{
+    if (mRegionAlpha == alpha)
+        return;
+
+    mRegionAlpha = alpha;
+    emit regionAlphaChanged();
+}
+
 QList<QPolygonF> RegionOverlay::validPolygons() const
 {
     QRegion insideMapRegion;
@@ -112,19 +127,7 @@ QList<QPolygonF> RegionOverlay::validPolygons() const
     else
         insideMapRegion = mRegion.intersected(mMapRect);
 
-    QPainterPath path;
-    for (const QRect &r : insideMapRegion.rects())
-        path.addRect(r);
-
-    QList<QPolygonF> polygons = path.simplified().toSubpathPolygons();
-
-    QTransform transform;
-    transform.scale(mTileSize.x(), mTileSize.y());
-
-    for (QPolygonF &polygon : polygons)
-        polygon = transform.map(polygon);
-
-    return polygons;
+    return polygons(insideMapRegion);
 }
 
 QList<QPolygonF> RegionOverlay::invalidPolygons() const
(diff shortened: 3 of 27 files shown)


src/tiled/editablemap.cpp
UltraDagon, line 1: I wanted to get these warnings out of the way, as with the new tool brush, they fill up the terminal quite quickly. I believe I have set them up to be functionally correct this time.

Our previous discussion: https://github.com/mapeditor/tiled/pull/4520#discussion_r3322563373

UltraDagon:
Currently, toolBrushMap is set using `dynamic_cast<AbstractTileTool*>(selectedTool())->brushItem()->map().get()`. However, AbstractTileTool::brushItem() is normally protected, and I've made it public for the time being without rearranging it in the header file.

Is this the recommended way for me to obtain this map? If so, I can go ahead and move it to be a public getter.

I plan on integrating the quickBrushUpdated signal tomorrow, and then the colored overlays for using the different tools.

src/tiled/editablemap.cpp
bjorn: Hmm, no, like this they will only emit when these properties are changed through this function, but only the scripting API uses these functions. All the native code makes changes by constructing undo commands, for example a `ChangeMapTileSize` instance when changing the map's tile size. This command in turn emits `Document::changed` with a `MapChangeEvent` value with `property` set to `Map::TileSizeProperty`.

You can check the existing `MapItem::documentChanged` function which handles such change events. The `EditableMap` instance would also need to connect to this `changed` signal (you can introduce an `EditableMap::documentChanged` member to connect to it) and there it needs to `emit tileSizeChanged()` and other signals as appropriate.

It's a little unusual in Qt Quick, because usually the classes exposed there manage their own properties entirely, but the `EditableMap` is a wrapper API that provides just one of the ways to access and modify the data.

src/tiled/mapeditor.cpp
```diff
@@ -415,6 +417,14 @@ void MapEditor::setCurrentDocument(Document *document)
         mCurrentMapDocument->disconnect(this);
     }
 
+#ifdef TILEDQUICK_LIB
+    {
+    QQuickWidget *oldQuickWidget = mViewForMap.value(mapDocument)->quickWidget();
+    disconnect(oldQuickWidget->rootObject(), SIGNAL(mouseCoordsChanged(QVariant)),
+               this, SLOT(onQuickMouseCoordsChanged(QVariant)));
```
bjorn: When the last map document is closed, `setCurrentDocument` is called with `nullptr`, in which case `oldQuickWidget` will be `nullptr` and this `disconnect` call will crash.

Btw, we really don't want to be managing such connections from C++. Instead, expose a writable property (maybe on the `MapViewInterface`, but can also be on the `MapEditor` if it's easier for now) such that the mouse coordinates can be set directly from Qt Quick rather than going through a signal.

(btw, this `disconnect` has another bug in that it disconnects for the new current document rather than the previous current one - anyway it should just be removed)

src/tiled/resources/qml/mapview.qml
```diff
@@ -55,19 +55,39 @@ Rectangle {
...
+
+            Tiled.MapItem { // Tool Brush
+                id: toolBrush
+                anchors.left: mapItem.left
+                anchors.top: mapItem.top
+
+                map: toolBrushMap;
+
+                visibleArea: {
+                    var scale = mapContainer.scale
```
bjorn: Unused variable can be removed.

src/tiled/resources/qml/mapview.qml
```diff
@@ -55,19 +55,39 @@ Rectangle {
...
+                    else
+                        Qt.rect(0, 0, 0, 0);
+                }
+            }
         }
     }
 
     DragArea {
         id: singleFingerPanArea
+        objectName: "singleFingerPanArea"
```
bjorn: Should be no need for setting `objectName`.

src/tiled/mapeditor.cpp
```diff
@@ -1106,7 +1173,10 @@ EditableMap *MapEditor::currentBrush() const
 
     auto map = stamp.variations().first().map->clone();
     auto editableMap = new EditableMap(std::move(map));
-    QQmlEngine::setObjectOwnership(editableMap, QQmlEngine::JavaScriptOwnership);
+
+    // Disable JavaScriptOwnership for QtQuick rendering
+    if (!Preferences::instance()->useNewHardwareRenderer())
+        QQmlEngine::setObjectOwnership(editableMap, QQmlEngine::JavaScriptOwnership);
```
bjorn: What's the idea here? Leaving C++ ownership on it means C++ stays responsible for the lifetime of this instance. That might be a good idea, but to do so we'll have to keep a reference to this instance (say, add an `EditableMap *mCurrentBrush` member) and then make sure to replace it as needed.

I think I left ownership to JS because this function is part of the JS API, where user might write code like:

```js
var b = tiled.mapEditor.currentBrush;
tiled.mapEditor.currentBrush = new TileMap();
// still do something with b here
```

If the `EditableMap` instance was instead owned by `MapEditor`, then assigning something else to it would invalidate `b` kind of unexpectedly.

We'll also want to avoid such behavioral differences just based on which renderer is used.

src/tiled/mapeditor.cpp
```diff
@@ -708,14 +735,54 @@ void MapEditor::onSelectedToolChanged(AbstractTool *tool)
...
+        QVariant brushMapContext = rootContext->contextProperty(QStringLiteral("toolBrushMap"));
+
+        if (brushMapContext.isValid()) {
+            EditableMap *oldBrushMap = brushMapContext.value<EditableMap*>();
+
+            if (oldBrushMap && oldBrushMap->map() == atTool->brushItem()->map().get())
+                return;
+
+            EditableMap *newBrushMap = new EditableMap(atTool->brushItem()->map().get());
+            rootContext->setContextProperty(QStringLiteral("toolBrushMap"), newBrushMap);
```
bjorn: You should try not to interact like this with Qt Quick from C++.

Really what we're looking for is a way to set the preview for tile edits. The `BrushItem` implements preview rendering for the `QGraphicsView`. The QML `MapItem` can form the basis of the preview rendering in Qt Quick.

I think the way to expose the preview to Qt Quick will be through a property on `MapEditor` similar to the `currentBrush`, maybe called `tileEditPreview`. Its value is basically what `toolBrushMap` is currently being set to.

src/tiled/stampbrush.cpp
```diff
@@ -676,6 +676,10 @@ void StampBrush::updatePreview(QPoint tilePos)
     }
 
     brushItem()->setMap(mPreviewMap, tileRegion);
+
+#ifdef TILEDQUICK_LIB
+    emit quickBrushChanged();
```
bjorn: Let's call this `previewChanged` (or `brushMapChanged`, which is a little more consistent with `brushItem`), and I don't think it needs to be compiled out when `TILEDQUICK_LIB` is not defined.

src/tiled/mapeditor.cpp
```diff
@@ -708,14 +735,54 @@ void MapEditor::onSelectedToolChanged(AbstractTool *tool)
     if (tool) {
         connect(tool, &AbstractTool::cursorChanged,
                 this, &MapEditor::cursorChanged);
+#ifdef TILEDQUICK_LIB
+        AbstractTileTool *atTool = dynamic_cast<AbstractTileTool*>(tool);
+        if (atTool) {
+            connect(atTool, &AbstractTileTool::quickBrushChanged,
+                       this, &MapEditor::onQuickBrushChanged);
```
bjorn: I think when you implement `tileEditPreview` as a property that creates the `EditableMap` when it is accessed, then here we can forward directly to a `MapEditor::tileEditPreviewChanged` signal.

src/tiled/abstracttiletool.h
```diff
@@ -112,8 +117,9 @@ class AbstractTileTool : public AbstractTool
      * hidden based on whether the mouse is in the scene and whether the
      * currently selected layer is a tile layer.
      */
+public: // TODO: MOVE
```
bjorn: Yeah, I think we only really want access to `brushItem()->map()`, so instead of changing this to public, introduce the necessary accessor, maybe just `const SharedMap &brushMap() const` for now.

src/tiled/mapeditor.cpp
```diff
@@ -690,6 +710,13 @@ void MapEditor::onSelectedToolChanged(AbstractTool *tool)
     if (mSelectedTool) {
         disconnect(mSelectedTool, &AbstractTool::cursorChanged,
                    this, &MapEditor::cursorChanged);
+#ifdef TILEDQUICK_LIB
+        AbstractTileTool *atTool = dynamic_cast<AbstractTileTool*>(mSelectedTool);
```
bjorn: Please prefer `qobject_cast` when dealing with classes derived from `QObject`:
```suggestion
        AbstractTileTool *atTool = qobject_cast<AbstractTileTool*>(mSelectedTool);
```
This mechanism exists in Qt for any type derived from `QObject` and does not rely on RTTI (though we don't disable that and we're using `dynamic_cast` in plenty of other places, I'm not sure if I have been very strict on this either).

src/libtiledquick/mapitem.cpp
```diff
@@ -42,6 +42,13 @@ void MapItem::setMap(Tiled::EditableMap *editableMap)
 
     mEditableMap = editableMap;
     mMap = editableMap ? editableMap->map() : nullptr;
+
+    if (editableMap && editableMap->mapDocument())
+    {
+        Tiled::MapDocument *mapDocument = editableMap->mapDocument();
+        connect(mapDocument, &Tiled::MapDocument::regionChanged, this, &MapItem::repaintRegion);
```
bjorn, line 51: You need to also disconnect from `regionChanged`.

src/tiled/mapeditor.cpp
```diff
@@ -708,14 +735,54 @@ void MapEditor::onSelectedToolChanged(AbstractTool *tool)
...
     updateActiveUndoStack();
+
     emit selectedToolChanged(tool);
 }
 
+#ifdef TILEDQUICK_LIB
+void MapEditor::onQuickMouseCoordsChanged(QVariant coords)
+{
+    QPointF mouseCoords = coords.toPointF();
+    selectedTool()->mouseMoved(mouseCoords, Qt::NoModifier);
```
bjorn: Careful, I think `selectedTool()` could be `nullptr`.

src/libtiledquick/tilelayeritem.cpp
```diff
@@ -201,6 +199,10 @@ QSGNode *TileLayerItem::updatePaintNode(QSGNode *node,
     return node;
 }
 
+Tiled::TileLayer *TileLayerItem::getLayer() {
```
bjorn: Should be called just `layer` instead of `getLayer`.

bjorn (changes requested):
Nice to see things are coming together, but I had a lot of feedback. :-)

src/tiled/mapeditor.cpp
UltraDagon: I added this toggle because if the setObjectOwnership acts normally while the preference is enabled, the toolBrushMap gets cleared after moving the cursor around a bit. I'll ask Adam and look into this closer to see why that behavior occurs.

bjorn: Hmm, but the `toolBrushMap` context property was set with a different `EditableMap` instance without custom ownership set (in which case I'm not currently sure about the behavior). But it doesn't use `currentBrush` at least.

UltraDagon: My apologies, you are totally right. I just tested removing it, and there were no issues. I think the last time I tested it was when I was halfway between moving the used brush from currentBrush() to the brushItem()->map(). I'll make sure to pay extra attention to these renderer-specific differences in the future.

UltraDagon:
After linking a lot of these abstractTileTool signals, it might be better to link within the brushitems instead. Let me know what you think.

Edit: I did this with the region changes, which, to do so, led to me having to change brushItem from a QGraphicsItem to a QGraphicsObject.

UltraDagon:
After adding the regionOverlay and rendering it using QML geometry, I think it might be better to change the griditem from being shader-based to QML geometry-based, as the current shader floating-point errors cause the regionOverlay to be slightly offset from the grid. Let me know what you think 😄

src/tiled/stampbrush.cpp
```diff
@@ -96,14 +96,18 @@ void StampBrush::tilePositionChanged(QPoint pos)
...
                 brushItem()->setMap(mPreviewMap);
+            }
 
             doPaint(Mergeable, &paintedRegions);
         }
 
         QHashIterator<TileLayer*, QRegion> ri(paintedRegions);
         while (ri.hasNext()) {
             ri.next();
+            emit brushRegionChanged(ri.value());
```
bjorn: This doesn't look like the right place to emit a brush region changed signal. The `ri.value()` is the region in which an edit was applied.

UltraDagon: That makes sense. I've moved the brushRegionChanged to be mPreviewMap->modifiedTileRegion(), moved it out of the while loop, and changed the brushMapChanged to nullptr while dragging to paint.

UltraDagon:
I'm not super familiar with working with smart pointers and wanted to make sure that my most recent commit (https://github.com/mapeditor/tiled/pull/4547/commits/a75bf94d5be970ef79f0953314f412728227fca3) is done correctly. From testing, I believe that it is properly freeing the EditableMap and SharedMap memory when they go out of scope, and mTileEditPreview should be the final owner of the toolBrush SharedMap.

Let me know if I misunderstood anything while making this. Thanks!

UltraDagon:
I also went ahead and undrafted the PR and updated the description since, once this is touched up and finalized, it should be a good stopping point before creating new PRs for other features.

UltraDagon:
> I'm not super familiar with working with smart pointers and wanted to make sure that my most recent commit ([a75bf94](https://github.com/mapeditor/tiled/commit/a75bf94d5be970ef79f0953314f412728227fca3)) is done correctly. From testing, I believe that it is properly freeing the EditableMap and SharedMap memory when they go out of scope, and mTileEditPreview should be the final owner of the toolBrush SharedMap.
> 
> Let me know if I misunderstood anything while making this. Thanks!

Talked a bit with Adam about this and added some safeguards to make sure this new constructor isn't misused in the future. I was also thinking that it could be possible to give MapItem the option to work how it used to, taking in a standard map, to avoid having to cast to an EditableMap. I'm not entirely sure, but I don't believe that the ToolBrush MapItem uses any EditableMap-specific features.

src/libtiledquick/libtiledquick.qbs
```diff
@@ -71,6 +73,8 @@ DynamicLibrary {
         cpp.includePaths: exportingProduct.sourceDirectory
     }
 
+    Depends { name: "Qt.core" }
```
bjorn: Why was this added here? I think that should be already covered by this line above:
```
Depends { name: "Qt"; submodules: ["quick","shadertools"]; versionAtLeast: "6.5" }
```

src/libtiledquick/regionoverlay.cpp
```diff
@@ -0,0 +1,158 @@
...
+#include <QApplication>
+#include <QPalette>
+
+using namespace TiledQuick;
+
+RegionOverlay::RegionOverlay(QQuickItem *parent)
+    : QQuickItem(parent)
+    , mTileSize(0, 0)
+    , mRegion(QRegion())
+    , mMapRect(QRect())
```
bjorn: No need to explicitly write the initialization if it's the same as the default initialization, just leave it out in that case.

src/libtiledquick/tilelayeritem.cpp
```diff
@@ -201,6 +199,10 @@ QSGNode *TileLayerItem::updatePaintNode(QSGNode *node,
     return node;
 }
 
+Tiled::TileLayer *TileLayerItem::layer() {
+    return mLayer;
```
bjorn: A good candidate for inlining. :-)

src/tiled/mapeditor.cpp
```diff
@@ -1106,10 +1142,40 @@ EditableMap *MapEditor::currentBrush() const
...
+EditableMap *MapEditor::tileEditPreview() const
+{
+    return mTileEditPreview.get();
+}
+
+void MapEditor::setTileEditPreview(const SharedMap &map)
+{
+    if (!map) {
+        mTileEditPreview.reset();
+        return;
```
bjorn, line 1169: I think this branch is missing the `emit tileEditPreviewChanged()`? You can also consider:
```c++
mTileEditPreview = map ? std::make_unique<EditableMap>(map) : nullptr;
emit tileEditPreviewChanged();
```
(not sure if it works exactly like that, might need `std::unique_ptr<EditableMap>()` instead of `nullptr` to match the type)

src/tiled/mapeditor.h
```diff
@@ -135,10 +146,20 @@ class MapEditor final : public Editor
     AbstractTool *selectedTool() const;
     void setSelectedTool(AbstractTool *tool);
 
+#ifdef TILEDQUICK_LIB
+    Q_INVOKABLE void setQuickMouseCoords(QPointF coords);
```
bjorn: I guess as implemented this function should just be called `quickMouseMoved` for consistency.

src/tiled/editablemap.h
```diff
@@ -240,6 +243,7 @@ class TILED_EDITOR_EXPORT EditableMap final : public EditableAsset
     MapRenderer *renderer() const;
 
     std::unique_ptr<Map> mDetachedMap;
+    SharedMap mSharedMap = nullptr;
```
bjorn:
```suggestion
    SharedMap mSharedMap;
```
It will be null by default.

src/tiled/mapeditor.cpp
```diff
@@ -708,14 +726,32 @@ void MapEditor::onSelectedToolChanged(AbstractTool *tool)
...
+
     emit selectedToolChanged(tool);
 }
 
+QRegion MapEditor::selectedRegion() const
+{
+    if (!mCurrentMapDocument)
+        return QRegion();
+
+    return mCurrentMapDocument->selectedArea();
```
bjorn, line 752: This property is already available from `EditableMap::selectedArea`, or was there a problem with the way that is exposed? I guess it's currently a problem to support `RegionValueType` as value in `RegionOverlay` since the former lives in the libtilededitor...

src/tiled/brushitem.cpp
```diff
@@ -236,3 +233,19 @@ void BrushItem::updateBoundingRect()
...
+
+void BrushItem::setRegion(const QRegion &region)
+{
+    if (mRegion == region)
+        return;
+
+    mRegion = region;
+}
+
+void BrushItem::translateRegion(const QPoint &p)
```
bjorn: What is this function for? Also `setRegion` seems superfluous.

src/tiled/resources/qml/RegionOverlay.qml
```diff
@@ -0,0 +1,32 @@
+import QtQuick
+import QtQuick.Shapes
+import org.mapeditor.Tiled as Tiled
+
+Tiled.RegionOverlay {
+    id: regionOverlay
+
+    property var scale
+    // TODO: Draw only visible area
```
bjorn: I guess we can remove that TODO for now, because it's unclear how that's supposed to work and also whether it will be more efficient anyway.

Hopefully the `Shape` item will already do some optimizations, but even if it doesn't, adjusting its shapes based on the visible area would also have a certain performance overhead.

src/tiled/resources/qml/mapview.qml
```diff
@@ -130,4 +195,19 @@ Rectangle {
             mapContainer.y = (mapView.height / 2) - ((mapItem.height * scale) / 2)
         }
     }
+
+    function cursorTileCoords() {
+        var tileCoords = mapItem.screenToTileCoords(mapRelativeCoords.x, mapRelativeCoords.y)
+
+        return tileCoords
```
bjorn: Let's shorten this a bit. :-)
```suggestion
        return mapItem.screenToTileCoords(mapRelativeCoords.x, mapRelativeCoords.y)
```

src/tiled/resources/qml/mapview.qml
```diff
@@ -130,4 +195,19 @@ Rectangle {
...
+    }
+
+    property var mapRelativeCoords: {
+        var mapRelativeCoords = singleFingerPanArea.mapToItem(mapItem, singleFingerPanArea.mouseX, singleFingerPanArea.mouseY)
+        return mapRelativeCoords
```
bjorn:
```suggestion
        return singleFingerPanArea.mapToItem(mapItem, singleFingerPanArea.mouseX, singleFingerPanArea.mouseY)
```

src/tiled/mapeditor.h
```diff
@@ -210,6 +234,9 @@ class MapEditor final : public Editor
     WangBrush *mWangBrush;
     EditPolygonTool *mEditPolygonTool;
 
+    std::unique_ptr<EditableMap> mTileEditPreview;
```
bjorn: I think it's a problem to use `std::unique_ptr<EditableMap>` here and also pass the `EditableMap*` into the QML context, from where it is passed on to `MapItem`. We're probably kind of lucky when the `unique_ptr` deletes the `EditableMap` instance, that the `MapItem` no longer references that pointer.

When we create the `EditableMap` instance, actually its `EditableObject` parent constructor will set it as the `Object::mEditable`. That means it probably suffices to just store a `SharedMap mTileEditPreview` here, and we can create its `EditableMap` instance on-demand in `MapEditor::tileEditRegion`, using `EditableObject::find(mTileEditPreview.get())` to reuse an existing editable instance when possible.

bjorn (changes requested):
@UltraDagon It's really great to see this is actually working and to be able to edit with all those existing tools while using the new renderer!

I left a bunch of minor and some larger points of feedback, but it seems to me you're on track to get this merged by the end of the week. :-)

src/libtiledquick/libtiledquick.qbs
UltraDagon: Sorry about that, Qt Creator auto adds it whenever I add new files and I must have forgot to remove it.

src/tiled/mapeditor.cpp
UltraDagon: It seems like it should do the same thing. I used your suggested change.

src/tiled/mapeditor.cpp
UltraDagon: It looks like EditableMap::selectedArea is flagged as a CONSTANT property. I'd probably be able to get it to notify correctly if that's the route we want to go. Otherwise, we could just keep it as coming from mCurrentMapDocument.

src/tiled/brushitem.cpp
UltraDagon: setRegion and translateRegion were remnants from when I was getting signals from brushItem changes, so they're no longer needed, but I had left them in anyway in case they were needed in the future to more easily send signals. I'll remove them since they don't seem necessary or hard to re-add if needed in the future.

src/tiled/mapeditor.cpp
bjorn: That's `CONSTANT` because the returned value never changes, it's an `EditableSelectedArea` instance that provides functions to inspect and modify the selected area. If you're going to add any signals, they should be added to the `EditableSelectedArea`.

But the problem with relying on it as-is I think is that its `get()` function returns a `RegionValueType`, which is currently only available in libtilededitor, which libtiledquick can't use (it only depends on libtiled). So either `RegionValueType` would need to be moved into to libtiled, or `EditableSelectedArea` would need to provide a property that returns `QRegion` directly (which would be preferred I think because it's the smaller change).

src/tiled/mapeditor.cpp
```diff
@@ -1149,17 +1149,27 @@ EditableMap *MapEditor::currentBrush() const
...
-    if (!map) {
-        mTileEditPreview.reset();
+    if (map == mTileEditPreview)
         return;
-    }
 
-    mTileEditPreview = std::make_unique<EditableMap>(map);
+    EditableObject* oldPreview = EditableObject::find(mTileEditPreview.get());
+    if (oldPreview)
+        delete oldPreview;
```
bjorn: Hmm, two things:
* You never need to check a pointer for nullptr before calling `delete` on it, that check is built into `delete`.
* Are you sure you need to delete the `oldPreview` manually? It should get deleted by this line eventually:
```c++
Object::~Object()
{
    delete mEditable;
}
```
(since `Map` derives from `Object`)

src/tiled/mapeditor.cpp
```diff
@@ -1149,17 +1149,27 @@ EditableMap *MapEditor::currentBrush() const
...
-    return mTileEditPreview.get();
+    EditableMap* preview;
+    EditableObject* existingPreview = EditableObject::find(mTileEditPreview.get());
+
+    if (existingPreview)
+        preview = qobject_cast<EditableMap*>(existingPreview);
+    else
+        preview = mTileEditPreview ? new EditableMap(mTileEditPreview) : nullptr;
+
+    return preview;
```
bjorn: That could be written a little shorter:
```c++
    if (EditableObject *preview = EditableObject::find(mTileEditPreview.get()))
        return qobject_cast<EditableMap*>(preview);
    return mTileEditPreview ? new EditableMap(mTileEditPreview) : nullptr;
```
:-)

bjorn (changes requested):
Almost there. :-)

src/libtiledquick/libtiledquick.qbs
bjorn: Hmm, I never noticed this problem. If it really does that I guess we can work it around by adding it explicitly to the existing `Depends` item. But that lists "quick" and "shadertools" modules, which I think pulls in the "core" dependency automatically.

bjorn:
> I was also thinking that it could be possible to give MapItem the option to work how it used to, taking in a standard map, to avoid having to cast to an EditableMap.

I think we'd need to reintroduce the `MapRef` wrapper for that, but at the moment I'm having difficulties thinking about the use-case. We generally want the `EditableMap` instance anyway because it provides the access to the data for the QML side, whereas the `MapRef` could only be passed around.

src/tiled/mapeditor.cpp
UltraDagon: This might be the wrong way to check it, but by putting qDebug()s in the EditableMap and Object destructors, it doesn't look like the EditableMaps created by MapEditor::tileEditPreview() are ever destroyed. I think it's because the *EditableMap owns the last remaining SharedMap reference, so it keeps itself alive if the EditableMap is not destroyed.

bjorn: Hmm, you're right the `EditableMap` instance keeps the `Map` alive through its shared reference. However, what keeps the `EditableMap` instance around?

In this case I guess that should be owned by the JS engine and should eventually be garbage collected when it is no longer referenced. We may need to set JS ownership on it explicitly, like in `MapEditor::currentBrush`.

UltraDagon: I think I was able to fully fix this by replacing mSharedMap with bool mIsSharedMap, which, if true, skips the layer detach section upon destruction. Skipping this section should probably be fine since `auto editableLayer = EditableLayer::find(layer);` would be nullptr, leading to the rest of the function's code not executing regardless of map()->layers() "finding" the layers that have already been deleted once the SharedMap loses all references.

bjorn:
@UltraDagon Thanks for your efforts on this! You've reached another great milestone and I'm looking forward to the support for object groups!

PR #4550, opened 2026-06-22
Update the mime types to support more of the Tiled extensions

### LINKED ISSUE #4523: Please add missing extensions and MIME types (file associations) so these files can be clicked to open from the explorer, Unity, etc.
Currently, only `*.tsx` and `*.tmx` files are automatically associated with the app once installed (at least on Linux/Flatpak, probably on Windows too). So only these files can be double-clicked to be opened in the Tiled app. The other extensions are not registered:

https://github.com/mapeditor/tiled/blob/f511930c1b6f816b3ec5a98cd4e740205f4fcf2e/mime/org.mapeditor.Tiled.xml#L1-L15
 This is particularly inconvenient for the project file, which IMO would be the primary way one may want to open stuff in Tiled: we typically want to open an entire project from the explorer (or from Unity), and then open other files from there.

These extensions (and their corresponding MIME types) are missing an association on Linux (so they're recognized as plain text documents and I can't set Tiled to open that extension, or else it would open all plain text documents):
- `*.tiled-project` - **Tiled Project**
- `*.tsj` - **Tiled Tileset as JSON** (useful for those who want to avoid conflict with tsx, which is already associated with Typescript React because of IDEs like VSCode; maybe consider adding something like `*.tiled-set` as an alternative to .tsx to avoid conflict?)
- `*.tmj` - **Tiled Tilemap as JSON**

Please add those MIME associations to the Flathub version. Thanks!

---
There's also `*.tiled-session` but that one cannot be opened directly, I think, so it's fine to leave it as plain text.
(side note: I wish there was an option to not write such machine-local files next to the project, e.g. put it on some `~/.cache` or equivalent, or even an option to not write it at all... so as to avoid accidental garbage committed on version control or shared between machines via pendrive, etc)

miffe:
This updates the mime database and the desktop file to add suport for:
 - .tmj: Tiled map json files
 - .tsj: Tiled tileset json files
 - .tiled-project: Tiled project files
 - .world: Tiled world files

Closes #4523

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,7 @@
 * Scripting: Added 'tiled.cell' function, 'cell.flags' property and 'TileLayerEdit.setCell' function (#4538)
 * Scripting: Added MapObject.resolvedClassName() (by MatusGuy, #4529)
 * Fixed crash when the selection becomes empty while starting a move (#4536)
+* Linux: Added file associations for .tmj, .tsj, .tiled-project and .world files (by miffe, #4550)
 * snap: Updated to Qt 6
 * Raised the minimum Qt version to 6.2 (drops Qt 5 support)
 
--- a/mime/org.mapeditor.Tiled.xml
+++ b/mime/org.mapeditor.Tiled.xml
@@ -12,4 +12,28 @@
     <glob pattern="*.tsx"/>
     <sub-class-of type="application/xml"/>
   </mime-type>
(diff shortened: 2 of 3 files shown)


bjorn:
@miffe Thanks a lot, that's really appreciated!

(now we just need to do the same for macOS (#4469) and Windows)

PR #4552, opened 2026-06-26
Add manual section for resizing maps and the world grid

kanishka0411:
This is the PR for the manual update from #4545 . Adds short sections under "Editing Worlds" for resizing maps and the world grid.

### DIFF
--- a/docs/manual/worlds.rst
+++ b/docs/manual/worlds.rst
@@ -94,21 +94,26 @@ Moving Maps
 
 Resizing Maps
     Hover over a map with the World Tool to show resize handles on its corners
-    and edges, then drag a handle to resize the map within the world. The size
-    snaps to the world grid; hold 'Ctrl' while dragging to resize freely. You
-    can abort a resize by hitting 'Escape' or by right-clicking.
-
-Using the World Grid
-    A grid can be shown across the world to help align maps. Toggle it with
-    *View > Show World Grid*, and toggle snapping with *View > Snapping > Snap
-    Maps to World Grid*. When snapping is enabled, maps snap to the grid while
-    being moved or resized (hold 'Ctrl' to override). The grid size is set per
-    world in the *World > World Properties* dialog.
+    and edges, then drag a handle to resize the map within the world. The map's
+    existing content keeps its position within the world, with the map's world
+    position adjusted when necessary to achieve this. You can abort a resize
+    by hitting 'Escape' or by right-clicking.
 
 Saving World files
     You can save manipulated world files by using the *World > Save World*
     menu. Worlds will also automatically be saved if you launch any external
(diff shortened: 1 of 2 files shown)


docs/manual/worlds.rst
```diff
@@ -92,6 +92,19 @@ Moving Maps
...
+    Hover over a map with the World Tool to show resize handles on its corners
+    and edges, then drag a handle to resize the map within the world. The size
+    snaps to the world grid; hold 'Ctrl' while dragging to resize freely. You
+    can abort a resize by hitting 'Escape' or by right-clicking.
+
+Using the World Grid
+    A grid can be shown across the world to help align maps. Toggle it with
+    *View > Show World Grid*, and toggle snapping with *View > Snapping > Snap
+    Maps to World Grid*. When snapping is enabled, maps snap to the grid while
+    being moved or resized (hold 'Ctrl' to override). The grid size is set per
```
bjorn: I guess it's enough to mention the snapping only here, expand a little on what "override" means and remove the note at "Resizing Maps". Otherwise for consistency we should also mention the snapping at "Moving Maps", but it feels like too much duplication.

I also think the grid should get its own section between "Editing Worlds" and "Using Pattern Matching".

docs/manual/worlds.rst
```diff
@@ -92,6 +92,19 @@ Moving Maps
     Alternatively you can use the arrow keys to move the current selected map
     - holding Shift will perform bigger steps.
 
+Resizing Maps
+    Hover over a map with the World Tool to show resize handles on its corners
+    and edges, then drag a handle to resize the map within the world. The size
```
bjorn: I think it would be interesting to mention that resizing a map keeps its existing content at the same position within the world, by adjusting the map's world position when necessary.

Btw, I don't think we need to explain this in detail, but we currently have the following unfortunately side-effect of having separate undo stacks:

- Resize map in world
- Undo
  - Expected: Resize undone
  - Actual: Only the position adjustment is undone

To undo the resize, the user currently has to switch away from the world tool. At least that will still also undo the position adjustment, but we probably need to think about linking these undo commands in both directions somehow.

docs/manual/worlds.rst
kanishka0411: DONE:-)

docs/manual/worlds.rst
kanishka0411: Yeah I see it now, the resize and the position end up on different undo stacks. I can dig into it as a follow-up Pr.

bjorn:
@kanishka0411 I did a small follow-up to make sure those new features are properly highlighted as being new in Tiled 1.13. Thanks for the updates! :-)

PR #4560, opened 2026-07-01
Ignore generated docs/_build output

bjorn:
The Sphinx build writes the HTML, images and the intermediate `.doctrees` cache under `docs/_build`. Only the gettext `.pot` files there are meant to be tracked, so ignore everything else under `docs/_build` while keeping those.


bjorn:
The files I had lying around in `docs/_build/.doctrees/` turned out to be stale leftovers from 3 years ago, so those should be removed instead of ignored, but I found some other cruft worth cleaning up...

PR #4561, opened 2026-07-01
Persisted collapsed state of the properties groups in the session

bjorn:
The expanded/collapsed state of the object properties groups and of the Custom Properties group is now stored in the session, so it survives both switching the current object and restarting the application.

The two properties docks are kept in sync at runtime through the session, so collapsing or expanding a group in one view is reflected in the other. States are stored per group index, with only the changed index being written, so a view showing fewer groups does not clobber the states of groups only shown in another view.

To support this, `toSettingsValue` and `fromSettingsValue` were generalized to handle `QList<T>` and `QSet<T>` via a `SettingsValue` trait struct, replacing the previous per-type specializations.

Also activate the affected layouts when expanding a properties group, not just when collapsing, to avoid flickering caused by repainting before the relayout is done.


Aboracon:
Nice, it works flawlessly on my end!

PR #4562, opened 2026-07-01
Add a default world for standalone maps

kanishka0411:
This makes the resize handles of the world tool work on maps that are not part of any world.
Resizing only changes the map itself so it doesn't really need a world, and undo just works since it goes through the map's own undo stack.


bjorn:
I pushed a few fixes for problems I found during review:

* New Worlds created in response to adding a map were not getting added to the session. This was more an issue with existing code, since the list of worlds stored in the session was manually updates for some actions in MainWindow. Now it is updated based on the appropriate signals.

* Fixed the jumping of the view after resize by taking into account that the map stays at 0,0.

* Reduced code duplication by calling into the MainWindow for creating the new world. Not entirely sure if it's the right design, but it was a quick way to avoid two "New World" functions.

There are a few other potential issues:

* If the user selects the file of a world that's already loaded in the New World save dialog, `addEmptyWorld` fails with "World already loaded" in an error box. Probably we could just add the map to that world instead?
* In `addAnotherMapToWorld`, if the user creates the world but then cancels the "Load Map" dialog, they end up with a world containing only the current map. Maybe it's better to not do anything in this case?
* "Add another map to the current world" is not the correct naming for this action when there is no world yet.

In general, I wonder if it would be better to separate the creation of the world. I think automatically creating the world when using the "add map" action might be too confusing. When the "New World" dialog pops up I think many people see just a file dialog and interpret it as the dialog to select the map from. What do you think?

kanishka0411:
I think the confusion mostly comes from the save dialog showing up right after clicking "add map", since it's easy to mistake it for the map picker. Maybe instead of separating the world creation completely, we could show a small confirmation first, like "This map is not part of a world. Create a new world containing it?", then let the user pick the map, and only create the world file at the end. That way cancelling anywhere leaves nothing behind, which also fixes the cancel case you mentioned.
For the already-loaded world I agree, just adding the map to it makes more sense than showing an error. And for the naming, I'd change the text when there's no world yet to something like "Add map to a new world".
But if you think a fully separate "create world first" step is clearer, I can go with that also :-)

bjorn:
> Maybe instead of separating the world creation completely, we could show a small confirmation first, like "This map is not part of a world. Create a new world containing it?", then let the user pick the map, and only create the world file at the end.

I think that's worth a try. You can implement it and see how it feels. :-)

bjorn:
@kanishka0411 I've pushed a number of fixes and simplifications. One of the fixes was for issues I had previously introduced with the change to how the loaded worlds were remembered in the session. I'm sorry if you had rather done these things yourself based on feedback, but especially the smaller fixes and cleanups were easier for me this way.

I think the change looks good now, possibly apart from missing a dedicated icon, but this can be improved later. Please go over the changes, do a final pass of testing and let me know whether you think this is ready to be merged. :-)

bjorn (approved):

kanishka0411:
I tested it and it seems to work well ,I think it is ready to merge :)

bjorn:
@kanishka0411 Thanks for the nice additions!

PR #4569, opened 2026-07-16
New Hardware Renderer: Add object layer support

UltraDagon:
Adds support for the object layer type and its associated tools.

## Primary features:

Added objectgroupitem and objectsnode classes, which are similar to the existing tilelayeritem and tilesnode classes, for containing and rendering the map's objects.

Added the objectgroupmaterial class and objectgroup shader. These are needed to render the object texturepoint2d materials affected by opacity and tint modifiers.

Added the ObjectInteractionItem QML element and cpp class. This is similar to the tile regionoverlay QML element, but instead renders the various GUI elements needed to interact with objects. Examples: Object selection rect, polygon edit tool, and resize and rotation handles.

## Secondary features/changes:

Moved the tilesethelper struct to be its own class in its own file and added geometry-based object support. Geometry-based objects are ones like rectangles, polygons, ellipses, etc which have simple palettes (gray border, black shadow, and transparent gray fill) that the tileset helper holds and can use in place of an actual texture.

Several classes were modified to emit signals to support the primary features. These *could* be contained in `#ifdef TILEDQUICK_LIB`s, but it's not super necessary.

### DIFF
--- a/src/libtiledquick/mapitem.cpp
+++ b/src/libtiledquick/mapitem.cpp
@@ -72,17 +72,17 @@ void MapItem::setVisibleArea(const QRectF &visibleArea)
     emit visibleAreaChanged();
 }
 
-void MapItem::setScale(const qreal &scale)
+void MapItem::setZoom(const qreal &zoom)
 {
-    if (mScale == scale)
+    if (mZoom == zoom)
         return;
 
-    mScale = scale;
+    mZoom = zoom;
 
     for (auto objectGroup : std::as_const(mObjectGroupItems))
-        objectGroup->setMapScale(scale);
+        objectGroup->setZoom(zoom);
 
-    emit scaleChanged();
+    emit zoomChanged();
 }
 
 QRectF MapItem::boundingRect() const
@@ -205,6 +205,7 @@ void MapItem::refresh()
         if (Tiled::ObjectGroup *og = layer->asObjectGroup()) {
             ObjectGroupItem *groupItem = new ObjectGroupItem(og, mRenderer.get(), this);
             mObjectGroupItems.append(groupItem);
+            groupItem->setZoom(mZoom);
         }
     }
 
--- a/src/libtiledquick/mapitem.cpp
+++ b/src/libtiledquick/mapitem.cpp
@@ -21,6 +21,7 @@
 #include "mapitem.h"
 
 #include "tilelayeritem.h"
+#include "objectgroupitem.h"
 
 #include "map.h"
 #include "maprenderer.h"
@@ -43,12 +44,14 @@ void MapItem::setMap(Tiled::EditableMap *editableMap)
     if (mEditableMap && mEditableMap->mapDocument()) {
         Tiled::MapDocument *oldMapDocument = mEditableMap->mapDocument();
         disconnect(oldMapDocument, &Tiled::MapDocument::regionChanged, this, &MapItem::repaintRegion);
+        disconnect(oldMapDocument, &Tiled::MapDocument::mapObjectsChanged, this, &MapItem::repaintObjects);
         disconnect(oldMapDocument, &Tiled::MapDocument::mapResized, this, &MapItem::refresh);
     }
 
     if (editableMap && editableMap->mapDocument()) {
         Tiled::MapDocument *mapDocument = editableMap->mapDocument();
         connect(mapDocument, &Tiled::MapDocument::regionChanged, this, &MapItem::repaintRegion);
+        connect(mapDocument, &Tiled::MapDocument::mapObjectsChanged, this, &MapItem::repaintObjects);
         connect(mapDocument, &Tiled::MapDocument::mapResized, this, &MapItem::refresh);
     }
 
@@ -166,6 +169,9 @@ void MapItem::refresh()
     qDeleteAll(mTileLayerItems);
     mTileLayerItems.clear();
 
+    qDeleteAll(mObjectGroupItems);
+    mObjectGroupItems.clear();
+
     mRenderer = nullptr;
 
     if (!mMap)
@@ -177,6 +183,11 @@ void MapItem::refresh()
         if (Tiled::TileLayer *tl = layer->asTileLayer()) {
             TileLayerItem *layerItem = new TileLayerItem(tl, mRenderer.get(), this);
             mTileLayerItems.append(layerItem);
+        }
+
+        if (Tiled::ObjectGroup *og = layer->asObjectGroup()) {
+            ObjectGroupItem *groupItem = new ObjectGroupItem(og, mRenderer.get(), this);
+            mObjectGroupItems.append(groupItem);
         }
     }
 
@@ -191,6 +202,22 @@ void MapItem::repaintRegion(const QRegion &, Tiled::TileLayer *tileLayer)
             // TODO: Update only the region edited
             tileLayerItem->update();
             break;
+        }
+    }
+}
+
+void MapItem::repaintObjects(const QList<Tiled::MapObject*> &objects)
+{
+    QSet<Tiled::ObjectGroup*> changedGroups;
+    for (auto *object : objects)
+        changedGroups.insert(object->objectGroup());
+
+    // TODO: Update only subgroups containing affected tiles rather than entire layers
+    for (auto *objectGroupItem : std::as_const(mObjectGroupItems)) {
+        if (changedGroups.contains(objectGroupItem->group()))
+        {
+            // objectGroupItem->syncWithObjectGroup();
+            objectGroupItem->update();
         }
     }
 }
--- a/src/libtiledquick/mapitem.cpp
+++ b/src/libtiledquick/mapitem.cpp
@@ -26,6 +26,7 @@
 #include "map.h"
 #include "maprenderer.h"
 #include "tilelayer.h"
+#include "tilesethelper.h"
 
 using namespace TiledQuick;
 
@@ -34,7 +35,10 @@ MapItem::MapItem(QQuickItem *parent)
 {
 }
 
-MapItem::~MapItem() = default;
+MapItem::~MapItem()
+{
+    TilesetHelper::deleteInstance();
+};
 
 void MapItem::setMap(Tiled::EditableMap *editableMap)
 {
@@ -66,6 +70,19 @@ void MapItem::setVisibleArea(const QRectF &visibleArea)
 {
     mVisibleArea = visibleArea;
     emit visibleAreaChanged();
+}
+
+void MapItem::setScale(const qreal &scale)
+{
+    if (mScale == scale)
+        return;
+
+    mScale = scale;
+
+    for (auto objectGroup : std::as_const(mObjectGroupItems))
+        objectGroup->setMapScale(scale);
+
+    emit scaleChanged();
 }
 
 QRectF MapItem::boundingRect() const
--- a/src/libtiledquick/mapitem.cpp
+++ b/src/libtiledquick/mapitem.cpp
@@ -230,7 +230,10 @@ void MapItem::repaintObjects(const QList<Tiled::MapObject*> &objects)
     for (auto *object : objects)
         changedGroups.insert(object->objectGroup());
 
-    // TODO: Update only subgroups containing affected tiles rather than entire layers
+    if (changedGroups.size() == 0)
+        return;
+
+    // TODO: Update only nodes containing affected objects rather than entire layers
     for (auto *objectGroupItem : std::as_const(mObjectGroupItems)) {
         if (changedGroups.contains(objectGroupItem->group()))
         {
@@ -238,4 +241,6 @@ void MapItem::repaintObjects(const QList<Tiled::MapObject*> &objects)
             objectGroupItem->update();
         }
     }
+
+    emit mapObjectsChanged();
 }
--- a/src/libtiledquick/mapitem.cpp
+++ b/src/libtiledquick/mapitem.cpp
@@ -21,17 +21,16 @@
 #include "mapitem.h"
 
 #include "tilelayeritem.h"
-#include "objectgroupitem.h"
+#include "tilesethelper.h"
 
 #include "map.h"
 #include "maprenderer.h"
-#include "tilelayer.h"
-#include "tilesethelper.h"
 
 using namespace TiledQuick;
 
 MapItem::MapItem(QQuickItem *parent)
     : QQuickItem(parent)
+    , mNewObjectsPreviewItem(nullptr, nullptr, this)
 {
 }
 
@@ -81,8 +80,28 @@ void MapItem::setZoom(const qreal &zoom)
 
     for (auto objectGroup : std::as_const(mObjectGroupItems))
         objectGroup->setZoom(zoom);
+    mNewObjectsPreviewItem.setZoom(zoom);
+
+    // Since the tool brush is always fully rendered and does not
+    // receive the visibleAreaChanged signal upon zoom changes,
+    // we need to update it here.
+    mNewObjectsPreviewItem.update();
 
     emit zoomChanged();
+}
+
+void MapItem::setNewObjectsPreview(Tiled::ObjectGroup *objects)
+{
+    if (mNewObjectsPreviewItem.group() == objects)
+        return;
+
+    mNewObjectsPreviewItem.setObjectGroup(objects);
+    emit newObjectsPreviewChanged();
+}
+
+void MapItem::repaintPreview()
+{
+    mNewObjectsPreviewItem.update();
 }
 
 QRectF MapItem::boundingRect() const
@@ -195,6 +214,8 @@ void MapItem::refresh()
         return;
 
     mRenderer = Tiled::MapRenderer::create(mMap);
+    mNewObjectsPreviewItem.setRenderer(mRenderer.get());
+    mNewObjectsPreviewItem.setZoom(mZoom);
 
     for (Tiled::Layer *layer : mMap->layers()) {
         if (Tiled::TileLayer *tl = layer->asTileLayer()) {
@@ -224,18 +245,18 @@ void MapItem::repaintRegion(const QRegion &, Tiled::TileLayer *tileLayer)
     }
 }
 
-void MapItem::repaintObjects(const QList<Tiled::MapObject*> &objects)
+void MapItem::repaintObjects(const QList<Tiled::MapObject*> &objects, ObjectGroup *group)
 {
     QSet<Tiled::ObjectGroup*> changedGroups;
-    for (auto *object : objects)
+    for (auto object : objects)
         changedGroups.insert(object->objectGroup());
 
-    if (changedGroups.size() == 0)
+    if (changedGroups.size() == 0 && !group)
         return;
 
     // TODO: Update only nodes containing affected objects rather than entire layers
-    for (auto *objectGroupItem : std::as_const(mObjectGroupItems)) {
-        if (changedGroups.contains(objectGroupItem->group()))
+    for (auto objectGroupItem : std::as_const(mObjectGroupItems)) {
+        if (objectGroupItem->group() == group || changedGroups.contains(objectGroupItem->group()))
         {
             // objectGroupItem->syncWithObjectGroup();
             objectGroupItem->update();
--- a/src/libtiledquick/mapitem.cpp
+++ b/src/libtiledquick/mapitem.cpp
@@ -46,16 +46,18 @@ void MapItem::setMap(Tiled::EditableMap *editableMap)
 
     if (mEditableMap && mEditableMap->mapDocument()) {
         Tiled::MapDocument *oldMapDocument = mEditableMap->mapDocument();
+        disconnect(oldMapDocument, &Tiled::MapDocument::mapResized, this, &MapItem::refresh);
         disconnect(oldMapDocument, &Tiled::MapDocument::regionChanged, this, &MapItem::repaintRegion);
         disconnect(oldMapDocument, &Tiled::MapDocument::mapObjectsChanged, this, &MapItem::repaintObjects);
-        disconnect(oldMapDocument, &Tiled::MapDocument::mapResized, this, &MapItem::refresh);
+        disconnect(oldMapDocument, &Tiled::MapDocument::tileLayerChanged, this, &MapItem::onTileLayerChanged);
     }
 
     if (editableMap && editableMap->mapDocument()) {
         Tiled::MapDocument *mapDocument = editableMap->mapDocument();
+        connect(mapDocument, &Tiled::MapDocument::mapResized, this, &MapItem::refresh);
         connect(mapDocument, &Tiled::MapDocument::regionChanged, this, &MapItem::repaintRegion);
         connect(mapDocument, &Tiled::MapDocument::mapObjectsChanged, this, &MapItem::repaintObjects);
-        connect(mapDocument, &Tiled::MapDocument::mapResized, this, &MapItem::refresh);
+        connect(mapDocument, &Tiled::MapDocument::tileLayerChanged, this, &MapItem::onTileLayerChanged);
     }
 
     mEditableMap = editableMap;
@@ -273,4 +275,10 @@ void MapItem::repaintObjects(const QList<Tiled::MapObject*> &objects, ObjectGrou
     }
 
     emit mapObjectsChanged();
+}
+
+void MapItem::onTileLayerChanged(Tiled::TileLayer *layer, Tiled::MapDocument::TileLayerChangeFlags) {
+    for (auto layerItem : std::as_const(mTileLayerItems))
+        if (layerItem->layer() == layer && layerItem->isVisible() != layer->isVisible())
+            layerItem->layerVisibilityChanged();
 }
--- a/src/libtiledquick/mapitem.cpp
+++ b/src/libtiledquick/mapitem.cpp
@@ -99,6 +99,15 @@ void MapItem::setNewObjectsPreview(Tiled::ObjectGroup *objects)
     emit newObjectsPreviewChanged();
 }
 
+void MapItem::setMousePos(const QPointF &pos)
+{
+    if (mNewObjectsPreviewItem.mousePos() == pos)
+        return;
+
+    mNewObjectsPreviewItem.setMousePos(pos);
+    emit mousePosChanged();
+}
+
 void MapItem::repaintPreview()
 {
     mNewObjectsPreviewItem.update();
--- a/src/libtiledquick/objectgroupitem.cpp
+++ b/src/libtiledquick/objectgroupitem.cpp
@@ -149,31 +149,68 @@ QSGNode *ObjectGroupItem::updatePaintNode(QSGNode *node,
                 objectData.clear();
             }
 
-            helper.setTileset(cell.tileset());
+            if (!cell.isEmpty())
+                helper.setTileset(cell.tileset());
         }
 
         ObjectData data;
+
+        switch (object->shape())
+        {
+        case MapObject::Shape::Rectangle:
+            if (object->isTileObject())
+                data.type = ObjectGroupMaterial::ObjectType::Tile;
+            else
+                data.type = ObjectGroupMaterial::ObjectType::Rectangle;
+            break;
+        case MapObject::Shape::Polygon:
+            data.type = ObjectGroupMaterial::ObjectType::Polygon;
+            break;
+        case MapObject::Shape::Polyline:
+            data.type = ObjectGroupMaterial::ObjectType::Polyline;
+            break;
+        case MapObject::Shape::Ellipse:
+            data.type = ObjectGroupMaterial::ObjectType::Ellipse;
+            break;
+        case MapObject::Shape::Capsule:
+            data.type = ObjectGroupMaterial::ObjectType::Capsule;
+            break;
+        case MapObject::Shape::Text:
+            data.type = ObjectGroupMaterial::ObjectType::Text;
+            break;
+        case MapObject::Shape::Point:
+            data.type = ObjectGroupMaterial::ObjectType::Point;
+            break;
+        }
+
         data.rotation = object->rotation();
         data.x = object->x() - x();
-        data.y = object->y() - y() - object->height();
+        data.y = object->y() - y() - ((data.type == ObjectGroupMaterial::ObjectType::Tile) ? object->height() : 0);
         data.width = object->width();
         data.height = object->height();
-        data.twidth = cell.tile()->height();
-        data.theight = cell.tile()->width();
-        data.flippedHorizontally = cell.flippedHorizontally();
-        data.flippedVertically = cell.flippedVertically();
-        data.shape = object->shape();
-        data.isTileObject = object->isTileObject();
+
+        if (!cell.isEmpty()) {
+            data.twidth = cell.tile()->height();
+            data.theight = cell.tile()->width();
+            data.flippedHorizontally = cell.flippedHorizontally();
+            data.flippedVertically = cell.flippedVertically();
+        } else {
+            data.twidth = 0;
+            data.theight = 0;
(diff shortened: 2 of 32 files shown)


UltraDagon:
Pretty big initial commit! A lot of it is adding the shader files and re-using the TileLayerItem code as a template. After confirmation on how this looks so far, I plan on adding the various shapes.

UltraDagon:
With this commit, I added `alignas(4)` to the objectgroup uniform buffers to avoid the manual padding that was previously used. I think it will be a good standard to keep, and plan on adding it to the mapgrid ubuf when that is revisited for the non-orthogonal map support.

src/libtiledquick/objectgroupitem.cpp
```diff
@@ -0,0 +1,291 @@
...
+
+        data.rotation = object->rotation();
+        data.x = object->x() - x();
+        data.y = object->y() - y() - ((data.type == ObjectGroupMaterial::ObjectType::Tile) ? object->height() : 0);
+        data.width = object->width();
+        data.height = object->height();
+
+        if (!cell.isEmpty()) {
+            data.twidth = cell.tile()->height();
+            data.theight = cell.tile()->width();
```
bjorn: width/height are swapped here.

src/libtiledquick/objectgroupitem.cpp
```diff
@@ -0,0 +1,291 @@
...
+            data.flippedHorizontally = cell.flippedHorizontally();
+            data.flippedVertically = cell.flippedVertically();
+        } else {
+            data.twidth = 0;
+            data.theight = 0;
+            data.flippedHorizontally = false;
+            data.flippedVertically = false;
+        }
+
+        data.alpha = object->opacity() * mGroup->opacity() * mGroup->tintColor().alpha();
```
bjorn: Layer opacity is already applied on the item (and its children) in the constructor, so `mGroup->opacity()` should be removed here.

src/libtiledquick/objectsnode.cpp
```diff
@@ -0,0 +1,141 @@
...
+        // BottomLeft                   // BottomRight
+        v[3].x = data.x;                v[4].x = data.x + data.width;
+        v[3].y = data.y + data.height;  v[4].y = data.y + data.height;
+        v[3].tx = s_tx;                 v[4].tx = s_tx + s_width;
+        v[3].ty = s_ty + s_height;      v[4].ty = s_ty + s_height;
+
+        // ObjectGroupMaterial
+        for (int i = 0; i < 6; i++) {
+            v[i].local_x = (v[i].x - data.x) / data.width;
+            v[i].local_y = (v[i].y - data.y) / data.width;
```
bjorn:
```suggestion
            v[i].local_y = (v[i].y - data.y) / data.height;
```

src/libtiledquick/objectgroupitem.cpp
```diff
@@ -0,0 +1,291 @@
...
+{
+    const bool visible = mGroup->isVisible();
+    setVisible(visible);
+
+    MapItem *parent = qobject_cast<MapItem*>(parentItem());
+    if (visible) {
+        updateVisibleObjects();
+
+        if (parent)
+            connect(parent, &MapItem::visibleAreaChanged, this, &ObjectGroupItem::updateVisibleObjects);
```
bjorn, line 260: Is the plan to only create nodes for objects within the visible area? Right now, this connection causes the rebuild of all nodes on every little view position or scale change. I think it's better to leave that out until it is would be actually used.

src/libtiledquick/objectgroupmaterial.cpp
```diff
@@ -0,0 +1,101 @@
...
+public:
+    ObjectGroupShader()
+    {
+        setShaderFileName(VertexStage, QStringLiteral(":/objectgroup.vert.qsb"));
+        setShaderFileName(FragmentStage, QStringLiteral(":/objectgroup.frag.qsb"));
+    }
+
+    bool updateUniformData(RenderState &state, QSGMaterial *, QSGMaterial *) override
+    {
+        if (!state.isMatrixDirty())
```
bjorn: This early if means an opacity-only change (`state.isOpacityDirty()`) never reaches the uniform buffer. See similar thing we dealt with in `MapGridShader`.

src/libtiledquick/objectgroupitem.cpp
```diff
@@ -0,0 +1,291 @@
...
+    QSGTexture *texture = cache.value(tileset);
+    if (!texture) {
+        const QString imagePath(Tiled::urlToLocalFileOrQrc(tileset->imageSource()));
+        texture = window->createTextureFromImage(QImage(imagePath));
+        cache.insert(tileset, texture);
+    }
+    return texture;
+}
+
+struct TilesetHelper
```
bjorn: Of course we'll need to make sure to use only a single `TilesetHelper` implementation, and even only a single instance and single texture cache, to avoid creating multiple textures for the same tileset.

src/libtiledquick/objectsnode.h
```diff
@@ -0,0 +1,83 @@
...
+    unsigned char tint_r, tint_g, tint_b, alpha;
+    float object_type;
+};
+
+class TILEDQUICK_SHARED_EXPORT ObjectsNode : public QSGGeometryNode
+{
+public:
+    enum {
+        // TODO: Make actual max count
+        MaxObjectCount = 100
```
bjorn: `TilesNode::MaxTileCount = 65536 / 6` exists because of 16-bit index buffers; `ObjectsNode` is non-indexed, so I think we can just drop this ceiling.

src/libtiledquick/objectsnode.h
```diff
@@ -0,0 +1,83 @@
+/*
+ * tilesnode.h
```
bjorn:
```suggestion
 * objectsnode.h
```

src/libtiledquick/objectgroupmaterial.h
```diff
@@ -0,0 +1,75 @@
...
+ * this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include <QSGMaterial>
+
+#include "tiledquick_global.h"
+
+struct alignas(4) ObjectGroupUniformBuffer {
```
bjorn: Doesn't need to be in the header.

src/libtiledquick/objectsnode.cpp
```diff
@@ -0,0 +1,141 @@
...
+ * this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "objectsnode.h"
+
+namespace TiledQuick {
+
+static const QSGGeometry::Attribute ObjectAttributes[] = {
+    QSGGeometry::Attribute::create(0, 2, QSGGeometry::FloatType, QSGGeometry::PositionAttribute), // Global Position
+    QSGGeometry::Attribute::create(1, 2, QSGGeometry::FloatType, QSGGeometry::PositionAttribute), // Local Position
```
bjorn: I think we should only have one `PositionAttribute`.

bjorn (changes requested):
Some rough feedback while this is still in draft:

Some checks are missing that might lead to crashes (to fix these, only call `setTextureCoordinates` for non-empty cells, never construct an `ObjectsNode` batch that needs a texture it does not have, and make `processObjectData` tolerate a null texture, since shape objects need no texture at all). Also `cell.tile()` might return `nullptr` and needs to be checked.

I'm not sure exactly what is your plan for rendering shape objects with the `ObjectGroupMaterial`, but my first thought is that the [`Shape`](https://doc.qt.io/qt-6/qml-qtquick-shapes-shape.html) QML item should be suitable for rendering those. Of course that'd require instantiating each object from QML instead.

UltraDagon:
I should have mentioned it last night, but Adam and I talked for a while, and we're going to try to render the objects by mapping the vertices in each object's process[type]ObjectData. With my previous push, I had planned on rendering the various object types using shaders, so I'll be commenting out a lot of code related to that.

src/libtiledquick/objectgroupitem.cpp
UltraDagon: I moved this to tilesethelper.h so that it could share an instance between classes. This is my first time creating one of these types of classes, so let me know if anything looks wrong.

UltraDagon:
With this last commit, I have brought back the visibleAreaChanged listener, as the object borders will need to be redrawn to remain the same thickness depending on the camera zoom. I'm not sure if there is a way to avoid this, but if there is I would be happy to implement it.

Also, in ObjectsNode::processObjectData(), there is a const int precision that changes the number of object vertices used to draw the curves of capsules and ellipses. Precision is currently based on the camera zoom, but it should also be based on the object's size. I'm not sure if there is an entirely "correct" way to find this precision number, but I thought I'd bring it up for discussion before setting it to something that looked good to me and leaving it at that.

~Finally, with these objects, there was an unintended quirk that I thought could be useful as a real feature. In the current Tiled renderer, geometry objects are not affected by tint. The new renderer, as of now, does apply this tint to its drawn objects, since they use the ObjectTexturedPoint2D vertex data. This can be reverted to default behavior pretty easily, but I thought it could be a useful feature for helping users categorize different objects based on layer.~
Nevermind! I found that the Object Layer Color is what determines this, will get it fixed in the next push.

UltraDagon:
Just pushed the text and polygon objects! And with those come a few more notes. 

1. The text objects currently redraw their textures using QPainter every update. We could either cache them, perhaps using the tilesethelper or the objectGroupItem, or just draw them on update (eventually only drawing them when they're visible)

2. The text objects need to know the zoom/scale in order to be drawn at the proper resolution (fontDetail in objectgroupitem.cpp). This isn't a very big issue if we draw them every frame, but if we cache, then they will still need to be either redrawn or cached at a high resolution.

3. The polygons are not finished and are only able to draw convex polygons. I talked with Adam about this for a bit, but I was wondering if you had any ideas on how to best draw the fill triangles for concave or self-intersecting polygons. I did have the idea to use the QPainterPath in order to split self-intersecting polygons into a list of non-self-intersecting polygons, but it was looking like the concave polygon problem would take a relatively large amount of code to implement by hand. Just let me know what you think.

src/libtiledquick/objectgroupitem.cpp
```diff
@@ -0,0 +1,303 @@
...
+        }
+
+        data.zoom = mZoom;
+        data.polygon = object->polygon();
+        if (data.type == ObjectGroupMaterial::ObjectType::Polyline)
+            data.polygon.append(mousePos() - object->position());
+
+        if (!cell.isEmpty()) {
+            data.twidth = cell.tile()->width();
+            data.theight = cell.tile()->height();
```
bjorn: This should probably be something as follows:
```suggestion
        if (Tile *tile = cell.tile()) {
            data.twidth = tile->width();
            data.theight = tile->height();
```
Because just checking `!cell.isEmpty()` does not suffice to make sure the tile actually exists in the tileset.

src/libtiledquick/objectgroupitem.cpp
```diff
@@ -0,0 +1,303 @@
...
+    node->setFlag(QSGNode::OwnedByParent);
+
+    if (!mGroup)
+        return node;
+
+    TilesetHelper helper = TilesetHelper::instance(static_cast<MapItem*>(parentItem()));
+
+    QVector<ObjectData> objectData;
+
+    for (auto object : *mGroup) {
```
bjorn, line 97: How do you plan to avoid rendering hidden objects? Probably here we could insert an `object->isVisible()` check.

src/libtiledquick/objectsnode.cpp
```diff
@@ -0,0 +1,609 @@
...
+    v[b+3].tx = v[b+1].tx;                  v[b+4].tx = border_tx;
+
+    // ObjectGroupMaterial
+    for (int i = 0; i < 6; i++) {
+        v[b+i].tint_r = data.tint_r;
+        v[b+i].tint_g = data.tint_g;
+        v[b+i].tint_b = data.tint_b;
+        v[b+i].alpha = data.alpha;
+
+        v[i].ty = 0;
```
bjorn: Looks like this should be:
```suggestion
        v[b+i].ty = 0;
```

src/libtiledquick/objectsnode.cpp
```diff
@@ -0,0 +1,609 @@
...
+        case ObjectGroupMaterial::ObjectType::Capsule: {
+            const int edges = 2*(precision/2 + 1);
+            totalVertices += 3*(edges-2) + 12*(edges);
+            break;
+        }
+        case ObjectGroupMaterial::ObjectType::Text:
+            totalVertices += 6;
+            break;
+        case ObjectGroupMaterial::ObjectType::Point:
+            totalVertices += 3*8 + (3+6+6)*(8+3);
```
bjorn: This doesn't match the amount of vertices initialized by `processPointData` for point objects (which uses `precision`). I wonder if we could bring these calculations closer to the code that they need to match?

src/libtiledquick/objectsnode.cpp
```diff
@@ -0,0 +1,609 @@
...
+                           qBound(12, int(18 / sqrt(objectData[0].zoom)), 56));
+
+    int totalVertices = 0;
+    for (const ObjectData &data : objectData) {
+        switch (data.type) {
+        case ObjectGroupMaterial::ObjectType::Rectangle:
+            totalVertices += 6 + 24 + 24;
+            break;
+        case ObjectGroupMaterial::ObjectType::Polygon:
+            totalVertices += 3*(data.polygon.size()-2) + 12*data.polygon.size();
```
bjorn, line 471: We should make sure to explicitly ignore polygons with size < 2 rather than risking corrupting memory.

src/libtiledquick/objectsnode.cpp
```diff
@@ -0,0 +1,609 @@
...
+/**
+ * Draws a border and shadow between two points.
+ *
+ * If totalBorderSegments is 0, no shadow will be drawn.
+ */
+static void processBorderSegment(const ObjectData &data, ObjectTexturedPoint2D *&v, const QPointF start, const QPointF end, const float thickness, const float border_tx, const int totalBorderSegments = 1, const float shadow_tx = 2.5, const QPointF shadowOffset = {1, 1}, const bool extendCorners = false)
+{
+    const float lineLength = sqrt(pow(end.x() - start.x(), 2) + pow(end.y() - start.y(), 2));
+    const QPointF thicknessOffset = thickness/2 * QPointF((start.y() - end.y())/lineLength,
+                                               (end.x() - start.x())/lineLength);
```
bjorn: Zero `lineLength` can happen in several cases and should be handled rather than dividing by zero and producing NaN geometry.

src/tiled/mapeditor.cpp
```diff
@@ -451,6 +452,12 @@ void MapEditor::setCurrentDocument(Document *document)
 //                this, &MapEditor::updateActions);
         connect(mapDocument, &MapDocument::selectedAreaChanged,
                 this, &MapEditor::selectedRegionChanged);
+        connect(mapDocument, &MapDocument::selectedObjectsChanged,
+                this, &MapEditor::selectedObjectsChanged);
```
bjorn: The act of switching documents itself should also emit `selectedObjectsChanged` to trigger a refresh of the QML bindings.

Btw, why introduce another property/signal for this rather than using `EditableMap.selectedObjects`?

bjorn (changes requested):
Great to see object support becoming functional in the new renderer! Though I must say, for some reason I don't actually see any object render in my tests so far. Could it be related to these warnings I see printed in the terminal?

```
Vertex output name 'out_tc' does not match fragment input 'in_tc'. This should be avoided because it causes problems with older GLSL versions.
Vertex output name 'out_tint_alpha' does not match fragment input 'in_tint_alpha'. This should be avoided because it causes problems with older GLSL versions.
```

I only see the selection outlines for objects for now (which look great, though). I wrote some further feedback in a few comments, though I couldn't look at the entire patch in detail yet.

UltraDagon:
> Great to see object support becoming functional in the new renderer! Though I must say, for some reason I don't actually see any object render in my tests so far. Could it be related to these warnings I see printed in the terminal?
> 
> ```
> Vertex output name 'out_tc' does not match fragment input 'in_tc'. This should be avoided because it causes problems with older GLSL versions.
> Vertex output name 'out_tint_alpha' does not match fragment input 'in_tint_alpha'. This should be avoided because it causes problems with older GLSL versions.
> ```
> 
> I only see the selection outlines for objects for now (which look great, though). I wrote some further feedback in a few comments, though I couldn't look at the entire patch in detail yet.

Hmm, in this next patch I'll have changed the vert output names to match the frag input names (which is what the grid does), so hopefully that will fix it.

src/tiled/mapeditor.cpp
UltraDagon: I had previously tried using EditableMap::selectedObjects by qobject_casting to EditableMapObject, but that led to crashes due to linking errors, so I thought it would be better to just get the MapObjects directly. After revisiting, I found that I could use dynamic_cast just fine to get the MapObjects from the EditableMap.

Edit: I did leave the old implementation as comments in case my patch has issues, but they will be removed once the patch is approved

Edit 2: I forgot about TILED_EDITOR_EXPORT! Fixed this whole issue in my latest push

UltraDagon:
Un-drafted and finished up the pr's description. I think we're in the final stretch for the object layer, so I'm going to get work started on the non-orthogonal support.

src/libtiledquick/objectsnode.cpp
bjorn: I see the double switch is necessary because you're processing all `ObjectData` instances in a loop and calculating the total memory necessary to allocate. It's good for rendering large amount of objects fast, but I wonder a little about the scalability of adjusting to changes. We'll see. :-)

src/tiled/mapeditor.cpp
bjorn: Nice that you've found the solution to the linker issue! I've removed the leftover commented out code.

src/libtiledquick/objectsnode.cpp
```diff
@@ -0,0 +1,611 @@
...
+    if (data.flippedVertically) {
+        std::swap(v[0].ty, v[4].ty);
+        std::swap(v[1].ty, v[5].ty);
+        std::swap(v[2].ty, v[3].ty);
+    }
+
+    if (data.rotation != 0) {
+        const double r = data.rotation * M_PI / 180;
+        const double origin_x = data.x;
+        const double origin_y = data.y + data.height;
```
bjorn, line 598: This is not always the right origin. For tiles, there's an alignment option that affects tile objects on the tileset and text objects have their origin in the top-left instead of bottom-left.

src/libtiledquick/objectgroupitem.cpp
```diff
@@ -0,0 +1,308 @@
...
+            font.setPixelSize(font.pixelSize()*fontDetail);
+            QImage textImage(object->width()*fontDetail, object->height()*fontDetail, QImage::Format_RGBA8888);
+            textImage.fill(Qt::transparent);
+
+            QPainter painter(&textImage);
+            painter.setFont(font);
+            painter.setPen(textData.color);
+            painter.drawText(textImage.rect(), textData.text, textData.textOption());
+
+            textTexture = window()->createTextureFromImage(textImage);
```
bjorn, line 161: This is not optimal, as you noted, but the `textTexture` will at least need to be deleted rather than leaked. It could be done by having `ObjectsNode` optionally take ownership over the texture, like:
```c++
class ObjectsNode : public QSGGeometryNode
{
public:
    ObjectsNode(QSGTexture *texture, const QVector<ObjectData> &objectData,
                bool ownsTexture = false);
    ~ObjectsNode() override { if (mOwnsTexture) delete mMaterial.texture(); }
```
Though we should probably raise the minimum Qt version for the hardware renderer to Qt 6.7 so we can use `QSGTextNode` (can be a follow-up change).

src/tiled/mapdocument.cpp
```diff
@@ -1475,6 +1475,26 @@ void MapDocument::onChanged(const ChangeEvent &change)
 
         break;
     }
+    case ChangeEvent::MapObjectsAdded:
+    case ChangeEvent::MapObjectsChanged: {
+        const auto &mapObjects = static_cast<const MapObjectsChangeEvent&>(change).mapObjects;
```
bjorn, line 1480: In case of `MapObjectsAdded` this is undefined behavior. Should be:
```suggestion
        const auto &mapObjects = static_cast<const MapObjectsEvent&>(change).mapObjects;
```

src/tiled/mapdocument.cpp
```diff
@@ -1475,6 +1475,26 @@ void MapDocument::onChanged(const ChangeEvent &change)
...
+        emit mapObjectsChanged(mapObjects);
+        break;
+    }
+    case ChangeEvent::MapObjectRemoved:
```
bjorn, line 1484: This case should be moved along with the other map object related change events handled above, it's not an `ObjectGroupChangeEvent`.

src/libtiledquick/objectgroupitem.cpp
```diff
@@ -0,0 +1,308 @@
...
+            data.flippedHorizontally = cell.flippedHorizontally();
+            data.flippedVertically = cell.flippedVertically();
+        } else {
+            data.twidth = 0;
+            data.theight = 0;
+            data.flippedHorizontally = false;
+            data.flippedVertically = false;
+        }
+
+        data.alpha = object->opacity() * mGroup->tintColor().alpha();
```
bjorn, line 202: I think that should be `effectiveTintColor`, and please only call it once in that case since it can be a little more expensive.

src/libtiledquick/objectgroupitem.cpp
```diff
@@ -0,0 +1,308 @@
...
+
+        if (mGroup->tintColor().isValid() && data.type == ObjectGroupMaterial::ObjectType::Tile) {
+            data.tint_r = mGroup->tintColor().red();
+            data.tint_g = mGroup->tintColor().green();
+            data.tint_b = mGroup->tintColor().blue();
+        } else if (mGroup->color().isValid()) {
```
bjorn, line 208: This color should not apply to tile objects, regardless of whether the tint color is set.

src/libtiledquick/tilesethelper.cpp
```diff
@@ -0,0 +1,108 @@
+/*
+ * tilesethelper.cpp
+ * Copyright 2026, UltraDagon
```
bjorn, line 3: I think you should keep my copyright line here as well because it was moved code:
```
 * Copyright 2014, Thorbjørn Lindeijer <bjorn@lindeijer.nl>
```

src/libtiledquick/tilesethelper.cpp
```diff
@@ -0,0 +1,108 @@
...
+    if (cell.isEmpty())
+        return;
+
+    const int tileId = cell.tileId();
+    const int column = tileId % mTilesPerRow;
+    const int row = tileId / mTilesPerRow;
+
+    tx = column * mTileHSpace + mMargin;
+    ty = row * mTileVSpace + mMargin;
+}
```
bjorn, line 108: Please make sure the file ends with a newline character.

src/libtiledquick/objectgroupitem.cpp
```diff
@@ -0,0 +1,308 @@
...
+                                          QQuickItem::UpdatePaintNodeData *)
+{
+    delete node;
+    node = new QSGNode;
+    node->setFlag(QSGNode::OwnedByParent);
+
+    if (!mGroup || !mGroup->isVisible())
+        return node;
+
+    TilesetHelper helper = TilesetHelper::instance(static_cast<MapItem*>(parentItem()));
```
bjorn, line 93: This copies the instance, if you want to use it as singleton it should be:
```suggestion
    TilesetHelper &helper = TilesetHelper::instance(static_cast<MapItem*>(parentItem()));
```
Though I haven't tested whether this introduces issues. Same issue in `TileLayerItem::updatePaintNode`.

src/libtiledquick/mapitem.cpp
```diff
@@ -21,19 +21,23 @@
...
 MapItem::MapItem(QQuickItem *parent)
     : QQuickItem(parent)
+    , mNewObjectsPreviewItem(nullptr, nullptr, this)
 {
 }
 
-MapItem::~MapItem() = default;
+MapItem::~MapItem()
+{
+    TilesetHelper::deleteInstance();
```
bjorn, line 39: I don't think we can delete the `TilesetHelper` when destroying any of the `MapItem` instances, and leak the palette texture?

src/tiled/mapeditor.cpp
```diff
@@ -451,6 +454,10 @@ void MapEditor::setCurrentDocument(Document *document)
 //                this, &MapEditor::updateActions);
         connect(mapDocument, &MapDocument::selectedAreaChanged,
                 this, &MapEditor::selectedRegionChanged);
+        connect(mapDocument, &MapDocument::aboutToBeSelectedObjectsChanged,
+                this, &MapEditor::aboutToBeSelectedObjectsChanged);
+        connect(mapDocument, &MapDocument::hoveredMapObjectChanged,
+                this, [this, mapDocument] {emit aboutToBeSelectedObjectsChanged(QList<MapObject*>{mapDocument->hoveredMapObject()}); });
```
bjorn, line 460: We'll want to avoid emitting this signal with a `{ nullptr }` value when no longer hovering an object.

src/libtiledquick/objectinteractionitem.cpp
bjorn, line 1: Copyright header missing on this file and its header.

src/libtiledquick/objectgroupitem.cpp
```diff
@@ -0,0 +1,308 @@
...
+        const QSize tileSize = mapItem->map()->map()->tileSize();
+        const QSize size = tile->size();
+        const QPoint offset = tileset->tileOffset();
+
+        QVector<ObjectData> data(1);
+        data[0].x = mPosition.x() * tileSize.width() + offset.x();
+        data[0].y = (mPosition.y() + 1) * tileSize.height() - tileset->tileHeight() + offset.y();
+        data[0].width = size.width();
+        data[0].height = size.height();
+        helper.setTextureCoordinates(data[0].ty, data[0].ty, mCell);
```
bjorn, line 302: Probably should be:
```suggestion
        helper.setTextureCoordinates(data[0].tx, data[0].ty, mCell);
```
But it seems `ObjectItem` is not used anywhere, so maybe it should be removed?

src/tiled/resources/qml/mapview.qml
```diff
@@ -108,13 +114,47 @@ Rectangle {
                 mapRect: Qt.rect(0, 0, mapItem.map.width, mapItem.map.height)
                 tileSize: Qt.point(mapItem.map.tileWidth, mapItem.map.tileHeight)
             }
+
+            ObjectInteractionItem {
+                id: objectInteractionItem
```
bjorn, line 119: Not that it matters at the moment, but this item should probably also have `anchors.fill: mapItem`.

src/libtiledquick/objectgroupitem.h
```diff
@@ -0,0 +1,95 @@
...
+                    MapItem *parent);
+
+    void setObjectGroup(Tiled::ObjectGroup *group);
+    void setRenderer(Tiled::MapRenderer *renderer);
+
+    QPointF mousePos() const;
+    void setMousePos(const QPointF &pos);
+
+    void syncWithObjectGroup();
+    void setZoom(const qreal &zoom);
```
bjorn, line 52: Please pass (and return) `qreal` by value in all cases. It's no larger than a pointer so passing a reference just complicates things.
```suggestion
    void setZoom(qreal zoom);
```

src/tiled/mapeditor.cpp
```diff
@@ -418,6 +419,8 @@ void MapEditor::setCurrentDocument(Document *document)
         mCurrentMapDocument->disconnect(this);
     }
 
+    if (mCurrentMapDocument)
+        mCurrentMapDocument->setSelectedObjects({});
```
bjorn, line 423: We shouldn't clear the selection when switching documents. Why was this added?

bjorn (commented):
Some further comments based on a closer look. :-)

src/libtiledquick/objectsnode.cpp
```diff
@@ -0,0 +1,611 @@
...
+    const float thickness = 2*data.zoom;
+    const QPointF shadowOffset = QPointF(thickness * 2/3, thickness * 2/3);
+
+    QList<QPointF> points;
+    for (QPointF p : data.polygon)
+        points.append(p + QPointF(data.x, data.y));
+    points.append(points[0]);
+
+    // Fill
+    // TODO: Add convex and intersecting polygon support
```
bjorn, line 244: Do you mean concave? I noticed the fill is not working correctly in that case:

Image

But it works fine for convex polygons.

src/libtiledquick/objectgroupitem.cpp
```diff
@@ -0,0 +1,308 @@
...
+            data.x -= 10;
+            data.y -= 10;
+            data.width += 20;
+            data.height += 20;
+        }
+
+        data.zoom = mZoom;
+        data.polygon = object->polygon();
+        if (data.type == ObjectGroupMaterial::ObjectType::Polyline)
+            data.polygon.append(mousePos() - object->position());
```
bjorn, line 188: This preview point gets appended for every polyline object in every object layer, not just the one being created.

Image

So this should be gated on the item actually being the preview. Either a flag on `ObjectGroupItem`, or have the tool append the point to the preview object itself so this path stays generic.

src/libtiledquick/objectinteractionitem.cpp
```diff
@@ -0,0 +1,241 @@
...
+    if (mSelectedToolId != QStringLiteral("EditPolygonTool"))
+        return points;
+
+    for (auto object : std::as_const(mSelectedObjects)) {
+        auto mapObject = qobject_cast<Tiled::EditableMapObject*>(object)->mapObject();
+        if (!mapObject)
+            continue;
+
+        for (QPointF point : mapObject->polygon())
+            points.append(point + mapObject->position());
```
bjorn, line 165: `EditPolygonTool` places the real handles with `rotateAt(objectScreenPos, object->rotation())` and `absolutePositionForLayer(...)`. So on a rotated polygon, or one in a group with an offset, the plain dots sit at the unrotated positions while the selected and highlighted ones are correct, since those come from `handle->pos()`.

Image

Rather than repeating the transform here, you could probably expose the full handle point list from the tool, the same way `selectedHandlePoints` and `highlightedHandlePoints` already do.

src/libtiledquick/tilesethelper.h
```diff
@@ -0,0 +1,74 @@
...
+    };
+
+    TilesetHelper(const MapItem *mapItem);
+
+    static TilesetHelper &instance(const MapItem *mapItem);
+
+    static void deleteInstance();
+
+    inline Tileset *tileset() const { return mTileset; }
+    inline QSGTexture *texture() const { return (mTexture ? mTexture : mObjectTexture); }
```
bjorn, line 56: The `mTexture ? mTexture : mObjectTexture` fallback makes the shape/tile batching work, but it also means `texture()` can never return null, so the `if (!helper.texture()) return;` guard in `tilelayeritem.cpp:91` is useless at the moment.

I'd rather see the object palette move out of `TilesetHelper`: it is a global constant, not something cached per tileset, and `TilesetHelper::Fill` and friends can move along. It could go into `MapItem` or its own small shared resource. The object loop can choose between the two explicitly before deciding whether it can continue batching.

Something like:

```c++
QSGTexture *wanted = cell.tile() ? helper.tilesetTexture() : objectPalette;
if (wanted != batchTexture) { flush(); batchTexture = wanted; }
```

bjorn (changes requested):
A few more things I ran into.

PR #4572, opened 2026-07-19
Fixed view jumping when undoing map resizes and moves

kanishka0411:
When undoing a resize from the left/top edge, the resize gets undone but the whole map jumps in the view. Same problem when undoing a map move in a world, the other maps move in the view instead of the moved one going back. This happens because the view was only adjusted when the change was made through the tool, so undo/redo never adjusted it.

Now the ResizeMap command reports the contents shift through the mapResized signal and MapScene adjusts the view when it happens. For maps that are in a world, refreshScene compensates for the change in world position instead. So the view stays steady no matter where the change came from.

### DIFF
--- a/src/tiled/mapscene.cpp
+++ b/src/tiled/mapscene.cpp
@@ -31,6 +31,7 @@
 #include "map.h"
 #include "mapobject.h"
 #include "maprenderer.h"
+#include "mapview.h"
 #include "objectgroup.h"
 #include "objecttemplate.h"
 #include "preferences.h"
@@ -117,6 +118,8 @@ void MapScene::setMapDocument(MapDocument *mapDocument)
                 this, [this] { update(); });
         connect(mMapDocument, &MapDocument::tilesetReplaced,
                 this, &MapScene::tilesetReplaced);
+        connect(mMapDocument, &MapDocument::mapResized,
+                this, &MapScene::mapResized);
     }
 
     refreshScene();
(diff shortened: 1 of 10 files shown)


src/tiled/mapscene.cpp
```diff
@@ -285,6 +288,32 @@ QPointF MapScene::parallaxOffset(const Layer &layer) const
...
+            return;
+
+    const MapRenderer *renderer = mMapDocument->renderer();
+    const QPointF pixelOffset = renderer->tileToPixelCoords(QPointF())
+                              - renderer->tileToPixelCoords(-offset);
+
+    const auto sceneViews = views();
+    for (QGraphicsView *view : sceneViews)
+        if (auto mapView = qobject_cast<MapView*>(view))
+            mapView->forceCenterOn(mapView->viewCenter() + pixelOffset);
```
bjorn: Consider introducing a `translateViews(const QPointF &delta)` helper function and using it here as well as in `refreshScene`.

src/tiled/mapscene.cpp
```diff
@@ -285,6 +288,32 @@ QPointF MapScene::parallaxOffset(const Layer &layer) const
...
+    // when the map has its own entry in a world, the resize also moves its
+    // world position and refreshScene handles the view, so nothing to do here
+    const QString &fileName = mMapDocument->fileName();
+    if (auto worldDocument = WorldManager::instance().worldForMap(fileName))
+        if (worldDocument->world()->mapIndex(fileName) >= 0)
+            return;
+
+    const MapRenderer *renderer = mMapDocument->renderer();
+    const QPointF pixelOffset = renderer->tileToPixelCoords(QPointF())
+                              - renderer->tileToPixelCoords(-offset);
```
bjorn: I know the `tileToPixelCoords` was moved from existing code, but this might be a good time to make it use `tileToScreenCoords` instead, which should make the offset adjustment work correctly for isometric maps as well.

bjorn (commented):
I noticed a remaining issue when resizing isometric maps, which can shift around when their size changes even when the offset might be 0, due to their origin being in the top-center.

I've addressed this by calculating the screen-space offset in the `ResizeMap::swapSize` function and emitting it from there to `MapScene`, because `MapScene` would not be able to calculate the correct offset without also knowing the old size.

Now in my tests it's working well, except for hexagonal and staggered maps, which can "walk around" when undoing/redoing their size changes. Tricky case because it depends on whether the change is odd or even. I think this change is a nice improvement already anyway, but this may be worth investigating. I also get an odd behavior when resizing `hexagonal-mini.tmx` in that the initial size when grabbing a resize handle does not match up with the map.

bjorn:
I still had one issue on staggered maps, for stand-alone maps (positioned at the origin) the right edge was never snapping correctly. Fixed it by snapping the delta rather than absolute position.

Thanks for this nice improvement!

PR #4585, opened 2026-08-12
New Hardware Renderer: Non-Orthogonal Orientation Support

UltraDagon:
Adds non-orthogonal map orientation support to the MapGridItem and RegionOverlay components.

The grid shader has a switch in it that changes the posIn[OrientationType]Grid function used to determine whether or not the pixel is within the grid dash line, which is used to color in the pixel.

The RegionOverlay utilizes the [OrientationType]Renderers to draw the selected/preview region. This also adds a MapDocument dependency to the RegionOverlays, which could be used to replace other Q_PROPERTYs.

### DIFF
--- a/src/libtiled/isometricrenderer.h
+++ b/src/libtiled/isometricrenderer.h
@@ -84,10 +84,11 @@ public:
     using MapRenderer::pixelToScreenCoords;
     QPointF pixelToScreenCoords(qreal x, qreal y) const override;
 
+    QPolygonF tileRectToScreenPolygon(const QRect &rect) const;
+
 private:
     QTransform transform() const;
-    QPolygonF pixelRectToScreenPolygon(const QRectF &rect) const;
-    QPolygonF tileRectToScreenPolygon(const QRect &rect) const;
+    QPolygonF pixelRectToScreenPolygon(const QRectF &rect) const;\
 };
 
 } // namespace Tiled
--- a/src/libtiledquick/grid.frag
+++ b/src/libtiledquick/grid.frag
@@ -9,33 +9,170 @@ layout(std140, binding = 0) uniform buf {
     float scale;
     float pixelWidth;
     float pixelHeight;
+    float mapWidth;
+    float mapHeight;
     float tileWidth;
     float tileHeight;
     vec4 color;
+    vec2 skew;
+    int orientation;
 } ubuf;
+
+const float sqrt3 = 1.7320508;
 
 const float thickness = 1;
 const float dashLength = thickness * 2;
 const float spaceLength = dashLength;
 
-bool posInGrid(vec2 pixelPos)
+bool posInOrthogonalGrid(vec2 pixelPos)
 {
+    pixelPos = pixelPos * ubuf.scale + 0.5 * thickness;
+
     if ((mod(pixelPos.x, ubuf.tileWidth * ubuf.scale) < thickness &&
          mod(pixelPos.y, dashLength + spaceLength) < dashLength) ||
         (mod(pixelPos.y, ubuf.tileHeight * ubuf.scale) < thickness &&
          mod(pixelPos.x, dashLength + spaceLength) < dashLength))
-    {
         return true;
-    }
+
     return false;
+}
+
+bool posInIsometricGrid(vec2 pixelPos, bool staggered)
+{
+    pixelPos = pixelPos * ubuf.scale;
+
+    const float leftSideLength = ubuf.mapHeight * ubuf.tileHeight;
+
+    // If not staggered and out of bounds, don't draw
+    if (!staggered &&
+        (pixelPos.x + pixelPos.y < 0.5 * leftSideLength * ubuf.scale ||
+         pixelPos.x + pixelPos.y > (ubuf.pixelHeight + ubuf.pixelWidth - 0.5 * leftSideLength) * ubuf.scale + thickness ||
+         abs(pixelPos.x - pixelPos.y) > 0.5 * leftSideLength * ubuf.scale + thickness))
+        return false;
+
+    if (staggered)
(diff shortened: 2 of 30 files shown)


src/libtiled/isometricrenderer.h
```diff
@@ -84,10 +84,11 @@ class TILEDSHARED_EXPORT IsometricRenderer final : public MapRenderer
     using MapRenderer::pixelToScreenCoords;
     QPointF pixelToScreenCoords(qreal x, qreal y) const override;
 
+    QPolygonF tileRectToScreenPolygon(const QRect &rect) const;
+
 private:
     QTransform transform() const;
-    QPolygonF pixelRectToScreenPolygon(const QRectF &rect) const;
-    QPolygonF tileRectToScreenPolygon(const QRect &rect) const;
+    QPolygonF pixelRectToScreenPolygon(const QRectF &rect) const;\
```
bjorn, line 91:
```suggestion
    QPolygonF pixelRectToScreenPolygon(const QRectF &rect) const;
```

src/libtiledquick/grid.frag
```diff
@@ -9,33 +9,170 @@ layout(std140, binding = 0) uniform buf {
...
+}
+
 void main()
 {
-    vec2 pixelPos = vTexCoord * vec2(ubuf.pixelWidth, ubuf.pixelHeight) * ubuf.scale + ceil(thickness * 0.5);
+    vec2 pixelPos = vTexCoord * vec2(ubuf.pixelWidth, ubuf.pixelHeight);
+
+    bool posInGrid = false;
+    // Synchroniezed with Map::Orientation
+    switch (ubuf.orientation) {
```
bjorn, line 154: Since this switch is basically the whole shader, I wonder if it would make more sense to have separate shaders for these orientations so we'd move this switch off the GPU and also no longer need the `orientation` shader parameter.

src/libtiledquick/grid.frag
```diff
@@ -9,33 +9,170 @@ layout(std140, binding = 0) uniform buf {
...
+    if ((mod(pixelPos.x, ubuf.tileWidth * ubuf.scale) < thickness &&
+         mod(pixelPos.y, dashLength + spaceLength) < dashLength) ||
+        (mod(pixelPos.y, ubuf.tileHeight * ubuf.scale) < thickness &&
+         mod(pixelPos.x, dashLength + spaceLength) < dashLength))
+        return true;
+
     return false;
 }
 
+bool posInHexagonalGrid(vec2 pixelPos)
```
bjorn, line 95: I see this works quite nicely for a specific hexagonal configuration, but unfortunately Tiled has several properties that affect the grid which are not taken into account here:

* Stagger Axis (X / Y)
* Stagger Index (Odd / Even)
* Hex Side Length

I can imagine it could get a little complicated, unless you have a good idea about how to support these with the current approach.

src/libtiledquick/grid.frag
```diff
@@ -9,33 +9,170 @@ layout(std140, binding = 0) uniform buf {
...
         (mod(pixelPos.y, ubuf.tileHeight * ubuf.scale) < thickness &&
          mod(pixelPos.x, dashLength + spaceLength) < dashLength))
-    {
         return true;
-    }
+
+    return false;
+}
+
+bool posInIsometricGrid(vec2 pixelPos, bool staggered)
```
bjorn, line 40: What's going wrong with the isometric grid?

Image

src/libtiledquick/regionoverlay.cpp
```diff
@@ -140,17 +126,62 @@ QList<QPolygonF> RegionOverlay::invalidPolygons() const
...
-        path.addRect(r);
+    if (!mMapDocument)
+        return {};
 
-    QList<QPolygonF> polygons = path.simplified().toSubpathPolygons();
+    QPainterPath path;
 
-    QTransform transform;
-    transform.scale(mTileSize.x(), mTileSize.y());
+    switch (mMapDocument->map()->orientation())
```
bjorn, line 134: Looks like a good time for adding another virtual function to the `MapRenderer` interface, which returns a `QPainterPath` for a given tile `QRegion`. The code for this is mostly already in the renderers, in their `drawTileSelection` function, and could be factored out.

src/tiled/editablemap.cpp
```diff
@@ -604,12 +614,17 @@ void EditableMap::setParallaxOrigin(const QPointF &parallaxOrigin)
...
+
     if (auto doc = mapDocument()) {
         push(new ChangeMapOrientation(doc, static_cast<Map::Orientation>(value)));
     } else if (!checkReadOnly()) {
         map()->setOrientation(static_cast<Map::Orientation>(value));
         mRenderer.reset();
     }
+
+    if (oldOrientation != orientation())
+    emit orientationChanged();
```
bjorn, line 627: These signals should really be emitted based on change events, these ones based on `MapChangeEvent` in particular (except, actually, in case there is no `mapDocument`, so an emit could still be part of the above `else if`.

src/tiled/editablemap.cpp
```diff
@@ -580,18 +580,28 @@ void EditableMap::setStaggerIndex(StaggerIndex value)
...
     else if (!checkReadOnly())
         map()->setSkewX(value);
+
+    if (oldSkewX != skewX())
+        emit skewChanged();
 }
 
 void EditableMap::setSkewY(int value)
 {
+    int oldSkewY = skewY();
```
bjorn, line 596: Rather than remembering `oldSkewY` for comparison later, I would suggest to just compare here and `return` immediately, when `skewY() == value`.

bjorn (changes requested):
It's quite impressive to see this mostly working! I noticed a few cases with issues though, and I added some feedback on the implementation.

Issue #590, opened 2014-01-12
Tiled doesn't remember widget size


Gornova:
My test case: create a new map with tiled, then resize tileset preview to bigger than levels (for example) then close.
On next tiled running, my adjustment on tiled widget size are gone, so I need to do it again and again.
Could you add a preference to save widget height (and not just position?)

Here my 2 screens to show my problem:

1) widget with custom height
![1](https://f.cloud.github.com/assets/24104/1897071/5569e738-7bca-11e3-8a0c-790de7738f8d.PNG)

2) then after close and re-open Tiled:
![2](https://f.cloud.github.com/assets/24104/1897073/5a6a321a-7bca-11e3-813f-5112f95d5793.PNG)

my 2 cents: seems to be a problem only on default-right widgets, on left there is no problem O_O

bjorn:
This issue stalled a long time ago because I could not reproduce the problem. Tiled already remembers both the position and the size of these views. I'm still not sure what the problem could be here.

Could @Gornova and @CyberMew maybe try the current version of Tiled, to see if the problem has somehow been fixed in the meantime?

CyberMew:
Haven't been using Tiled for 2 years :o 

Just installed it and it seems fixed :dancer: 

Just a side note, I found an app named `Enfusoft` that could let you tiled patterns easily and quickly. Do you think Tiled will have some of the basic features as well? After all, this is `Tiled` :heart_eyes:

Gornova:
My test system is windows 8.1 64 bit

1) downloaded 0.16.2 for windows 32bit
2) downloaded this tileset for example http://opengameart.org/content/hearts-platformer-tileset-32-pixels
3) created a new map and resized tiled window

![before](https://cloud.githubusercontent.com/assets/24104/16757313/5f7c1ac4-480b-11e6-85b2-3a45293d7fa0.png)

4) closed and reopened the same map

![after](https://cloud.githubusercontent.com/assets/24104/16757321/6b71ea98-480b-11e6-86ba-2ddcdb1b1fdb.png)

For me this issue is not fixed! As you can see properties panel on left is not readable when I reopen tiled

CyberMew:
It worked for me on a windows 7 x64 machine, but I believe it was a brand new install and not an upgrade over existing installation. Not sure if that affects anything. And in my test I only tried to adjust the right panels and not the left panels. Maybe I'll try to test again when I am at home.

CyberMew:
[...]

PR #4586, opened 2026-08-12
fix: restore layout synchronously to prevent extreme defaults

### LINKED ISSUE #4580: After restarting Tiled, the views are resized to (extreme) defaults, every time. 🧙
### Describe the bug

When restarting Tiled, then views are resized to an extreme default (especially on ultra-wide monitors like 38 inch), so I always need to the resize it by hand.

### Steps to reproduce

Close Tiled.
Open it. 

### Expected behavior

The exact positioned views and widgets should have the same size and position (incl. maximized / windowed) after restarting Tiled. 

### OS version

Windows 11

### Tiled version

Version 2026.07.17 (Portable). Same on installed version.

shubhrai23:
Resolves #4580

**What changed:** Removed the deferred layout restoration logic (the \mHasRestoredLayout\ flag and the 200ms \QTimer::singleShot\) in \MainWindow::readSettings()\. The layout is now restored synchronously immediately after \
estoreGeometry()\ finishes.

**Why:** In modern Qt versions, restoring a \QMainWindow\'s state (which includes dock widget geometry and sizes) asynchronously during or after the window is shown can cause the layout engine to calculate extreme or corrupted defaults for dock widgets, especially on ultra-wide displays.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,6 +11,7 @@
 * Scripting: Added MapObject.resolvedClassName() (by MatusGuy, #4529)
 * Fixed crash when the selection becomes empty while starting a move (#4536)
 * Fixed Properties view update on 'Reset Template Instance' and 'Replace With Template' actions
+* Fixed restoring of the layout for maximized windows on startup (#4580)
 * Linux: Added file associations for .tmj, .tsj, .tiled-project and .world files (by miffe, #4550)
 * Linux: Fixed the window icon on some Wayland compositors (by Balló György, #4567)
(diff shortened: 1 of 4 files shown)


bjorn:
> In modern Qt versions

Since when is that? The change you're reverting (a49d87f9b) was at least tested against Qt 6.3. If the original issue reported in #590 was fixed in Qt it would be good to know since when (and I believe this also affected Linux). The minimum supported Qt version is now 6.2.

Please remove the unrelated CI changes. I'm not sure why you'd need them since `master` has been building fine, but if any fixes are indeed necessary please submit them separately.

PR #4590, opened 2026-08-13
Added drag-to-create maps in the world view

kanishka0411:
With the World Tool active, a new map can now be created by dragging out arectangle in empty world space. The rectangle snaps to the world grid, with ctrl bypassing the snapping and Escape or right-click cancelling the drag.
The new map takes its orientation, tile size, render order and tilesets from the nearest map in the world, and is saved immediately as `untitled-N.tmx` alongside the `.world` file, since worlds refer to their maps by file name.

### DIFF
--- a/src/tiled/abstractworldtool.cpp
+++ b/src/tiled/abstractworldtool.cpp
@@ -113,18 +113,6 @@ MapDocumentPtr nearestMapDocument(const World *world, const QRect &rect)
     return DocumentManager::instance()->loadDocument(nearestFileName).objectCast<MapDocument>();
 }
 
-// Picks the first free untitled-N.tmx alongside the world file
-QString newMapFileName(const World *world)
-{
-    const QDir dir = QFileInfo(world->fileName).dir();
-
-    for (int i = 1; ; ++i) {
-        const QString fileName = dir.filePath(QStringLiteral("untitled-%1.tmx").arg(i));
-        if (!QFileInfo::exists(fileName))
-            return fileName;
-    }
-}
-
 } // namespace
 
 Qt::CursorShape cursorForHandle(int handle)
@@ -464,8 +452,7 @@ void AbstractWorldTool::addAnotherMapToWorld(QPoint insertPos)
 }
 
 // Creates a new map covering the given world rect and adds it to the current
-// world. The map is saved right away, since a world refers to its maps by
-// file name.
+// world. The map stays unsaved until the user saves it.
 void AbstractWorldTool::createMapAt(const QRect &worldRect)
 {
     auto worldDocument = worldForMap(mapDocument());
@@ -507,13 +494,10 @@ void AbstractWorldTool::createMapAt(const QRect &worldRect)
     QRect entryRect = MapRenderer::create(map.get())->mapBoundingRect();
     entryRect.moveTo(worldRect.topLeft());
 
-    const QString fileName = newMapFileName(world);
+    auto mapDocument = MapDocumentPtr::create(std::move(map));
 
     QUndoStack *undoStack = worldDocument->undoStack();
-    undoStack->beginMacro(tr("Create Map"));
-    undoStack->push(new CreateMapFileCommand(std::move(map), fileName));
-    undoStack->push(new AddMapCommand(worldDocument, fileName, entryRect));
-    undoStack->endMacro();
+    undoStack->push(new AddUnsavedMapCommand(worldDocument, mapDocument, entryRect));
 }
 
 // Asks for a file name and creates a new world containing the current map.
--- a/src/tiled/abstractworldtool.cpp
+++ b/src/tiled/abstractworldtool.cpp
@@ -619,7 +619,7 @@ QPoint AbstractWorldTool::snapPoint(QPoint point, MapDocument *document) const
 QPoint AbstractWorldTool::currentMapWorldPos() const
 {
     if (auto worldDocument = worldForMap(mapDocument()))
-        return worldDocument->world()->mapRect(mapDocument()->fileName()).topLeft();
+        return worldDocument->mapRect(mapDocument()).topLeft();
     return QPoint();
 }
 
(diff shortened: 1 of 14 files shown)


kanishka0411:
While working on this i ran into a few things i wasn't sure about-

- Right now undoing a create also moves the new file to the trash, so you don't end up with leftover untitled files lying around. but that means the undo command is doing file operations and nothing else in the codebase doesthat. Is that okay or would you rather the file gets written outside the undo stack?
- Should the new map open in a tab right away or is it fine that it just shows up in the world view?
- The new file gets saved next to the `.world` file, should it go next to the nearest map instead? like in sticker-knight all the maps live in `map/`

Ablu:
> * Right now undoing a create also moves the new file to the trash, so you don't end up with leftover untitled files lying around. but that means the undo command is doing file operations and nothing else in the codebase doesthat. Is that okay or would you rather the file gets written outside the undo stack?

I think the most user-friendly solution would be to allow an unsaved map to be added to an (unsaved) world view. That would allow creating the new map just like the "New Map" option (which creates a new map without saving it yet). It would also save the user from ending up with many `untitled-{}.tmx` maps as there does not seem to be a way within tiled to rename maps? But adding unsaved maps to the world view probably complicates the saving procedure (maybe the whole project always would have to be saved at once?).

I think this needs @bjorn to decide :)

> * Should the new map open in a tab right away or is it fine that it just shows up in the world view?

I think it would make sense to open it right away. But the open is just a click away anyway. So not sure how important it it.

> * The new file gets saved next to the `.world` file, should it go next to the nearest map instead? like in sticker-knight all the maps live in `map/`

Yeah, this one seems tricky too... I think next to the `.world` file is probably the only reasonable choice one can make. `map/` seems highly opinionated and is unlikely to match what other projects want. Again, this problem would go away if we would allow unsaved maps to be added. In that case the user would get prompted for the location on the first save of the map...

bjorn:
Due to the way `WorldMapEntry` refers to maps by their file name, I'd suggest to keep the automatic saving in this PR (and deleting, but I'd say only when the map has no content).

But consider making the necessary changes to allow maps to stay unsaved and be saved only on-demand (either when user triggers save on map, or tries to save the world). I think that could be done by keeping a `MapDocumentPtr` in the `WorldMapEntry` to keep the map loaded along with the world. You can make it as separate commit, and if that change seems small enough, you can make it part of this PR.

src/tiled/worlddocument.h
```diff
@@ -69,14 +70,27 @@ class WorldDocument final : public Document
 
     void swapWorld(std::unique_ptr<World> &other);
 
+    // A map that is part of this world but does not have a file yet
+    struct UnsavedMap {
+        MapDocumentPtr mapDocument;
+        QRect rect;
+    };
```
bjorn, line 77: Mainly curious for now, but why did you decide to add a separate collection for this, rather than adding a `MapDocumentPtr` member to `WorldMapEntry`?

bjorn:
When I click on an unsaved map in the world, it opens in its own tab, no longer showing the world, because `worldManager.worldForMap` works based on a file name only, it currently can't return the world.

src/tiled/worlddocument.h
kanishka0411: Because WorldMapEntry is in libtiled and MapDocument is editor side, so the pointer can't go in there. I agree the separate list isn't ideal though,that's also why move/resize of unsaved maps are still open. A middle way could be making them normal entries with an empty fileName and a non-owningMap* just for identity, while WorldDocument keeps the` MapDocumentPtr` alive.would that work?

kanishka0411:
> When I click on an unsaved map in the world, it opens in its own tab, no longer showing the world, because `worldManager.worldForMap` works based on a file name only, it currently can't return the world.

yeah I think this also comes from the same root cause, the scene looks up the current map's world by file name. so if we give the entry a `Map*`identity like i mentioned above, the scene can fall back to looking up unsaved maps by pointer and this would get fixed

src/tiled/worlddocument.h
bjorn: Hmm, right, I had missed that. I don't think non-owning `Map*` is a good idea, it sounds too fragile to make sure those pointers stay valid.

I think we can stick with the separate collection in `WorldDocument` for these maps, but patch up `WorldManager::worldForMap` to take the `MapDocument*` instead of just its file name, so that it can look for maps by pointer as well.

bjorn:
I haven't looked very deeply at the code yet, but in testing I noticed a few problems with the unsaved maps:

* Pressing them immediately selects them, whereas for saved maps this happens on release. It seems abrupt and I think it should better be consistent.

* Creating more new maps with an unsaved map selected does not create add new maps in the right position in the world.

* The first time you switch to a new map in a world, the view displays it fitting in the view. This is the normal behavior when a map becomes visible the first time, but in the case of switching to a new map within a world, we should suppress or override this somehow. I kind of wonder now why this doesn't happen when switching to saved maps which were not opened before yet.

But it's nice to be able to add a map like this, especially in combination with the world grid! I do worry a little that it becomes too easy to trigger by accident, since adding a new map to a world can be a pretty rare thing to do.

src/tiled/abstractworldtool.cpp
```diff
@@ -533,6 +614,25 @@ QPoint AbstractWorldTool::snapPoint(QPoint point, MapDocument *document) const
     return point;
 }
 
+// The scene puts the current map at 0,0, so converting between scene and
+// world coordinates is just a shift by the current map's world position
+QPoint AbstractWorldTool::currentMapWorldPos() const
+{
+    if (auto worldDocument = worldForMap(mapDocument()))
+        return worldDocument->world()->mapRect(mapDocument()->fileName()).topLeft();
```
bjorn: Here's probably our bug with new maps being created at an offset when an unsaved map is selected:
```suggestion
        return worldDocument->mapRect(mapDocument()).topLeft();
```

bjorn (changes requested):

PR #4603, opened 2026-09-01
Add option to export selected area

leenattress:
## Summary

Adds an **Only export selected area** option to the Export As Image dialog.

When enabled, Tiled exports only the currently selected map area instead of the full map. The crop calculation accounts for renderer scaling and centering so the saved image matches the visible selection.

## How It Works

- Use **Rectangular Select** (`R`) to select the tiles you want to export.
- Open **Export As Image**.
- Enable **Only export selected area**.
- Tiled exports only the selected tiles.

## When Nothing Is Selected

- The option is disabled, so it cannot be turned on.
- Tiled remembers your previous choice, so it may appear ticked but greyed out.
- Tiled checks again before exporting and shows a message instead of creating an incorrect image.

## Why This Is Needed

For projects targeting constrained platforms such as the Game Boy (my personal use case), exported maps often need to stay within strict size limits.

A useful workflow is to design a large world as one connected map in Tiled: individual levels, rooms, and areas can be laid out together so their connections are visible and easy to verify. This is particularly helpful for tools such as GB Studio and NES studio, where seeing the whole world at once makes it easier to manage transitions and keep areas wired together correctly.

However, the game still needs each area exported separately. The new option allows selecting one area of the larger map and exporting only that selection as a PNG, without splitting the source world into many disconnected map files. This keeps the overview and editing workflow in Tiled while producing smaller images suitable for the target platform.

## Testing

- Added regression tests for selected-area image crop geometry.
- Built and ran `test_export` successfully.

### DIFF
--- a/src/tiled/exportasimagedialog.cpp
+++ b/src/tiled/exportasimagedialog.cpp
@@ -43,12 +43,38 @@ using namespace Tiled;
 namespace session {
 static SessionOption<bool> visibleLayersOnly { "exportAsImage.visibleLayersOnly", true };
 static SessionOption<bool> useCurrentScale { "exportAsImage.useCurrentScale", false };
+static SessionOption<bool> selectedAreaOnly { "exportAsImage.selectedAreaOnly", false };
 static SessionOption<bool> drawTileGrid { "exportAsImage.drawTileGrid", false };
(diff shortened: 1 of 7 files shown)


leenattress:
Manually tested the Linux AppImage (I'm on arch today) built by CI. Selected-area export works as expectedfrom your build

PR #4612, opened 2026-09-16
Fix world scripting APIs without the editor

### LINKED ISSUE #4584: api: tiled.loadWorld() fails in scripts
### Describe the bug

Tiled does not load world files in scripts evaluated at the CLI.

### Steps to reproduce

inside any js file (not in an extensions folder):

```js
tiled.loadWorld(filename);
console.log(tiled.worlds); // []
```

then in the cli:

```bash
tiled --project project.tiled-project -e script.js
```

observe that `tiled.worlds` does not contain a world. no error is printed, no matter the validity of `filename`.

### Expected behavior

ideally, the world is loaded, and added to `tiled.worlds` for further scripting. at the very least, an error message would be appreciated.

### OS version

Arch Linux

### Tiled version

1.12.2

Shubham-Padkonde:
Fixes #4584.

Command-line scripts can load worlds through `WorldManager`, but `tiled.worlds` returns an empty list because it unnecessarily requires a GUI `DocumentManager`. The same GUI guard prevents `worldsChanged` from being forwarded. `loadWorld` also discards loading errors.

Read worlds and connect their change notification independently of the GUI, and propagate load failures as script errors. Document the error behavior. Add a scripting test target that runs the public JavaScript API without creating a `DocumentManager`, covering load/unload visibility, change notifications, and a missing world file.

Validation on Ubuntu 22.04, Qt 6.2.4, GCC 11:
- All three new regression cases fail on the original code and pass with the fix (QtTest reports 5 passes including setup/cleanup).
- Built the Tiled application and all five test executables.
- A real `tiled -e` script verified load/unload visibility, notifications, and a catchable missing-file error.
- Across the five test executables: 60 passed, 1 failed, including setup/cleanup. The failure is `test_Properties::loadCircularReference` at its `membersA.contains("b")` assertion. That executable and its `libtiled` dependency are unchanged and do not link the modified `libtilededitor` library. Map reader, staggered renderer, automapping, and the new scripting suite pass.

The Qbs aggregate test runner could not locate the local shared libraries in this build environment. The counts above come from running the same built executables with explicit library paths and each suite's expected working directory.

Upstream CI: Ubuntu, macOS, Windows Qt 6, documentation, and CodeQL checks passed. The ARM64 snap job failed while installing `kde-qt6-core24-sdk` from `latest/stable`, before compiling this change ([job log](https://github.com/mapeditor/tiled/actions/runs/35140588768/job/104943856270)). AppVeyor is still pending.

Prepared with Codex assistance, including implementation, regression tests, and this description.


bjorn:
Thanks, fix looks good! :-)

PR #4619, opened 2026-09-20
docs: update pixi-tiledmap entry in framework support list

riebel:
I maintain [pixi-tiledmap](https://github.com/riebel/pixi-tiledmap). Its entry in the framework support list dates from #4499 and now understates what the library does, describing only loading and rendering.

Since then it also supports runtime tile editing, procedural map generation, and export back to Tiled JSON, along with all layer types and map orientations including the 1.12 oblique projection. This updates the entry to match.

The rewrite stays about the same size: I dropped "typed API" (implied by "written in TypeScript") and "ESM/CJS dual output" (packaging detail) to make room. It also corrects `Pixi.JS` to `PixiJS` and removes a trailing space.

No other entries touched.


bjorn:
@riebel Thanks for the update! Seems you're doing amazing work with this extension and are on track to replace Tiled entirely. The showcase is mind-blowing! :-D

riebel:
Wow, thank you, that means a lot to me! Don't worry, Tiled stays the editor, for now... 😉 Still fiddling around with the showcase.

